PR #35, opened 2015-07-20
Limit callstack frames in breakpoint tests

chuckries:
Due to differences in callstack frames of systems calls on different
platforms/sdks for android, limit the callstacks frames to user frames.


gregg-miskelly:
:+1:

PR #37, opened 2015-07-21
Make MI changes for compatibilit with lldb:

paulmaybee:
1. eliminate unnecessary quotes on variable names
  2. use '-' as frame parameter when using "--frame" option on var-create; spec does not define usage
 Changes to be committed:
	modified:   src/MICore/MICommandFactory.cs


chuckries:
:+1:

PR #39, opened 2015-07-22
Fix lldb shared lib loading processing.

paulmaybee:
Changes to be committed:
	modified:   src/MICore/MICommandFactory.cs
	modified:   src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs

### DIFF
--- a/src/MICore/MICommandFactory.cs
+++ b/src/MICore/MICommandFactory.cs
@@ -617,16 +617,6 @@ namespace MICore
             return false;
         }
 
-        public override Results IsModuleLoad(string cmd)
-        {
-            Results results = null;
-            if (cmd.StartsWith("shlibs-added,", StringComparison.Ordinal))
-            {
-                results = MIResults.ParseResultList(cmd.Substring(13));
-            }
-            return results;
-        }
-
         public override bool AllowCommandsWhileRunning()
         {
             return false;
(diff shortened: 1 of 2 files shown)


src/MICore/MICommandFactory.cs
```diff
@@ -617,16 +617,6 @@ public override bool SupportsStopOnDynamicLibLoad()
             return false;
         }
 
-        public override Results IsModuleLoad(string cmd)
```
gregg-miskelly: Should we make the base method non-virtual now that we don't have any overrides of it?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -101,23 +101,33 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
                         _waitDialog.ShowWaitDialog(file);
                     }
                 }
-                else if (results.Results.Contains("shlib-info")) //lldb
+                else if (this.MICommandFactory.Mode == MIMode.Clrdbg)
```
gregg-miskelly: NOTE: Now that the lldb library load event is more applicable, we should update clrdbg to match lldb. I added a cleanup item to the clrdbg list.

gregg-miskelly:
:+1:

src/MICore/MICommandFactory.cs
paulmaybee: done

PR #41, opened 2015-07-24
Support 32-bit apps on 64-bit Android devices

gregg-miskelly:
Address the following problems we had when targetting a 64-bit Android
device:

- Our ABI detection logic fell down because the device's ABI is arm64
instead of one of the arm values. Changed the code to check the
'abilist' property as weell.

- At least on Galaxy S6 thet we tested, run-as was adding extra debug
spew which broke our processing of the output from pwd. Modified this
code to ignore the extra spew.

- The 32-bit app_process is now called 'app_process32' and at least on
this device, ADB wasn't handling the redirect

### DIFF
--- a/src/AndroidDebugLauncher/Launcher.cs
+++ b/src/AndroidDebugLauncher/Launcher.cs
@@ -161,6 +161,7 @@ namespace AndroidDebugLauncher
                 string workingDirectory = null;
                 string gdbServerRemotePath = null;
                 string gdbServerSocketDescription = null;
+                string exePath = null;
                 Task taskGdbServer = null;
                 int gdbPortNumber = 0;
                 int progressCurrentIndex = 0;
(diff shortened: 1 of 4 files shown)


src/AndroidDebugLauncher/Launcher.cs
```diff
@@ -333,7 +330,33 @@ private LocalLaunchOptions SetupForDebuggingWorker(CancellationToken token)
...
+                    try
+                    {
+                        fileSystem.Download(@"/system/bin/" + app_process, exePath, true);
+                    }
+                    catch (AdbException)
+                    {
+                        // Older devices don't have an 'app_process32', only an 'app_process', so retry
+                        // NOTE: libadb doesn't have an error code property to verify that this is caused
+                        // by the file not being found.
+                        retry = is32BitTarget;
```
gregg-miskelly: Need to change this to only catch the execption if it is a 32-bit target.

gregg-miskelly: Fixed.

chuckries:
:+1:

PR #43, opened 2015-07-24
Fix PipeTransport crash

gregg-miskelly:
This resolves a non-deterministic crash I ran into in OnReadStreamAborted which was taking down Visual Studio.


chuckries:
:+1:

PR #47, opened 2015-07-30
Resolve NullReferencException in AndroidDebugLauncher.Launcher.Dispose

gregg-miskelly:
Dispose wasn't checking to see if _launchOptions was non-null


chuckries:
:+1:

PR #49, opened 2015-08-05
Fix official build to drop files

gregg-miskelly:
The official build is no longer dropping the build files to the build share. The problem is that this used to work for setting 'OutDir' but BuildAndTest.proj forces OutDir to a particular value. The other problem we had was that the output directory was getting filled with VS reference assemblies.

To resolve this, I created a new 'DropFiles' target in MIDebugPackage.csproj which takes care of copying the files.

I also marked our various VS reference assemblies with '<Private>False</Private>' which mostly fixed our build to stop dropping VS assemblies into the output directory.

### DIFF
--- a/src/MIDebugPackage/MIDebugPackage.csproj
+++ b/src/MIDebugPackage/MIDebugPackage.csproj
@@ -6,8 +6,6 @@
     <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">14.0</VisualStudioVersion>
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <!--When called from TeamBuild, change OutDir to be the directory we want copied to the drop server-->
-    <OutDir Condition="'$(TF_BUILD_BINARIESDIRECTORY)'!=''">$(TF_BUILD_BINARIESDIRECTORY)\$(Configuration)</OutDir>
   </PropertyGroup>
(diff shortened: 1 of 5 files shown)


src/MIDebugPackage/MIDebugPackage.csproj
```diff
@@ -193,7 +155,23 @@
...
+    <DropSignedFile Include="$(OutDir)\Microsoft.IOSDebugLauncher.dll"/>
+    <DropUnsignedFile Include="$(OutDir)\Microsoft.IOSDebugLauncher.pkgdef"/>
+    <DropSignedFile Include="$(OutDir)\Microsoft.JDbg.dll"/>
+    <DropSignedFile Include="$(OutDir)\Microsoft.MICore.dll"/>
+    <DropSignedFile Include="$(OutDir)\Microsoft.MIDebugEngine.dll"/>
+    <DropUnsignedFile Include="$(OutDir)\Microsoft.MIDebugEngine.pkgdef"/>
+    <DropSignedFile Include="$(OutDir)\Microsoft.MIDebugPackage.dll"/>
+    <DropUnsignedFile Include="$(OutDir)\Microsoft.MIDebugPackage.pkgdef"/>
+    <DropSignedFile Include="$(OutDir)\Microsoft.AndroidDebugLauncher.dll"/>
+    <DropUnsignedFile Include="$(OutDir)\Microsoft.AndroidDebugLauncher.dll"/>
```
chuckries: dropping AndroidDebugLauncher twice?

chuckries:
:+1:  looks good, added comments.

src/MIDebugPackage/MIDebugPackage.csproj
gregg-miskelly: Thanks, good catch.

gregg-miskelly: I also forgot PDBs from my list. Fixing that too.

PR #50, opened 2015-08-06
Run CodeFormatter

chuckries:
I ran codeformatter over this just for housekeepings sake. one wrinkle is that codeformatter likes to fix up a bunch of things in the autogenerated LaunchOptions.xsd.types.desginer.cs. I ran codeformatter, then built again to regenerated, and left the generated version as is.


gregg-miskelly:
:+1:

PR #53, opened 2015-08-11
Add support for executing console interface commands

wiizzard:
Non-MI commands specified via SetupCommands/CustomLaunchSetupCommands are currently ignored silently. This patch adds the capability to execute console interface commands. It is especially useful for remote debugging situations in which you want to pass commands to the gdbserver via the "monitor ..." command in order to set it up properly for debugging.


chuckries:
:+1:

chuckries:
Thanks for the contribution!

PR #60, opened 2015-08-25
Remove extra '()' from CLRDBG frames

gregg-miskelly:
CLRDBG will format frames itself, so the MIEngine doesn't need to append arguments. In the initial bring up, I disabled fetching the arguments, but MIEngine was still adding an extra '()' at the end of frames. This removes that.


gregg-miskelly:
@orablu FYI

chuckries:
:+1:

PR #61, opened 2015-08-25
Address FxCop error in Natvis.cs

gregg-miskelly:
The new version of FxCop we are now using in official builds is issuing an error in the use of XmlReader by the natvis code. The error is because we were not providing a settings object which set the XmlResolver property set to null. With this checkin we are now doing that.

I also made the code a bit closer to our launch option reader code by having the XmlReader ignore a bunch of things that we will not care about (comments, processing instructions, whitespace).


chuckries:
:+1:

PR #62, opened 2015-08-26
Add MI debugger working directory to PATH

Andrew-MSFT:
For some reason GDB locally requires that it's working directory be on the PATH, so adding that to the PATH variable before launching GDB


chuckries:
:+1:

PR #65, opened 2015-09-05
Fix LLDB/GDB Frame formatting

gregg-miskelly:
Commit e2c73817b9db22e7cc04600e311d9732089c3752 regressed LLDB/GDB Frame formatting because I accentially returned the wrong value in the base virtual method. This fixes it.


chuckries:
:+1:

PR #66, opened 2015-09-08
Fixes for glass tests

gregg-miskelly:
This fixes two issues that I found while running the Android glass tests.

### DIFF
--- a/src/JDbg/Jdwp.cs
+++ b/src/JDbg/Jdwp.cs
@@ -168,12 +168,12 @@ namespace JDbg
 
             internal void Abort()
             {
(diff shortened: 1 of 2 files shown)


chuckries:
:+1: 

Thank you for improving the tests! Should these be expected events in every test?

gregg-miskelly:
Short answer: this test is the only one we need to fix. I wrote up an explanation as to why this is in https://github.com/Microsoft/MIEngine/wiki/Testing-with-Glass#idebugbreakpointboundevent2 let me know if that is clear or not.

chuckries:
Excellent, thank you!

PR #67, opened 2015-09-09
CLRDBG exception setting support for the MIEngine

gregg-miskelly:
This checkin contains changes to the MIEngine to support exception
settings. Currently this only supports exception settings for CLRDBG,
but I added notes for how we could make this work for GDB/LLDB in
the future if we need this.


gregg-miskelly:
FYI @orablu , here are the MIEngine changes for this.

chuckries:
:+1:

orablu:
:+1:

PR #68, opened 2015-09-09
64 bit support for android debugging

chuckries:
Implement 64 bit debugging support for native android apps. This work
required:

1. Adding the paths for the 64 bit targeted files on windows (gdb.exe,
linker, libc.so, etc.)
2. Adding ARM64 and x64 launch options support.
3. Working around 64 bit android issues with deployment.
4. Updating libadb.il for new abi's

### DIFF
--- a/src/AndroidDebugLauncher/Launcher.cs
+++ b/src/AndroidDebugLauncher/Launcher.cs
@@ -244,8 +244,16 @@ namespace AndroidDebugLauncher
                                 allowedAbis = new DeviceAbi[] { DeviceAbi.armeabi, DeviceAbi.armeabiv7a };
                                 break;
 
+                            case TargetArchitecture.ARM64:
+                                allowedAbis = new DeviceAbi[] { DeviceAbi.arm64v8a };
+                                break;
+
                             case TargetArchitecture.X86:
                                 allowedAbis = new DeviceAbi[] { DeviceAbi.x86 };
+                                break;
+
+                            case TargetArchitecture.X64:
+                                allowedAbis = new DeviceAbi[] { DeviceAbi.x64 };
                                 break;
 
                             default:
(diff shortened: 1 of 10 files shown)


chuckries:
The x64 remote gdbserver path logic needs a little more testing...

src/AndroidDebugLauncher/Launcher.cs
```diff
@@ -244,10 +244,18 @@ private LocalLaunchOptions SetupForDebuggingWorker(CancellationToken token)
                                 allowedAbis = new DeviceAbi[] { DeviceAbi.armeabi, DeviceAbi.armeabiv7a };
                                 break;
 
+                            case TargetArchitecture.ARM64:
+                                allowedAbis = new DeviceAbi[] { DeviceAbi.arm64v8a };
+                                break;
+
                             case TargetArchitecture.X86:
                                 allowedAbis = new DeviceAbi[] { DeviceAbi.x86 };
```
gregg-miskelly: Should this allow DeviceAbi.x64 as well?

src/AndroidDebugLauncher/Launcher.cs
```diff
@@ -341,7 +356,7 @@ private LocalLaunchOptions SetupForDebuggingWorker(CancellationToken token)
                     {
                         fileSystem.Download(@"/system/bin/" + app_process, exePath, true);
                     }
-                    catch (AdbException) // 64-bit TODO: add 'when (is32BitTarget)'
+                    catch (AdbException)
```
gregg-miskelly: Don't you want --
  when (app_process_suffix == "32")

Here

src/AndroidDebugLauncher/Launcher.cs
```diff
@@ -496,6 +521,37 @@ private void VerifySdkVersion()
             }
         }
 
+        private string VerifyGdbServerPath(string workingDirectory)
```
gregg-miskelly: nit: Might want to call this 'GetGdbServerPath' since we are doing more than just verifying it.

src/AndroidDebugLauncher/Launcher.cs
```diff
@@ -496,6 +521,37 @@ private void VerifySdkVersion()
...
+            string output = ExecCommand(lsCommand);
+            if (string.Compare(output, gdbServerPath, StringComparison.OrdinalIgnoreCase) == 0)
+            {
+                //gdbserver is symlinked correctly
+                return gdbServerPath;
+            }
+            else if (_launchOptions.TargetArchitecture == TargetArchitecture.X64)
+            {
+                //start looking other places, only do this on 64 bit
+                //TODO: This needs some additional testsing
```
gregg-miskelly: Should we open an Android bug for this and link to it?

gregg-miskelly:
Signing off with comments.

src/AndroidDebugLauncher/Launcher.cs
chuckries: I originally had this, but a 64 bit should report both abi's as compatible. My thinking was that if a device supported ONLY the 64 bit abi, then it would actually be incompatible with 32 bit. What do you think?

src/AndroidDebugLauncher/Launcher.cs
chuckries: The behavior would be that if app_process_suffix is 64 and we fail to download, the AdbException would propogate, correct? And if app_process_suffix is 32, we set retry, and then if we fail to download "app_process", that AdbException will propogate? I will add this then.

gregg-miskelly: Correct.

src/AndroidDebugLauncher/Launcher.cs
gregg-miskelly: Ah yes. You are correct.

src/AndroidDebugLauncher/Launcher.cs
chuckries: Done

src/AndroidDebugLauncher/Launcher.cs
chuckries: Done

PR #70, opened 2015-09-14
CLRDBG message event + output window improvements

gregg-miskelly:
commit 5b082abc997235668acbe7a892425b08d6efcd70
Date:   Mon Sep 14 08:21:49 2015 -0700

    Support for CLRDBG 'message' events
    
    CLRDBG uses an MI Extension to send message events. This adds support for it to the MI Engine.

commit be9b8ca39044e5af13287ec4d54bb2aa4afecbcb
Date:   Mon Sep 14 09:29:23 2015 -0700

    Output message improvements
    
    1. Using plink to send messages to the target debugger and back results in an echo of the command that we sent. We were treating this echo as debuggee output instead of ignoring it. Now we ignore it.
    
    2. We were echoing the output of message events, thread create events, and thread exit events. Not sure why. Now we aren't.
    
    3. The debugger was waiting for the first run mode transition in the underlying debugger to decide that AD7 was ready to receive output string events. This doesn't really make much sense since it is really the AD7 side which needs to be ready, and that has nothing to do with run mode events from the underlying debugger. Now we wait for the resume process call instead.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -28,7 +28,7 @@ namespace MICore
         public event EventHandler ProcessExitEvent;
         public event EventHandler DebuggerExitEvent;
         public event EventHandler DebuggerAbortedEvent;
-        public event EventHandler<string> MessageEvent;
+        public event EventHandler<string> OutputStringEvent;
         public event EventHandler EvaluationEvent;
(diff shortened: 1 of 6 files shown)


chuckries:
:+1:

PR #71, opened 2015-09-18
Start of MIEngine refactor to support VS Code

gregg-miskelly:
This PR creates a new 'VSCodeRefactor' branch where we can workout changes needed to support VS Code in the MIEngine. This initial work is to create a new dll (Microsoft.DebugEngineHost) that will provide host services to debug engines. This hopefully will allow all the assemblies of the MIEngine to work in either VS or VS Code without needing to define yet-another-API-layer by swapping out just Microsoft.DebugEngineHost.

There is still a lot of work to go, but here is a start.


gregg-miskelly:
FYI @caslan @orablu

src/MIDebugEngine/AD7.Impl/AD7ProgramProvider.cs
```diff
@@ -28,6 +28,8 @@ int IDebugProgramProvider2.GetProviderProcessData(enum_PROVIDER_FLAGS Flags, IDe
         {
             processArray[0] = new PROVIDER_PROCESS_DATA();
 
+            Debug.Fail("Will this be called?");
```
chuckries: Should this stay in here?

src/AndroidDebugLauncher/AndroidDebugLauncher.csproj
```diff
@@ -137,6 +137,10 @@
     </Content>
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\DebugEngineHost\DebugEngineHost.csproj">
```
chuckries: Do the launchers use reference the wait loop? Is that why they need these project references?

chuckries:
LGTM

src/AndroidDebugLauncher/AndroidDebugLauncher.csproj
gregg-miskelly: Yes.

src/MIDebugEngine/AD7.Impl/AD7ProgramProvider.cs
gregg-miskelly: Good catch.

PR #75, opened 2015-09-25
Remove SerialPortLaunchOptions support

gregg-miskelly:
The SerialPortLaunchOptions support required functionality from the .NET Framework which is missing from CoreCLR. Since we aren't currently using this code for anything, and haven't in many years, this removes it.


paulmaybee:
:+1:

PR #76, opened 2015-09-28
Enables attach for applications and services.

rajkumar42:
Enables attach for applications and services.

- JDWP Port forwarding is required for both launch and attach.
- Moving the "run-as" code inside the TargetEngine.Native inside the block.


chuckries:
:+1:

PR #79, opened 2015-10-06
Sending ProcessName on JavaLaunchOptions.

rajkumar42:


src/MICore/LaunchOptions.cs
```diff
@@ -199,11 +199,12 @@ public override string ExePath
 
     public sealed class JavaLaunchOptions : LaunchOptions
     {
-        public JavaLaunchOptions(string jvmHost, int jvmPort, string[] sourceRoots)
+        public JavaLaunchOptions(string jvmHost, int jvmPort, string[] sourceRoots, string processName)
```
gregg-miskelly: One slightly confusing thing -- the package name on Android is not the name of the process executable. Might be worth a comment that this is more of a logical name than necessarily the true process name.

gregg-miskelly:
:+1:

src/MICore/LaunchOptions.cs
rajkumar42: Agreed!

PR #81, opened 2015-10-06
Remove quotes in -var-list-children

chuckries:
lldb does not accept quotes by default and gdb does not required them.


gregg-miskelly:
:+1:

paulmaybee:
:+1:

PR #83, opened 2015-10-06
Replace MIDebugEngine, MICore with PCLs

chlowell:
This replaces MIDebugEngine and MICore with PCLs targeting .NET 4.6 and ASP.NET Core 5.0. The new PCLs build but won't run because they depend on a stub version of DebugEngineHost pending its reorganization. There are some noteworthy changes to the solution:
* Visual Studio requires Universal Windows App Development Tools 1.1 to load the PCLs
* It remains necessary to add a local package source containing the Interop nupkg
* One test is broken by a project system bug (I left a TODO describing this and its workaround)
* xsd.exe is no longer used because the code it generates isn't portable
* A workaround must be applied each time the new libraries' resource string code is generated (see commit comment)


src/MICore/LaunchOptions.cs
```diff
@@ -433,7 +435,6 @@ public static XmlReader OpenXml(string content)
             settings.IgnoreProcessingInstructions = true;
             settings.IgnoreWhitespace = true;
             settings.NameTable = new NameTable();
-            settings.XmlResolver = null;
```
gregg-miskelly: Do you know if there is something else we should do to set this to null? There is a Microsoft-internal FXCop rule in the SDL bar that requires it to be set to null?

gregg-miskelly:
Signing off with comments. Let me know when you want me to merge this.

src/MICore/LaunchOptions.cs
chlowell: No, the XmlResolver property doesn't exist in the portable contract.

gregg-miskelly: That is really unfortunate. I guess the best solution then is to invoke the property via reflection (or 'dynamic') if it exists, and otherwise ignore it. Would you mind doing that?

gregg-miskelly:
Looks good to me.

PR #88, opened 2015-10-08
DebugEngineHost Fixes

gregg-miskelly:
This fixes the assembly version of Microsoft.DebugEngineHost to be set to '1.0.0.0' so that it is consistent accross the various implementations.

This also fixes the VSIX to exclude the contract version of this assembly and instead use the VS version. Still need to figure out how to exclude a few other assemblies, but this is a good start.


gregg-miskelly:
@chlowell can you take a look as well?

chlowell:
Looks good to me :+1:

PR #89, opened 2015-10-10
Add device udid to appRemotePath call for vcremote

chuckries:
We need to send the device udid to both the seutp for debugging call and
the appremotepath call. In a perfect world, we would make this a global
setting in the header of all vcremote packets, but that fix will have to
wait for another day.

Additionally, fix a URI issue in setupForDebugging where the second '?'
needed to be an &.

### DIFF
--- a/src/IOSDebugLauncher/VcRemoteClient.cs
+++ b/src/IOSDebugLauncher/VcRemoteClient.cs
@@ -58,7 +58,7 @@ namespace IOSDebugLauncher
         public Launcher.RemotePorts StartDebugListener()
         {
             string remotePortsJsonString;
-            CallVcRemote(new Uri("debug/setupForDebugging?target=" + _launchOptions.IOSDebugTarget.ToString() + "?deviceUdid=" + _launchOptions.DeviceUdid, UriKind.Relative), LauncherResources.Info_StartingDebugListener, out remotePortsJsonString);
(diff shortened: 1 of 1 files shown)


src/IOSDebugLauncher/VcRemoteClient.cs
```diff
@@ -58,7 +58,7 @@ public static VcRemoteClient GetInstance(IOSLaunchOptions options)
         public Launcher.RemotePorts StartDebugListener()
         {
             string remotePortsJsonString;
-            CallVcRemote(new Uri("debug/setupForDebugging?target=" + _launchOptions.IOSDebugTarget.ToString() + "?deviceUdid=" + _launchOptions.DeviceUdid, UriKind.Relative), LauncherResources.Info_StartingDebugListener, out remotePortsJsonString);
+            CallVcRemote(new Uri("debug/setupForDebugging?target=" + _launchOptions.IOSDebugTarget.ToString() + "&deviceUdid=" + _launchOptions.DeviceUdid, UriKind.Relative), LauncherResources.Info_StartingDebugListener, out remotePortsJsonString);
```
gregg-miskelly: Do you still want to send the DeviceUdid if you didn't get one in the launch request?

gregg-miskelly:
Signing off with a comment.

src/IOSDebugLauncher/VcRemoteClient.cs
chuckries: It doesn't hurt us at all. The parameter will just be empty, and vcremote will default to an empty string if there is no deviceudid parameter.

PR #92, opened 2015-10-14
Get MIEngine working after the refactor

gregg-miskelly:
- Update the name of the managed resource to match what is embedded in the module
- Update the MIDebugPackage and MICoreUnitTests projects to copy required facade assemblies
- Update the MIDebugPackage to exclude extra assemblies it was copying which we don't plan to ship with the MIEngine


gregg-miskelly:
@chlowell please take a look. @chuckries you might want to look also, as we will need to follow this pattern as we add more dependencies moving forward.

src/MICoreUnitTests/MICoreUnitTests.csproj
```diff
@@ -97,6 +98,9 @@
     <Compile Include="MIResultsTests.cs" />
     <Compile Include="NdkVersionTests.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <None Include="@(RequiredNetFX46FacadeAssemblies)">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
gregg-miskelly: This should probably be 'Always' instead

chuckries:
:+1:

chlowell:
Looks good to me :+1:

PR #93, opened 2015-10-14
Pull Microsoft.VisualStudio.Debugger.Interop.Portable from nuget.org

gregg-miskelly:
I published Microsoft.VisualStudio.Debugger.Interop.Portable to nuget.org, so we no longer need a checked in copy.


gregg-miskelly:
@chlowell please take a look

chuckries:
:+1:

PR #94, opened 2015-10-15
I fought the build, and I won (package management fix)

gregg-miskelly:
Summary: With this checkin, I am changing how we restore packages in
the MI Engine, and also, on a much more minor note, how we compile our
contract IL.

Problem: I couldn't figure out any normal way to set the package output
directory when NuGet restores of our project.json projects in the IDE.
I didn't like this, but could live with it if this was the only
problem. But on the PR machines packages were getting restored to some
unknown directory, so it wasn't working at all.

Solution:
1. Move package restore to the same spot in both the IDE build and the
build.cmd build by creating 'prebuild.csproj', and making all projects
depend on this project.

_NOTE_ when adding a new project, we need to add this new dependency.
To do so, bring up the project context menu and invoke:
  Build Dependencies->Project Dependencies

This will add a new item in the .sln file to declare the dependency, so
as long as we review .sln changes, we should be good.

2. Have prebuild.csproj set the package output switch (which actually
is respected)

3. Set the NuGetPackagesDirectory property in all projects, so projects
know where to look for their dependencies.

4. For good measure, I moved the .il compilation from MICore to
prebuild.csproj since that is really where it belongs.

### DIFF
--- a/src/prebuild/prebuild.csproj
+++ b/src/prebuild/prebuild.csproj
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\build\all_projects.settings.targets" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{95C0421F-D4C1-46E8-9177-6DAA4D71D0F2}</ProjectGuid>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <GenerateAssembly Include="$(ILDir)*.il">
+      <Visible>false</Visible>
+    </GenerateAssembly>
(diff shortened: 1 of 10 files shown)


src/prebuild/prebuild.csproj
```diff
@@ -0,0 +1,44 @@
...
+    </GenerateAssembly>
+  </ItemGroup>
+
+  <Target Name="GenerateAssemblies"
+        Condition="'@(GenerateAssembly)' != ''"
+        Inputs="@(GenerateAssembly)"
+        Outputs="@(GenerateAssembly->'$(GeneratedAssembliesDir)%(FileName).dll')"
+        AfterTargets="Build">
+    <PropertyGroup>
+      <IlAsmCommand>"$(windir)\Microsoft.NET\Framework\v4.0.30319\ilasm.exe"</IlAsmCommand>
```
chuckries: Is this path safe to use?

chuckries:
Signing off with a comment. This should also fix some issues we had with parallel build.

src/prebuild/prebuild.csproj
gregg-miskelly: On my box, there is no other copy of ilasm.exe under Program Files, and since it ships with the desktop .NET Framework, I can't think of any reason it wouldn't be there.

PR #98, opened 2015-10-17
Add handling for '\r'

gregg-miskelly:
The c string decoding code in the MI engine didn't handle '\r'. This adds code to handle that.


caslan:
LGTM

PR #103, opened 2015-10-20
Fix Windows VS Code debugging with new CoreCLR-based OpenDebugAD7

gregg-miskelly:
This checkin fixes the setup script so that OpenDebugAD7 works again now that we have moved to CoreCLR.

TODO: Need to work with the VS Code team to understand a 'coreclr' runtime so that we don't need to hardcode the path to it.


gregg-miskelly:
@caslan please take a look.

caslan:
LGTM

orablu:
:+1:

PR #105, opened 2015-10-21
MIEngine: Add support for logging in VS Code

gregg-miskelly:
This checking changes how we enable logging in the MIEngine so that the details of how we enable logging and where the logging winds up moves out of Microsoft.MICore.dll and into Microsoft.DebugEngineHost.dll. This way we can use console logging in VS Code, and log to a temp file in VS.


gregg-miskelly:
@caslan @orablu please review

caslan:
LGTM

PR #106, opened 2015-10-23
Re-enable XML tools in VSCodeRefactor

gregg-miskelly:
This checkin contains two commits to re-enable the two XML tools which were disable when converting to PCLs.

commit 593578ef1ed4f6f545296f3202fba3d381c2150f
Date:   Thu Oct 22 08:42:26 2015 -0700

    Create Microsoft.MICore.XmlSerializers.dll
    
    After an investigation, I determined that there is no practicle way to use sgen code with CoreCLR - the types aren't in a contract, so therefore they are subject to change. So this makes changes to run sgen in a more traditional way where the sgen code is in its own dll. This way MICore can stay portable but we will still get the benefits of a static XML deserializer in the VS IDE, which is where we care about it anyway.

commit caa8c4af269998ea0dcbdc76f3d0c17f71b67746
Date:   Wed Oct 21 19:49:15 2015 -0700

    Re-enable xsd.exe
    
    During the work to make MICore CoreCLR compatible, the code to run xsd.exe during the build was disabled because xsd.exe added extra attributes which are missing from the contract assemblies. This brings back xsd.exe, now with a post-processing step to strip out the extra attributes.


gregg-miskelly:
@chuckries can you review? No rush.

chuckries:
LGTM :+1:

PR #107, opened 2015-10-24
Don't crash VS on non-UInt32 debugger exit code

yacoder:
This fixes a crash reported for VisualRust extension: PistonDevelopers/VisualRust#190 (to reproduce the crash, just install the Visual Rust extension in Visual Studio and follow the steps from the linked issue).

Let me know if there's a better way to address this. The assumption in several places across MIEngine seems to be that the exit code always fits into UInt32.

Alternatively here we could truncate the string to fit into uint, or report a hardcoded non-zero exit code for VS, but I'm not sure it adds any value, it may still be confusing that in the Output window VS shows a different exit code than what's in the GDB output...

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -192,8 +192,11 @@ namespace Microsoft.MIDebugEngine
                     // GDB sometimes returns exit codes, which don't fit into uint, like "030000000472".
                     // And we can't throw from here, because it crashes VS.
                     // Full exit code will still usually be reported in the Output window,
-                    // but let VS think it was "0" in this case.
-                    uint.TryParse(results.Results.FindString("exit-code"), out processExitCode);
+                    // but here let's return "uint.MaxValue" just to indicate that something went wrong.
+                    if (!uint.TryParse(results.Results.FindString("exit-code"), out processExitCode))
+                    {
+                        processExitCode = uint.MaxValue;
+                    }
                 }
 
                 // quit MI Debugger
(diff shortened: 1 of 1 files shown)


gregg-miskelly:
I wonder if it would make sense to use:
if (UInt64.TryParse(..., out u64ExitCode))
  processExitCode = (uint)u64ExitCode;

Which clearly isn't ideal, but it seems like it might be better to truncate then drop it entirely. What do you think?

gregg-miskelly:
Or thinking about it more, maybe just set the exit code to uint.MaxValue if TryParse fails? That way it at least will not show up as '0' since that often means success.

yacoder:
I'm not sure any of these alternatives are so much better than returning zero. The truncated number might accidentally map into some code (including zero) that has some other meaning for the user or for the extension. Same with `uint.MaxValue`, it could mean something else for the receiver.

Nevertheless, I'm willing to change to any of these alternatives if it gets this merge accepted, any of them are better than exception+crash :-)

gregg-miskelly:
Totally agree -- the main thing is to not crash :). I think my preference would be uint.MaxValue - not ideal, but seems better than zero.

yacoder:
Ok, here's the Output in VS with the latest version of the fix (no crash):

    Breakpoint 1, 0x0000000000402d9a in main ()
    =thread-created,id="5",group-id="i1"[New Thread 8432.0x24f8]
    [Thread 8432.0x1a2c exited with code 3221225786]
    =thread-exited,id="4",group-id="i1"[Thread 8432.0x19f0 exited with code 3221225786]
    =thread-exited,id="2",group-id="i1"[Thread 8432.0x2dc0 exited with code 3221225786]
    =thread-exited,id="3",group-id="i1"[Inferior 1 (process 8432) exited with code 030000000472]
    =thread-exited,id="5",group-id="i1"=thread-exited,id="1",group-id="i1"=thread-group-exited,id="i1",exit-code="01777777777770000000472"The program 'rust_application3.exe' has exited with code -1 (0xffffffff).

gregg-miskelly:
Looks reasonable to me. Do you want to push your latest changes?

gregg-miskelly:
Oh, never mind. I missed your new changes. Looks good to me.

PR #108, opened 2015-10-24
OutDir / build configuration fixes

gregg-miskelly:
1: Before, for official/PR/command line build, we would define OutDir in BuildAndTest.proj. We defined in in a way that it wasn't possible to override in any sub project. This worked until VSCodeRefactor. In VSCodeRefactor this caused a theoritical build race because we actually have two assemblies with the same name (Microsoft.DebugEngineHost) that were both going to the same folder.

FIX: Remove the setting from BuildAndTest.proj, and instead fix all of our other projects to output into that directory. This has the other nice property that build in and out of the IDE work the same.

2: Not all of our projects had a Build.Lab configuration. With this checkin they all do.


gregg-miskelly:
@chuckries can you take a look when you get a chance?

chuckries:
:+1:

chuckries:
I would try a lab build after this is merged to confirm that the drop still looks normal

PR #109, opened 2015-10-26
Update Microsoft.VisualStudio.Debugger.Interop.Portable to 1.0.1

gregg-miskelly:
Microsoft.VisualStudio.Debugger.Interop.Portable v1.0.0 had a bug that
we block types from loading on MacOS/Linux. This switches to the fixed
version.


chuckries:
:+1:

Issue #44, opened 2015-07-27
Integration with BlackBerry development plugin


phofman:
Hi,
I am the author of the [plugin that enables development targetting BlackBerry devices](/phofman/vs-plugin) directly from Visual Studio (or being more precise - I forked originally abandoned BlackBerry project and made it usable).
The mentioned plugin already has a GDB integration and as a starting point I wish to write some glue layer, that would allow me to have both (original and MIEngine) available. Do you see any problem with my integration with the MIEngine? Should I create a user story for that propose or any architecture drawings? All ideas are welcome.
This plugin has also some nice features in it, that I would love to port back, if possible. At the end of the day I would love to have MIEngine as the main debugger, but this could take longer time as I estimate ;)

gregg-miskelly:
Sure, it should definitely be possible to have your plugin use either your current GDB integration, or MIEngine. A few questions:
1. Do you run gdbserver on the Blackberry, and GDB on the PC?
2. Can you issue GDB commands to start the app on the black berry? Or do you have some sort of ‘side channel’ where you connect to the blackberry, start the app, and then attach to the app using GDB and/or gdbserver?

So the switching back and forth should be relatively easy - in your project system (the bit of code that calls IVsDebugger[2/3/4].LaunchDebugTargets[2/3/4] you swap out the engine GUID of whatever engine you have with the MIEngine’s guid (Microsoft.MIDebugEngine.EngineConstants.EngineId). You can see some example code to start debugging with the MIEngine in the 'LaunchDebugTarget' function in MIEngine\src\MIDebugPackage\MIDebugPackagePackage.cs. The harder part (or depending on how your code is structured, this could be pretty easy) is to set things up so that you can share the code for setting up GDB. We can provide better advice once we know what you need to do.

phofman:
Great! Here is some brief architecture view. I described it below in more details.

![gdb](https://cloud.githubusercontent.com/assets/475630/8920786/ea68212a-34cd-11e5-8e0e-4876cf48cf22.png)
(copied from [pull-request #663](/BlackBerry/VSPlugin/pull/663))

ad1. There are already all components in place, ie:
 * the GDBServer already running on the device (all the time when 'development mode' is on)
 * and GDB (ntoarm-gdb.exe) available on PC (when dedicated NDK is installed, separately than the VS package), started when needed (when the actual debugging begins).

[...]

Issue #19, opened 2015-06-11
Target process isn't killed during stop debugigng on Linux+GDB


gregg-miskelly:
On the version of GDB that I have on Linux (Ubuntu 7.7.1-0ubuntu5~14.04.2), gdb doesn't support an -exec-abort command. I went through all the mi commands using strings on it. It appears that the only way to kill the process is to use the non-mi 'kill' command.

PR #110, opened 2015-10-27
Support of BlackBerry devices

phofman:
Hi,

I finally managed to finalize the integration of BlackBerry-specific GDB.
The discussion about it was initially started in [issue #44](/Microsoft/MIEngine/issues/44), but I went a little bit further with my implementation. You need to have installed [BlackBerry plugin v3.7.0 for Visual Studio 2015](https://visualstudiogallery.msdn.microsoft.com/2b21e9d1-a54b-4d58-878f-2d53604e85b1) to see it running. All previous versions won't let run (and serialize options for) MIEngine at all.

Big picture:
 * I created dedicated launch options and launcher not to mess to much with existing _generic_ implementations

* BB GDB has a dedicated set of startup commands

 * GDB used by BlackBerry comes from QNX environment and even if docs say differently, the reality is it **doesn't support** async MI execution; to overcome this situation I use a dedicated 'host' application that generates CTRL+C signal to stop running GDB and start accepting MI commands; in code it's accessible through IBreakHandler interface

 * I added some very basic tests

 * this GDB doesn't implement `-exec-abort` command, I use `kill` instead

 * I changed the way `UseUnixSymbolPaths` was evaluated

Let me know, what do you think about it.
Thanks!


gregg-miskelly:
Hey Paweł, 

First off, it is very cool that you got this to work and we are really happy that you are making Visual Studio better by adding Blackberry support. However we want to make sure the official MIEngine repo contains only code that Visual Studio can commit to being the long term owner of.  While there are parts of these changes that we would be happy to accept, there are also a lot of Blackberry-specific code in this PR that we cannot commit to own.

So that leaves us with two options, and we will leave it up to you which one you go with –
#1 – You could refactor your code to have the blackberry launcher + blackberry launch options XSD in its own repo, and you could submit a PR for adding the hooks you need to MICore. I think that would mean the changes to support IBreakHandler, support for UseUnixSymbolPaths, and the ‘BlackBerryLaunchOptions’->CLSID mapping.
#2 – You could maintain your own fork of the MIEngine so that you can continue to ship with your own ship schedule. See https://github.com/Microsoft/MIEngine/wiki/Redistribution for more information if you go down this route.

Thanks,
Gregg

phofman:
Hi Gregg,

Thanks and sorry for the delayed response.
All true, I understand your point of view about rejection and respect it. Especially I don't want to ask you to support BlackBerry-specific, 5-years old, non-asynchronous GDB (nor buy their devices etc.), if this doesn't match your project's assumptions. I would offer my help on that front, but that just adds another layer of potential complications to the way your project is driven.

Anyway... I don't want to release MIEngine on my own. That's why I would vote for option #1. It also matches the approach I took about dedicated port-supplier. It could stay inside the [BlackBerry plugin](/phofman/vs-plugin) repo directly, as this plugin must be installed to make the whole stack to work.

I can handle the refactoring. Feel free to reject current pull-request. I will start another branch, at the moment I started the implementation and patch MIEngine without any BlackBerry-specific code. Your history will stay clean that way. Although few thing just keep bothering me, if I move the code to my repo. Maybe you have some thoughts about it. They are:
 - I would need to create an interface for a service to parse the *LaunchOptions* content and produce the instance; simple mapping LaunchOptions <-> CLSID is not enough; interface for that service would need to be defined in MIEngine Core, while implementation would stay next to the BB-LaunchOptions; there must be a way to pass me the XML (content or reader) for processing
 - MIEngine should of course handle situation, when this launcher-args-parser (no matter, how named) is not registered in VS2015-hive registry node
 - how could I reference the MIEngine library? my whole project would depend on it as source of all the required interfaces? it's not installed in GAC, rather somewhere inside VS folder; not mentioning it should be possible to correctly start my BB-plugin even, if MIEngine is not installed, as this is possible combination; how should I handle it in my dev-environment, where I could introduce changes on both fronts (BB-plugin and MIEngine) and ci-environment, to just produce "official releases"
 - what about the GDB termination? BB-GDB uses totally different command, will you merge it into master repo or I need to come up with a different way of invoking it?

Cheers,
 Paweł

gregg-miskelly:
Thanks Paweł,
RE: there must be a way to pass me the XML (content or reader) for processing -
I would recommend in LaunchOptions.cs doing something like this:

    case "BlackBerryLaunchOptions":  
    {  
        // NOTE: external customer launchers implement their own XML deserialization
        launcherXmlOptions = options;
        clsidLauncher = new Guid("43BC8C7F-5184-4FE8-9ECF-F33A498375EE");  
    }  
    break;

This should allow the launch options to be passed to your launcher as a string, and you can deserialize them yourself. This is actually how we used to implement the Android and iOS launchers as well.

RE: MIEngine should of course handle situation, when this launcher-args-parser (no matter, how named) is not registered in VS2015-hive registry node
Agreed. You should be able to throw an exception in this case.

RE: how could I reference the MIEngine library?
You could either -
1. Publish a Microsoft.MICore.dll binary that you can build against - checkin and/or fetch from someplace else (ex: OneDrive/DropBox share, or some HTTP endpoint that you have access to), Microsoft.MICore.dll.
-or-
2. You could build a 'contract' version of Microsoft.MICore.dll - this doesn't work quite yet, but pretty soon we expect to have the Microsoft.MICore build produce a 'contract' C# file, which is a C# file that has the same shape as the real thing, but all the methods just throw NotImplementedException. You can then checkin+build this file in another repo, such as yours, in some sort of 'prebuild' step.
-or-
3. You could have some sort of 'prebuild' project which kicks off a build of the MIEngine. In the VSCodeRefactor branch we have added a ['prebuild' project](https://github.com/Microsoft/MIEngine/blob/VSCodeRefactor/src/prebuild/prebuild.csproj) to do some steps, and you could do something similar - in your case, building the MIEngine sln.

gregg-miskelly:
<Sorry, GitHub stopped letting me edit the above comment. Continuing...>

RE: what about the GDB termination?
We actually have an (issue)[https://github.com/Microsoft/MIEngine/issues/19] about this on other platforms too. I _think_ you could safely change the MICommand for GDB to always use 'kill'.

phofman:
Thanks a lot Gregg! I will follow your instructions.

PR #112, opened 2015-11-06
Error handling improvements

gregg-miskelly:
commit 41d14acb3618aabfcf3943a0f6de3b70dd673f70
Date:   Fri Nov 6 10:31:05 2015 -0800

    Fix issues with reliably reporting the MI debugger exited on MacOS
    
    1. Sending the 'logout' command would raise an IOException because the pipe was already closed.
    2. The pipe would close before the process was reported as terminated, so we would often report the process without the exit code. I increased the timeout to prevent this.

commit 3108d04aa2690b4bdbeb1ee27150174313071c55
Date:   Thu Nov 5 17:07:45 2015 -0800

    MIEngine error enhancements
    
    - Many of our error strings still had 'GDB' in them even if the underlying debugger wasn't GDB. Now we plumb through the name of the debugger.
    - For CLRDBG we stop ignoring the failure code from kill. We should probably eventually do this for LLDB also. But I don't have things setup to test LLDB at the second.
    - When gdb/lldb/clrdbg exits were wern't displaying the exit code. Now we do.
    - Stop catching exceptions in [Can]TerminateProcess. For TerminateProcess, I wanted to be able to get the message in OpenDebugAD7. For CanTerminateProcess, there didn't seem to be a reason to have a filter at all.

### DIFF
--- a/src/MICore/Transports/PipeTransport.cs
+++ b/src/MICore/Transports/PipeTransport.cs
@@ -72,7 +72,14 @@ namespace MICore
         {
             if (_writer != null)
             {
-                Echo("logout");
+                try
+                {
+                    Echo("logout");
+                }
+                catch (Exception)
+                {
+                    // Ignore errors if logout couldn't be written
+                }
             }
 
             base.Close();
@@ -98,7 +105,7 @@ namespace MICore
 
             try
             {
-                if (_process.WaitForExit(50))
+                if (_process.WaitForExit(1000))
                 {
                     // If the pipe process has already exited, or is just about to exit, we want to send the abort event from OnProcessExit
                     // instead of from here since that will have access to stderr
--- a/src/MICore/Transports/PipeTransport.cs
+++ b/src/MICore/Transports/PipeTransport.cs
@@ -176,14 +176,13 @@ namespace MICore
                 }
                 catch (InvalidOperationException)
                 {
-                    exitCode = "Unknown";
                 }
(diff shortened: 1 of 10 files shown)


gregg-miskelly:
@caslan please take a look.

chuckries:
:+1:

src/MICore/Transports/PipeTransport.cs
```diff
@@ -72,7 +72,14 @@ public override void Close()
...
             if (_writer != null)
             {
-                Echo("logout");
+                try
+                {
+                    Echo("logout");
+                }
+                catch (Exception)
+                {
+                    // Ignore errors if loggout couldn't be written
```
gregg-miskelly: Spelling

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -516,27 +505,23 @@ int IDebugEngineLaunch2.TerminateProcess(IDebugProcess2 process)
...
             }
-            catch (MIException e)
-            {
-                return e.HResult;
-            }
-            catch (Exception e)
+            catch (ObjectDisposedException)
             {
-                return EngineUtils.UnexpectedException(e);
+                // if we have already shutdown, ignore the failure
```
caslan: what is the 'if' for? Isn't it always the case if you are here?

caslan:
LGTM, just had one comment

PR #114, opened 2015-11-11
Cleanup contract definition of DebugEngineHost

gregg-miskelly:
DebugEngineHost had a bunch of different source files, and no XML document comments. This follows the same pattern that CoreFX does where they have one C# file with everything in it for the assembly (since it doesn't take much code to throw NotImplementedException :)). I also added XML doc comments and fixed some spelling mistakes in the comments.

I also enabled FxCop for the contract assembly to make sure 'we' don't mispell anything in an API that will someday be public (and of course by 'we' I am really just thinking of myself :)).


src/DebugEngineHost.Stub/Contract.ruleset
```diff
@@ -0,0 +1,238 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="Copy of Microsoft All Rules" Description="This rule set contains all rules. Running this rule set may result in a large number of warnings being reported. Use this rule set to get a comprehensive picture of all issues in your code. This can help you decide which of the more focused rule sets are most appropriate to run for your projects." ToolsVersion="14.0">
```
gregg-miskelly: I don't like how the IDE created this file. Let me change it.

gregg-miskelly:
@caslan can you take a look?

caslan:
LGTM

PR #117, opened 2015-11-13
Fix FxCop warning in Natvis.cs

gregg-miskelly:
For some reason the previous CodeAnalysis run didn't report the XmlResolver issue in natvis.cs, even though the issue was in both spots. This fixes the issue there also.

I also updated the comment in LaunchOptions.cs, as when I was copy/pasting the code, I realized that the comment wasn't accurate.


gregg-miskelly:
@caslan please take a look.

caslan:
LGTM

PR #120, opened 2015-11-17
Drop additional files for Java debugger

gregg-miskelly:
This checkin outputs additional files that we will want to use in the java debugger from our official builds.


gregg-miskelly:
@chuckries can you take a look when you have a chance?

chuckries:
:+1:

PR #124, opened 2015-11-29
put _internalName in quotes in VariableInformation.InternalFetchChildren

RYDB3RG:
_internalName can contain spaces (template typenames), which leads to
invalid usage error in gdb's `mi_cmd_var_list_children()` because it thinks
its additional parameters

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -483,7 +483,7 @@ namespace Microsoft.MIDebugEngine
         }
         private async Task InternalFetchChildren()
         {
-            Results results = await _engine.DebuggedProcess.CmdAsync(string.Format("-var-list-children --simple-values {0}", _internalName), ResultClass.None);
+            Results results = await _engine.DebuggedProcess.CmdAsync(string.Format("-var-list-children --simple-values \"{0}\"", _internalName), ResultClass.None);
 
             if (results.ResultClass == ResultClass.done)
             {


chuckries:
@paulmaybee do we know if this will work against both gdb and lldb?

chuckries:
Thank you for the PR, @RYDB3RG! We originally included the quotes here but removed them because gdb did not require them and lldb did not accept them. I am going to try to get the necessary change into lldb before mergning this into MIEngine. In the meantime, could you provide repro instructions so I can take a look at this? We didn't know internal names could contain spaces.

RYDB3RG:
Hi,

Thanks for looking into this issue.

Repro:

main.cpp:
```
#include <vector>
#include <string>

int main(int argc, const char **args)
{
    std::string s = "hAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAllo";
    std::vector<std::string> v = {s, s, s, s, s, s};
    return 0;         // breakpoint here
}
```

compile with:
`g++ -std=c++0x -O0 -g main.cpp`


Launch XML:
```
<PipeLaunchOptions xmlns="http://schemas.microsoft.com/vstudio/MDDDebuggerOptions/2014" 
    PipePath="plink.exe" 
    PipeArguments="-i ... -batch -t gdb --interpreter=mi" 
    ExePath="..../a.out"
/> 
```



Drilling into `v` in debugger will lead to this error:
```
MS_MIDebug: (26415) <-1021-var-list-children --simple-values var5.std::_Vector_base<std::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::basic_string<char, std::char_traits<char>, std::allocator<char> > > >
MS_MIDebug: (26416) ->1021-var-list-children --simple-values var5.std::_Vector_base<std::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::basic_string<char, std::char_traits<char>, std::allocator<char> > > >
MS_MIDebug: (26417) ->
MS_MIDebug: (26417) ->1021^error,msg="-var-list-children: Usage: [PRINT_VALUES] NAME [FROM TO]"
```

chuckries:
Here is the lldb commit: http://reviews.llvm.org/D15168

@gregg-miskelly is cldbg ready for this change?

chuckries:
@gregg-miskelly it looks like it is. This PR looks good to me then.

gregg-miskelly:
Sounds good to me.

PR #125, opened 2015-11-30
Handle invalid threads in Continue

orablu:
VS Code is giving a thread Id of 0 when continuing from async break. Update Continue method to better handle cases where we can't get a debugged thread from the given pThread.

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7Engine.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Engine.cs
@@ -570,7 +570,7 @@ namespace Microsoft.MIDebugEngine
         public int Continue(IDebugThread2 pThread)
         {
             AD7Thread thread = pThread as AD7Thread;
-            _pollThread.RunOperation(() => _debuggedProcess.Continue(thread != null ? thread.GetDebuggedThread() : null));
+            _pollThread.RunOperation(() => _debuggedProcess.Continue(thread?.GetDebuggedThread()));
 
             return Constants.S_OK;
         }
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -569,9 +569,8 @@ public int CauseBreak()
         // and the debugger does not want to actually enter break mode.
         public int Continue(IDebugThread2 pThread)
         {
-            AD7Thread thread = (AD7Thread)pThread;
-
-            _pollThread.RunOperation(() => _debuggedProcess.Continue(thread.GetDebuggedThread()));
+            AD7Thread thread = pThread as AD7Thread;
+            _pollThread.RunOperation(() => _debuggedProcess.Continue(thread != null ? thread.GetDebuggedThread() : null));
```
chuckries: this can just be thread?.GetDebuggedThread();

chuckries:
:+1:

orablu:
Thanks!

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -569,9 +569,8 @@ public int CauseBreak()
         // and the debugger does not want to actually enter break mode.
         public int Continue(IDebugThread2 pThread)
         {
-            AD7Thread thread = (AD7Thread)pThread;
-
-            _pollThread.RunOperation(() => _debuggedProcess.Continue(thread.GetDebuggedThread()));
+            AD7Thread thread = pThread as AD7Thread;
```
gregg-miskelly: Might want to add a note here indicating why we are doing this. Something like:
// VS Code currently has a bug where a thread isn't provided in some cases. Work
// around this by passing zero as the thread for now.

PR #129, opened 2015-12-05
Gdb-mi add support python pretty printing

olizit:
I take a look into well designed Mi-engine source code. Unfortunaltely Mi-engine is not fully compliante with python base pretty pretting. Gdb mi-interface is a tricky API regarding python-base pretty printing. I add support of python-base pretty printing.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -213,7 +213,14 @@ namespace Microsoft.MIDebugEngine
             TypeName = results.TryFindString("type");
             Value = results.TryFindString("value");
             Name = results.FindString("exp");
-            CountChildren = results.FindUint("numchild");
+            if (results.Contains("dynamic"))
+            {
+                CountChildren = 1;
+            }
+            else
+            {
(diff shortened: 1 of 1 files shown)


paulmaybee:
Hi olizit,

I have just one issue with your change. I am afraid that it may conflict with natvis processing, especially when used on Android where the C++ std library may have both pretty-printers and natvis available. It might be best if you disabled natvis processing if the value was already pretty-printed by gdb.

-Paul

olizit:
I understand your fear but first of all pretty printer feature is not enabled by default inside MIEngine. An additional command must be send at startup (-enable-pretty-printing) to load pretty printer. 
I am wondering something about lldb debugger engine. According to my poor knowledge in lldb , pretty printer seems to be always enable. Am I right ?
Moreover, I don't know how to disable natvis. Someone can help me ?

olizit

olizit:
Thanks for your precisions. Now CLA is signed.

olizit

paulmaybee:
Merged into master

gregg-miskelly:
Hi Olizit, Paul merged this on the command line instead of through the tool. I will close this request since it has been merged.

PR #131, opened 2015-12-09
Rename Debug.Lab and Release.Lab configs to Lab.Debug and Lab.Release

chuckries:
The default configuration for the Solution appears to be selected by
choosing the first one alphabetically. The Debug.Lab configuration was
being chosen, which was causing F5 to be broken for a fresh clone. This
rename will cause Debug to be the default configuration and F5 should work
from a brand new clone of the project. This change requires altering
official builds to understand the new configuration.


gregg-miskelly:
:+1:

PR #134, opened 2015-12-09
support reading RegistryRoot\Debugger settings

rick111111:
support reading RegistryRoot\Debugger settings


gregg-miskelly:
:+1:

PR #135, opened 2015-12-09
Fix InvalidOperationException in JDbg 'OnDisconnect'

chuckries:
This fixes an issue we saw come in through Watson. On disconnect, JDbg
attempts to abort all existing WaitingOperations. The abort inside of a
foreach loop was causing the dictionary to be altered, which causes an
InvalidOperationException. Fix this by copying each
WaitingOperationDescriptor into a temporary list, clearing the main list,
and then aborting each one out of the temporary list.

### DIFF
--- a/src/JDbg/Jdwp.cs
+++ b/src/JDbg/Jdwp.cs
@@ -22,8 +22,6 @@ namespace JDbg
         private TcpTransport _transport;
         private JdwpCommand.IDSizes _IDSizes;
 
-
-
         private Jdwp(string hostname, int port)
         {
             _transport = new TcpTransport(hostname, port, OnPacketReceived, OnDisconnect);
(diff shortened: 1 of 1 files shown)


gregg-miskelly:
Does Kofe have the same issue?

src/JDbg/Jdwp.cs
```diff
@@ -66,21 +66,26 @@ private void OnDisconnect(/*OPTIONAL*/ SocketException socketException)
...
                 {
-                    if (socketException == null)
-                    {
-                        operation.Value.Abort();
-                    }
-                    else
-                    {
-                        operation.Value.OnSocketError(socketException);
-                    }
+                    operationsToAbort.Add(operation.Value);
```
gregg-miskelly: nit: use operationsToAbort = new List<WaitingOperationDescriptor>(_waitingOperations.Values);

So that List can be smart about settings its size based on the size of the collection being copied.

gregg-miskelly:
Signing off with comments.

chuckries:
It looks like Kofe does potentially have the same issue. I will submit a PR to them with this fix and your fix that changes SetException -> TrySetException.

chuckries:
@gregg-miskelly please take another glance at this if you can.

gregg-miskelly:
:+1

gregg-miskelly:
:+1:

PR #140, opened 2015-12-17
Remove global DebuggedProcess variable

chuckries:
This does not completely pave the way for multiple process debugging from
MIEngine, but it sets the stage. This removes the evil global variable
storing the one and only DebuggedProcess object.

TODO: remove the DebuggedProcess property from AD7Engine and allow for
multiple DebuggedProcess objects to be created and managed.


gregg-miskelly:
:+1:

gregg-miskelly:
BTW: Having DebuggedProcess as a singleton is mostly okay. I believe the only time that would pose a problem is if we wanted to support some sort of child process debugging scenario. But that scenario would have lots of problems beyond just that.

PR #145, opened 2016-01-05
Merging master to release

rajkumar42:


chuckries:
Why are we merging release into master? It looks like this just adds a bunch of merge commits and doesn't actually change any files except alters the -var-list-children quotes change. I don't believe this should get merged.

rajkumar42:
my bad "master from release" is the wrong direction.

PR #150, opened 2016-01-07
CLRDBG / VSCode telemetry start

gregg-miskelly:
This checkin makes four changes:
1. Define telemetry API in DebugEngineHost. This will eventually allow us to
send telemetry for both VS and VS Code.
2. Now only the 'lab' configurations send telemetry.
3. Moved the Visual Studio specific code for sending telemetry out of the
IOS/Android Launchers and into the Visual Studio version of DebugEngineHost.
4. CLRDBG is going to soon output a 'telemetry' event. This adds support for the
event to the MI Engine.

Additional changes that we also need, but are not part of this commit:
* Send telemetry if the underlying debugger fails to start
* Send telemetry if the underyling debugger crashes
* Performance related telemetry?

### DIFF
--- a/src/DebugEngineHost/HostTelemetry.cs
+++ b/src/DebugEngineHost/HostTelemetry.cs
@@ -1,0 +1,71 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.Internal.VisualStudio.Shell;
+using Microsoft.Internal.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.DebugEngineHost
+{
+    /// <summary>
+    /// Static class providing telemetry reporting services to debug engines. Telemetry 
+    /// reports go to Microsoft, and so in general this functionality should not be used 
+    /// by non-Microsoft implemented debug engines.
+    /// </summary>
+    public static class HostTelemetry
+    {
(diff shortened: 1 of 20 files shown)


src/MICore/Debugger.cs
```diff
@@ -38,6 +38,7 @@ public class Debugger : ITransportCallback
         public event EventHandler ThreadCreatedEvent;
         public event EventHandler ThreadExitedEvent;
         public event EventHandler<ResultEventArgs> MessageEvent;
+        public event EventHandler<ResultEventArgs> TelemeetryEvent;
```
chuckries: [nit] typo "Telemeetry"

chuckries:
:+1:

src/DebugEngineHost/HostTelemetry.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace Microsoft.DebugEngineHost
+{
+    /// <summary>
+    /// Static class providing telemetry reporting services to debug engines. Telemetry 
+    /// reports go to Microsoft, and so in general this functionality should not be used 
+    /// by non-Microsoft implemented debug engines.
+    /// </summary>
+    public static class HostTelemetry
+    {
+        private static bool s_isDisabled;
```
rajkumar42: Build warning s_isDisabled is not used when built in non-LAB build.

gregg-miskelly: Fixed

src/MICore/Debugger.cs
gregg-miskelly: Fixed.

rajkumar42:
:+1:

PR #152, opened 2016-01-08
Support server proxy startup in local debugging mode

paulmaybee:
Change Description:

     1. Add new ServerTransport to manage the server proxy process
     2. Add new ClientServerTransport to wrap the connections to both client and server in one transport object
     3. Plumb new local debugging launch options into new transports


gregg-miskelly:
:+1:

PR #156, opened 2016-01-14
Dev/rajkumar42/thirdpartynotices

rajkumar42:
Adding thirdpartynotices.txt and copying it to output folder.

### DIFF
--- a/src/prebuild/prebuild.csproj
+++ b/src/prebuild/prebuild.csproj
@@ -18,6 +18,11 @@
       <Visible>false</Visible>
     </GenerateAssembly>
   </ItemGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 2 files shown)


src/prebuild/prebuild.csproj
```diff
@@ -18,6 +18,11 @@
       <Visible>false</Visible>
     </GenerateAssembly>
   </ItemGroup>
+
+  <ItemGroup>
```
gregg-miskelly: It seems like we shouldn't really need to copy this to the output folder - the ThirdPartNotices.txt for the MIEngine isn't actually what you want for the MSI as a whole.

gregg-miskelly:
Can you squash your commits together?

PR #157, opened 2016-01-14
Adding ThirdPartyNotices and copying it to output.

rajkumar42:
Adding ThirdPartyNotices and copying it to output.


rajkumar42:
Clarified. Don't need to copy to output. Will be sending a different pull request.

PR #159, opened 2016-01-20
Mac's ps is different from ubuntu's: doesn't support "h" (no headers)…

paulmaybee:
… option, different output formatting specification.


paulmaybee:
reviewed in codeflow

gregg-miskelly:
:+1:

PR #160, opened 2016-01-21
1) Fix issue that breakpoints were only matched against bkptno result…

jacdavis:
…ing in bugs when multple bps bind to the same place such as the entrypoint bp. 2) Use the trap command to setup a signal that cleans up the fifos in the local linux case if the user closes the debuggee terminal

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/BreakpointManager.cs
+++ b/src/MIDebugEngine/Engine.Impl/BreakpointManager.cs
@@ -4,6 +4,7 @@
 using System;
 using Microsoft.VisualStudio.Debugger.Interop;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using System.Diagnostics;
 using MICore;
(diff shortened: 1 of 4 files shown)


src/MIDebugEngine/Engine.Impl/BreakpointManager.cs
```diff
@@ -141,29 +141,66 @@ private AD7PendingBreakpoint BindToAddress(string bkptno, ulong addr, /*OPTIONAL
...
-        public AD7BoundBreakpoint FindHitBreakpoint(string bkptno, ulong addr, /*OPTIONAL*/ TupleValue frame, out bool fContinue)
+        // Return all bound breakpoints bound to an address regardless of bptno
+        AD7BoundBreakpoint[] FindBoundBreakpointsAtAddress(ulong addr)
+        {
+            lock (_pendingBreakpoints)
+            {
+                List<AD7BoundBreakpoint> matchingBoundBreakpoints = new List<AD7BoundBreakpoint>();
+                foreach (AD7PendingBreakpoint currPending in _pendingBreakpoints)
+                {
+                    matchingBoundBreakpoints.AddRange(Array.FindAll(currPending.EnumBoundBreakpoints(), (b) => b.Addr == addr));
```
gregg-miskelly: Do we need to exclude Disabled/Deleted/PendingDeleted breakpoints?

src/MIDebugEngine/Engine.Impl/BreakpointManager.cs
```diff
@@ -141,29 +141,66 @@ private AD7PendingBreakpoint BindToAddress(string bkptno, ulong addr, /*OPTIONAL
...
+                // However, clrdbg does not support addresses on its breakpoints, so the bkptno code path is
+                // preserved.
+                if (addr != 0)
                 {
-                    return null;
+                    hitBps = FindBoundBreakpointsAtAddress(addr);
                 }
-                fContinue = true;
-                if (!pending.Enabled || pending.Deleted || pending.PendingDelete)
+                else
```
gregg-miskelly: Do we need to lookup by breakpoint number in the case that we previously didn't know the address of the breakpoint? I think that can happen if the breakpoint binds to multiple addresses.

gregg-miskelly:
Signing off with comments.

paulmaybee:
:+1:

gregg-miskelly:
:+1:

PR #161, opened 2016-01-25
Launch fixes

gregg-miskelly:
commit c7354755afa79e539055fa9064ec420e51b370ac
Date:   Mon Jan 25 13:22:45 2016 -0800

    VS Code: Fix launch.json attributes for CLRDBG
    
    - Rename 'runtimeExecutable'->'program'
    - Rename 'runtimeArgs'->'args'
    - Add definition for 'stopAtEntry'. This doesn't work yet, but it will soon.
    - Add 'symbolPath'
    - Define attributes for attach
    - Update the template

commit cebd6b8ea730b39d0951d6c95a4181989c703d8b
Date:   Mon Jan 25 13:18:57 2016 -0800

    Fix UseUnixSymbolPaths for non-local launch options
    
    Back before update 1, UseUnixSymbolPaths would return true for any non-local target. This was wrong for LocalLaunchOptions once we added support for running on non-Windows, so we fixed that. But we broke the scenario for non-LocalLaunchOptions. So this restores that code path. This still isn't completely correct since technically you could use Pipe/Tcp launch options to connect to Windows. But we don't have any scenarios yet where we want to do that.

### DIFF
--- a/tools/InstallToVSCode/coreclr/package.json
+++ b/tools/InstallToVSCode/coreclr/package.json
@@ -19,47 +19,104 @@
 
         "configurationAttributes": {
           "launch": {
-            "required": [ "runtimeExecutable", "cwd" ],
+            "required": [ "program", "cwd" ],
             "properties": {
+              "program": {
+                "type": "string",
+                "description": "Workspace relative path to the program (executable file) to launch. On Windows, a '.exe' suffix is appended if not specified already.",
+                "default": "bin/Debug/dnxcore50/<My-Project-Name>"
+              },
               "cwd": {
                 "type": "string",
-                "description": "Specifies the workspace relative or absolute path to the webserver root.",
+                "description": "Workspace relative or absolute path to the working directory of the program being debugged. Default is the current workspace.",
                 "default": "."
               },
-              "runtimeExecutable": {
-                  "type": "string",
-                  "description": "Workspace relative or absolute path to the runtime executable to be used.",
-                  "default": null
+              "args": {
+                "type": "array",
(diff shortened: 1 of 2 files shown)


chuckries:
:+1:

gregg-miskelly:
@caslan please have a look.

tools/InstallToVSCode/coreclr/package.json
```diff
@@ -19,47 +19,104 @@
...
+                "default": { }
+              },
+              "justMyCode": {
+                "type": "boolean",
+                "description": "Optional flag to only show user code.",
+                "default": true
+              },
+              "symbolPath": {
+                "type": "array",
+                "description": "Array of directories to use to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originially dropped to. Example: '[ \"~/symbols\" ]",
```
gregg-miskelly: Spelling error

tools/InstallToVSCode/coreclr/package.json
```diff
@@ -19,47 +19,104 @@
...
+              // --- End common attributes. Copy these down to 'attach' if modified
+            }
+          },
+          "attach": {
+            "required": [ ],
+            "properties": {
+              "processName": {
+                "type": "string",
+                "description": "",
+                "default": "For a 'request'='attach', this species the process name to attach to. If this is used, 'processId' should not be used."
```
gregg-miskelly: Remove the 
   For a 'request'='attach'
prefix

PR #163, opened 2016-02-02
Prevent insertion into debugged pid map for pid 0

chuckries:
This prevents an exception occuring during iOS debugging. We see two
thread-group create events, one with pid 0 and one with the actual pid.
Ignore the pid 0 for entry into the debuggee pid map.


chuckries:
@jacdavis please take a look at this.

jacdavis:
Apart from my comment above, LGTM

PR #168, opened 2016-02-04
Move PackageNuGet target to its own file, add package to drop folder

chuckries:
With the last commit, the nuget package was being created but not ending
up in the drop folder so it was not getting placed on the drop server.
This commit moves the target into it's own file and includes it in
MIDebugPackage.csproj so that we can make use of the DropFiles properites.


gregg-miskelly:
Looks good to me, except I would avoid dropping the extra files that don't need to be dropped.

PR #169, opened 2016-02-04
Add telemetry for debugger abort

gregg-miskelly:
This checkin updates MIEngine to fire telemetry when the underlying debugger exits unexpectedly.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -90,6 +90,10 @@ namespace MICore
         private CommandLock _commandLock = new CommandLock();
 
         private string _lastResult;
+        /// <summary>
+        /// The last command we sent over the transport. This includes both the command name and arguments.
+        /// </summary>
+        private string _lastCommandText;
         private uint _lastCommandId;
         private bool _isClosed;
 
@@ -629,6 +633,7 @@ namespace MICore
 
                 id = ++_lastCommandId;
                 _waitingOperations.Add(id, waitingOperation);
+                _lastCommandText = command;
             }
 
             SendToTransport(id.ToString(CultureInfo.InvariantCulture) + command);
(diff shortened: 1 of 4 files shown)


src/MICore/Debugger.cs
```diff
@@ -90,6 +90,10 @@ public ResultEventArgs(Results results)
         private CommandLock _commandLock = new CommandLock();
 
         private string _lastResult;
+        /// <summary>
+        /// The last command we sent over the transport. This includes both the string and arguments.
```
gregg-miskelly: 'command name' instead of 'string'

src/MIDebugEngine/Engine.Impl/EngineTelemetry.cs
```diff
@@ -0,0 +1,96 @@
...
+    internal class EngineTelemetry
+    {
+        private const string Event_DebuggerAborted = @"VS/Diagnostics/Debugger/MIEngine/DebuggerAborted";
+        private const string Property_DebuggerName = @"VS.Diagnostics.Debugger.MIEngine.DebuggerName";
+        private const string Property_LastSentCommandName = @"VS.Diagnostics.Debugger.MIEngine.LastSentCommandName";
+        private const string Property_DebuggerExitCode = @"VS.Diagnostics.Debugger.MIEngine.DebuggerExitCode";
+        
+        KeyValuePair<string, object>[] _clrdbgProcessCreateProperties;
+
+        public bool DecodeTelemetryEvent(Results results, out string eventName, out KeyValuePair<string, object>[] properties)
```
gregg-miskelly: NOTE: This function was just moved, except for the 'if' on line 70.

chuckries:
Is there any info we would want for this event besides simply the last sent command? Does this have VS Sequence type bad correlation where the most recent command may not be what caused the underlying debugger to die?

gregg-miskelly:
Last command is certainly not perfect information -- as you say, the crash may be unrelated to the last command, and certainly it isn't enough information to really investigate. But I can't think of anything better we can do than last command and exit code. This will at least be enough to give us some sort of idea as to what type of crash, and give us some way of bucketing the crash a bit.

chuckries:
sounds good :+1:

caslan:
LGTM

PR #170, opened 2016-02-05
Add prerelease suffix for MIEngine nuget package

chuckries:
Without a suffix, multiple daily builds will produce the same version
nuget package which will break automatic deployment. This can be very
annoying in development scenarios. However, adding a '-pb-0', etc. to a
nuget package version makes it 'prerelease'. This changes adds an optional
prerelease suffix that we can turn on in daily builds and turn off in the
release build. It is on by default. To turn off the prerelease suffix, add
NuGetPrerelease=false to the msbuil paremeters.


gregg-miskelly:
:+1:

PR #171, opened 2016-02-05
Changing entry points for System.Native calls.

wiktork:
- Newer version of the CLR uses different names.

### DIFF
--- a/src/MICore/NativeMethods.cs
+++ b/src/MICore/NativeMethods.cs
@@ -8,15 +8,15 @@ namespace MICore
 {
     internal class NativeMethods
     {
-        // TODO: It would be better to route these to the correct .so files directly rather than pasing through system.native. 
+        private const string Libc = "libc";
 
-        [DllImport("System.Native", SetLastError = true)]
+        [DllImport(Libc, EntryPoint = "kill", SetLastError = true)]
         internal static extern int Kill(int pid, int mode);
 
-        [DllImport("System.Native", SetLastError = true)]
+        [DllImport(Libc, EntryPoint = "mkfifo", SetLastError = true)]
         internal static extern int MkFifo(string name, int mode);
 
-        [DllImport("System.Native", SetLastError = true)]
+        [DllImport(Libc, EntryPoint = "geteuid", SetLastError = true)]
         internal static extern uint GetEUid();
     }
 }
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -485,7 +485,8 @@ namespace MICore
 
         public Task CmdBreakInternal()
         {
-            if (IsLocalGdbAttach())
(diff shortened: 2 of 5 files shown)


jacdavis:
We should switch to direct pinvokes on the real target versions I possible.

Sent from my Windows Phone

wiktork:
Updated to pinvoke directly

chuckries:
:+1:

src/MICore/NativeMethods.cs
```diff
@@ -8,15 +8,15 @@ namespace MICore
 {
     internal class NativeMethods
     {
-        // TODO: It would be better to route these to the correct .so files directly rather than pasing through system.native. 
+        private const string Libc = "libc";
```
gregg-miskelly: Maybe I am missing something, but I don't see an explicit check that we are on Linux. It seems like we may need some additional calls to RuntimeInformation.IsOSPlatform(OSPlatform.Linux).
Also, it seems like we might want to rename this class to 'UnixNativeMethods' or maybe 'LinuxNativeMethods' if this code doesn't work on OSX either.

gregg-miskelly:
See my other comment.

src/MICore/NativeMethods.cs
wiktork: These functions are called by LocalLinuxTransport, which is only used after checking the platform.  I will rename the class.

jacdavis: Would it be best to make the pinvokes internal and write wrapper methods that verified the platform?

wiktork:
I think there's a bigger work item here to make sure this works on all 3 platforms. I will make further changes to the pinvoke once we start that work.

gregg-miskelly:
Maybe I am being dumb, but I think in Debugger.cs(488) we are missing a check that we are actually on Linux before we decide to call CmdLinuxBreak instead of the regular async-break processing. Aside from that, looks good.

wiktork:
You are correct, we just check IsLocalGdbAttach. I'm going to test this on OSX and see what happens; assuming attach works with normal exec-interrupt, I can do the OS check. Otherwise, I'll need to make sure the pinvoke works on a mac. Note this must have been an issue before, unless System.Native offered some abstraction around the syscalls.

jacdavis:
IsLocalGdbAttach is specific to the Linux local transport isn't it? Regardless of any of this, the wrappers that verify the platform seem like a good idea to me designwise

Sent from my Windows Phone

wiktork:
I think IsLocalGdbAttach is agnostic to the transport itself; it's on MICore.Debugger. I think the wrappers are a good idea, but I'd like to unblock Linux debugging support with this change first, and then work on any gaps in OSX support for local gdb.

gregg-miskelly:
I _think_ before we would have been okay on OSX, but blow up on Windows.

PR #175, opened 2016-02-08
Support gdb pretty-printing displayHint type "map" by processing map …

paulmaybee:
…children as name-value pairs.


src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -561,6 +573,21 @@ private Task FetchChildren()
                         i++;
                     }
                 }
+                else if (IsMapType())
+                {
+                    //
+                    // support for gdb's pretty-printing built-in displayHint "map"
+                    //
+                    List<VariableInformation> listChildren = new List<VariableInformation>();
+                    for(int p = 0; p+1 < children.Length; p+=2)
```
gregg-miskelly: Might want to add a comment explaining this format a bit, and why we are skipping two at a time.

gregg-miskelly:
Signing off with one minor comment.

src/MIDebugEngine/Engine.Impl/Variables.cs
jacdavis: Yeah, this is quite confusing for someone with no experience with these mi records.

paulmaybee:
Comments added

PR #177, opened 2016-02-11
Merge master into release to prep for VS release

chuckries:
@gregg-miskelly I made sure that this includes quotes for -var-list-children.


gregg-miskelly:
:+1:

chuckries:
I ran all the android tests and things look as good as those tests can tell us. (They all passed but they don't cover much).

PR #178, opened 2016-02-11
Set hint path Microsoft.VisualStudio.Debugger.Engine.dll

gregg-miskelly:
The official build is failing because it is using the VS 12 version of Microsoft.VisualStudio.Debugger.Engine.dll. This makes changes to DebugEngineHost so that it should force the 14 version to be used.


wiktork:
Looks good to me.

PR #181, opened 2016-02-15
InstallToVSCode and package.json fixes

gregg-miskelly:
commit abeebec39221c654bd69a0d2bcadca6a4a0d0392
Date:   Mon Feb 15 12:18:05 2016 -0800

    InstallToVSCode fixes
    
    Fix package.json so that:
    * It doesn't use CoreRun
    * It has a better label name in the template
    * It doesn't assume that 'program' and 'cwd' are relative paths
    * It uses 'build' instead of 'compile' for dotnet
    
    Fix InstallToVSCode to pull CoreCLR dependencies from nuget, and to make the path to clrdbg a required argument.

commit 7e1d1ee10859487186bd81f1aa3601c3005aa2ae
Date:   Mon Feb 15 13:52:57 2016 -0800

    Add work around for CoreCLR 3164
    
    The CoreCLR has a bug where abort is called if multiple versions of libcoreclrtraceptprovider load. This works around it by removing our copy.

### DIFF
--- a/tools/InstallToVSCode/CLRDependencies/NuGet.Config
+++ b/tools/InstallToVSCode/CLRDependencies/NuGet.Config
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <!--To inherit the global NuGet package sources remove the <clear/> line below -->
+    <clear />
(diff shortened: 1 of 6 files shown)


caslan:
LGTM

PR #186, opened 2016-02-18
Fix strings in registry lookup for NDK/SDK

chuckries:
Noticed this while I was looking at our current registry fallback behavior
if the NDK/SDK are not installed.


chuckries:
I think this means any telemetry we've gotten for missing NDK/SDK is actually reversed.

gregg-miskelly:
:+1:

PR #187, opened 2016-02-20
InstallToVSCode: Update to CoreCLR/CoreFx 23819

gregg-miskelly:
This fixes the InstallToVSCode script to use CoreCLR/FX 23819


chuckries:
:+1:

PR #196, opened 2016-02-22
[Bug fix] Closing debuggee window leaves VSCode in a debug state

xingshenglu:
1. Add HandleThreadGroupExited() to handle GDB/MI "thread-group-exited" async record
=thread-exited,id="id",group-id="gid"
2. This handler removes exited thread group ids from _debuggeePids
3. This handler calls OnDebuggerProcessExit() to leave debug mode when debuggeePids reaches zero count.
4. Clean up logic error from OnStopped() that removes thread-group-id. The reason is that the "thread-id " property in this async record is a thread-id, not an thread-group-id.
*stopped,reason="reason",thread-id="id",stopped-threads="stopped",core="core"

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -174,15 +174,6 @@ namespace MICore
 
             if (reason.StartsWith("exited"))
             {
-                string threadGroupId = results.TryFindString("id");
-                if (!String.IsNullOrEmpty(threadGroupId))
-                {
-                    lock (_debuggeePids)
-                    {
-                        _debuggeePids.Remove(threadGroupId);
-                    }
-                }
-
                 if (IsLocalGdbAttach())
                 {
                     CmdExitAsync();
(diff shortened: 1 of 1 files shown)


src/MICore/Debugger.cs
```diff
@@ -1118,11 +1114,34 @@ private void HandleThreadGroupStarted(Results results)
...
+                    if (_debuggeePids.Remove(threadGroupId))
+                    {
+                        shouldExit = _debuggeePids.Count == 0;
+                    }
+                }
+            }
+
+            if (shouldExit)
+            {
+                OnDebuggerProcessExit(null);
```
gregg-miskelly: So this is the code path that acts like gdb just disappeared. I think this probably isn't want you want. It seems like we should handle it like:

https://github.com/Microsoft/MIEngine/blob/master/src/MICore/Debugger.cs#L194

If that doesn't work for some reason, we likely need to plumb something else along to OnDebuggerProcessExit to make sure we don't report telemetry that GDB exited.

gregg-miskelly:
Looks good. Please squash your changes.

jacdavis:
Merged

PR #199, opened 2016-02-24
Add support for debugging core dumps.

jander-msft:
1. Changed local launch options schema to support specifying CoreDumpPath field and added additional LaunchCompleteOption enum value "core-dump".
2. Update DebuggedProcess initialization commands to debug core dumps.
3. Update DebuggedProcess transport initialization to use normal local transport since core dump debugging does not require a new terminal upon launch.
4. Update DebuggedProcess.ResumeFromLaunch to set initial state of debug engine to stopped with generated MI result record.
5. Block operations that are not allow during core dump debugging (e.g. continue, step, breakpoint binding, etc) by throwing exception. Catch exception at COM boundary and convert to appropriate result code.

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -55,7 +55,7 @@ namespace MICore
         /// <summary>
         /// No command should be executed. This is useful if the target is already ready to go.
         /// </summary>
-        None
+        None,
     };
 
     /// <summary>
@@ -148,6 +148,8 @@ namespace MICore
     /// </summary>
     public sealed class LocalLaunchOptions : LaunchOptions
     {
+        private string _coreDumpPath;
+
         private const int DefaultLaunchTimeout = 10 * 1000; // 10 seconds
 
         public LocalLaunchOptions(string MIDebuggerPath, string MIDebuggerServerAddress, int processId, Xml.LaunchOptions.EnvironmentEntry[] environmentEntries)
@@ -184,8 +186,21 @@ namespace MICore
                 {
                     FilterStdout = true;    // no pattern source specified, use stdout
                 }
-                ServerLaunchTimeout = source.ServerLaunchTimeoutSpecified ? source.ServerLaunchTimeout : DefaultLaunchTimeout; 
+                ServerLaunchTimeout = source.ServerLaunchTimeoutSpecified ? source.ServerLaunchTimeout : DefaultLaunchTimeout;
             }
+        }
+
+        /// <summary>
+        /// Checks that the path is valid, exists, and is rooted.
+        /// </summary>
+        private static bool CheckPath(string path)
+        {
+            return path.IndexOfAny(Path.GetInvalidPathChars()) < 0 && File.Exists(path) && Path.IsPathRooted(path);
+        }
+
+        public bool IsCoreDump
+        {
+            get { return !String.IsNullOrEmpty(this.CoreDumpPath); }
         }
 
         public bool ShouldStartServer()
@@ -202,6 +217,7 @@ namespace MICore
                 source.Environment);
             options.InitializeCommonOptions(source);
             options.InitializeServerOptions(source);
+            options.CoreDumpPath = source.CoreDumpPath;
 
             return options;
         }
@@ -232,7 +248,7 @@ namespace MICore
             }
             set
             {
-                if (string.IsNullOrWhiteSpace(value) || value.IndexOfAny(Path.GetInvalidPathChars()) >= 0 || !File.Exists(value) || !Path.IsPathRooted(value))
+                if (String.IsNullOrEmpty(value) || !LocalLaunchOptions.CheckPath(value))
                     throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, MICoreResources.Error_InvalidLocalExePath, value));
 
                 base.ExePath = value;
@@ -273,6 +289,25 @@ namespace MICore
         /// [Optional] Log strings written to stderr and examine for server started pattern
         /// </summary>
         public int ServerLaunchTimeout { get; private set; }
+
+        /// <summary>
+        /// [Optional] Path to a core dump file for the specified executable.
+        /// </summary>
+        public string CoreDumpPath
+        {
+            get
+            {
+                return _coreDumpPath;
+            }
+            private set
+            {
+                // CoreDumpPath is allowed to be null/empty
+                if (!String.IsNullOrEmpty(value) && !LocalLaunchOptions.CheckPath(value))
+                    throw new ArgumentException(String.Format(CultureInfo.CurrentCulture, MICoreResources.Error_InvalidLocalExePath, value));
+
+                _coreDumpPath = value;
+            }
+        }
     }
 
     public sealed class SourceRoot
@@ -803,7 +838,7 @@ namespace MICore
                     deviceAppLauncher.Initialize(configStore, eventCallback);
                     deviceAppLauncher.SetLaunchOptions(exePath, args, dir, launcherXmlOptions, targetEngine);
                 }
-                catch (Exception e) when (!(e is InvalidLaunchOptionsException) && ExceptionHelper.BeforeCatch(e, reportOnlyCorrupting:true))
+                catch (Exception e) when (!(e is InvalidLaunchOptionsException) && ExceptionHelper.BeforeCatch(e, reportOnlyCorrupting: true))
                 {
                     throw new InvalidLaunchOptionsException(e.Message);
                 }
--- a/src/MICore/LaunchOptions.xsd
+++ b/src/MICore/LaunchOptions.xsd
@@ -256,6 +256,11 @@
               <xs:documentation>How long, in milliseconds, to wait for the DebugServer executable to start and ServerStarted pattern to be found.</xs:documentation>
             </xs:annotation>
           </xs:attribute>
+          <xs:attribute name="CoreDumpPath" type="xs:string" use="optional">
+            <xs:annotation>
+              <xs:documentation>Path to a core dump file for the specified executable.</xs:documentation>
+            </xs:annotation>
+          </xs:attribute>
         </xs:extension>
       </xs:complexContent>
     </xs:complexType>
--- a/src/MICore/MICoreResources.resx
+++ b/src/MICore/MICoreResources.resx
@@ -204,4 +204,7 @@ Error: {1}</value>
   <data name="Status_BreakpointPending" xml:space="preserve">
     <value>Module containing this breakpoint has not yet loaded or the breakpoint address could not be obtained.</value>
   </data>
+  <data name="Error_InvalidLocalCoreDumpPath" xml:space="preserve">
+    <value>Invalid path to core dump file '{0}'. File must be a valid file name that exists on the computer.</value>
+  </data>
 </root>
--- a/src/MICore/MIResults.cs
+++ b/src/MICore/MIResults.cs
@@ -403,6 +403,16 @@ namespace MICore
         {
             return FindAll(name).OfType<T>().ToArray();
         }
+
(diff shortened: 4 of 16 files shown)


src/MICore/LaunchOptions.cs
```diff
@@ -53,9 +53,14 @@ public enum LaunchCompleteCommand
         ExecContinue,
 
         /// <summary>
+        /// Starts the debugger in the stopped state for debugging a core dump.
+        /// </summary>
+        CoreDump,
```
jacdavis: I'm not sure this is the right place for this. The launch complete command is designed for users to manually insert gdb commands into the launch sequence. I would expect the core dump support to be a first class citizen and detected by the presense of a coredump element or attribute in the local launch options.

src/MICore/LaunchOptions.cs
```diff
@@ -203,6 +227,11 @@ static internal LocalLaunchOptions CreateFromXml(Xml.LaunchOptions.LocalLaunchOp
             options.InitializeCommonOptions(source);
             options.InitializeServerOptions(source);
 
+            if (LaunchCompleteCommand.CoreDump == options.LaunchCompleteCommand)
```
jacdavis: same as above. LaunchCompleteCommand doesn't seem like the right place for this state

src/MICore/LaunchOptions.xsd
```diff
@@ -340,6 +345,11 @@
                 the result of setting up the debugger is that the debuggee is in break state.</xs:documentation>
               </xs:annotation>
             </xs:enumeration>
+            <xs:enumeration value="core-dump">
```
jacdavis: Remove (once issue above is resolved)

src/MICore/MICommandFactory.cs
```diff
@@ -501,6 +501,34 @@ public virtual void DecodeExceptionReceivedProperties(Results miExceptionResult,
...
+        /// </summary>
+        public async Task<Results> GenerateStoppedRecordResults()
+        {
+            Results threadInfo = await this.ThreadInfo();
+            // Get the current thread identifier
+            string currentThreadId = threadInfo.FindString("current-thread-id");
+            Debug.Assert(!String.IsNullOrEmpty(currentThreadId), "Could not determine current thread.");
+            
+            ValueListValue threads = threadInfo.Find<ValueListValue>("threads");
+            Debug.Assert(null != threads, "Could not get thread information from debuggee.");
```
jacdavis: You will get a null ref on the next line after the assert. Is that what you want here?

src/MICore/MICommandFactory.cs
```diff
@@ -501,6 +501,34 @@ public virtual void DecodeExceptionReceivedProperties(Results miExceptionResult,
...
+                
+            TupleValue currentThread = threads.AsArray<TupleValue>().FirstOrDefault(tv => currentThreadId.Equals(tv.FindString("id"), StringComparison.Ordinal));
+            Debug.Assert(null != currentThread, String.Concat("Unable to find thread with ID ", currentThreadId, "."));
```
jacdavis: same as above. Assert but then null ref if not found.

src/MICore/MICommandFactory.cs
```diff
@@ -501,6 +501,34 @@ public virtual void DecodeExceptionReceivedProperties(Results miExceptionResult,
 
         #region Helpers
 
+        /// <summary>
+        /// Generates results that represent an emulated MI stopped record.
+        /// </summary>
+        public async Task<Results> GenerateStoppedRecordResults()
```
jacdavis: Is the command factory the right place for this code to live? Isn't command factory normally used to send commands to the debuggee?

src/MICore/MICoreResources.resx
```diff
@@ -204,4 +204,7 @@ Error: {1}</value>
   <data name="Status_BreakpointPending" xml:space="preserve">
     <value>Module containing this breakpoint has not yet loaded or the breakpoint address could not be obtained.</value>
   </data>
+  <data name="Error_InvalidLocalCoreDumpPath" xml:space="preserve">
+    <value>Invalid path to core dump file path '{0}'. File must be a valid file name that exists on the Visual Studio computer.</value>
```
jacdavis: Visual Studio isn't the only product hosting this. I think I would remove the reference to visual studio here.

src/MICore/MIResults.cs
```diff
@@ -403,6 +409,28 @@ public ResultValue[] FindAll(string name)
...
+            outTuple.Append("{");
+            outTuple.Append(String.Join(",", this.Content.Select(selector)));
+            outTuple.Append("}");
+            return outTuple.ToString();
+        }
+
+        public TupleValue Subset(params string[] names)
+        {
+            List<NamedResultValue> values = new List<NamedResultValue>();
+            foreach (string name in names)
```
jacdavis: style nitpick. Please use braces on single line statements.

src/MICore/MIResults.cs
```diff
@@ -31,7 +31,7 @@ public enum ResultClass
         exit
     }
 
-    public class ResultValue
+    public abstract class ResultValue
```
jacdavis: What's the purpose of all the changes to MIResults.cs? I don't see any motivation behind these changes except style.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -572,13 +572,20 @@ public int Continue(IDebugThread2 pThread)
...
+            try
             {
-                _debuggedProcess.Continue(thread?.GetDebuggedThread());
+                if (_pollThread.IsPollThread())
+                {
+                    _debuggedProcess.Continue(thread?.GetDebuggedThread());
+                }
+                else
+                {
+                    _pollThread.RunOperation(() => _debuggedProcess.Continue(thread?.GetDebuggedThread()));
```
jacdavis: Does pullthread.runoperation carry the exception back here if the exception is thrown from the other thread?

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
```diff
@@ -86,11 +86,15 @@ private static MIBreakpointState StringToBreakpointState(string addr)
 
         internal static async Task<BindResult> Bind(string functionName, DebuggedProcess process, string condition, AD7PendingBreakpoint pbreak)
         {
+            process.VerifyNotDebuggingCoreDump();
```
jacdavis: what happens if this throws from the user's perspective in vscode? In vs, we'd display an error breakpoint in the gutter that says that bps aren't supported when dump debugging.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -572,13 +572,20 @@ public int Continue(IDebugThread2 pThread)
...
+                }
             }
-            else
+            catch (InvalidCoreDumpOperationException)
             {
-                _pollThread.RunOperation(() => _debuggedProcess.Continue(thread?.GetDebuggedThread()));
+                return AD7_HRESULT.E_CRASHDUMP_UNSUPPORTED;
```
wiktork: Looking at the vscode side, it doesn't look like we always check the return codes to all the ad7 calls, such as Continue and Bind.

src/MICore/LaunchOptions.cs
```diff
@@ -273,6 +302,25 @@ public override string ExePath
...
+        public string CoreDumpPath
+        {
+            get
+            {
+                return this.coreDumpPath;
+            }
+            private set
+            {
+                // CoreDumpPath is allowed to be null/empty
+                if (!String.IsNullOrEmpty(value) && !LocalLaunchOptions.CheckPath(value))
```
paulmaybee: The core file needs to be on the machine where gdb is executing, which is not necessarily the local machine.

src/MICore/MICoreResources.resx
jander-msft: I copied it from the Error_InvalidLocalExePath resource string and modified it. Other strings in this resx file have references to Visual Studio. I'll log a bug to remove those references.

src/MICore/MIResults.cs
jander-msft: I originally wrote it so that I could compose the MI record again. The ToString implementations did not adhere to the MI result format, so I created a separate property to return what I needed at the time. Eventually, the consume was rewritten so that I didn't need the properly formatted record anymore and forgot to remove the implementation. I'll remove it.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
jander-msft: Yes, it does.

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
jander-msft: Everywhere that I use the VerifyNotDebuggingCoreDump method, I've made sure all upstream callers catch the InvalidCoreDumpOperationException and return AD7_HRESULT.E_CRASHDUMP_UNSUPPORTED. Code on the VSCode side should be checking the hresult when calling these types of operations. In our case, we'll show a user-friendly error message stating something like "Unable to step next. This operation is not supported when debugging dump files". In the case of breakpoints, I was planning of having them show as set but unbound, so that the user doesn't have an expectation that they will be hit for some reason.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
jander-msft: I have separate changes that are not out for review yet that do exactly this.

src/MICore/LaunchOptions.cs
jander-msft: I'll remove the CoreDump enum value and just key off of the CoreDumpPath on the LocalLaunchOptions class.

src/MICore/MICommandFactory.cs
jander-msft: I dug a little more to see what would actually happen if it couldn't fine something via the Find* methods. Those methods will throw MIResultFormatException. AD7Engine will handle any exception thrown during the call to ResumeFromLaunch, report the error, and stops debugging. The error message would read as "Unable to start debugging. Unrecognized format of field 'current-thread-id' in result: <pseudo-MI-record-string-for-thread-info-call>".

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
jacdavis: Since vscode uses the OpenDebug prototocol and not com, I suspect we'll need to make sure opendebugad7 catches this and displays an error. If you've already tested that an error is displayed then it is doing the right thing. The behavior you describe for breakpoints sounds reasonable.

jander-msft: I meant that all upstream callers within the MIEngine assemblies catch the exception and then return the result code. Thus the exception should never reach OpenDebugAD7. Sorry for the ambiguity.

src/MICore/LaunchOptions.cs
```diff
@@ -148,6 +148,8 @@ public EnvironmentEntry(Xml.LaunchOptions.EnvironmentEntry xmlEntry)
     /// </summary>
     public sealed class LocalLaunchOptions : LaunchOptions
     {
+        private string coreDumpPath;
```
gregg-miskelly: Can you run https://github.com/dotnet/codeformatter on this?

src/MICore/LaunchOptions.cs
gregg-miskelly: Support is only present is LocalLaunchOptions with this change. So, for now at least, they are the same.

gregg-miskelly:
:+1:

jacdavis:
LGTM

PR #200, opened 2016-02-24
Cleaned up remote debugger output

devilman3d:
Two simple changes to address most, if not all the issues with missing newlines and whitespace in debug output from the remote debugger.

1. Return the original untrimmed string when MIResults.ParseCString does not parse a quoted string. This preserves newlines and whitespace from remote printf statements.
2. Append a newline as needed to unhandled notification messages ("=thread-group-added", "=library-unloaded", etc.)

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -1074,6 +1074,11 @@ namespace MICore
             }
             else
             {
+                // append a newline if the message didn't come with one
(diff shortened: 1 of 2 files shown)


gregg-miskelly:
:+1:

chuckries:
:+1:

orablu:
:+1:

PR #201, opened 2016-02-24
Support coreclr launch

orablu:
Clrdbg sends the entry breakpoint with a reason of "entry-point-hit" instead of "".


orablu:
:+1: From CodeFlow

PR #202, opened 2016-02-25
[Bug Fix] Make sure termination sequence for attach works as expected

xingshenglu:
1. In attaching mode, send "-gdb-exit" to detach from debuggee.
2. In launching mode, send "kill" to stop debuggee. 
3. In attaching mode, send "-execute-interrupt" only when debuggee is in run mode before detaching from debuggee.

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7Engine.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Engine.cs
@@ -590,10 +590,7 @@ namespace Microsoft.MIDebugEngine
         {
             _breakpointManager.ClearBoundBreakpoints();
 
-            _pollThread.RunOperation(new Operation(delegate
-            {
-                _debuggedProcess.Detach();
-            }));
+            _pollThread.RunOperation(() => _debuggedProcess.CmdDetach());
 
             return Constants.S_OK;
         }
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -466,10 +466,16 @@ namespace MICore
         {
             await MICommandFactory.Terminate();
 
-            if (IsLocalGdbAttach())
+            return new Results(ResultClass.done);
+        }
+
+        public async Task<Results> CmdDetach()
+        {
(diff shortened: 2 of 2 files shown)


gregg-miskelly:
Is the goal here to ensure that we detach from a process that we attached to instead of termining? Or is it that terminate doesn't work the way it was originally implemented for local GDB attach?

jacdavis:
The goal is to detach rather than terminate

gregg-miskelly:
I think this fix is in the wrong spot then -- we should have OpenDebugAD7 remember if it wants to detach or terminate and call the appropriate API on the MIEngine.

jacdavis:
Is that because we still need to terminate in the android case?

gregg-miskelly:
No, its just because we are abusing the AD7 API - we are making 'Terminate' do a detach when the API is spec'ed to be Terminate.

jacdavis:
Good point. I didn't think about the process window terminate case under attach.

xingshenglu:
Fix revised:
1. Add CmdDetach() alongside CmdTerminate() in Debugger.cs
2. Call CmdDetach() from Detach() in AD7Enging.
3. Move Launch/Attach logic to OpenDebugAD7, see https://mseng.visualstudio.com/DefaultCollection/vsplatdbgmdd/_git/OpenDebugAD7/pullrequest/85169

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -590,10 +590,8 @@ public int Detach()
         {
             _breakpointManager.ClearBoundBreakpoints();
 
-            _pollThread.RunOperation(new Operation(delegate
-            {
-                _debuggedProcess.Detach();
-            }));
+            _pollThread.RunOperation(() => _debuggedProcess.CmdDetach());
+            _debuggedProcess.Detach();
```
gregg-miskelly: Can you remove the Detach method from _debuggedProcess while you are in here? Otherwise looks good to me.

PR #203, opened 2016-02-25
Add missing StringComparison

josesimoes:
Couldn't build solution compiler was issuing the following error:

Error	CA1307	Because the behavior of 'string.EndsWith(string)' could vary based on the current user's locale settings, replace this call in 'Debugger.OnNotificationOutput(string)' with a call to 'string.EndsWith(string, StringComparison)'. If the result of 'string.EndsWith(string, StringComparison)' will be displayed to the user, such as when sorting a list of items for display in a list box, specify 'StringComparison.CurrentCulture' or 'StringComparison.CurrentCultureIgnoreCase' as the 'StringComparison' parameter. If comparing case-insensitive identifiers, such as file paths, environment variables, or registry keys and values, specify 'StringComparison.OrdinalIgnoreCase'. Otherwise, if comparing case-sensitive identifiers, specify 'StringComparison.Ordinal'.	MICore	E:\GitHub\MIEngine\src\MICore\Debugger.cs	1074	Active

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -1071,7 +1071,7 @@ namespace MICore
             else
             {
                 // append a newline if the message didn't come with one
-                if (!cmd.EndsWith("\n"))
+                if (!cmd.EndsWith("\n", StringComparison.Ordinal))
                 {
                     cmd += "\n";
                 }


gregg-miskelly:
:+1:

chuckries:
Does this mean our PR builds aren't catching these?

gregg-miskelly:
@chuckries You can submit a PR that should fail the check to test..

chuckries:
@gregg-miskelly it looks like the PR build will catch things like this. The PR build that introduced this issue never completed a PR build and was still merged. How this hasn't been causing every PR build to fail since then I'm not sure. mmitche just did some work on it today so maybe it's different than it used to be.

chuckries:
@gregg-miskelly I just realized that this is from months ago. Idk why I'm just seeing it now.

PR #205, opened 2016-02-29
Ensure that both core dump and attach are not specified at the same time.

jander-msft:
Prevent launch if both core dump path and process id are specified at the same time in order to disallow the ambiguity of attempting to debug a dump and attach to a process. If both are specified and the user attempts to launch, an error is shown stating "Unable to start debugging. Launch options string provided by the project system is invalid. Both 'CoreDumpPath' and 'ProcessId' cannot be specified at the same time.".


jacdavis:
LGTM

gregg-miskelly:
:+1:

PR #209, opened 2016-03-01
immediate return on breakpoint bind

rick111111:
We don't need to wait 250ms for breakpoint bind since we are not handling break point error events in OpenDebugAD7


gregg-miskelly:
Yes, but we should be right? And we want to wait in the case of Visual Studio.

PR #210, opened 2016-03-02
readd missing closing quote to linux local launch options. Note that …

jacdavis:
…this didn't have any impact on coreclr but broke on some other runtimes we're investigating


gregg-miskelly:
:+1:

PR #212, opened 2016-03-04
Recognize coreclr embedded breakpoints

orablu:
Embedded breakpoints in coreclr (e.g. `Debugger.Break()` commands) are sent with an `<EMBEDDED>` id, as they don't correlate to any bound breakpoints. Add a case to check for them and send an empty breakpoint list instead of defaulting to the `else` clause and immediately continuing from them.


chuckries:
:+1:

caslan:
LGTM

PR #215, opened 2016-03-04
Child debugging: phase 1, remove/protect writable static vars

paulmaybee:
The logger is the primary shared object that changes. Each process will now get a Logger object which supports the WriteLine functionality. Writeline prefaces each output line with the number of the process ("1:" below).
Example 
1: (550) <-1001-gdb-set target-async on


jacdavis:
LGTM

PR #216, opened 2016-03-05
Bug fixes for closing terminal window

xingshenglu:
Fixing bugs related to closing terminal window by improving the Process State transition.
    Bug 194992:[C++ on Linux] Console Window opened by VSCode isn't closed after stopping debugging
    Bug 196499:[C++ on Linux] Closing VSCode won't close debuggee window
    Bug 197013:[C++ on Linux] Attaching: Run to the end leaves the console window open

Descriptions:
1. OnStopped()
    a. Prevent reentry of handling "*stopped,reason="exited"" event
    b. Remove CmdExitAsync() since it will be called by ProcessExitEvent event handler, this is a duplicate.
2. CmdDetach()
    a. Use "-target-detach" to make sure gdb is in detach mode, this is good enough. "-gdb-exit" should be called by OnStopped(), see 1.
3. CmdBreakInternal()
    a. "-exec-interrupt" works with attach and launch mode both.
    b. Unlike CmdAsync("-exec-interrupt", ResultClass.done) which receives a confirmation from gdb, CmdLinuxBreak() doesn't verify that gdb is really in break mode. This sometimes fail detach.
    c. "-target-detach"requires that gdb is in break mode; if not, it will core dump the debuggee. 
 4. HandleThreadGroupExited()
    a. Post a "*stopped,reason=""exited""" event, let OnStopped() handle it. OnStopped() does call OnDebuggerProcessExit(), and it doe more cleanup work.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -178,15 +178,13 @@ namespace MICore
 
             if (reason.StartsWith("exited"))
             {
-                if (IsLocalGdbAttach())
+                if (this.ProcessState != ProcessState.Exited)
                 {
-                    CmdExitAsync();
-                }
-
-                this.ProcessState = ProcessState.Exited;
-                if (ProcessExitEvent != null)
-                {
-                    ProcessExitEvent(this, new ResultEventArgs(results));
+                    this.ProcessState = ProcessState.Exited;
+                    if (ProcessExitEvent != null)
+                    {
+                        ProcessExitEvent(this, new ResultEventArgs(results));
+                    }
                 }
                 return;
             }
@@ -498,7 +496,7 @@ namespace MICore
             {
                 await CmdBreak();
             }
-            CmdExitAsync();
+            await CmdAsync("-target-detach", ResultClass.done);
 
             return new Results(ResultClass.done);
         }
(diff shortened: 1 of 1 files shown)


src/MICore/Debugger.cs
```diff
@@ -494,37 +492,13 @@ protected bool IsCoreDump
...
-            CmdExitAsync();
+            await CmdAsync("-target-detach", ResultClass.done);
 
             return new Results(ResultClass.done);
         }
 
         public Task CmdBreakInternal()
         {
-            //TODO May need to fix attach on windows and osx.
-            if (IsLocalGdbAttach() && RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
gregg-miskelly: If you haven't already done so, please double check with @jacdavis what scenarios this was required for.

gregg-miskelly:
:+1:

jacdavis:
I've spent the entire day trying to repro the issues that were caused from exec-interrupt not working. So far, the only issues I've repro'd have been when I disable the launching of the debugger terminal. I was also a bit concerned about the change that waits the debugging to exit, but Freddie has addressed my earlier feedback.

src/MICore/Debugger.cs
```diff
@@ -1161,15 +1136,14 @@ private void HandleThreadGroupExited(Results results)
...
+                        isThreadGroupEmpty = _debuggeePids.Count == 0;
                     }
                 }
             }
 
-            if (ProcessState == ProcessState.Exited)
+            if (isThreadGroupEmpty)
             {
-                OnDebuggerProcessExit(null);
+                ScheduleStdOutProcessing(@"*stopped,reason=""exited""");
```
paulmaybee: Does the HandleThreadGroupExited path still lead to Close() eventually being invoked without the direct call to OnDebuggerProcessExit?

xingshenglu: Yes, the code path is "stopped, reason='exited'" >> Onstopped() >> ProcessExitEvent >> CmdExitAsync() >> "-gdb-exit" >> OnDebugProcessExit() >> Close()

xingshenglu: Tested scenarios:
1. Attach to a process, set a BP, run debuggee to the end, make sure the terminal window opened by VSCode is closed (bug 197013)
2. Launch VSCode, start debugging, then close VSCode, make sure the terminal window opened by VSCode closed (Bug 196499)
3. Run as root, launch VSCode, start debugging, then stop debugging, make sure the terminal window closed. (Bug 194992)

PR #217, opened 2016-03-08
[Bug Fix] Closing VSCode leaves a zombie process

xingshenglu:
The real cause is that StreamReader.ReadLine() hangs even if the file it is reading gets deleted. The fix is to use StreamReader.ReadLineAync(TimeOut) and check the status if thread is quitting.

Tested Scenarios:
For both Root and Normal User:
1. Launch VSCode
2. Start debugging
3. Close VSCode while it is still in debugging mode
4. Make sure  no zombie process left, especially "corerun OpenDebugAD7.dll" process.

### DIFF
--- a/src/MICore/Transports/StreamTransport.cs
+++ b/src/MICore/Transports/StreamTransport.cs
@@ -17,9 +17,12 @@ namespace MICore
         private ITransportCallback _callback;
         private Thread _thread;
         private bool _bQuit;
+        private CancellationTokenSource _streamReadCancellationTokenSource = new CancellationTokenSource();
         protected StreamReader _reader;
         protected StreamWriter _writer;
         private bool _filterStdout;
+        private Object _locker = new object();
+
         protected Logger Logger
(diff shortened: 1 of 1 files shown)


chuckries:
:+1:

src/MICore/Transports/StreamTransport.cs
```diff
@@ -114,7 +114,15 @@ private string GetLine()
         {
             try
             {
-                return _reader.ReadLine();
+                Task<string> task = _reader.ReadLineAsync();
+                while (!task.Wait(100))
```
wiktork: Can we use a cancellation token here instead and cancel during Close?

jacdavis:
Del had another suggestion which may or may not work. We should check to see if a call to dispose or close on the stream works event when blocked in this call. If it does, we can cancel this that way when _bQuit becomes true. Note that I'm not certain this will work so we'd need to verify via experimentation.

PR #218, opened 2016-03-09
Fix for bug 197616. Issue is this debuggee has excercised an edge cas…

jacdavis:
…e in pthreads that allows the app to continue running after the main thread exits. GDB has a bug that it is reporting a stopping event for a bp on a background thread which then fails to report thread info as it believes the debuggee should be ripped down. While I can't fix that, I can stop the miengine from crashing which is what i've done here

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -602,6 +602,17 @@ namespace Microsoft.MIDebugEngine
             DebuggedThread thread = await ThreadCache.GetThread(tid);
             await ThreadCache.StackFrames(thread);  // prepopulate the break thread in the thread cache
             ThreadContext cxt = await ThreadCache.GetThreadContext(thread);
+
+            if (cxt == null)
+            {
+                // Something went seriously wrong. For instance, this can happen when the primary thread
+                // of an app exits on linux while background threads continue to run with pthread_exit on the main thread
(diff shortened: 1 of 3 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -602,6 +602,17 @@ private void Dispose()
...
             ThreadContext cxt = await ThreadCache.GetThreadContext(thread);
+
+            if (cxt == null)
+            {
+                // Something went seriously wrong. For instance, this can happen when the primary thread
+                // of an app exits on linux while background threads continue to run with pthread_exit on the main thread
+                // See https://devdiv.visualstudio.com/DefaultCollection/DevDiv/VS%20Diag%20IntelliTrace/_workItems?_a=edit&id=197616&triage=true
+                // for a repro
+                Debug.Fail("Failed to find thread on break event.");
+                throw new MIException(Constants.E_FAIL);
```
gregg-miskelly: I wonder if we want to provide a better description here. It looks like we will send an error event that the user might see.

gregg-miskelly:
Aside from that, looks good to me.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
jacdavis: Sure, I can add something. I didn't see a lot of exceptions in mi that had more detailed error messages except for the miexception type which couldn't really be used here. I'll throw something with a message that describes that the thread couldn't be found.

PR #219, opened 2016-03-09
Validate the midebugtarget isn't a folder or non-existant file. Since…

jacdavis:
… this

is passed to bash we don't directly receive an error from the
proces.start. Note that doesn't protect valid exes that aren't gdb but the
name of gdb and lldb, and clrdbg can change drastically in different
scenarios so I don't see a good way to do that. (for instance, gdb can
have wildly different names in cross compiled scenarios)

### DIFF
--- a/src/MICore/MICoreResources.Designer.cs
+++ b/src/MICore/MICoreResources.Designer.cs
@@ -185,6 +185,15 @@ namespace MICore {
         }
         
         /// <summary>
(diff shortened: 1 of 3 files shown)


jacdavis:
@chuckries

chuckries:
:+1:

gregg-miskelly:
:+1:

PR #220, opened 2016-03-11
Collect source info for core dumps for top frame only if available.

jander-msft:
This fixes an issue that occurs while attempting to debug a core dump where the stack frame that causes the crash belongs to a binary that does not appropriate symbol information. In this situation, an exception would be throw with a message similar to "Unrecognized format of field "file" in result: {level=0,addr=0x000000000040070a,func=main,args=[]}".

Previously, it required that the file, filename, and line fields of the frame to exist, but they don't exist if the binary doesn't have the debugging information in it. These fields are now included only if they are available.

### DIFF
--- a/src/MICore/MIResults.cs
+++ b/src/MICore/MIResults.cs
@@ -404,13 +404,35 @@ namespace MICore
             return FindAll(name).OfType<T>().ToArray();
         }
 
-        public TupleValue Subset(params string[] names)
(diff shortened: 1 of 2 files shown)


gregg-miskelly:
:+1:

jacdavis:
LGTM. Please let me know when you're ready for merge

PR #222, opened 2016-03-15
Add support for running GDB with sudo to attach

andrewcrawley:
Many Linux systems have a default configuration that disallows non-root users from attaching to arbitrary processes.  On those systems, we currently require the MIEngine host to run as root, but that is undesirable for several reasons.  This fixes the LocalLinuxTransport to detect this case and use either pkexec or sudo to run GDB as root.

@jacdavis @jander-msft


paulmaybee:
:+1:

PR #225, opened 2016-03-17
Fix VSCode C++ debug F10

rick111111:
add a launch option to opt out "stop on dynamic module load"

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -444,7 +444,7 @@ namespace MICore
             }
         }
 
-        private bool _waitDynamicLibLoad;
+        private bool _waitDynamicLibLoad = true;
         /// <summary>
         /// If true, wait for dynamic library load to finish.
         /// </summary>
@@ -453,7 +453,7 @@ namespace MICore
             get { return _waitDynamicLibLoad; }
             set
             {
-                VerifyCanModifyProperty("StopOnDynamicLibLoad");
+                VerifyCanModifyProperty("WaitDynamicLibLoad");
                 _waitDynamicLibLoad = value;
             }
         }
@@ -798,6 +798,7 @@ namespace MICore
                 this.VisualizerFile = source.VisualizerFile;
 
             this.ShowDisplayString = source.ShowDisplayString;
+            this.WaitDynamicLibLoad = source.WaitDynamicLibLoad;
 
             this.SetupCommands = LaunchCommand.CreateCollectionFromXml(source.SetupCommands);
 
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -444,6 +444,20 @@ namespace MICore
             }
         }
 
+        private bool _waitDynamicLibLoad;
+        /// <summary>
+        /// If true, wait for dynamic library load to finish.
+        /// </summary>
+        public bool WaitDynamicLibLoad
+        {
+            get { return _waitDynamicLibLoad; }
+            set
+            {
+                VerifyCanModifyProperty("StopOnDynamicLibLoad");
+                _waitDynamicLibLoad = value;
+            }
+        }
+
         /// <summary>
         /// If true, instead of showing Natvis-DisplayString value as a child of a dummy element, it is shown immediately.
         /// Should only be enabled if debugger is fast enough providing the value.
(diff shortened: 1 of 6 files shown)


src/MICore/LaunchOptions.cs
```diff
@@ -444,6 +444,20 @@ public string VisualizerFile
...
+        private bool _waitDynamicLibLoad;
+        /// <summary>
+        /// If true, wait for dynamic library load to finish.
+        /// </summary>
+        public bool WaitDynamicLibLoad
+        {
+            get { return _waitDynamicLibLoad; }
+            set
+            {
+                VerifyCanModifyProperty("StopOnDynamicLibLoad");
```
gregg-miskelly: Wrong name of the property

src/MICore/LaunchOptions.cs
```diff
@@ -444,6 +444,20 @@ public string VisualizerFile
             }
         }
 
+        private bool _waitDynamicLibLoad;
```
gregg-miskelly: Default to true?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -289,7 +289,12 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
                     // This is to work around a GDB bug of warning "Failed to set controlling terminal: Operation not permitted"
                     // Reset debuggee terminal after the first module load.
                     // The clear is done by sending reset string (ESC, c) to terminal STDERR
-                    await ConsoleCmdAsync(@"shell echo -e \\033c 1>&2");
+                    await ConsoleCmdAsync(@"shell echo -e \\033c 1>&2");                    
+                }
+
+                if (this.MICommandFactory.SupportsStopOnDynamicLibLoad() && !_launchOptions.WaitDynamicLibLoad)
```
gregg-miskelly: Isn't off the default state for GDB? Any reason we need to explicitly turn this off instead of just leaving it default?

gregg-miskelly:
Looks good aside from those minor comments.

gregg-miskelly:
You need a check here too:
MIDebugEngine\Engine.Impl\DebuggedProcess.cs - (96, 69) : if (!string.IsNullOrEmpty(file) && MICommandFactory.SupportsStopOnDynamicLibLoad())

gregg-miskelly:
Or maybe you should just fix the code path that decides if it should assign to _waitDialog since then you could easily fix all the code paths that use that.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -485,7 +490,7 @@ private List<LaunchCommand> GetInitializeCommands()
             if (this.MICommandFactory.SupportsStopOnDynamicLibLoad())
             {
                 // Do not stop on shared library load/unload events while debugging core dump
-                if (!this.IsCoreDump)
+                if ((_needTerminalReset || _launchOptions.WaitDynamicLibLoad) && !this.IsCoreDump)
```
jacdavis: May want a comment on why terminal reset requires module load.

src/MICore/LaunchOptions.cs
jacdavis: What happens if this is false? Do we still go through this codepath in the default scenarios? Does this throw and block debugging?

rick111111: Not sure I get the question. If the property is false, for linux we might still wait for the first module load to clear the terminal. But other than that we simple not sending the "set stop-on-solib-events 1" command. It should not block debugging.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
rick111111: Yes off is the default setting for GDB. However for Linux C++ debugging, we need to work around a GDB bug and call "ConsoleCmdAsync(@"shell echo -e \\033c 1>&2")" to clear a terminal error. This clear operation can't be done too early (because GDB only generate this message after start debugging) or too late (otherwise we might clear debuggee's output). The stop cause by first module load seems to be the perfect timing to clear the terminal, that's why we still need to initially turn stop-on-solib-events on then turn it off after the first stop.

gregg-miskelly: Ah. Got it. Misread the change.

jacdavis:
LGTM if you address Gregg's remaining two comments.

paulmaybee:
lgtm

chuckries:
@dotnet-bot build please

PR #227, opened 2016-03-18
Fixes for InstallToVSCode

gregg-miskelly:
- With the latest drop of the .NET CLI, the CLR was unhappy that we were using the version of CLR files that were dropped from CLRDBG instead of using the ones created from a dummy exe restore. This changes the script to only copy over CLR files from CLRDBG if we didn't have them already.
- InstallToVSCode.cmd was correctly using the passed in configuration for OpenDebugAD7 files, but was always using 'Debug' for MIEngine files. Now it correctly honor's the command line argument.


gregg-miskelly:
@caslan please review.

tools/InstallToVSCode/InstallToVSCode.cmd
```diff
@@ -84,7 +84,7 @@ for %%f in (%OpenDebugAD7BinDir%\*.dll) do call :InstallFile "%%f"
 
 echo.
 echo Installing clrdbg bits from %CLRDBGBITSDIR%...
-for %%f in (%CLRDBGBITSDIR%\*.dll) do call :InstallFile "%%f"
+for %%f in (%CLRDBGBITSDIR%\*.dll) do call :InstallNewFile "%%f"
```
caslan: Do you want to add a comment here why we are doing so?

caslan:
LGTM with one minor comment.

gregg-miskelly:
Added comment per review.

PR #228, opened 2016-03-18
Add support for non-absolute paths in 'ExePath'

gregg-miskelly:
The .NET CLI is making changes where we want to tell clrdbg to launch 'dotnet'. This makes some MIEngine tweaks so that we let clrdbg/gdb/lldb decide if the passed in program path is valid instead of having the MIEngine pre-validate it, and then do a reasonable job handling failures from the underlying debugger.

### DIFF
--- a/src/MICore/Transports/LocalLinuxTransport.cs
+++ b/src/MICore/Transports/LocalLinuxTransport.cs
@@ -80,7 +80,19 @@ namespace MICore
                 throw new Exception(MICoreResources.Error_InvalidMiDebuggerPath);
             }
 
-            string debuggeeDir = System.IO.Path.GetDirectoryName(options.ExePath);
+            // Default working directory is next to the app
+            string debuggeeDir;
+            if (Path.IsPathRooted(options.ExePath) && File.Exists(options.ExePath))
+            {
+                debuggeeDir = System.IO.Path.GetDirectoryName(options.ExePath);
+            }
+            else
+            {
+                // If we don't know where the app is, default to HOME, and if we somehow can't get that, go with the root directory.
+                debuggeeDir = Environment.GetEnvironmentVariable("HOME");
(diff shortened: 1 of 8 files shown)


gregg-miskelly:
@jacdavis @chuckries please review.

src/MICore/Transports/LocalLinuxTransport.cs
```diff
@@ -80,7 +80,19 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
...
+            {
+                debuggeeDir = System.IO.Path.GetDirectoryName(options.ExePath);
+            }
+            else
+            {
+                // If we don't know where the app is, default to HOME, and if we somehow can't get that, go with the root directory.
+                debuggeeDir = Environment.GetEnvironmentVariable("HOME");
+                if (string.IsNullOrEmpty(debuggeeDir))
+                    debuggeeDir = "/";
+            }
```
jacdavis: We have a launch option that allows the user to specify the working folder. Is it not getting consumed?

gregg-miskelly: Yeah, this code seems a bit weird to me too. That option setting will work because we issue a '-environment-cd' to GDB to set the working directory. This just establishes kind of an initial working directory. I think honestly this code should probably be removed and replaced with something (see debuggedprocess.cs(556)). But I didn't have a good way of proving out all the cases, so I was hesitant to just remove it. So I just wanted to avoid breaking what was here.

jacdavis: I'll file a bug in the morning to track this

Sent from my Windows Phone

jacdavis: Bug filed on our side to resolve this before build.

chuckries:
:+1:

gregg-miskelly:
Rebased on top of current master.

PR #230, opened 2016-03-21
Update VS Code metrics for 'coreclr' mode

gregg-miskelly:
OpenDebugAD7 is being changed to read more configuration from the metrics. This adds
these metrics to the .json file.


chuckries:
:+1:

PR #231, opened 2016-03-21
Added fallback to xterm if gnome-terminal doesn't exist

pieandcakes:
@jacdavis @gregg-miskelly This pull request is to allow the terminal fallback to xterm if gnome-terminal isn't available

### DIFF
--- a/src/MICore/Transports/LocalLinuxTransport.cs
+++ b/src/MICore/Transports/LocalLinuxTransport.cs
@@ -16,6 +16,7 @@ namespace MICore
         private const string PKExecPath = "/usr/bin/pkexec";
         private const string SudoPath = "/usr/bin/sudo";
         private const string GnomeTerminalPath = "/usr/bin/gnome-terminal";
+        private const string XTermPath = "/usr/bin/xterm";
 
         private void MakeGdbFifo(string path)
         {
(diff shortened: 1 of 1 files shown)


src/MICore/Transports/LocalLinuxTransport.cs
```diff
@@ -110,6 +111,19 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
...
             bool requiresRootAttach = this.GetPtraceScope() != 0;
 
+            // Check and see if gnome-terminal exists. If not, fall back to xterm
+            string terminalCmd, bashCommandPrefix;
+            if (File.Exists(GnomeTerminalPath))
+            {
+                terminalCmd = GnomeTerminalPath;
+                bashCommandPrefix = "--title DebuggerTerminal -x";
+            }
+            else
```
gregg-miskelly: Should we have some failure message if both are unavailable?

gregg-miskelly:
Looks good aside from comment.

src/MICore/Transports/LocalLinuxTransport.cs
```diff
@@ -144,15 +158,16 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
                 }
             }
 
-            string argumentString = string.Format(System.Globalization.CultureInfo.InvariantCulture,
-                    "--title DebuggerTerminal -x bash -c \"cd {0}; DbgTerm=`tty`; trap 'rm {2}; rm {3}' EXIT; {1} --interpreter=mi --tty=$DbgTerm < {2} > {3};\"",
+            string argumentString = string.Format(CultureInfo.InvariantCulture,
+                    "{4} bash -c \"cd {0}; DbgTerm=`tty`; trap 'rm {2}; rm {3}' EXIT; {1} --interpreter=mi --tty=$DbgTerm < {2} > {3};\"",
```
jacdavis: I would normally nitpick the order of the params, but given how complex this is (and the fact that I too am guilty of this) I shall keep my mouth shut.

jacdavis:
LGTM. Let me know when you're ready for us to merge.

src/MICore/Transports/LocalLinuxTransport.cs
pieandcakes: Probably, but right now it just says that gnome-terminal (or xterm) isn't available

PR #232, opened 2016-03-21
Fix InstallToVSCode on OSX

gregg-miskelly:
On OSX InstallToVSCode wasn't working because 'wc' was outputting whitespace
before the number of lines. This changes to use a regex to handle this.


caslan:
LGTM

chuckries:
:+1:

PR #235, opened 2016-03-22
Update GetClrDbg.sh script to understand version constants

chuckries:
Introduce named constants for versions of clrdbg to make the script
easier to use. We still accept 'latest' and any specific version number.


chuckries:
@gregg-miskelly

gregg-miskelly:
:+1:

PR #237, opened 2016-03-23
Be more eager about cleaning up GDB variable objects

andrewcrawley:
Waiting for VariableInfo objects to be GC'ed before cleaning up the corresponding object on the GDB side means that GDB can end up with a lot of memory usage due to unused objects before the MIEngine side experiences enough memory pressure to trigger a GC.

We can fix this by keeping a list of active VariableInfo objects and disposing any that exist when we enter break mode, as they must have been created during an earlier break and are no longer valid anyway.

@jacdavis @gregg-miskelly @chuckries @paulmaybee

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -23,6 +23,7 @@ namespace Microsoft.MIDebugEngine
         public AD_PROCESS_ID Id { get; private set; }
         public AD7Engine Engine { get; private set; }
         public List<string> VariablesToDelete { get; private set; }
+        public List<IVariableInformation> ActiveVariables { get; private set; }
 
         public SourceLineCache SourceLineCache { get; private set; }
         public ThreadCache ThreadCache { get; private set; }
@@ -74,6 +75,7 @@ namespace Microsoft.MIDebugEngine
             ExceptionManager = new ExceptionManager(MICommandFactory, _worker, _callback, configStore);
 
             VariablesToDelete = new List<string>();
+            this.ActiveVariables = new List<IVariableInformation>();
 
             OutputStringEvent += delegate (object o, string message)
             {
@@ -620,6 +622,17 @@ namespace Microsoft.MIDebugEngine
                 tid = results.Results.FindInt("thread-id");
             }
 
+            // Any existing variable objects at this point are from the last time we were in break mode, and are
+            //  therefore invalid.  Dispose them so they're marked for cleanup.
+            lock(this.ActiveVariables)
+            {
+                foreach (IVariableInformation varInfo in this.ActiveVariables)
+                {
+                    varInfo.Dispose();
+                }
+                this.ActiveVariables.Clear();
+            }
+
             ThreadCache.MarkDirty();
             MICommandFactory.DefineCurrentThread(tid);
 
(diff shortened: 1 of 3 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -74,6 +75,7 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
             ExceptionManager = new ExceptionManager(MICommandFactory, _worker, _callback, configStore);
 
             VariablesToDelete = new List<string>();
+            this.ActiveVariables = new List<IVariableInformation>();
```
gregg-miskelly: You are accessing this from multiple threads. Probably should have a lock+make this private, or you a collection with locking built in.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -620,6 +622,14 @@ private void Dispose()
                 tid = results.Results.FindInt("thread-id");
             }
 
+            // Any existing variable objects at this point are from the last time we were in break mode, and are
+            //  therefore invalid.  Dispose them so they're marked for cleanup.
+            foreach (IVariableInformation varInfo in this.ActiveVariables)
```
gregg-miskelly: Might just be my bias as this is how VS does it, but it might make sense to do this when we continue instead of when we reenter break.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -173,6 +173,11 @@ private VariableInformation(ThreadContext ctx, AD7Engine engine, AD7Thread threa
             _attribsFetched = false;
             Access = enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_NONE;
             _fullname = null;
+
+            lock(this._debuggedProcess.ActiveVariables)
+            {
+                this._debuggedProcess.ActiveVariables.Add(this);
```
gregg-miskelly: Does it make sense to add all the children since it doesn't do much of anything?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
gregg-miskelly: I wonder if we should make these WeakReferences so variables can still be GC'ed

gregg-miskelly:
Otherwise sounds reasonable.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -703,5 +691,38 @@ public void Assign(string expression)
...
         }
+
+        #region IDisposable Implementation
+
+        public void Dispose()
+        {
+            this.Dispose(true);
+
+            GC.SuppressFinalize(this);
+        }
```
jacdavis: Typical dispose pattern is to throw object disposed after this call for everything except the finalizer. How much extra work would that be?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
andrewcrawley: Added missing lock

src/MIDebugEngine/Engine.Impl/Variables.cs
andrewcrawley: Verified object not disposed in all methods that invoke MI commands

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -697,11 +695,58 @@ public bool IsReadOnly()
...
+
+        private void Dispose(bool isDisposing)
+        {
+            this._isDisposed = true;
```
jacdavis: race on this?

andrewcrawley: Potentially, but due to the use of async/await, I think this is the best we can reasonably do.

PR #239, opened 2016-03-23
Stop debugging when in or out fifo is deleted

rick111111:


chuckries:
Please rebase on top of current master.

src/MICore/Transports/LocalLinuxTransport.cs
```diff
@@ -187,6 +195,15 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
             reader = new StreamReader(gdbStdOutStream);
         }
 
+        private void FifoWatcher_Deleted(object sender, FileSystemEventArgs e)
+        {
+            if (e.FullPath.Contains(_gdbStdInName) || e.FullPath.Contains(_gdbStdOutName))
+            {
+                _fifoWatcher.Deleted -= FifoWatcher_Deleted;
+                this.Callback.OnDebuggerProcessExit(null);
```
jacdavis: Since the fifos are always deleted on debugging stop, does this now mean we're trigging exit twice?

src/MICore/Transports/LocalLinuxTransport.cs
```diff
@@ -187,9 +195,29 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
...
+        }
+
         protected override string GetThreadName()
         {
             return "MI.LocalLinuxTransport";
         }
+
+        public override void Close()
+        {
+            base.Close();
```
rick111111: Jackson's question seems disappear after I rebase the branch, the question is "Since the fifos are always deleted on debugging stop, does this now mean we're trigging exit twice?"
I added the code to remove file watcher on Close(), or normal exit Close will be called before the fifo is actually deleted.

jacdavis: But then isn’t there still a race condition?

src/MICore/Transports/LocalLinuxTransport.cs
```diff
@@ -95,15 +99,19 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
...
-            string gdbStdOutName = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+            _gdbStdInName = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+            _gdbStdOutName = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+
+            MakeGdbFifo(_gdbStdInName);
+            MakeGdbFifo(_gdbStdOutName);
 
-            MakeGdbFifo(gdbStdInName);
-            MakeGdbFifo(gdbStdOutName);
+            _fifoWatcher = new FileSystemWatcher(Path.GetTempPath());
```
gregg-miskelly: Can we specify a file path instead of a directory path? Or stick a prefix on our file? Watching for any delete for any file under /tmp seems wasteful.

src/MICore/Transports/LocalLinuxTransport.cs
```diff
@@ -187,9 +195,29 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
...
         }
 
+        private void FifoWatcher_Deleted(object sender, FileSystemEventArgs e)
+        {
+            if (e.FullPath == _gdbStdInName || e.FullPath == _gdbStdOutName)
+            {
+                Logger?.WriteLine("Gdb fifo deleted, stop debugging");
+                _fifoWatcher.Deleted -= FifoWatcher_Deleted;
+                this.Callback.OnDebuggerProcessExit(null);
+            }
```
gregg-miskelly: If you throw an exception here, will it take down the OpenDebugAD7 process? If so, I would probably suggest having a try/cacth

gregg-miskelly:
Otherwise looks good to me.

src/MICore/Transports/LocalLinuxTransport.cs
rick111111: I double checked this, there won't be a race condition. When user stop debugging from the UI, we either send "kill" command or "exec-detach", which will make the thread-group exit which will in turn cause the transport to Close(). and the file watcher will be removed.
After all this, we'll send "exec-quit" to quit Gdb and only then the fifos will be removed.

src/MICore/Transports/LocalLinuxTransport.cs
rick111111: Addressed by adding filter

src/MICore/Transports/LocalLinuxTransport.cs
jacdavis: But let's consider this scenario: the user clicks stop debugging and at that very moment, gdb crashes. The file system watcher wakes up and notifies us that the fifos were deleted before the watcher was removed so the code path is taken. Then, the call to stop also enters this code on another thread and also enters this function. Are you sure this can't happen?

src/MICore/Transports/LocalLinuxTransport.cs
jacdavis: How does the filter work? Isn't the watcher still triggered if another file is added (even though we don't do anything with it?) Would it be better to place the fifos in a sub folder under tmp and then watch that?

src/MICore/Transports/LocalLinuxTransport.cs
rick111111: I see your point, in that case Debugger._exit is guarding against this situation and making sure the exit only happens once per session.

chuckries:
Not a huge deal but please squash your multiple 'take review comments' commits before merging.

PR #240, opened 2016-03-23
Use awk for converting version paremeter to lower case

chuckries:
The old method required bash 4.0 or above, which is not available by default on OSX.

### DIFF
--- a/scripts/GetClrDbg.sh
+++ b/scripts/GetClrDbg.sh
@@ -47,12 +47,12 @@ elif [ "$1" == "-h" ]; then
     exit 1
 fi
 
-version_string=$1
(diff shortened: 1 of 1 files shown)


gregg-miskelly:
Should you commit this to master and then fast-forward master to getclrdbg-release?

gregg-miskelly:
:+1:

chuckries:
For posterity's sake, to answer Gregg's question:
I believe the correct thing to do is merge the PR into the getclrdbg-release branch, and then merge the getclrdbg-release branch into master. getclrdbg-release should always be able to merge into master, and there is no need for getclrdbg-release to have newer commits from master.

PR #241, opened 2016-03-24
Add launch option for controlling external console

gregg-miskelly:
This checkin adds a new launch option for controlling if an external console should be
used and modifies the external console support so that it can be used with CLRDBG.

Remaining work:
* I am currently sometimes seeing an access violation which leads to a hang inside of
clrdbg during shutdown. This is inside of the JIT, so I want to try upgrading my CoreCLR
before investigating more.
* Support OSX

### DIFF
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -504,7 +504,7 @@ namespace MICore
 
         #endregion
 
-        #region Abstract Methods
+        #region Other
 
         abstract protected Task<Results> ThreadFrameCmdAsync(string command, ResultClass expectedResultClass, int threadId, uint frameLevel);
         abstract protected Task<Results> ThreadCmdAsync(string command, ResultClass expectedResultClass, int threadId);
@@ -522,6 +522,16 @@ namespace MICore
         public virtual bool IsAsyncBreakSignal(Results results)
         {
             return (results.TryFindString("reason") == "signal-received" && results.TryFindString("signal-name") == "SIGINT");
+        }
+
+        /// <summary>
+        /// Determines if a new external console should be spawned on non-Windows platforms for the debugger+app
+        /// </summary>
+        /// <param name="localLaunchOptions">[required] local launch options</param>
+        /// <returns>True if an external console should be used</returns>
+        public virtual bool UseExternalConsoleForLocalLaunch(LocalLaunchOptions localLaunchOptions)
+        {
+            return localLaunchOptions.UseExternalConsole && String.IsNullOrEmpty(localLaunchOptions.MIDebuggerServerAddress) && !localLaunchOptions.IsCoreDump;
         }
 
         public Results IsModuleLoad(string cmd)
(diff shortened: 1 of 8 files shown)


gregg-miskelly:
@caslan @jacdavis please review

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -524,6 +524,16 @@ public virtual bool IsAsyncBreakSignal(Results results)
...
         }
 
+        /// <summary>
+        /// Determines if a new external console should be spawned on non-Windows platforms for the debugger+app
+        /// </summary>
+        /// <param name="localLaunchOptions">[required] local launch options</param>
+        /// <returns>True if an external console should be used</returns>
+        public virtual bool UseExternalConsoleForLocalLaunch(LocalLaunchOptions localLaunchOptions)
+        {
+            return localLaunchOptions.UseExternalConsole && String.IsNullOrEmpty(localLaunchOptions.MIDebuggerServerAddress) && !localLaunchOptions.IsCoreDump;
```
jacdavis: Is the default on this true? gdb will deadlock for us if this isn't set. Eventually, we'd probably not want to throw the console on attach, but again, today gdb deadlocks.

jacdavis:
Assuming the default is true, this looks good. @andrewcrawley Please review as well.

src/MICore/Transports/LocalLinuxTransport.cs
```diff
@@ -213,6 +197,42 @@ private void FifoWatcher_Deleted(object sender, FileSystemEventArgs e)
             }
         }
 
+        private bool GetRequiresRootAttach()
+        {
+            if (_miMode != MIMode.Clrdbg)
```
andrewcrawley: This check shouldn't be necessary - we already check in InitStreams to be sure the target is GDB before checking this.

src/MICore/Transports/LocalLinuxTransport.cs
```diff
@@ -17,11 +17,17 @@ public class LocalLinuxTransport : StreamTransport
...
         private const string XTermPath = "/usr/bin/xterm";
-        private const string FifoPrefix = "cpp-debug-fifo-";
+        private const string FifoPrefix = "Microsoft-MIEngine-fifo-";
 
         private string _gdbStdInName;
         private string _gdbStdOutName;
         private FileSystemWatcher _fifoWatcher;
+        private readonly MIMode _miMode;
+
+        public LocalLinuxTransport(MIMode miMode)
```
andrewcrawley: I don't think this constructor is necessary - the value is only used in InitStreams, and we can get it from localLaunchOptions.DebuggerMIMode instead of having it passed in at instantiation.

src/MICore/Transports/LocalLinuxTransport.cs
gregg-miskelly: Not any more it doesn't

src/MICore/Transports/LocalLinuxTransport.cs
gregg-miskelly: Yeah, that would work too.

src/MICore/CommandFactories/MICommandFactory.cs
gregg-miskelly: Yes, see gdb.cs

caslan:
LGTM

src/MICore/Transports/LocalLinuxTransport.cs
andrewcrawley: The other check is on line 137 - I don't see that it was removed?  In any case, we probably only need one check or the other.

gregg-miskelly: Thanks. I missed that one. I thought you were talking about the one before we entered. I will take the one out on line 137.

PR #242, opened 2016-03-24
Update error message

rick111111:
Some VSCode scenario will share this error message, update the message to work for both scenarios.

### DIFF
--- a/src/MICore/MICoreResources.resx
+++ b/src/MICore/MICoreResources.resx
@@ -126,7 +126,7 @@
 {1}</value>
   </data>
   <data name="Error_DebuggerInitializeFailed_NoStdErr" xml:space="preserve">
-    <value>Unable to establish a connection to {0}. Go to the 'Debug' pane of the Visual Studio output window for more information.</value>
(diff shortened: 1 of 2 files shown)


jacdavis:
LGTM

src/MICore/MICoreResources.resx
```diff
@@ -154,7 +154,7 @@ Error: {1}</value>
     <value>Launch options string provided by the project system is invalid. {0}</value>
   </data>
   <data name="Error_InvalidLocalExePath" xml:space="preserve">
-    <value>Invalid path to executable file path '{0}'. File must be a valid file name that exists on the Visual Studio computer.</value>
+    <value>Invalid path to executable file path '{0}'. File must be a valid file name that exists on the computer.</value>
```
gregg-miskelly: This string is only used for core dumps anymore. So I would suggest just --
Invalid path to executable file path '{0}'. File must be a valid file name that exists.

gregg-miskelly:
Looks good aside from that comment.

PR #243, opened 2016-03-25
Add the type to the name of the variable for base class variables.

edumunoz:
This is required for VSCode because it makes sure all variables are unique
by the name.

screenshot from 2016-03-24 17 24 43

/cc @jacdavis @gregg-miskelly @chuckries

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -239,7 +239,7 @@ namespace Microsoft.MIDebugEngine
             {
                 // base classes show up with no value and exp==type 
                 // (sometimes underlying debugger does not follow this convention, when using typedefs in templated types so look for "::" in the field name too)
-                Name = "base";
+                Name = TypeName + " (base)";
                 Value = TypeName;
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -239,7 +239,7 @@ private VariableInformation(TupleValue results, VariableInformation parent, stri
             {
                 // base classes show up with no value and exp==type 
                 // (sometimes underlying debugger does not follow this convention, when using typedefs in templated types so look for "::" in the field name too)
-                Name = "base";
+                Name = TypeName + " (base)";
```
jacdavis: We may need to put this in an if(gdb.... to not impact clrdbg.

jacdavis: @caslan

gregg-miskelly:
This seems fine. I don't think we really need to special case clrdbg. :+1:

src/MIDebugEngine/Engine.Impl/Variables.cs
caslan: I think it's fine

PR #244, opened 2016-03-26
Add handling for async return stack frames

caslan:
This fixes a crash in MIEngine when there are managed return stack frames in the callstack. These frames have IL offsets and therefore text positions but no native offsets or CPU instruction addresses. The crash happens due to trying to set document context on a null code context object. In order to be able to properly show these frames and allow showing the source for them, changed the code to store the document context separately on the stack frame object and use that for GetDocumentContext call.


caslan:
@jacdavis @gregg-miskelly please review

gregg-miskelly:
Did you try this in VS? We should make sure that VS isn't unhappy with there being a document context without a code context.

gregg-miskelly:
:+1:

caslan:
VS works as well, you can view the return frame, navigate to it etc.

PR #246, opened 2016-03-29
Fix regression caused by earlier async frame fix

caslan:
Previously the MIEngine code would mark a frame as “annotated” if there was no code context for it. I changed that logic to look for a document context to decide whether a frame was annotated. This was so that we can switch to the virtual async frame. However, this brought up the issue where the frame for an exception thrown in external code would also be marked as annotated since it has a code context but no document context. 

Changed the code so that a frame is marked annotated if there is no code context and no document context.


caslan:
@gregg-miskelly

gregg-miskelly:
:+1:

PR #248, opened 2016-03-31
Update CoreCLR version

gregg-miskelly:
commit cc670121793cef49c452fb2274768cef3600dd46
Date:   Thu Mar 31 12:09:23 2016 -0700

    Update CoreCLR version
    
    This updates the CoreCLR version when using InstallToVSCode.

commit e4dc11aa48edea4b51f0194a8c3022ff83cd2aae
Date:   Thu Mar 31 12:07:55 2016 -0700

    Fix intermitent crash in SteamTransport
    
    On Unix, the stream transport was sometimes crashing on debugger
    abort as it still had unflushed data. This catches the exception.

commit 7cb30a6ecdb6fe8ad3ada63052684e6f3f513108
Date:   Thu Mar 31 12:03:37 2016 -0700

    LocalTransport: Stop corrupting the PATH environment variable
    
    LocalTransport had some code meant for working around a bug with GDB on
    Windows where the GDB directory needed to be on the path. However,
    it was doing this in all scenarios.
    
    It's possible that there are other scenarios where this code might be useful,
    but in that case it needs to use the ':' separator on UNIX instead of
    corrupting the path. For now, I just went with the easy fix of turning it off
    except for where we know we need it.

### DIFF
--- a/src/MICore/Transports/LocalTransport.cs
+++ b/src/MICore/Transports/LocalTransport.cs
@@ -9,6 +9,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Collections.Specialized;
 using System.Collections;
+using System.Runtime.InteropServices;
 
 namespace MICore
 {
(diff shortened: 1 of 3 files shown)


gregg-miskelly:
@wiktork you will want to do something like this.

chuckries:
:+1:

PR #249, opened 2016-03-31
Send MIEngine version to clrdbg

caslan:
Send MIEngine version to clrdbg on initialize so that it can set capabilities correctly.


caslan:
@gregg-miskelly

gregg-miskelly:
:+1:

PR #250, opened 2016-04-01
Fix InstallToVSCode for breaking CLI change

gregg-miskelly:
The CLI team previously indicated that a native host executable will not be dropped
unless the project.json explicitly indicates that the application wants to be
platform depentant. But builds from a few days ago were actually behaving nicer than
the way I told it should work.

Now things are actually behaving this way. So I am fixing up InstallToVSCode to generate
project.json on the fly so we can write out a runtime id.

### DIFF
--- a/tools/InstallToVSCode/InstallToVSCode.sh
+++ b/tools/InstallToVSCode/InstallToVSCode.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-
 script_dir=`dirname $0`
 
 print_help()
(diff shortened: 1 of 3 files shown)


gregg-miskelly:
@chuckries please review, @wiktork FYI

tools/InstallToVSCode/InstallToVSCode.sh
```diff
@@ -199,8 +198,31 @@ hash dotnet 2>/dev/null
...
+cp -r $script_dir/CLRDependencies/* $DESTDIR/CLRDependencies
+[ $? -ne 0 ] && echo "ERROR: unable to create destination copy CLRDependencies directory." && exit 1
+
+pushd $DESTDIR/CLRDependencies 1>/dev/null 2>/dev/null
+[ $? -ne 0 ] && echo "ERROR: Unable to change to CLRDependencies directory???" && exit 1
+
+OSName=$(uname -s)
+if [ "$OSName" == "Darwin" ]; then
+    # On OSX, hard code the runtime id so we don't need to worry about version.
+    runtime_id="osx.10.11-x64"
```
caslan: No yosemite anymore?

caslan: I mean you assume we won't have yosemite on our boxes?

caslan:
LGTM except a minor comment about yosemite support

tools/InstallToVSCode/InstallToVSCode.sh
gregg-miskelly: I think my Mac is still Yosemite, and it seems to work. This is just the runtime id they use on OSX.

wiktork: Do we know if the CLI team plans on removing the --infer-runtimes switch? That brings back the old behavior.

gregg-miskelly: Interesting. Never would have expected that.

PR #251, opened 2016-04-01
Fix bound breakpoints to resolve to correct line when using gdb

richardstanton:
GDB doesn't return the bound breakpoint line after creating a line breakpoint, this code will request the address to line mapping for the breakpoint and get the correct line if required.

I used the existing SourceLineCache code for doing the lookup, this gets the address mapping of every line in the file. Unfortunately gdb does not actually implement the mi call for "symbol-info-symbol" to request a single address lookup.

LLDB does not have this issue and returns the bound line from "break-insert".

@gregg-miskelly @jacdavis @chuckries @paulmaybee

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/Breakpoints.cs
+++ b/src/MIDebugEngine/Engine.Impl/Breakpoints.cs
@@ -96,7 +96,21 @@ namespace Microsoft.MIDebugEngine
             process.VerifyNotDebuggingCoreDump();
 
             string basename = System.IO.Path.GetFileName(documentName);     // get basename from Windows path
-            return EvalBindResult(await process.MICommandFactory.BreakInsert(process.EscapePath(basename), line, condition, ResultClass.None), pbreak);
+            basename = process.EscapePath(basename);
+            BindResult bindResults = EvalBindResult(await process.MICommandFactory.BreakInsert(basename, line, condition, ResultClass.None), pbreak);
+
+            // On GDB, the returned line information is from the pending breakpoint instead of the bound breakpoint.
+            // Check the address mapping to make sure the line info is correct.
+            if (process.MICommandFactory.Mode == MIMode.Gdb &&
+                bindResults.BoundBreakpoints != null)
+            {
+                foreach (var boundBreakpoint in bindResults.BoundBreakpoints)
+                {
+                    boundBreakpoint.Line = await process.LineForStartAddress(basename, boundBreakpoint.Addr);
+                }
+            }
+
+            return bindResults;
         }
 
         private static BindResult EvalBindResult(Results bindResult, AD7PendingBreakpoint pbreak)
@@ -324,6 +338,26 @@ namespace Microsoft.MIDebugEngine
             }
 
             return new AD7DocumentContext(_textPosition, new AD7MemoryAddress(engine, Addr, this.FunctionName));
+        }
+
+        /// <summary>
+        /// Returns the start line of the breakpoint.
+        /// NOTE: If set this overwrites any column or multiline information.
+        /// </summary>
+        internal uint Line
+        {
+            get
+            {
+                return _textPosition.BeginPosition.dwLine;
+            }
+            set
+            {
+                if (this.Line == value || value == 0)
+                {
+                    return;
+                }
+                _textPosition = new MITextPosition(_textPosition.FileName, value);
+            }
         }
     }
 }
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -1422,29 +1422,53 @@ namespace Microsoft.MIDebugEngine
             await _breakpointManager.EnableAfterFuncEvalAsync();
         }
 
-        public async Task<List<ulong>> StartAddressesForLine(string file, uint line)
+        /// <summary>
+        /// Finds the line associated with a start address.
+        /// </summary>
+        public async Task<uint> LineForStartAddress(string file, ulong startAddress)
         {
             List<ulong> addresses = new List<ulong>();
-            var srcLines = await SourceLineCache.GetLinesForFile(file);
-            if (srcLines == null || srcLines.Length == 0)
+            SourceLineMap srcLines = await SourceLineCache.GetLinesForFile(file);
+            if (srcLines == null || srcLines.Count == 0)
             {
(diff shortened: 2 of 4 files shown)


src/MIDebugEngine/Engine.Impl/Breakpoints.cs
```diff
@@ -96,7 +96,20 @@ private static MIBreakpointState StringToBreakpointState(string addr)
...
+            basename = process.EscapePath(basename);
+            BindResult bindResults = EvalBindResult(await process.MICommandFactory.BreakInsert(basename, line, condition, ResultClass.None), pbreak);
+
+            // On GDB, the returned line information is from the pending breakpoint instead of the bound breakpoint.
+            // Check the address mapping to make sure the line info is correct.
+            if (process.MICommandFactory.Mode == MIMode.Gdb)
+            {
+                foreach (var boundBreakpoint in bindResults.BoundBreakpoints)
+                {
+                    boundBreakpoint.Line = await process.LineForStartAddress(basename, boundBreakpoint.Addr);
```
jacdavis: need to do the same thing for end address?

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
```diff
@@ -325,5 +338,25 @@ internal AD7DocumentContext DocumentContext(AD7Engine engine)
 
             return new AD7DocumentContext(_textPosition, new AD7MemoryAddress(engine, Addr, this.FunctionName));
         }
+
+        /// <summary>
+        /// Returns the start line of the breakpoint.
+        /// NOTE: If set this overwrites any column or multiline information.
```
jacdavis: what is the impact of this statement?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1422,6 +1422,32 @@ public ReadOnlyCollection<RegisterGroup> GetRegisterGroups()
...
+            if (srcLines == null || srcLines.Length == 0)
+            {
+                srcLines = await SourceLineCache.GetLinesForFile(System.IO.Path.GetFileName(file));
+            }
+            if (srcLines == null || srcLines.Length == 0)
+            {
+                return 0;
+            }
+
+            foreach (var l in srcLines)
```
jacdavis: there can be a *LOT* of lines in a source file. a dictionary by line number would be better as you could do lookup in constant time.

src/MIDebugEngine/Engine.Impl/MITextPosition.cs
```diff
@@ -25,6 +25,13 @@ public MITextPosition(string filename, TEXT_POSITION beginPosition, TEXT_POSITIO
             this.EndPosition = endPosition;
         }
 
+        public MITextPosition(string filename, uint line)
+        {
+            this.FileName = filename;
+            this.BeginPosition = new Microsoft.VisualStudio.Debugger.Interop.TEXT_POSITION() { dwLine = line - 1 };
+            this.EndPosition = this.BeginPosition;
```
jacdavis: why no real end position?

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
richardstanton: Unfortunately we only have a single address to work with.

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
richardstanton: Potentially the break-insert could have provided a range of information, however in practice I haven't seen it happen for gdb.

src/MIDebugEngine/Engine.Impl/MITextPosition.cs
richardstanton: Unfortunately we don't have that info

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
```diff
@@ -96,7 +96,20 @@ private static MIBreakpointState StringToBreakpointState(string addr)
...
             string basename = System.IO.Path.GetFileName(documentName);     // get basename from Windows path
-            return EvalBindResult(await process.MICommandFactory.BreakInsert(process.EscapePath(basename), line, condition, ResultClass.None), pbreak);
+            basename = process.EscapePath(basename);
+            BindResult bindResults = EvalBindResult(await process.MICommandFactory.BreakInsert(basename, line, condition, ResultClass.None), pbreak);
+
+            // On GDB, the returned line information is from the pending breakpoint instead of the bound breakpoint.
+            // Check the address mapping to make sure the line info is correct.
+            if (process.MICommandFactory.Mode == MIMode.Gdb)
+            {
+                foreach (var boundBreakpoint in bindResults.BoundBreakpoints)
```
gregg-miskelly: I think this collection will be null if the error message constructor is used (see line 58)

richardstanton: I will add a null check.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1422,29 +1422,53 @@ public ReadOnlyCollection<RegisterGroup> GetRegisterGroups()
...
+            {
+                return srcLine.Line;
+            }
+            return 0;
+        }
+
         public async Task<List<ulong>> StartAddressesForLine(string file, uint line)
         {
             List<ulong> addresses = new List<ulong>();
             var srcLines = await SourceLineCache.GetLinesForFile(file);
```
gregg-miskelly: nit: code is easier to read if it uses the real type (SourceLineMap in this case) instead of 'var' in cases where the type isn't obvious. Especially since it is used both here and on line 1460.

gregg-miskelly:
Otherwise looks good to me.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
richardstanton: I agree. I was trying to not use it in the code I wrote, but left it in existing code. I can be more aggressive about changing it in related code.
I still use var in foreach's statements though, I think I've found a few cases in foreach when it implicitly tries to cast to the type specified resulting in a runtime error instead of a compile error (I think they added this in the .NET 1.0 days to make working with collection easier).

PR #253, opened 2016-04-04
Use the new location of gdb.exe (multi-arch) and parse source.properties

edumunoz:
This PR has fixes to work with the new path to gdb in NDK  >= r11.

Also note that the RELEASE.txt file that used to be in the root of the NDK has been replaced by a source.properties file.

### DIFF
--- a/src/AndroidDebugLauncher/InstallPaths.cs
+++ b/src/AndroidDebugLauncher/InstallPaths.cs
@@ -45,38 +45,57 @@ namespace AndroidDebugLauncher
                 ndkRoot = GetDirectoryFromRegistry(RegistryRoot.Value + @"\Setup\VS\SecondaryInstaller\AndroidNDK", "NDK_HOME", checkBothBitnesses: false, externalProductName: LauncherResources.ProductName_NDK);
             }
 
+            NdkReleaseId ndkReleaseId = new NdkReleaseId();
             string ndkReleaseVersionFile = Path.Combine(ndkRoot, "RELEASE.TXT");
-            if (!File.Exists(ndkReleaseVersionFile))
+            string ndkSourcePropertiesFile = Path.Combine(ndkRoot, "source.properties");
+
+            // NDK releases >= r11 have a source.properties file
+            if (File.Exists(ndkSourcePropertiesFile))
+            {
+                NdkReleaseId.TryParsePropertiesFile(ndkSourcePropertiesFile, out ndkReleaseId);
+            }
+            // NDK releases < r11 have a RELEASE.txt file
+            else if (File.Exists(ndkReleaseVersionFile))
+            {
+                NdkReleaseId.TryParseFile(ndkReleaseVersionFile, out ndkReleaseId);
+            }
+            else
             {
                 ThrowExternalFileNotFoundException(ndkReleaseVersionFile, LauncherResources.ProductName_NDK);
             }
 
-            NdkReleaseId ndkReleaseId;
-            NdkReleaseId.TryParseFile(ndkReleaseVersionFile, out ndkReleaseId);
             logger.WriteLine("Using NDK '{0}' from path '{1}'", ndkReleaseId, ndkRoot);
 
+            // 32 vs 64-bit doesn't matter when comparing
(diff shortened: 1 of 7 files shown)


src/AndroidDebugLauncher/NdkReleaseId.cs
```diff
@@ -189,7 +272,7 @@ public override string ToString()
                 }
 
                 object subRelease = string.Empty;
-                if (this.SubRelease != (char)0)
+                if (this.SubRelease != (char)0 && this.SubRelease != 'a')
```
edumunoz: NDK versioning never has a 'a'. The character for the subversion is only included if it's not 'a'.

src/MICoreUnitTests/NdkVersionTests.cs
```diff
@@ -52,14 +52,14 @@ public void TestNdkToolVersion2()
...
-        public void TestNdkReleaseId1()
+        public void TestNdkReleaseIdReleaseFile()
         {
             NdkReleaseId r;
             Assert.True(NdkReleaseId.TryParse("r1", out r));
             Assert.Equal(r.ToString(), "r1");
 
             Assert.True(NdkReleaseId.TryParse("r1a", out r));
-            Assert.Equal(r.ToString(), "r1a");
+            Assert.Equal(r.ToString(), "r1");
```
edumunoz: This is the corresponding change in the tests to get rid of 'a'.

BTW all these tests have the arguments passed to Assert.Equal the wrong way (it's (expected, actual)). I didn't want to introduce more changes. I can change them if you all agree.

src/AndroidDebugLauncher/NdkReleaseId.cs
```diff
@@ -49,6 +49,89 @@ public bool IsValid
...
+            char subReleaseChar = 'a';
+            int subReleaseNumber;
+            if (!int.TryParse(subReleaseString, out subReleaseNumber))
+            {
+                return false;
+            }
+
+            subReleaseChar += Convert.ToChar(subReleaseNumber);
+
+            result = new NdkReleaseId(releaseNumber, subReleaseChar, true);
```
edumunoz: The new source.properties file doesn't have any indication of whether it's a 32 or 64-bit NDK.

As far as I can tell, the `Is32Bit` property is only used in the `toString()` method of `NdkReleaseId`.

Let me know if knowing 32 vs 64 is needed, and then we can look into exploring the directory structure of the NDK to figure it out.

edumunoz:
cc: @gregg-miskelly @chuckries @jacdavis

src/AndroidDebugLauncher/InstallPaths.cs
```diff
@@ -45,38 +45,57 @@ public static InstallPaths Resolve(CancellationToken token, AndroidLaunchOptions
...
             string ndkReleaseVersionFile = Path.Combine(ndkRoot, "RELEASE.TXT");
-            if (!File.Exists(ndkReleaseVersionFile))
+            string ndkSourcePropertiesFile = Path.Combine(ndkRoot, "source.properties");
+
+            // NDK releases < r11 have a RELEASE.txt file
+            if (File.Exists(ndkReleaseVersionFile))
+            {
+                NdkReleaseId.TryParseFile(ndkReleaseVersionFile, out ndkReleaseId);
+            }
+            // NDK releases >= r11 have a source.properties file
```
gregg-miskelly: Might want to check for the new one first in case someone has a crufty install that left the old file around.

src/AndroidDebugLauncher/NdkReleaseId.cs
gregg-miskelly: Yeah, looks like it is barely used. You could probably remove it if you want...

gregg-miskelly:
:+1:

src/AndroidDebugLauncher/InstallPaths.cs
edumunoz: Fair enough. Will do. Although all hell will break lose if ndk's are combined.

PR #254, opened 2016-04-04
Update projects to support Microbuild v2

pieandcakes:
@jacdavis @gregg-miskelly @wiktork @chuckries 

Adding lines to support MicroBuild v2

### DIFF
--- a/src/DebugEngineHost.Stub/project.json
+++ b/src/DebugEngineHost.Stub/project.json
@@ -4,6 +4,7 @@
     "dnxcore50.app": {}
   },
   "dependencies": {
+    "MicroBuild.Core": "0.2.0",
     "Microsoft.NETCore": "5.0.0",
     "Microsoft.NETCore.Portable.Compatibility": "1.0.0",
     "Microsoft.VisualStudio.Debugger.Interop.Portable": "1.0.1"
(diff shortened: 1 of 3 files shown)


src/DebugEngineHost.Stub/project.json
```diff
@@ -4,6 +4,7 @@
     "dnxcore50.app": {}
   },
   "dependencies": {
+    "MicroBuild.Core": "0.2.0",
```
wiktork: You'll also need FilesToSign in each proj file.

gregg-miskelly:
:+1:

chuckries:
:+1:

src/DebugEngineHost.Stub/project.json
pieandcakes: @wiktork Are you talking about in the project.json or in the .proj file?

wiktork: So it looks like FilesToSign is already part of DropFiles.targets, (which is included in MIDebugPackage.csproj). However, the final binaries do not appear to be signed.

gregg-miskelly: My guess: This adds MicroBuild.Core dependency to three projects that probably don't actually need the dependency at all, and doesn't add it to the project that actually needs the dependency (MIDebugPackage)

pieandcakes:
I will follow up with Fangliang and Wiktor. Cancelling this pull request for now.

PR #255, opened 2016-04-04
Fix issue that signal based exceptions in lldb were not working. 

jacdavis:
This isbecause a hack in miengine that adds frame info to the results class
wasn't accounting for signal-received events resulting in engine assuming
this is a module load event. Note that this hack (assuming any event
without a frame is a mod-load) seems wrong but removing it would be a bit
risky. It may be better to account for mod loads by directly checking for
them. Also, the existing hack no longer works as results.add returns a new 
results class. Fixed that as well.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -190,16 +190,17 @@ namespace MICore
             }
 
             //if this is an exception reported from LLDB, it will not currently contain a frame object in the MI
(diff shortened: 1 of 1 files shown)


jacdavis:
@gregg-miskelly @chuckries @paulmaybee

chuckries:
:+1:

gregg-miskelly:
:+1:

PR #256, opened 2016-04-05
Update GetClrDbg.sh to dynamically generate a runtime id

chuckries:
Tested on ubuntu.14.04-x64 and osx.10.10-x64

### DIFF
--- a/scripts/GetClrDbg.sh
+++ b/scripts/GetClrDbg.sh
@@ -12,15 +12,54 @@ print_help()
     echo '<version> can be "latest" or a version number such as 14.0.25109-preview-2865786'
 }
 
+# Sets __RuntimeID as given by 'dotnet --info'
+get_dotnet_runtime_id()
+{
+    # example output of dotnet --info
+    # .NET Command Line Tools (1.0.0-beta-002194)
+    #
+    # Product Information:
+    #  Version:     1.0.0-beta-002194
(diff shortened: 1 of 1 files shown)


chuckries:
@gregg-miskelly

scripts/GetClrDbg.sh
```diff
@@ -95,8 +134,17 @@ fi
 
 # For the rest of this script we can assume the working directory is the install path
 
+echo 'Info: Determinig Runtime ID'
+__RuntimeID=
+get_dotnet_runtime_id
+if [ -z $__RuntimeID ]; then
+    echo "Error: Unable to determine dotnet Runtime ID"
```
gregg-miskelly: Should we indicate that a new version of .NET CLI is required?

gregg-miskelly:
Otherwise LGTM

scripts/GetClrDbg.sh
chuckries: Sure. What is the lowest version of .net cli we need?

gregg-miskelly: I believe it would be: 1.0.0-beta-002035

gregg-miskelly: Thinking about it more: debugging probably doesn't work in 2035, so I would suggest using 1.0.0-beta-002173

PR #257, opened 2016-04-05
Enable attach in lldb

jacdavis:
Enable attach in lldb
Attach wasn't working with lldb on linux because lldb ends the
attach process in break mode whereas gdb ends in run mode. Issue
a continue after the attach in lldb to make the behavior match
gdb.

NOTE: if externalConsole is not enabled, and you issue an
attach, the elevation dialog will not be displayed. This will
need to be resolved at some point.


jacdavis:
@gregg-miskelly @chuckries @paulmaybee

gregg-miskelly:
👍

chuckries:
:+1:

PR #259, opened 2016-04-06
When the underlying debugger supports it, automatically attach to chi…

paulmaybee:
…ld processes created by fork and vfork.


paulmaybee:
@jacdavis @gregg-miskelly

src/DebugEngineHost/HostDebugger.cs
```diff
@@ -0,0 +1,47 @@
...
+    public static class HostDebugger
+    {
+        /// <summary>
+        /// Attach to a process using the provided options
+        /// </summary>
+        public static void Attach(string filePath, string options, Guid engineId)
+        {
+            try
+            {
+                ThreadHelper.Generic.BeginInvoke(() => Internal.LaunchDebugTarget(filePath, options, engineId));
```
jacdavis: Is it intentional that this is async and returns immediately? Are all the callers prepared for immediate return?

src/DebugEngineHost/HostMarshal.cs
```diff
@@ -68,5 +68,18 @@ public static IDebugEventCallback2 GetThreadSafeEventCallback(IDebugEventCallbac
         {
             return new VSImpl.VSEventCallbackWrapper(ad7Callback);
         }
+
+        /// <summary>
+        /// Allocate storage and copy the guid to the allocated bytes
+        /// </summary>
+        /// <param name="guid">guid to copy</param>
+        /// <returns>pointer to the allocated bytes</returns>
+        public static IntPtr AllocateGuid(Guid guid)
```
jacdavis: Is this still needed? I don't see a call to it anywhere while I was trying to verify it was freed.

src/DebugEngineHost/HostDebugger.cs
paulmaybee: Yes, intentional. Callers do not need to wait for this request to happen.

src/DebugEngineHost/HostMarshal.cs
paulmaybee: No not referenced any more. I will remove it.

src/MICore/Debugger.cs
```diff
@@ -1154,6 +1163,27 @@ private void HandleThreadGroupStarted(Results results)
...
+            foreach (var grp in _debuggeePids)
+            {
+                if ( grp.Value == pid)
+                {
+                    // Inferior names are of the form "iX" where X in the inferior number
+                    string name = grp.Key;
+                    if (name[0] == 'i')
+                    {
+                        int id;
+                        if (Int32.TryParse(name.Substring(1), out id))
```
jacdavis: Is signed int the correct type here? I've fixed a few signed vs unsigned ints in the miengine recently.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -824,10 +831,13 @@ public int ExecuteOnThread(IDebugThread2 pThread)
         // that is, not all threads should be required to be stopped before this method returns. The implementation of this method may be 
         // as simple as calling the IDebugProgram2::CauseBreak method on this program.
         //
-        // The sample engine only supports debugging native applications and therefore only has one program per-process
         public int Stop()
         {
-            throw new NotImplementedException();
+            DebuggedProcess.WorkerThread.RunOperation(async () =>
+            {
+                await _debuggedProcess.CmdBreak();
```
jacdavis: There is a known bug in during attach that CmdBreak doesn't currently wait until the signal is actually received by the debuggee (it returns as soon as the signal is sent but not before it is received). This await will therefore return too early if the original debuggee was attached to rather than launched. Is child process debugging support for attach on Linux?

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
```diff
@@ -0,0 +1,260 @@
...
+                if (inf != 0)
+                {
+                    await _process.ConsoleCmdAsync("inferior " + inf.ToString());
+                    if (!string.IsNullOrEmpty(_mainBreak))
+                    {
+                        await _process.MICommandFactory.BreakDelete(_mainBreak);
+                        _mainBreak = null;
+                    }
+                    _state = State.AtSignal;
+                    await _process.MICommandFactory.Signal("SIGSTOP");  // stop the child
```
jacdavis: does this call actually wait for the signal to be received or does it return as soon as the signal is sent?

paulmaybee: It does not wait for the signal to be received. The behavior of signal is identical to issuing a continue, except that is continues after delivering the named signal.

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
```diff
@@ -0,0 +1,260 @@
...
+            return true;   // parent is running
+        }
+
+        public async Task Enable(bool enable)
+        {
+            if (enable && _state == State.Init)
+            {
+                await _process.MICommandFactory.Set("detach-on-fork", "off");
+                await _process.MICommandFactory.Catch("fork");
+                await _process.MICommandFactory.Catch("vfork");
```
jacdavis: Need to handle Clone?

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
```diff
@@ -0,0 +1,260 @@
...
+                    await _process.MICommandFactory.BreakDelete(_vforkBp);
+                    _vforkBp = null;
+                }
+                _state = State.Init;
+            }
+        }
+
+        public async Task SetBreakAtMain()
+        {
+            Results results = await _process.MICommandFactory.BreakInsert("main", null);
```
jacdavis: cl.exe has an option to change the main function for a process. I wonder if gcc/clang also support this (minor issue, just curious)

src/MICore/CommandFactories/gdb.cs
```diff
@@ -194,5 +199,16 @@ public override Task EnableTargetAsyncOption()
             // that isn't actually supported by gdb. 
             await _debugger.CmdAsync("kill", ResultClass.None);
         }
+        public override async Task Signal(string sig)
+        {
+            string command = String.Format("-interpreter-exec console \"signal {0}\"", sig);
+            await _debugger.CmdAsync(command, ResultClass.running);
```
gregg-miskelly: I am assuming you didn't use ConsoleCmdAsync because you are calling this from run mode? My one concern is that we are going to have a tricky time understanding output from the signal command since we will not be able to tell what it is from. Maybe that is okay if there is no output.

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -251,6 +251,13 @@ public virtual void DefineCurrentThread(int threadId)
             await _debugger.CmdAsync(command, ResultClass.running);
         }
 
+        public abstract Task Signal(string sig);
```
gregg-miskelly: Can we add an XML doc comment indicating what this does and the fact that it is optional?

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -502,6 +511,15 @@ public virtual void DecodeExceptionReceivedProperties(Results miExceptionResult,
 
         #region Helpers
 
+        public virtual async Task<Results> Set(string variable, string value, ResultClass resultClass = ResultClass.done)
```
gregg-miskelly: How about 'SetOption' instead of 'Set'

src/MICore/CommandFactories/MICommandFactory.cs
paulmaybee: Done

src/MICore/CommandFactories/MICommandFactory.cs
paulmaybee: Done

src/MICore/CommandFactories/gdb.cs
paulmaybee: I didn't use ConsoleCmdAsync because I couldn't give it a ResultClass.Running parameter (it expects Done). Signal will only be called when the process is stopped, so it might be better to add an optional second ResultClass parameter to ConsoleCmdAsync. However in this case there is no other output except the ResultClass to look for.

src/MICore/Debugger.cs
paulmaybee: fixed

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
paulmaybee: I am not clear what the problem is for the attach case, but child process debugging should work for attached processes too.

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
paulmaybee: For now no support for clone. I cannot find any gdb support for catching the clone operation.

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
paulmaybee: Yeah, this could be an issue. I cannot find any way to discover the entry point name for a process. The gdb command 'start' seems to be independent of the name of the entry point, but it cannot be used in this ('exec') context.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
jacdavis: The problem is that for the attach case, we send the debuggee a signal and the api to send the signal (kill) returns before the debuggee actually processes the signal. There is a known bug because of that since the await will return early. There are no scenarios were really depending on the timing, but this one might actually need to be sure the debuggee is stopped

paulmaybee: I don't believe we depend on the timing in any way for the child processing case. As far as I can tell the debugger makes no assumption about process state based on this call. It would only be if the UI were to make some assumptions. But the ui seems to wait for notification from the separate processes correctly.

src/MICore/LaunchOptions.xsd
```diff
@@ -399,6 +389,23 @@
...
+        <xs:documentation>Path to a core dump file for the specified executable.</xs:documentation>
+      </xs:annotation>
+    </xs:attribute>
+    <xs:attribute name="DebugChildProcesses" type="xs:boolean" use="optional">
+      <xs:annotation>
+        <xs:documentation>
+          If true, automatically attach debugger to any child processes created.
+        </xs:documentation>
+      </xs:annotation>
+    </xs:attribute>
```
gregg-miskelly: This is going to wind up applying to Android and iOS also. Should this really be just pipe/tcp/local?

src/MICore/LaunchOptions.cs
```diff
@@ -239,9 +231,22 @@ static internal LocalLaunchOptions CreateFromXml(Xml.LaunchOptions.LocalLaunchOp
...
+                return base.ExePath;
+            }
+            set
+            {
+                if (String.IsNullOrEmpty(value) || !LocalLaunchOptions.CheckPath(value))
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, MICoreResources.Error_InvalidLocalExePath, value));
+
+                base.ExePath = value;
+            }
+        }
```
gregg-miskelly: I disabled this code because the file path can be an exe on the PATH. You don't want to bring it back.

paulmaybee: On Fri, Apr 8, 2016 at 6:15 PM -0700, "Gregg Miskelly" <notifications@github.com<mailto:notifications@github.com>> wrote:


In src/MICore/LaunchOptions.cs<https://github.com/Microsoft/MIEngine/pull/259#discussion_r59107215>:

>          /// </summary>
> -        public int ProcessId { get; private set; }
> +        public override string ExePath
> +        {
> +            get
> +            {
> +                return base.ExePath;
> +            }
> +            set
> +            {
> +                if (String.IsNullOrEmpty(value) || !LocalLaunchOptions.CheckPath(value))
> +                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, MICoreResources.Error_InvalidLocalExePath, value));
> +
> +                base.ExePath = value;
> +            }
> +        }


I disabled this code because the file path can be an exe on the PATH. You don't want to bring it back.

src/MICore/LaunchOptions.cs
```diff
@@ -360,6 +346,12 @@ public abstract class LaunchOptions
         public MIMode DebuggerMIMode { get; set; }
 
         private string _exePath;
+
+        /// <summary>
+        /// Hold on to options in serializable form to support child process debugging
+        /// </summary>
+        public Xml.LaunchOptions.BaseLaunchOptions BaseOptions { get; set; }
```
gregg-miskelly: This shouldn't be publicly settable at any time. If it needs to be publicly settable, we should use VerifyCanModifyProperty

src/MICore/LaunchOptions.cs
```diff
@@ -391,6 +383,32 @@ public string ExeArguments
...
+        /// [Optional] Path to a core dump file for the specified executable.
+        /// </summary>
+        public string CoreDumpPath
+        {
+            get
+            {
+                return _coreDumpPath;
+            }
+            set
+            {
```
gregg-miskelly: Use VerifyCanModifyProperty

src/MICore/LaunchOptions.cs
```diff
@@ -391,6 +383,32 @@ public string ExeArguments
             }
         }
 
+        /// <summary>
+        /// [Optional] If supplied, the debugger will attach to the process rather than launching a new one. Note that some operating systems will require admin rights to do this.
+        /// </summary>
+        public int ProcessId { get; set; }
```
gregg-miskelly: Use VerifyCanModifyProperty

src/MICore/LaunchOptions.cs
```diff
@@ -531,6 +549,44 @@ public LaunchCompleteCommand LaunchCompleteCommand
             }
         }
 
+        public bool DebugChildProcesses { get; set; }
```
gregg-miskelly: Same

src/MICore/LaunchOptions.cs
```diff
@@ -710,6 +769,22 @@ public static object Deserialize(XmlSerializer serializer, XmlReader reader)
             }
         }
 
+        public static void Serialize(XmlSerializer serializer, XmlWriter writer, object o)
```
gregg-miskelly: Should this be public?

src/MICore/LaunchOptions.cs
```diff
@@ -531,6 +549,44 @@ public LaunchCompleteCommand LaunchCompleteCommand
             }
         }
 
+        public bool DebugChildProcesses { get; set; }
+
+        public static string GetOptionsString(object o)
```
gregg-miskelly: Would this make more sense as an instance method on LaunchOptions?

src/MICore/Debugger.cs
```diff
@@ -1154,6 +1163,27 @@ private void HandleThreadGroupStarted(Results results)
...
 
+        public int InferiorByPid(int pid)
+        {
+            foreach (var grp in _debuggeePids)
+            {
+                if ( grp.Value == pid)
+                {
+                    // Inferior names are of the form "iX" where X in the inferior number
+                    string name = grp.Key;
+                    if (name[0] == 'i')
```
gregg-miskelly: Can name be zero length?

src/MICore/CommandFactories/gdb.cs
gregg-miskelly: Do you mean the 'signal' GDB commands outputs nothing? If so, sounds okay I suppose.

---
In reply to: [58890158](https://github.com/Microsoft/MIEngine/pull/259#discussion_r58890158) [](ancestors = 58890158)

src/MICore/Transports/LocalLinuxTransport.cs
```diff
@@ -23,7 +23,7 @@ public class LocalLinuxTransport : StreamTransport
         private string _gdbStdOutName;
         private FileSystemWatcher _fifoWatcher;
 
-        private void MakeGdbFifo(string path)
+        public static void MakeGdbFifo(string path, Logger logger)
```
gregg-miskelly: Maybe I will figure this out on my own in a moment, but why public?

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -56,6 +56,13 @@ sealed public class AD7Engine : IDebugEngine2, IDebugEngineLaunch2, IDebugProgra
 
         private IDebugSettingsCallback110 _settingsCallback;
 
+        public static List<int> ChildProcessLaunch;
```
gregg-miskelly: Since there is no locking on List, I would probably suggest either adding an accessor method instead of making the collection public.

src/MICore/Debugger.cs
paulmaybee: added check for length >= 2

src/MICore/LaunchOptions.cs
paulmaybee: Hmmm, I must have pulled this in with the merge conflict.

src/MICore/LaunchOptions.cs
paulmaybee: changed to 'protected set'

src/MICore/LaunchOptions.cs
paulmaybee: changed to 'protected set'

src/MICore/LaunchOptions.cs
paulmaybee: changed to 'protected set', added verify

src/MICore/LaunchOptions.cs
paulmaybee: changed to 'protected set', added verify

src/MICore/LaunchOptions.cs
paulmaybee: done

src/MICore/LaunchOptions.cs
paulmaybee: fixed

src/MICore/LaunchOptions.xsd
paulmaybee: Added an ExtendedBaseLaunchOptions as new base class for pipe/tcp/local

src/MICore/Transports/LocalLinuxTransport.cs
paulmaybee: This was part of the work for another fix related to Mac debugging. I'll remove it from this checkin.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -824,10 +831,13 @@ public int ExecuteOnThread(IDebugThread2 pThread)
...
         //
-        // The sample engine only supports debugging native applications and therefore only has one program per-process
         public int Stop()
         {
-            throw new NotImplementedException();
+            DebuggedProcess.WorkerThread.RunOperation(async () =>
+            {
+                await _debuggedProcess.CmdBreak();
+            });
+            return Constants.S_OK;
```
gregg-miskelly: S_OK says that we are already stopped. Generally we should be returning S_ASYNC_STOP. A complete implementation of this stuff is complicated. I would probably suggest adding a --
// TODO: this should be returning S_ASYNC_STOP
for now and fixing this in a subsequent checkin.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -516,6 +523,14 @@ private List<LaunchCommand> GetInitializeCommands()
                 }
             }
 
+            if (MICommandFactory.SupportsChildProcessDebugging())
+            {
+                if (_launchOptions.DebugChildProcesses || true)
```
gregg-miskelly: Remove before checkin?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -516,6 +523,14 @@ private List<LaunchCommand> GetInitializeCommands()
...
+                {
+                    _inProgress = new DebugUnixChild(this, this._launchOptions);  // TODO: let the user enable/disable this functionality
```
gregg-miskelly: Should this be called something more like '_childProcessHandler'?

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
paulmaybee: done

src/MICore/CommandFactories/gdb.cs
```diff
@@ -33,6 +33,11 @@ public override bool SupportsStopOnDynamicLibLoad()
             return true;
         }
 
+        public override bool SupportsChildProcessDebugging()
+        {
+            return true;
```
gregg-miskelly: Should we do something to block this feature for local Windows debugging?

gregg-miskelly: Or does GDB support it there too?

---
In reply to: [59229655](https://github.com/Microsoft/MIEngine/pull/259#discussion_r59229655) [](ancestors = 59229655)

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
paulmaybee: done

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -793,7 +813,27 @@ private void Dispose()
                     {
                         code = EngineUtils.SignalMap.Instance[sigName];
                     }
-                    _callback.OnException(thread, sigName, results.Results.TryFindString("signal-meaning"), code);
+                    bool _stoppedAtSIGSTOP = false;
```
gregg-miskelly: This shouldn't have an underscore since it isn't a member.

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -213,13 +213,20 @@ private ThreadContext CreateContext(TupleValue frame)
                             {
                                 thread.TargetId = tid;
                             }
-                            else if (targetId.StartsWith("Thread") &&
+                            else if (targetId.StartsWith("Thread", StringComparison.OrdinalIgnoreCase) &&
```
gregg-miskelly: Should this be "Thread<SPACE>" instead of just "Thread"?

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
```diff
@@ -0,0 +1,260 @@
+﻿using System;
+using System.Collections.Generic;
```
gregg-miskelly: Missing copyright header

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
paulmaybee: done

src/DebugEngineHost/HostDebugger.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using Microsoft.VisualStudio.Shell.Interop;
```
gregg-miskelly: Missing copyright header

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
gregg-miskelly: If I am reading the code right, this will throw if 'main' can't be found. Expected?

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
```diff
@@ -0,0 +1,260 @@
...
+
+            switch (_state)
+            {
+                case State.Enabled:
+                    reason = results.TryFindString("reason");
+                    if (reason == "fork")
+                    {
+                        int threadId = results.FindInt("thread-id");
+                        _newpid = results.FindInt("newpid");
+                        _state = State.AtFork;
```
gregg-miskelly: Will we have races if multiple threads are forking at the same time?

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
```diff
@@ -0,0 +1,260 @@
...
+            await _process.MICommandFactory.TargetDetach();     // detach from the child
+            await _process.ConsoleCmdAsync("inferior 1");
+            lock(AD7Engine.ChildProcessLaunch)
+            {
+                AD7Engine.ChildProcessLaunch.Add(_newpid);
+            }
+            string engineName;
+            Guid engineGuid;
+            _process.Engine.GetEngineInfo(out engineName, out engineGuid);
+            _launchOptions.BaseOptions.ProcessId = _newpid;
```
gregg-miskelly: This seems like a race here. Seems like we need to clone the original launch options and change these.

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
```diff
@@ -0,0 +1,260 @@
...
+            _launchOptions.BaseOptions.ProcessIdSpecified = true;
+            //_process.Callbacks.OnDebugLaunchedProcessAttachRequest(_newpid, _process.Engine);
```
gregg-miskelly: Remove?

src/MICore/CommandFactories/gdb.cs
paulmaybee: gdb (under Cygwin anyway) does not catch the fork. Unfortunately it also doesn't report an error of any kind when you try (except in an output string after run is executed). I added this check
_launchOptions is LocalLaunchOptions && RuntimeInformation.IsOSPlatform(OSPlatform.Windows)

src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
```diff
@@ -279,6 +289,20 @@ public static void FindNatvisInSolution(NatvisLoader loader)
     }
 
     /// <summary>
+    /// Provides interactions with the host's debugger
+    /// </summary>
+    public static class HostDebugger
+    {
+        /// <summary>
+        /// Attach to a process using the provided options
```
gregg-miskelly: How about - asks the host to async spin up a new instance of the debug engine and go through the launch sequence using the specified options.

Do we want to also maybe call this 'StartDebugChildProcess' or something like that?

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
paulmaybee: done

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
paulmaybee: done

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
paulmaybee: done

src/DebugEngineHost/HostDebugger.cs
paulmaybee: done

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
```diff
@@ -0,0 +1,260 @@
...
+                    }
+                    return false;
+                case State.AtFork:
+                    await ProcessChild();
+                    break;
+                case State.AtVfork:
+                    if ("exec" == results.TryFindString("reason"))
+                    {
+                        await _process.MICommandFactory.Set("schedule-multiple", "off");
+                        await RunChildToMain();
```
gregg-miskelly: Could you add a comment here explaining why we need to run to main? I am curious, and can't quite figure it out from the code.

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
```diff
@@ -0,0 +1,260 @@
...
+                    {
+                        // sometimes gdb misses the breakpoint at exec and execution will proceed to a breakpoint in the child
+                        _process.Logger.WriteLine("Missed catching the exec after vfork. Spawning the child's debugger.");
+                        _state = State.AtExec;
+                        goto missedExec;
+                    }
+                    break;
+                case State.AtSignal:    // both child and parent are stopped
+                    return await DetachAndContinue();
+                case State.AtExec:
```
gregg-miskelly: If we are stopped in the parent process, couldn't this be a stopping event that we want to actually stop on? (ex: breakpoint set in the code that calls vfork() within the parent) Or would we no longer be in the 'AtExec' state if that happened?

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
paulmaybee: The process doesn't handle the SIGSTOP correctly (if just ignores it) when the process is at the start of program (after exec). No idea why. I had to let it run some code before it would correctly respond to the SIGSTOP. Perhaps there is an earlier stopping point that would work too, but "main" was a well known symbols that would work in most cases.

gregg-miskelly:
That is all the comments that I have...

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
paulmaybee: if we are at atExec state then only the child should be executing. I have occasionally seen the spurious signal 0 break in the parent in the interval, but no code being executed in the parent.

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
paulmaybee: done

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
gregg-miskelly: Is that because we have left the parent suspended? Or because vfork should suspend the parent until the child calls an exec function? If the later, I wasn't clear on if Linux would allow other threads in the process to run or not.

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
paulmaybee: Shouldn't be an issue. This field is only used for child forking. The code only ever runs on the worker thread and the field is cloned at the call to StartDebugChildProcess below.

src/MIDebugEngine/Engine.Impl/DebugUnixChildProcess.cs
paulmaybee: gdb never throws if the symbol is not found. it just creates an unbound breakpoint. In which case we will continue to execute the child until another breakpoint is hit, or the user pauses execution. I have added code to detach the child if the user interrupts execution.

src/MICore/LaunchOptions.xsd
paulmaybee: I don’t see any reference to  BaseLaunchOptions from either IOSLaunchOptions or AndroidLaunchOptions. How do these end up applying?

PR #262, opened 2016-04-08
Fix ARM with new multi-arch gdb.exe from Android NDK >= r11

edumunoz:
The new version of gdb included in the new Android NDK (>=r11) has a couple of changes that MIEngine didn't account for:

- `sysroot` (aka `solib-absolute-prefix`) is now being set to `target:` by default. Previous versions had it set to `""`, which this change explicitly does for Android debugging sessions.
- the default ABI is now being set to `Cygwin`. Previous versions had it set to `GNU/Linux`, which this change explicitly does for Android debugging sessions.

I tested these changes with NDK r11c and r10e, as well as ARM, ARM64 and x86.

@gregg-miskelly @chuckries @jacdavis


jacdavis:
LGTM. Please wait for Chuck's signoff

gregg-miskelly:
:+1:

chuckries:
:+1:

PR #263, opened 2016-04-08
fix glass tests

faxue-msft:
This is to improve the current MIEngine Glass test

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
@@ -455,7 +455,8 @@ namespace Microsoft.MIDebugEngine
         // pointer associated with the stack frame. The debugger calls EnumProperties to obtain these values in the sample.
         int IDebugStackFrame2.GetDebugProperty(out IDebugProperty2 property)
         {
-            throw new NotImplementedException();
+            property = null;
+            return Constants.E_NOTIMPL;
         }
 
         // Gets the document context for this stack frame. The debugger will call this when the current stack frame is changed
--- a/test/Android/Breakpoint/TestScript.xml
+++ b/test/Android/Breakpoint/TestScript.xml
@@ -111,7 +111,7 @@
     <Event name="IDebugThreadCreateEvent2" expected="False" />
     <Event name="IDebugOutputStringEvent2" expected="False" />
     <Event name="IDebugBreakpointBoundEvent2" expected="True">
-      <BoundBreakpoint>24,0 to 24,0 in [Unknown]!break_test_1()</BoundBreakpoint>
+      <BoundBreakpoint RegEx="^24,0 to 24,0 in (\[Unknown\]|libbreakpoint.so)!break_test_1\(\)$"></BoundBreakpoint>
     </Event>
     <Event name="IDebugModuleLoadEvent2" expected="False" />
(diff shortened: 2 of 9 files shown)


src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
```diff
@@ -455,7 +455,8 @@ int IDebugStackFrame2.GetCodeContext(out IDebugCodeContext2 memoryAddress)
         // pointer associated with the stack frame. The debugger calls EnumProperties to obtain these values in the sample.
         int IDebugStackFrame2.GetDebugProperty(out IDebugProperty2 property)
         {
-            throw new NotImplementedException();
+            property = null;
+            return Constants.E_NOTIMPL;
```
faxue-msft: Glass test will try to call this, if fail it Glass turns to IDebugStackFrame2.EnumProperties

test/Android/Breakpoint/TestScript.xml
```diff
@@ -111,7 +111,7 @@
     <Event name="IDebugThreadCreateEvent2" expected="False" />
     <Event name="IDebugOutputStringEvent2" expected="False" />
     <Event name="IDebugBreakpointBoundEvent2" expected="True">
-      <BoundBreakpoint>24,0 to 24,0 in [Unknown]!break_test_1()</BoundBreakpoint>
+      <BoundBreakpoint RegEx="24,0 to 24,0 in (\[Unknown\]|libbreakpoint.so)!break_test_1()"></BoundBreakpoint>
     </Event>
```
faxue-msft: There is a timing issue here, the breakpoint bound and module load events usually happen very close to each other inside worker thread. Therefore when Glass try to get the breakpoint information from the main thread, MIEngine may or may not finished dealing with the module load event.
This won't cause any issue in product, but in Glass the BoundBreakpoint may or may not have the right module name.

gregg-miskelly: Might want to use '^<your-pattern>$' so that you make sure the pattern only matches the whole string.
Might want to add your comment as a comment to the code.

test/Android/androidtest.cmd
```diff
@@ -123,10 +123,19 @@ if "%_Verbose%"=="1" (
 if "%_TestsToRun%"=="" goto RunAll
 goto RunArgs
 
+:FindDeviceId
+pushd %_SdkRoot%
+for /f "tokens=1,6" %%i in ('platform-tools\adb devices -l ^| find /i "VS Emulator"') do (
+	echo Run on device: %%i %%j
+    set _DeviceId=%%i
+	popd
```
gregg-miskelly: Would it make sense to only set the Platform to x86 if you are using this for loop?

gregg-miskelly:
Otherwise LGTM

test/Android/Breakpoint/TestScript.xml
chuckries: is there any way to fix the race condition rather than have a regex in all these tests? As more tests will be added this seems like it will be annoying to maintain as the Regex will never be part of the session log.

chuckries:
:+1:  with one comment.

test/Android/Breakpoint/TestScript.xml
faxue-msft: There isn't an easy way to know whether the breakpoint bound is because of module load, so we don't now whether should we wait for it.

PR #267, opened 2016-04-09
Copy gdbserver ourselves if gdbserver can't be found

edumunoz:
This fixes debugging 64-bit native APKs. The previous solution wasn't
completely correct and required `root` on the device/emulator.

On Android, the files of an application are extracted to the system owned
directory `/data/app/<appname>-<int>/lib`

For 32-bit shared libraries, Android will create a symlink from the app owned
directory `/data/data/<appname>/lib` to maintain backwards compatibility with
previous API versions.

There is no such symlink for 64-bit libraries, since Android only supports
64-bit since API 21, at which point the app owned directory was no longer used
for libs.

@chuckries @gregg-miskelly @jacdavis


src/AndroidDebugLauncher/InstallPaths.cs
```diff
@@ -121,6 +117,14 @@ out matchedPath
                 }
             }
 
+            IEnumerable<INDKFilePath> gdbServerPath = NDKPrebuiltFilePath.GDBServerPaths(targetArchitectureName);
+            result.GDBServerPath = GetNDKFilePath(
+                string.Concat("Android-", targetArchitectureName, "-GDBServerPath"),
```
edumunoz: I couldn't find where this registry key is being set for `Android-<arch>-GDBPath`. I added the `Android-<arch>-GDBServerPath` query here, but I doubt this is being set anywhere.

src/AndroidDebugLauncher/Launcher.cs
```diff
@@ -556,37 +556,63 @@ private void VerifySdkVersion()
...
+                // Can't use cp since shell doesn't have permissions to /data/data/<app>
+                // and the app uid doesn't have permissions to /data/local/tmp
+                string copyGDBServerCommand = string.Format(CultureInfo.InvariantCulture,
+                    "cat {0} | run-as {1} /system/bin/sh -c \"cat > {2}\"", tmpGDBServer, _launchOptions.Package, dest);
+                int rc = ExecCommand(copyGDBServerCommand, out output);
+                RunAsOutputParser.ThrowIfRunAsErrors(output, _launchOptions.Package);
+                if (rc != 0)
+                {
+                    throw new LauncherException(
+                        Telemetry.LaunchFailureCode.NoGdbServer,
```
edumunoz: Here and in line 608 I used the `NoGdbServer` telemetry event since we only go down this code path if gdbserver can't be found. Let me know if I should create a different telemetry event.

src/AndroidDebugLauncher/InstallPaths.cs
```diff
@@ -121,6 +117,14 @@ out matchedPath
...
+                ndkRoot,
+                gdbServerPath,
+                out matchedPath
+                );
```
jacdavis: why do path outputs? What's the difference between result.GDBServerPath and out matchedPath?

edumunoz:
FYI the CI build doesn't seem to be checking unit tests anymore
http://dotnet-ci.cloudapp.net/job/microsoft_miengine_prtest/361/console

src/AndroidDebugLauncher/StringExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.IO;
+
+namespace AndroidDebugLauncher
+{
+    internal static class StringExtensions
+    {
+        public static IEnumerable<string> GetLines(this string content)
```
jacdavis: Why not just call Split on the strings?

src/AndroidDebugLauncher/InstallPaths.cs
edumunoz: This is used when finding a possible gdb.exe. There are some if-statements using it.
It's not used in the case of gdbserver.

I can add a comment `// not used`.

src/AndroidDebugLauncher/Telemetry.cs
```diff
@@ -40,6 +40,11 @@ public enum LaunchFailureCode
             RunAsFailure,
 
             /// <summary>
+            /// run-as returned 'Package 'bla' is not debuggable'
```
jacdavis: what is bla? be careful with works like blah and bla as they translate into swear words in russian

src/AndroidDebugLauncher/Launcher.cs
```diff
@@ -556,37 +556,63 @@ private void VerifySdkVersion()
...
-                //start looking other places, only do this on 64 bit
-                //TODO: This needs some additional testsing
-                lsCommand = string.Format(CultureInfo.InvariantCulture, "ls /data/app/{0}*/lib/x86_64/gdbserver", _launchOptions.Package);
-                output = ExecCommand(lsCommand);
-                return output;
-            }
-            else
-            {
-                throw new LauncherException(Telemetry.LaunchFailureCode.NoGdbServer, string.Format(CultureInfo.CurrentCulture, LauncherResources.Error_NoGdbServer, gdbServerPath));
+            if (string.Compare(output, gdbServerPath, StringComparison.OrdinalIgnoreCase) != 0)
```
jacdavis: is this path relative to the windows machine or to android? If android, probably don't want ignore case.

src/AndroidDebugLauncher/Telemetry.cs
edumunoz: I just used the same bogus example as RunAsPackageUnknown that was already there.

src/AndroidDebugLauncher/Launcher.cs
edumunoz: Android.
I agree it should be a case sensitive comparison. This is the way it was already being done and it doesn't really matter in this case since it's testing the output of an `ls` command.

src/AndroidDebugLauncher/Telemetry.cs
jacdavis: We should change it then as we may open source this in the future. I suggest using Squid but I'm a weirdo

src/AndroidDebugLauncher/Launcher.cs
jacdavis: Case insensitive is slower and may lead to bugs someday.... I personally would rather we fix it rather than leaving it as-is.

src/AndroidDebugLauncher/StringExtensions.cs
edumunoz: Memory concerns. This function was part of `PwdOutputParser`.

src/AndroidDebugLauncher/Launcher.cs
edumunoz: Done.

src/AndroidDebugLauncher/StringExtensions.cs
gregg-miskelly: Because this handles the various line endings for us already.

---
In reply to: [59105891](https://github.com/Microsoft/MIEngine/pull/267#discussion_r59105891) [](ancestors = 59105891)

src/AndroidDebugLauncher/TargetArchitectureExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using System;
+using System.Diagnostics;
+using MICore;
+
+namespace AndroidDebugLauncher
+{
+    internal static class TargetArchitectureExtensions
+    {
+        public static string ToNDKPath(this TargetArchitecture arch)
```
gregg-miskelly: nit: This isn't really a path. ToNDKArchitectureName?

src/AndroidDebugLauncher/Launcher.cs
```diff
@@ -556,37 +556,63 @@ private void VerifySdkVersion()
...
-            else
-            {
-                throw new LauncherException(Telemetry.LaunchFailureCode.NoGdbServer, string.Format(CultureInfo.CurrentCulture, LauncherResources.Error_NoGdbServer, gdbServerPath));
+            if (string.Compare(output, gdbServerPath, StringComparison.Ordinal) != 0)
+            {
+                // <app data>/lib is not symlinked to system's app data directory (/data/app/<app package>/lib). 
+                // We copy gdbserver ourselves because of 2 reasons:
+                // 1. the app user doesn't have permissions on /data/app/<app package>-<suffix>/lib (it's owned by system)
+                // 2. we can't deterministically figure out the location where gdbserver is in /data/app because of the suffix,
+                //    which is used by Android to have multiple copies of the APK extracted in the system.
```
chuckries: are we sure that there's no way to determine the correct suffix and we have to copy gdbserver ourselves?

chuckries:
:+1:

src/AndroidDebugLauncher/Launcher.cs
edumunoz: Depending on how many times the app has been installed, Android keeps around the previous directories, but appends an integer suffix.

Neither shell nor the app's user have read access on `/data/app`, so it's not possible to see what directories exist. We could try iterating say for `indices <= 10`, and try accessing gdbserver in there, but it seems way more hacky than just copying gdbserver ourselves.

src/AndroidDebugLauncher/TargetArchitectureExtensions.cs
edumunoz: Thanks for the suggestion. Changed it.

src/AndroidDebugLauncher/Launcher.cs
chuckries: This seems fine to me.

src/AndroidDebugLauncher/Launcher.cs
chuckries: any way to stat and use the newest folder? This still might not be a good heuristic, I'm not sure how Android handles these directories exactly.

src/AndroidDebugLauncher/Launcher.cs
```diff
@@ -556,37 +556,63 @@ private void VerifySdkVersion()
...
+                string arch = _launchOptions.TargetArchitecture.ToNDKPath();
+                string tmpGDBServer = string.Format(CultureInfo.InvariantCulture, "/data/local/tmp/{0}-gdbserver", arch);
+
+                device.FileSystem.Upload(_installPaths.GDBServerPath, tmpGDBServer, true);
+
+                // Files can't be executed from /data/local/tmp
+                string dest = string.Format(CultureInfo.InvariantCulture, "{0}/{1}-gdbserver", workingDirectory, arch);
+
+                // Can't use cp since shell doesn't have permissions to /data/data/<app>
```
gregg-miskelly: BTW: This is a really clever solution. Smart.

src/AndroidDebugLauncher/Launcher.cs
```diff
@@ -764,6 +790,21 @@ private bool HasGdbServerInvalidSocketError(/*OPTIONAL*/ StringBuilder errorOutp
...
 
+        private int ExecCommand(string command, out string response)
+        {
+            const string resultCodeSeparator = "rc=";
+            string resultCodeCheck = string.Format(CultureInfo.InvariantCulture, "; echo {0}$?", resultCodeSeparator);
+
+            response = ExecCommand(string.Concat(command, resultCodeCheck));
+
+            int resultCodeIndex =
+                response.LastIndexOf(resultCodeSeparator, StringComparison.Ordinal) + resultCodeSeparator.Length;
```
gregg-miskelly: Handle the case that 'rc=' can't be found?

gregg-miskelly:
:+1:

src/AndroidDebugLauncher/Launcher.cs
edumunoz: We would still need to iterate through them. My only concern with this approach is these are system-owned directories, that can change anytime in the future and break this.

PR #268, opened 2016-04-11
Fix glass tests for NDK r11c and fix androidtest.cmd

edumunoz:
The new gdb version in the NDK r11c formats the locals slightly different.

androidtest.cmd wouldn't work properly running all tests if any flag was
specified but `/Tests` was not because when using `shift`, the 0th argument was
being discarded, and hence `%~dp0` was changed to a different value than the
script's directory.

@faxue-msft @gregg-miskelly @chuckries

### DIFF
--- a/test/Android/Locals/TestScript.xml
+++ b/test/Android/Locals/TestScript.xml
@@ -265,11 +265,11 @@
     <LocalType>DerivedClass *</LocalType>
     <LocalIsExpandable>True</LocalIsExpandable>
     <Local>escaped</Local>
-    <LocalValue RegEx="0x[a-f0-9]{8} &lt;.L.str1&gt; &quot;Hello\\n\tWorld!\\n&quot;">0xa1b7fc26 &lt;.L.str1&gt; "Hello\n	World!\n"</LocalValue>
+    <LocalValue RegEx="0x[a-f0-9]{8} &lt;.L.str.?1&gt; &quot;Hello\\n(\\t|\t)World!\\n&quot;">0xa1b7fc26 &lt;.L.str.1&gt; "Hello\n\tWorld!\n"</LocalValue>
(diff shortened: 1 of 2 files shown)


test/Android/Locals/TestScript.xml
```diff
@@ -265,11 +265,11 @@
     <LocalType>DerivedClass *</LocalType>
     <LocalIsExpandable>True</LocalIsExpandable>
     <Local>escaped</Local>
-    <LocalValue RegEx="0x[a-f0-9]{8} &lt;.L.str1&gt; &quot;Hello\\n\tWorld!\\n&quot;">0xa1b7fc26 &lt;.L.str1&gt; "Hello\n	World!\n"</LocalValue>
+    <LocalValue RegEx="0x[a-f0-9]{8} &lt;.L.str.1&gt; &quot;Hello\\n\\tWorld!\\n&quot;">0xa1b7fc26 &lt;.L.str.1&gt; "Hello\n\tWorld!\n"</LocalValue>
```
faxue-msft: Could you make it work on both Ndk version 11 and 10, since it's still possible that we can run the test cases on Ndk 10 by passing the right parameter?

gregg-miskelly: We might want to have the regex's accept either so that the test still passes with the old stuff too..

gregg-miskelly:
Otherwise LGTM.

test/Android/Locals/TestScript.xml
edumunoz: Sure. I thought we only wanted the new NDK since they are flaky with r10. I'll change it.

edumunoz:
@gregg-miskelly can you merge it if it looks good now? Thanks.

PR #274, opened 2016-04-15
Add nuspec file for MDD

edumunoz:
This is needed for a release with the new process.

This is only used for the internal release process, but I can change the `author`, `owners`, etc. to match that of `MIEngine.clrdbg.nuspec`.

@gregg-miskelly @chuckries


gregg-miskelly:
The build lab wants us to include metadata.json in our package at the root. Otherwise this looks good.

edumunoz:
I updated the PR so that `metadata.json` is included at the root of the NuGet package.

PR #276, opened 2016-04-15
Change frame "-1" to frame "0".

paulmaybee:
Keeps the MIEngine from crashing on bad input. Seems like a gdb bug. Only seen it happen with an otherwise empty stack.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
@@ -171,7 +171,15 @@ namespace Microsoft.MIDebugEngine
             ulong? pc = frame.TryFindAddr("addr");
             MITextPosition textPosition = MITextPosition.TryParse(frame);
             string func = frame.TryFindString("func");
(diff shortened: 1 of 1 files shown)


paulmaybee:
@edumunoz @jacdavis @gregg-miskelly

gregg-miskelly:
👍

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -171,7 +171,15 @@ private ThreadContext CreateContext(TupleValue frame)
             ulong? pc = frame.TryFindAddr("addr");
             MITextPosition textPosition = MITextPosition.TryParse(frame);
             string func = frame.TryFindString("func");
-            uint level = frame.FindUint("level");
+            uint level;
+            try
+            {
+                level = frame.FindUint("level");
```
jacdavis: Is there a tryfindunit that doesn't throw?

edumunoz:
Does debugging work as expected after we hit the `catch` codepath?

PR #280, opened 2016-04-20
Add a script to launch emulator and clean up error output

faxue-msft:
I plan to use this script for automation test run

### DIFF
--- a/test/Android/PrepareEmulator.ps1
+++ b/test/Android/PrepareEmulator.ps1
@@ -1,0 +1,82 @@
+[CmdletBinding()]
+Param(
+    [switch]$killRunningEmulator
+)
+
+$xdeProcess = Get-Process | where {$_ -like "*XDE*"}
+if ($xdeProcess.count -gt 0)
+{
+    if ($killRunningEmulator)
+    {
+        <# Kill emulator if already started, otherwise the hanging issue is more likely to repro. #>
+        $xdeProcess | Stop-Process
+    }
+    else 
+    {
+        Write-Output "Emulator already started"
+        exit 0
+    }
+}
+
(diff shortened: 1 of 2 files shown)


test/Android/PrepareEmulator.ps1
```diff
@@ -0,0 +1,86 @@
...
+{
+    Write-Error "Emulator manager is not installed at $emulatorPath"
+    exit 1
+}
+
+pushd $emulatorPath
+
+$deviceList = ./emulatorcmd list /sku:Android /type:device
+
+<# try to use Lollipop emulator if there is one #>
```
edumunoz: Just curious: why Lollipop?

faxue-msft: I try use the same version of emulator for every run, so we won't randomly use a different version of emulator for each run. And Lollipop arguably is the most used version of Android.

edumunoz: OK. Maybe in the future we *may* consider having the tests run on different Android APIs.

test/Android/PrepareEmulator.ps1
```diff
@@ -0,0 +1,86 @@
...
+
+<# launch emulator #>
+Write-Output "Starting device: $device"
+
+$deviceId = $device.Substring(0, $device.IndexOf(' '))
+start-process powershell -argumentlist "emulatorcmd.exe launch /sku:android /id:$deviceId"
+popd
+
+<# wait for emulator to be ready #>
+$adbPath = ${Env:\ProgramFiles(x86)} + '/Android/android-sdk/platform-tools'
```
edumunoz: Most likely not worth it (given this will run on a test machine we control), but just in case you weren't aware: there are registry keys with the path to the Android SDK.

Software\Android SDK Tools
Software\Android Studio

edumunoz:
:+1:

PR #284, opened 2016-04-20
Fixing CodeAnalysis and FXCop errors

pieandcakes:
@wiktork @andrewcrawley @gregg-miskelly @jacdavis @chuckries

### DIFF
--- a/src/MICore/LinuxNativeMethods.cs
+++ b/src/MICore/LinuxNativeMethods.cs
@@ -13,8 +13,8 @@ namespace MICore
         [DllImport(Libc, EntryPoint = "kill", SetLastError = true)]
         internal static extern int Kill(int pid, int mode);
 
-        [DllImport(Libc, EntryPoint = "mkfifo", SetLastError = true)]
-        internal static extern int MkFifo(string name, int mode);
+        [DllImport(Libc, EntryPoint = "mkfifo", SetLastError = true, ThrowOnUnmappableChar = true)]
+        internal static extern int MkFifo([MarshalAs(UnmanagedType.LPWStr)]string name, int mode);
 
         [DllImport(Libc, EntryPoint = "geteuid", SetLastError = true)]
         internal static extern uint GetEUid();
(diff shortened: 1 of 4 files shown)


src/MICore/LinuxNativeMethods.cs
```diff
@@ -13,8 +13,8 @@ internal class LinuxNativeMethods
         [DllImport(Libc, EntryPoint = "kill", SetLastError = true)]
         internal static extern int Kill(int pid, int mode);
 
-        [DllImport(Libc, EntryPoint = "mkfifo", SetLastError = true)]
-        internal static extern int MkFifo(string name, int mode);
+        [DllImport(Libc, EntryPoint = "mkfifo", SetLastError = true, ThrowOnUnmappableChar = true)]
+        internal static extern int MkFifo([MarshalAs(UnmanagedType.LPWStr)]string name, int mode);
```
jacdavis: Did you actually test this marshalling on Linux? The Unicode mappings there behave differently than windows. Not sure if this marshalas is correct. Furthermore, I think this fxcop error is actually bogus off windows since there aren't A and W versions of the functions. Is the ThrowOnUnmappableChar warning actually in the divisional min bar?

pieandcakes: I can suppress it instead of changing what the code actually does (which might be better since this is non-windows)

wiktork: Do we need a CharSet as part of the DllImport?

jacdavis: If it works on Linux and mac, then lets keep it. Better to be clean than suppressed. Just wanted to make sure it was actually tested with these changes on both platforms.

gregg-miskelly: I wonder if we should actually be declaring this as a byte[] so that we can manually marshal this to a UTF8 string

pieandcakes: I changed it over based on Gregg's suggestion

pieandcakes:
Tested new bits on published VSCode instance. Verified that the FIFOs are created and I am able to debug.

jacdavis:
LGTM

gregg-miskelly:
👍

PR #287, opened 2016-04-21
Add DropFile step to copy .mdb files out to the output 'bin' directory

pieandcakes:
@jacdavis @andrewcrawley @wiktork @chuckries @gregg-miskelly  - this change is to make sure we are getting the mdb files also in the out directory. 

Tested locally on Mono and non-Mono builds to make sure it didn't break.

### DIFF
--- a/src/MIDebugPackage/MIDebugPackage.csproj
+++ b/src/MIDebugPackage/MIDebugPackage.csproj
@@ -211,6 +211,10 @@
     <DropUnsignedFile Include="$(OutDir)\Microsoft.DebugEngineHost.pdb" />
     <DropUnsignedFile Include="@(RequiredNetFX46FacadeAssemblies)" />
   </ItemGroup>
+  <ItemGroup Condition="'$(IsCoreCLR)' == 'false'">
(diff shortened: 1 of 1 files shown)


jacdavis:
LGTM

src/MIDebugPackage/MIDebugPackage.csproj
```diff
@@ -211,6 +211,12 @@
     <DropUnsignedFile Include="$(OutDir)\Microsoft.DebugEngineHost.pdb" />
     <DropUnsignedFile Include="@(RequiredNetFX46FacadeAssemblies)" />
   </ItemGroup>
+  <Target Name="DropMDBFiles" Condition="'$(IsCoreCLR)'=='false'" BeforeTargets="DropFiles">
```
andrewcrawley: You don't actually need a separate target here - that's only necessary if you're checking the IsMonoRuntime variable, which is set by an earlier target.  Otherwise LGTM.

pieandcakes: Removed and replaced it with just an ItemGroup

andrewcrawley:
LGTM.

gregg-miskelly:
👍

PR #290, opened 2016-04-22
Disable detach when using clrdbg

gregg-miskelly:
CLRDBG doesn't yet support detach. This checkin changes the MIEngine so that when it is
talking to CLRDBG it will not report that detach is supported.


caslan:
LGTM

PR #296, opened 2016-04-25
DD206020 - Break all on CentOS not working

pieandcakes:
Found an issue when using out fifos that causes -exec-interrupt to return
Quit from gdb. Change is to use SIGINT 2 instead. when on Linux in all
local situations instead of just Attach.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -455,11 +455,10 @@ namespace MICore
         }
 
 
-        private bool IsLocalGdbAttach()
+        private bool IsLocalGdb()
         {
             if (this.MICommandFactory.Mode == MIMode.Gdb &&
                this._launchOptions is LocalLaunchOptions &&
-               ((LocalLaunchOptions)this._launchOptions).ProcessId != 0 &&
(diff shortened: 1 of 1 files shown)


pieandcakes:
@jacdavis @gregg-miskelly @chuckries @edumunoz

chuckries:
@paulmaybee

pieandcakes:
Verified launch and attach on CentOS and Ubuntu 14.04 work after fix.

src/MICore/Debugger.cs
```diff
@@ -504,11 +503,11 @@ protected bool IsCoreDump
         public Task CmdBreakInternal()
         {
             //TODO May need to fix attach on windows and osx.
-            if (IsLocalGdbAttach() && PlatformUtilities.IsLinux())
+            if (IsLocalGdb() && PlatformUtilities.IsLinux())
```
gregg-miskelly: Do we need logic to explicitly ignore this for dumps?

gregg-miskelly:
👍

src/MICore/Debugger.cs
pieandcakes: I added a verify like we have in the CmdContinueAsync()

paulmaybee:
:+1:

chuckries:
:+1:

PR #297, opened 2016-04-26
Implement data breakpoints

paulmaybee:


paulmaybee:
@gregg-miskelly @jacdavis

src/DebugEngineHost/HostMarshal.cs
```diff
@@ -59,6 +59,16 @@ public static IDebugFunctionPosition2 GetDebugFunctionPositionForIntPtr(IntPtr l
         }
 
         /// <summary>
+        /// Obtain a string from a BSTR
+        /// </summary>
+        /// <param name="stringId"></param>
+        /// <returns></returns>
+        public static string GetStringForIntPtr(IntPtr stringId)
```
gregg-miskelly: I think I would suggest calling this something like 'GetDataBreakpointStringForIntPtr' since this shouldn't be used as a general purpose thing.

src/DebugEngineHost/HostMarshal.cs
```diff
@@ -59,6 +59,16 @@ public static IDebugFunctionPosition2 GetDebugFunctionPositionForIntPtr(IntPtr l
         }
 
         /// <summary>
+        /// Obtain a string from a BSTR
```
gregg-miskelly: How about:
Obtain the string expression from the bpLocation union for a BPLT_DATA_STRING breakpoint.

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -420,6 +420,11 @@ private StringBuilder BuildBreakInsert(string condition)
             return await _debugger.CmdAsync(cmd.ToString(), resultClass);
         }
 
+        public virtual Task<Results> BreakWatch(string address, uint size, ResultClass resultClass = ResultClass.done)
+        {
+            throw new NotImplementedException();
```
gregg-miskelly: Do you know if LLDB supports this?

paulmaybee: It does not

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -80,7 +80,8 @@ private bool CanBind()
             // The sample engine only supports breakpoints on a file and line number or function name. No other types of breakpoints are supported.
             if (_deleted ||
                 (_bpRequestInfo.bpLocation.bpLocationType != (uint)enum_BP_LOCATION_TYPE.BPLT_CODE_FILE_LINE
-                && _bpRequestInfo.bpLocation.bpLocationType != (uint)enum_BP_LOCATION_TYPE.BPLT_CODE_FUNC_OFFSET))
+                && _bpRequestInfo.bpLocation.bpLocationType != (uint)enum_BP_LOCATION_TYPE.BPLT_CODE_FUNC_OFFSET
+                && _bpRequestInfo.bpLocation.bpLocationType != (uint)enum_BP_LOCATION_TYPE.BPLT_DATA_STRING))
```
gregg-miskelly: Should we ask the command factor if data breakpoints are supported?

src/MICore/CommandFactories/MICommandFactory.cs
paulmaybee: That is, the MI interface does not support it.

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
```diff
@@ -91,6 +91,13 @@ private static MIBreakpointState StringToBreakpointState(string addr)
             return EvalBindResult(await process.MICommandFactory.BreakInsert(functionName, condition, ResultClass.None), pbreak);
         }
 
+        internal static async Task<BindResult> Bind(string address, uint size, DebuggedProcess process, string condition, AD7PendingBreakpoint pbreak)
+        {
+            process.VerifyNotDebuggingCoreDump();
+
+            return EvalBindWatchResult(await process.MICommandFactory.BreakWatch(address, size, ResultClass.None), pbreak, address, size);
```
gregg-miskelly: Do we need to do something if the breakpoint is set from run mode?

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
```diff
@@ -329,6 +377,12 @@ internal BoundBreakpoint(PendingBreakpoint parent, ulong addr, /*optional*/ Tupl
             }
         }
 
+        internal BoundBreakpoint(PendingBreakpoint parent, ulong addr, uint size)
+        {
+            Addr = addr;
```
gregg-miskelly: Aren't we using Addr for two kind of unrelated purposes here -- address of the instruction where the breakpoint is set vs. address of the data being watched.

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
paulmaybee: UI doesn't seem to allow that to happen.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -773,6 +773,36 @@ private void Dispose()
...
+                var wpt = results.Results.Find("wpt");
+                string bkptno = wpt.FindString("number");
+                ulong addr = cxt.pc ?? 0;
+
+                bool fContinue;
+                TupleValue frame = wpt.TryFind<TupleValue>("frame");
+                AD7BoundBreakpoint[] bkpt = _breakpointManager.FindHitBreakpoints(bkptno, addr, frame, out fContinue);
+                if (bkpt != null)
+                {
+                    List<object> bplist = new List<object>();
```
gregg-miskelly: I wouldn't think we should need to create the separate list. Can't we just do this?
new ReadOnlyCollection<object>(bkpt)

paulmaybee: Only constructor takes an IList argument

src/MICore/CommandFactories/gdb.cs
```diff
@@ -194,5 +194,28 @@ public override Task EnableTargetAsyncOption()
...
+            switch (size)
+            {
+                case 1:
+                    return "char";
+                case 2:
+                    return "short";
+                case 4:
+                    return "int";
+                case 8:
+                    return "double";
```
jacdavis: Why is 8 a double? Does gdb not support 64 bit integer data bps?

paulmaybee: I don't think the type matters, just the size.

paulmaybee:
Made data breakpoints more explicit rather than depending upon code breakpoint behavior. Implemented correct behavior for data breakpoints in BreakpointResolution. 
@gregg-miskelly @jacdavis

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
paulmaybee: Now uses IsDataBreakpoint to distinguish the difference, Also BreakpointManager explicitly deals with code and data breaks separately.

src/MICore/CommandFactories/MICommandFactory.cs
jacdavis: do we want to put it inside the gdb specific portion of the command factory given that clrdbg also can't support databps?

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -438,7 +474,7 @@ int IDebugPendingBreakpoint2.SetCondition(BP_CONDITION bpCondition)
             PendingBreakpoint bp = null;
             lock (_boundBreakpoints)
             {
-                if (!VerifyCondition(bpCondition))
+                if (!VerifyCondition(bpCondition) || IsDataBreakpoint)
```
jacdavis: is this necessary because gdb doesn't support conditions on data bps?

paulmaybee: yes

src/MICore/CommandFactories/MICommandFactory.cs
paulmaybee: gdb subclass overrides the default behavior (which just throws). How would you like to see this done differently?

jacdavis: that's exactly what I was looking for. Got confused by GitHub only showing partial diffs

src/MIDebugEngine/Engine.Impl/BreakpointManager.cs
```diff
@@ -24,6 +24,28 @@ public BreakpointManager(AD7Engine engine)
             _pendingBreakpoints = new System.Collections.Generic.List<AD7PendingBreakpoint>();
         }
 
+        private IEnumerable<AD7PendingBreakpoint> CodeBreakpoints
+        {
+            get
+            {
+                lock (_pendingBreakpoints)
```
gregg-miskelly: Since this code is doing deferred execution, the lock isn't going to do anything.

gregg-miskelly:
:+1:

PR #298, opened 2016-04-28
Determine debugger path from $PATH if not specified

andrewcrawley:
If the "miDebuggerPath" is omitted from the launch options, we will now attempt to determine the correct path based on the "MIMode" option and the user's $PATH.  "miDebuggerPath" can still be specified to override the path if necessary.

This change also adds some additional path checking to the LocalTransport, so we get better error messages on OSX if the path is invalid.

Please review: @jacdavis @gregg-miskelly @chuckries @paulmaybee

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -211,8 +211,36 @@ namespace MICore
 
         static internal LocalLaunchOptions CreateFromXml(Xml.LaunchOptions.LocalLaunchOptions source)
         {
+            string miDebuggerPath = source.MIDebuggerPath;
+
+            // If no path to the debugger was specified, look for the proper binary in the user's $PATH
+            if (String.IsNullOrEmpty(miDebuggerPath))
+            {
+                string debuggerBinary = null;
+                switch (source.MIMode)
+                {
+                    case Xml.LaunchOptions.MIMode.gdb:
+                        debuggerBinary = "gdb";
+                        break;
+
+                    case Xml.LaunchOptions.MIMode.lldb:
+                        debuggerBinary = "lldb-mi";
+                        break;
+                }
+
+                if (!String.IsNullOrEmpty(debuggerBinary))
+                {
+                    miDebuggerPath = LocalLaunchOptions.ResolveFromPath(debuggerBinary);
+                }
+
+                if (String.IsNullOrEmpty(miDebuggerPath))
+                {
+                    throw new InvalidLaunchOptionsException(MICoreResources.Error_NoMiDebuggerPath);
+                }
+            }
+
             var options = new LocalLaunchOptions(
-                RequireAttribute(source.MIDebuggerPath, "MIDebuggerPath"),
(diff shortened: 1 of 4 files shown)


src/MICore/LaunchOptions.cs
```diff
@@ -228,6 +256,25 @@ static internal LocalLaunchOptions CreateFromXml(Xml.LaunchOptions.LocalLaunchOp
...
+        {
+            string pathVar = System.Environment.GetEnvironmentVariable("PATH");
+
+            // Check each portion of the PATH environment variable to see if it contains the requested file
+            foreach (string pathPart in pathVar.Split(Path.PathSeparator))
+            {
+                string candidate = Path.Combine(pathPart, command);
+
+                // If the file exists and is not a directory, use it
+                if (File.Exists(candidate) && (File.GetAttributes(candidate) & FileAttributes.Directory) == 0)
```
gregg-miskelly: Will File.Exists return true for a directory?

src/MICore/Transports/LocalTransport.cs
```diff
@@ -19,9 +19,34 @@ public LocalTransport()
...
+            return true;
+        }
+
         public override void InitStreams(LaunchOptions options, out StreamReader reader, out StreamWriter writer)
         {
             LocalLaunchOptions localOptions = (LocalLaunchOptions)options;
+
+            if (!this.IsValidMiDebuggerPath(localOptions.MIDebuggerPath))
+            {
+                throw new Exception(MICoreResources.Error_InvalidMiDebuggerPath);
```
gregg-miskelly: Should this message include the path?

gregg-miskelly:
Otherwise LGTM

src/MICore/LaunchOptions.cs
```diff
@@ -211,8 +211,36 @@ public bool ShouldStartServer()
...
+                        debuggerBinary = "lldb-mi";
+                        break;
+                }
+
+                if (!String.IsNullOrEmpty(debuggerBinary))
+                {
+                    if (PlatformUtilities.IsWindows())
+                    {
+                        Path.ChangeExtension(debuggerBinary, "exe");
+                    }
```
jacdavis: not sure if this will be correct on Cygwin or the Linux subsystem.

src/MICore/Transports/LocalTransport.cs
andrewcrawley: The message is "The value of miDebuggerPath is invalid", which seems pretty clear and actionable.  It's the same message we already use in this case for the LocalLinuxTransport.

src/MICore/LaunchOptions.cs
andrewcrawley: Nope.  Removed the unnecessary check.

src/MICore/LaunchOptions.cs
andrewcrawley: Removed the Windows check entirely - all scenarios on Windows currently specify an explicit path anyway.

PR #300, opened 2016-05-02
Add null check in case thread is null;

pieandcakes:
In some instances, if a user steps over on the last line of main, it can throw a NullReferenceException in the Step method. The only place I can see this happening was either thread or the _debuggedthread being null.

Added checks that will return E_FAIL if it is null. 

@jacdavis @chuckries @gregg-miskelly

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7Engine.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Engine.cs
@@ -170,7 +170,7 @@ namespace Microsoft.MIDebugEngine
             {
                 return e.HResult;
             }
-            catch (Exception e) when (ExceptionHelper.BeforeCatch(e, Logger, reportOnlyCorrupting:true))
+            catch (Exception e) when (ExceptionHelper.BeforeCatch(e, Logger, reportOnlyCorrupting: true))
             {
                 return EngineUtils.UnexpectedException(e);
             }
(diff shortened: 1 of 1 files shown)


pieandcakes:
Tested by returning E_FAIL from Step and VSCode will return "Error: Unable to step next" if either object is null

gregg-miskelly:
Chuck made an OpenDebugAD7 change last week that may have made this unnecessary. But it certainly doesn't hurt anything.

gregg-miskelly:
👍

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -769,6 +769,11 @@ public int Step(IDebugThread2 pThread, enum_STEPKIND kind, enum_STEPUNIT unit)
 
             try
             {
+                if (null == thread || null == thread.GetDebuggedThread())
+                {
+                    return Constants.E_FAIL;
```
jacdavis: Kneejerk reaction to E_FAIL (like we spoke about). Concord winds returning a xapi failed hr from the dispatcher in this case (soo concord\impl\ad7\Programm.cpp step, so there is precedence here.

jacdavis: LGTM

chuckries:
:+1:

PR #301, opened 2016-05-03
Create powershell version of GetClrDbg script

chuckries:
This is the intial version of getClrDbg.ps1 which automates installing
clrdbg on windows. clrdbg version, runtime id, and install location can be
provided on the command line.

### DIFF
--- a/scripts/GetClrDbg.ps1
+++ b/scripts/GetClrDbg.ps1
@@ -1,0 +1,131 @@
+﻿<#
+.SYNOPSIS
+Downloads the given $Version of clrdbg for the given $RuntimeID and installs it to the given $InstallPath
+
+.DESCRIPTION
+The following script will generate a project.json and NuGet.config and use dotnet restore and publish to install clrdbg, the .NET Core Debugger
+
+.PARAMETER Version
+Specifies the version of clrdbg to install. Can be 'latest', VS2015U2, or a specific version string i.e. 14.0.25229-preview-2963841
+
+.PARAMETER RuntimeID
+Specifies the .NET Runtime ID of the clrdbg that will be downloaded. Example: ubuntu.14.04-x64. Defaults to the Runtime ID of the current machine.
+
+.Parameter InstallPath
+Specifies the path where clrdbg will be installed. Defaults to the directory containing this script.
+
+.INPUTS
+None. You cannot pipe inputs to GetClrDbg.
+
+.EXAMPLE
+C:\PS> .\GetClrDbg.ps1 -Version latest -RuntimeID ubuntu.14.04-x64 -InstallPath .\clrdbg
+
+.LINK
+https://github.com/Microsoft/MIEngine
+#>
+
+Param (
+    [Parameter(Mandatory=$true, ParameterSetName="ByName")]
+    [string]
+    [ValidateSet("latest", "VS2015U2")]
+    $Version,
+
+    [Parameter(Mandatory=$true, ParameterSetName="ByNumber")]
+    [string]
+    [ValidatePattern("\d+\.\d+\.\d+.*")]
+    $VersionNumber,
+
+    [string]$RuntimeID,
+    [string]$InstallPath = (Split-Path -Path $MyInvocation.MyCommand.Definition)
+)
+
+function GetDotNetRuntimeID() {
+    $ridLine = dotnet --info | findstr "RID"
+    
+    if ([System.String]::IsNullOrEmpty($ridLine)) {
+        throw [System.Exception] "Unable to determine runtime from dotnet --info. Make sure dotnet cli is up to date on this machine"
+    }
+
+    $rid = $ridLine.Split(":")[1].Trim();
+
+    if ([System.String]::IsNullOrEmpty($rid)) {
+        throw [System.Exception] "Unable to determine runtime from dotnet --info. Make sure dotnet cli is up to date on this machine"
+    }
+
+    return $rid
+}
+
+# Produces project.json in the current directory
+function GenerateProjectJson([string] $version, [string]$runtimeID) {
+    $projectJson = 
+"{
+    `"dependencies`": {
+       `"Microsoft.VisualStudio.clrdbg`": `"$version`"
+    },
+    `"frameworks`": {
+        `"netstandardapp1.5`": {
+          `"imports`": [ `"dnxcore50`", `"portable-net45+win8`" ]
+       }
+   },
+   `"runtimes`": {
+      `"$runtimeID`": {}
+   }
+}"
+
+    $projectJson | Out-File -Encoding utf8 project.json
+}
+
+# Produces NuGet.config in the current directory
+function GenerateNuGetConfig() {
+    $nugetConfig = 
+"<?xml version=`"1.0`" encoding=`"utf-8`"?>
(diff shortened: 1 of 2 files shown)


chuckries:
@gregg-miskelly

gregg-miskelly:
@kelltrick since you have more PowerShell experience than most of the rest of us, can you take a look also?

scripts/GetClrDbg.ps1
```diff
@@ -0,0 +1,118 @@
+﻿<#
+.SYNOPSIS
+Downloads the given $Version of clrdbg for the given $RuntimeID and installs it to the given $InstallPath
+
+.DESCRIPTION
+The following script will generate a project.json and NuGet.config and use dotnet resotre and publish to install clrdbg, the .NET Core Debugger
```
gregg-miskelly: 'resotre' is spelled wrong

scripts/GetClrDbg.ps1
```diff
@@ -0,0 +1,118 @@
...
+Specifies the .NET Runtime ID of the clrdbg that will be downloaded. Example: ubuntu.14.04-x64. Defaults to the Runtime ID of the current machine.
+
+.Parameter InstallPath
+Specifies the path where clrdbg will be installed. Defaults to the directory containing this script.
+
+.INPUTS
+None. You cannot pipe inputs to GetClrDbg.
+
+.EXAMPLE
+C:\PS> .\GetClrDbg.ps1 -Version latest -RuntimeID ubunut.14.04-x64 -InstallPath .\clrdbg
```
gregg-miskelly: Spelling: 'ubunut'

scripts/GetClrDbg.ps1
```diff
@@ -0,0 +1,118 @@
...
+
+Param (
+    [Parameter(Mandatory=$true)]
+    [string]$Version,
+    [string]$RuntimeID,
+    [string]$InstallPath = (Split-Path -Path $MyInvocation.MyCommand.Definition)
+)
+
+function GetDotNetRuntimeID() {
+    $ridLine = dotnet --info | findstr "RID"
```
gregg-miskelly: I think you either need to explicitly set the error mode to abort the script, or check for errors, because by default this code will happily continue on if dotnet cannot be found, or if it doesn't have a 'RID' line.

gregg-miskelly:
Otherwise LGTM

scripts/GetClrDbg.ps1
```diff
@@ -0,0 +1,118 @@
...
+      <add key=`"coreclr-debug`" value=`"https://www.myget.org/F/coreclr-debug/api/v3/index.json`" />
+  </packageSources>
+</configuration>"
+
+    $nugetConfig | Out-File -Encoding utf8 NuGet.config
+}
+
+# Add new version constants here
+# 'latest' version may be updated
+# all other version contstants i.e. 'vs2015u2' may not be updated after they are finalized
```
gregg-miskelly: Spelling: 'contstants'

scripts/GetClrDbg.ps1
```diff
@@ -0,0 +1,118 @@
...
+
+Write-Host "Info: Executing dotnet restore"
+dotnet restore
+
+Write-Host "Info: Executing dotnet publish"
+dotnet publish -r $RuntimeID -o .
+
+Pop-Location
+
+Write-Host "Succesfully installed clrdbg at '$InstallPath'"
```
gregg-miskelly: Spelling: 'Succesfully'

chuckries:
Apparently I can't spell.

scripts/GetClrDbg.ps1
chuckries: I decided to handle the errors myself by checking for null or empty output from my filter logic

caslan:
LGTM

scripts/GetClrDbg.ps1
```diff
@@ -0,0 +1,128 @@
...
+.EXAMPLE
+C:\PS> .\GetClrDbg.ps1 -Version latest -RuntimeID ubuntu.14.04-x64 -InstallPath .\clrdbg
+
+.LINK
+https://github.com/Microsoft/MIEngine
+#>
+
+Param (
+    [Parameter(Mandatory=$true)]
+    [string]$Version,
```
kelltrick: If you care about tab completion you can make this part work better by adding [Parameter Mandatory=$true, ParameterSet="ByName"][string][ValidateSet("latest","vs2015u2")]$VersionName, [Parameter Mandatory=$false, ParameterSet="ByNumber"][string][ValidatePattern("\d+\.\d+\.\d+(-.[a-zA-Z0-9-]{1,20})?"]$Version

chuckries: Why would $VersionName be mandatory here? Wouldn't this prevent you from being able to specify just a version number?

scripts/GetClrDbg.ps1
```diff
@@ -0,0 +1,128 @@
...
+
+    if ([System.String]::IsNullOrEmpty($rid)) {
+        throw [System.Exception] "Unable to determine runtime from dotnet --info. Make sure dotnet cli is up to date on this machine"
+    }
+
+    return $rid
+}
+
+# Produces project.json in the current directory
+function GenerateProjectJson([string] $version, [string]$runtimeID) {
```
kelltrick: Replace this with @{
    dependencies=@{'Microsoft.VisualStudio.clrdbg'="$version"}; 
    frameworks=@{
        'netstandardapp1.5'=@{
            imports=@('dnxcore50','portable-net45+win8')
        }
    }
    runtimes=@{"$runtimeID"=@{}}
}  | ConvertTo-Json

scripts/GetClrDbg.ps1
```diff
@@ -0,0 +1,128 @@
...
+   `"runtimes`": {
+      `"$runtimeID`": {}
+   }
+}"
+
+    $projectJson | Out-File -Encoding utf8 project.json
+}
+
+# Produces NuGet.config in the current directory
+function GenerateNuGetConfig() {
```
kelltrick: You can do the same thing for XML as with the json above, but its way more of a pain (because of attributes) so I wouldn't bother

scripts/GetClrDbg.ps1
chuckries: Since I am just outputting this JSON to a file a not actually manipulating it in powershell at all, I think I'm going to leave it as is. The spacing in the file output is wonky if I do it this way. Additionally the imports section needs to be a JSON array and this blob isn't getting translated like that.

scripts/GetClrDbg.ps1
kelltrick: Mandatory applies to each ParameterSet. ParameterSets are groups of parameters, when PS parses the parameters it falls into a single ParameterSet. If it gets params from multiple sets it throws an error. This way it would force the user to specify exactly one of "-Version 14.0.23232" or "-VersionName latest". I believe this is the behavior you are looking for but I'm not 100% sure on that.

chuckries: That is great. I was unaware of parameter sets.

kelltrick:
lgtm

chuckries:
thanks all!

PR #303, opened 2016-05-03
Add OS X terminal for debugging locally

edumunoz:
There is some refactoring.

Pending issue:
- There is a race condition when shutting down gdb. Because of it, a fix for unblocking a `close()` syscall on OS X is commented out (with a TODO). This will be fixed in an upcoming pull request.

I tested this on OS X 10.10 and Ubuntu 14.04. On Ubuntu, I verified there are no processes left running if debugging terminates "gracefully" (the program ends to completion or the debugging session is stopped) or the terminal window is closed.

@jacdavis @gregg-miskelly


src/MICore/LaunchOptions.cs
```diff
@@ -209,6 +209,25 @@ public bool ShouldStartServer()
...
+            if (!File.Exists(MIDebuggerPath))
+            {
+                return false;
+            }
+            else
+            {
+                // Verify the target is a file and not a directory
+                FileAttributes attr = File.GetAttributes(MIDebuggerPath);
+                if ((attr & FileAttributes.Directory) != 0)
+                {
```
gregg-miskelly: You shouldn't need this - File.Exists should return false for a directory.

src/MICore/LaunchOptions.cs
```diff
@@ -209,6 +209,25 @@ public bool ShouldStartServer()
...
+        public bool IsValidMiDebuggerPath()
+        {
+            if (!File.Exists(MIDebuggerPath))
+            {
+                return false;
+            }
+            else
+            {
+                // Verify the target is a file and not a directory
+                FileAttributes attr = File.GetAttributes(MIDebuggerPath);
```
andrewcrawley: This came up on one of my recent pull requests - File.Exists will return false if you pass the path of a directory, so you don't need the additional check.

src/MICore/osxlaunchhelper.scpt
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/osascript
+
+on run argv
+   set debuggerTitle to (item 1 of argv)
+   set executeCommand to (item 2 of argv)
+   # Note: if other tabs are open on the terminal window that is opened by this script, this won't behave properly.
+   set command to "clear; " & ¬
```
gregg-miskelly: There is kind of a weird character here. Just wanted to make sure -- this is intentional correct?

src/MICore/osxlaunchhelper.scpt
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/osascript
```
gregg-miskelly: Should we have a copyright header on this file?

src/MICore/MICore.csproj
```diff
@@ -126,6 +126,10 @@
   <ItemGroup>
     <None Include="LaunchOptions.xsd" />
     <None Include="SetMIDebugLogging.cmd" />
+    <ContentWithTargetPath Include="osxlaunchhelper.scpt">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <TargetPath>osxlaunchhelper.scpt</TargetPath>
+    </ContentWithTargetPath>
```
jacdavis: you'll need to put this in the extension and enabled execution permissions.

src/MICore/MICoreResources.resx
```diff
@@ -219,4 +219,7 @@ Error: {1}</value>
   <data name="Error_NoMiDebuggerPath" xml:space="preserve">
     <value>Unable to determine path to debugger.  Please specify the "MIDebuggerPath" option.</value>
   </data>
+  <data name="Error_InternalFileMissing" xml:space="preserve">
+    <value>An internal file could not be found.</value>
```
jacdavis: for diagnostic purposes, it would be nice to have the file name in this message.

src/MICore/Transports/LocalUnixTransport.cs
```diff
@@ -0,0 +1,98 @@
...
+                    debuggeeDir = "/";
+            }
+
+            _dbgStdInName = UnixUtilities.MakeFifo(Logger);
+            _dbgStdOutName = UnixUtilities.MakeFifo(Logger);
+
+            string watcherDirectory = _dbgStdInName.Substring(0, _dbgStdInName.IndexOf(UnixUtilities.FifoPrefix, StringComparison.OrdinalIgnoreCase));
+            _fifoWatcher = new FileSystemWatcher(watcherDirectory, UnixUtilities.FifoPrefix + "*");
+            _fifoWatcher.Deleted += FifoWatcher_Deleted;
+            _fifoWatcher.EnableRaisingEvents = true;
```
jacdavis: we need to test this since this was deadlocking before.

src/MICore/UnixUtilities.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Text;
+
+namespace MICore
+{
+    public static class UnixUtilities
+    {
+        internal const string FifoPrefix = "Microsoft-MIEngine-fifo-";
+        internal const string SudoPath = "/usr/bin/sudo";
+        private const string PtraceScopePath = "/proc/sys/kernel/yama/ptrace_scope";
+        private const string PKExecPath = "/usr/bin/pkexec";
```
jacdavis: All of these are all Linux specific except the fifo prefix.

src/MICore/MICore.csproj
edumunoz: This the reason why the script is being launched with `/usr/bin/osascript`, so that we don't need to worry about permissions.

src/MICore/osxlaunchhelper.scpt
edumunoz: It is, yeah. It's the character used to break a statement into multiple lines.

src/MICore/osxlaunchhelper.scpt
```diff
@@ -0,0 +1,22 @@
...
+                  "osascript -e 'tell application \"Terminal\" to close (every window whose tty is \"'\"$(tty)\"'\")' & exit"
+
+    tell application "Terminal"
+        if it is running then
+            activate
+            set newTab to do script command
+        else
+            activate
+            set newTab to do script command in window 1
+        end if
```
jacdavis: comment why this is written like this?

src/MICore/UnixUtilities.cs
edumunoz: `SudoPath` is also shared. The Linux-specific ones are private are only used internally in this class.

src/MICore/Transports/LocalUnixTransport.cs
edumunoz: Agreed. @jacdavis specifically is referring to `FileSystemWatcher` on OS X on mono.

src/MICore/MICoreResources.resx
edumunoz: I agree, but this message is being shown in VSCode. Are we still ok showing the file that couldn't be found there?

src/MICore/Transports/LocalUnixTransport.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public override void InitStreams(LaunchOptions options, out StreamReader reader, out StreamWriter writer)
+        {
+            LocalLaunchOptions localOptions = (LocalLaunchOptions)options;
+
+            // Default working directory is next to the app
+            string debuggeeDir;
+            if (Path.IsPathRooted(options.ExePath) && File.Exists(options.ExePath))
+            {
+                debuggeeDir = System.IO.Path.GetDirectoryName(options.ExePath);
```
pieandcakes: don't think you need System.IO anymore

src/MICore/Transports/LocalUnixTransport.cs
```diff
@@ -0,0 +1,98 @@
...
+            _dbgStdInName = UnixUtilities.MakeFifo(Logger);
+            _dbgStdOutName = UnixUtilities.MakeFifo(Logger);
+
+            string watcherDirectory = _dbgStdInName.Substring(0, _dbgStdInName.IndexOf(UnixUtilities.FifoPrefix, StringComparison.OrdinalIgnoreCase));
+            _fifoWatcher = new FileSystemWatcher(watcherDirectory, UnixUtilities.FifoPrefix + "*");
+            _fifoWatcher.Deleted += FifoWatcher_Deleted;
+            _fifoWatcher.EnableRaisingEvents = true;
+
+            // Setup the streams on the fifos as soon as possible.
+            System.IO.FileStream dbgStdInStream = new FileStream(_dbgStdInName, FileMode.Open);
```
pieandcakes: Same here

src/MICore/Transports/LocalUnixTransport.cs
```diff
@@ -0,0 +1,98 @@
...
+            string debuggeeDir;
+            if (Path.IsPathRooted(options.ExePath) && File.Exists(options.ExePath))
+            {
+                debuggeeDir = System.IO.Path.GetDirectoryName(options.ExePath);
+            }
+            else
+            {
+                // If we don't know where the app is, default to HOME, and if we somehow can't get that, go with the root directory.
+                debuggeeDir = Environment.GetEnvironmentVariable("HOME");
+                if (string.IsNullOrEmpty(debuggeeDir))
```
pieandcakes: Nit: Captial S "String" too please

src/MICore/TerminalLauncher.cs
```diff
@@ -0,0 +1,152 @@
...
+        protected override string GetProcessArgs()
+        {
+            string thisModulePath = typeof(MacTerminalLauncher).GetTypeInfo().Assembly.ManifestModule.FullyQualifiedName;
+            string launchScript = Path.Combine(Path.GetDirectoryName(thisModulePath), LaunchTerminalScript);
+            if (!File.Exists(launchScript))
+            {
+                throw new FileNotFoundException(MICoreResources.Error_InternalFileMissing);
+            }
+
+            return string.Format(CultureInfo.InvariantCulture, "{0} \"{1}\" \"{2}\"", launchScript, _title, _initScript);
```
pieandcakes: Nit: Capital S String please

src/MICore/Transports/LocalUnixTransport.cs
edumunoz: Where? The type `string` is always used as lowercase.

src/MICore/TerminalLauncher.cs
gregg-miskelly: Why?

---
In reply to: [61978609](https://github.com/Microsoft/MIEngine/pull/303#discussion_r61978609) [](ancestors = 61978609)

src/MICore/Transports/LocalUnixTransport.cs
```diff
@@ -0,0 +1,98 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Globalization;
+using System.IO;
+using System.Text;
+
+namespace MICore
+{
+    public class LocalUnixTransport : StreamTransport
```
gregg-miskelly: nit: LocalUnixTerminalTransport since what is special about this is that we are using the terminal?

edumunoz: Fair point.

gregg-miskelly:
:+1:

src/MICore/UnixUtilities.cs
pieandcakes: maybe rename the ones that are Linux specific so we know they are or leave a comment?

PR #304, opened 2016-05-04
Use '-gdb-set new-console' on Windows

gregg-miskelly:
On Windows, GDB has an option to launch using a new console. CLRDBG was changed to use the same syntax. This changes the MIEngine to use this syntax, and I removed the 'if clrdbg' check.

Note that I don't believe that LLDB supports this option yet. But this shouldn't blow up as we are ignoring failures.


gregg-miskelly:
@jacdavis @edumunoz please review.

jacdavis:
LGTM. Same code I had written.

PR #306, opened 2016-05-05
Support Detach.

rajkumar42:
Sending "stopped" message with reason as "disconnected" to avoid the confusion.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -1169,8 +1169,13 @@ namespace Microsoft.MIDebugEngine
 
         public void Detach()
         {
-            // Fake stopped message to send ProgramDestroyEvent.
-            ScheduleStdOutProcessing(@"*stopped,reason=""disconnected""");
+            // Special casing sending the fake stopped event for clrdbg. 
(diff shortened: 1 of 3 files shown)


rajkumar42:
@edumunoz

edumunoz:
Wouldn't `HandleThreadGroupExited` still be called, causing the `"*stopped,reason=""exited"""` string to be processed too?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1167,7 +1167,12 @@ public void Terminate()
             ScheduleStdOutProcessing(@"*stopped,reason=""exited"",exit-code=""42""");
         }
 
-        public void Detach() { }
+        public void Detach()
+        {
+            // Fake stopped message to send ProgramDestroyEvent.
+            ScheduleStdOutProcessing(@"*stopped,reason=""disconnected""");
```
edumunoz: I don't see the reason `"disconnected"` being handled on the `OnStopped` handler (in Debugger.cs).

Shouldn't it cause the debugger to exit (`-gdb-exit`)?

rajkumar42: Sorry, I missed the file during comment. Have added it now.

gregg-miskelly:
:+1:

PR #308, opened 2016-05-06
Detect and warn when source files are out of date

andrewcrawley:
When we hit a breakpoint, warn the user if the source file is newer than the module containing it.

Please review: @jacdavis @gregg-miskelly @chuckries @paulmaybee

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -49,6 +49,7 @@ namespace Microsoft.MIDebugEngine
         private ReadOnlyCollection<RegisterGroup> _registerGroups;
         private readonly EngineTelemetry _engineTelemetry = new EngineTelemetry();
         private bool _needTerminalReset;
+        private HashSet<Tuple<string, string>> _fileTimestampWarnings;
 
         public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngineCallback callback, WorkerThread worker, BreakpointManager bpman, AD7Engine engine, HostConfigurationStore configStore) : base(launchOptions, engine.Logger)
         {
@@ -79,6 +80,7 @@ namespace Microsoft.MIDebugEngine
 
             VariablesToDelete = new List<string>();
             this.ActiveVariables = new List<IVariableInformation>();
+            this._fileTimestampWarnings = new HashSet<Tuple<string, string>>();
 
             OutputStringEvent += delegate (object o, string message)
             {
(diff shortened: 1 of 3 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -744,8 +753,17 @@ private void Dispose()
...
                 ulong addr = cxt.pc ?? 0;
 
-                bool fContinue;
                 TupleValue frame = results.Results.TryFind<TupleValue>("frame");
+                if (frame != null && addr != 0)
+                {
+                    string sourceFile = frame.TryFindString("fullname");
+                    if (!String.IsNullOrEmpty(sourceFile))
+                    {
+                        await this.VerifySourceFileTimestamp(addr, sourceFile);
```
paulmaybee: I don't think you want to do this check unless the debugger is really going to stop. Maybe delay this check until the decision to send up a stopping event is made.

paulmaybee:
👍

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
gregg-miskelly: This is also only going to do the verification in the leaf stack frame. We might want to check other frames on the call stack.

---
In reply to: [62338268](https://github.com/Microsoft/MIEngine/pull/308#discussion_r62338268) [](ancestors = 62338268)

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -858,6 +876,40 @@ private void Dispose()
...
+            await this.EnsureModulesLoaded();
+
+            string targetModulePath = this._launchOptions.ExePath;
+            DebuggedModule targetModule = this._moduleList.FirstOrDefault(m => m.AddressInModule(addr));
+            if (targetModule != null)
+            {
+                targetModulePath = targetModule.Name;
+            }
+
+            if (!File.Exists(sourceFilePath) || !File.Exists(targetModulePath))
```
gregg-miskelly: Don't we need also need --
    targetModulePath == null ||

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -858,6 +876,40 @@ private void Dispose()
...
+            DateTime sourceFileTimestamp = File.GetLastWriteTimeUtc(sourceFilePath);
+            DateTime moduleFileTimestamp = File.GetLastWriteTimeUtc(targetModulePath);
+
+            if (sourceFileTimestamp > moduleFileTimestamp)
+            {
+                // Source file is newer than the module - warn the user if we haven't already
+				Tuple<string, string> key = Tuple.Create(sourceFilePath, targetModulePath);
+
+                if (!this._fileTimestampWarnings.Contains(key))
+                {
```
gregg-miskelly: Should we check this before asking the file system for the time?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -858,6 +876,40 @@ private void Dispose()
             }
         }
 
+        private async Task VerifySourceFileTimestamp(ulong addr, string sourceFilePath)
+        {
```
gregg-miskelly: No nothing if we are using pipe transport?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -858,6 +876,40 @@ private void Dispose()
...
+                targetModulePath = targetModule.Name;
+            }
+
+            if (!File.Exists(sourceFilePath) || !File.Exists(targetModulePath))
+            {
+                return;
+            }
+
+            DateTime sourceFileTimestamp = File.GetLastWriteTimeUtc(sourceFilePath);
+            DateTime moduleFileTimestamp = File.GetLastWriteTimeUtc(targetModulePath);
```
gregg-miskelly: Do we want a try/catch around these to guard against I/O errors?

gregg-miskelly:
Otherwise looks good to me.

jacdavis:
Gregg already covered everything I would have said on this.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
andrewcrawley: File.Exists will return false if you pass null / empty string, so this is handled.

gregg-miskelly: Cool. You learn something new everyday :)

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -858,6 +876,49 @@ private void Dispose()
...
+            {
+                if (!File.Exists(sourceFilePath) || !File.Exists(targetModulePath))
+                {
+                    return;
+                }
+
+                DateTime sourceFileTimestamp = File.GetLastWriteTimeUtc(sourceFilePath);
+                DateTime moduleFileTimestamp = File.GetLastWriteTimeUtc(targetModulePath);
+
+                if (sourceFileTimestamp > moduleFileTimestamp)
```
faxue-msft: Do we need to also warn user if the source file is older?

gregg-miskelly: No. That would be the normal case.

PR #310, opened 2016-05-07
Add osxlaunchhelper.scpt to the CLRDBG MIEngine package

gregg-miskelly:
This checkin adds osxlaunchhelper.scpt to the MIEngine package. With this I
was able to see the 'externalConsole' option work correctly on OSX.


gregg-miskelly:
@chuckries please review.

chuckries:
:+1:

chuckries:
I'm pretty sure this will cause the file to be dropped on all platforms, which isn't a big deal. The /any/any/ refer to language (such as cs or vb) and target framework moniker (such as net35 or wp81), so I do not think there's a way to only drop this on mac. LGTM as is.

PR #311, opened 2016-05-07
Fix gdb exit race condition

edumunoz:
This PR makes the fake `-gdb-exit` event conditional on whether the debugger has already exited or not. This allows to uncomment the TODO block that used to be in the StreamTransport class for OS X.

I also removed the need for `FileSystemWatchers` by adding another fifo. `FileSystemWatchers` were flaky on OS X.

Verified:

Linux:
Launch:
- Close the terminal window
- Stop execution
- Let the program run to completion

Attach:
- Canceling the pkexec prompt (main use of the `FileSystemWatchers`)
- Dettaching
- Close the terminal window

OS X
Launch:
- Close the terminal window
- Stop execution
- Let the program run to completion

Attach:
- Dettaching
- Close the terminal window

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -55,6 +55,22 @@ namespace MICore
             }
         }
 
+        public bool IsDebuggerRunning
+        {
+            get
+            {
+                if (_localDebuggerPid > 0)
+                {
+                    if (PlatformUtilities.IsLinux() || PlatformUtilities.IsOSX())
+                    {
+                        return UnixNativeMethods.Kill(_localDebuggerPid, 0) == 0;
+                    }
+                }
+
+                return false;
+            }
+        }
+
         public uint MaxInstructionSize { get; private set; }
         public bool Is64BitArch { get; private set; }
         public CommandLock CommandLock { get { return _commandLock; } }
@@ -72,6 +88,7 @@ namespace MICore
         private LinkedList<string> _initialErrors = new LinkedList<string>();
 
         protected bool _connected;
+        protected int _localDebuggerPid = -1;
 
         public class ResultEventArgs : EventArgs
         {
@@ -1176,10 +1193,13 @@ namespace MICore
             {
                 ScheduleStdOutProcessing(@"*stopped,reason=""exited""");
 
-                // Processing the fake "stopped" event sent above will normally cause the debugger to close, but if
-                //  the debugger process is already gone (e.g. because the terminal window was closed), we won't get
-                //  a response, so queue a fake "exit" event for processing as well, just to be sure.
-                ScheduleStdOutProcessing("^exit");
+                if (!IsDebuggerRunning)
+                {
+                    // Processing the fake "stopped" event sent above will normally cause the debugger to close, but if
+                    // the debugger process is already gone (e.g. because the terminal window was closed), we won't get
+                    // a response, so queue a fake "exit" event for processing as well.
+                    ScheduleStdOutProcessing("^exit");
+                }
             }
         }
 
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -55,22 +55,6 @@ namespace MICore
             }
         }
 
-        public bool IsDebuggerRunning
-        {
-            get
-            {
-                if (_localDebuggerPid > 0)
-                {
-                    if (PlatformUtilities.IsLinux() || PlatformUtilities.IsOSX())
-                    {
-                        return UnixNativeMethods.Kill(_localDebuggerPid, 0) == 0;
-                    }
-                }
-
-                return false;
-            }
-        }
-
         public uint MaxInstructionSize { get; private set; }
         public bool Is64BitArch { get; private set; }
         public CommandLock CommandLock { get { return _commandLock; } }
@@ -86,9 +70,9 @@ namespace MICore
         private TaskCompletionSource<object> _consoleDebuggerInitializeCompletionSource = new TaskCompletionSource<object>();
         private LinkedList<string> _initializationLog = new LinkedList<string>();
         private LinkedList<string> _initialErrors = new LinkedList<string>();
+        private int _localDebuggerPid = -1;
 
         protected bool _connected;
-        protected int _localDebuggerPid = -1;
 
         public class ResultEventArgs : EventArgs
         {
@@ -137,6 +121,32 @@ namespace MICore
             _debuggeePids = new Dictionary<string, int>();
             Logger = logger;
             _miResults = new MIResults(logger);
+        }
+
+        protected void SetDebuggerPid(int debuggerPid)
+        {
+            _localDebuggerPid = debuggerPid;
+        }
+
+        /// <summary>
+        /// Check if the local debugger process is running.
+        /// For Windows, it returns False always to avoid shortcuts taken when it returns True.
+        /// </summary>
+        /// <returns>True if the local debugger process is running and the platform is Linux or OS X.
+        /// False otherwise.</returns>
+        private bool IsUnixDebuggerRunning()
+        {
+            if (_localDebuggerPid > 0)
+            {
+                if (PlatformUtilities.IsLinux() || PlatformUtilities.IsOSX())
+                {
+                    // When 0 is passed as the signal to send in kill,
+                    // it will check the validity of the pid (e.g., does the pid exist?)
+                    return UnixNativeMethods.Kill(_localDebuggerPid, 0) == 0;
+                }
+            }
+
+            return false;
         }
 
         private void RetryBreak(object o)
@@ -1193,7 +1203,7 @@ namespace MICore
             {
                 ScheduleStdOutProcessing(@"*stopped,reason=""exited""");
 
-                if (!IsDebuggerRunning)
+                if (!IsUnixDebuggerRunning())
                 {
                     // Processing the fake "stopped" event sent above will normally cause the debugger to close, but if
                     // the debugger process is already gone (e.g. because the terminal window was closed), we won't get
--- a/src/MICore/Transports/ITransport.cs
+++ b/src/MICore/Transports/ITransport.cs
@@ -16,6 +16,12 @@ namespace MICore
         void Send(string cmd);
         void Close();
         bool IsClosed { get; }
+
+        /// <summary>
+        /// It is used to know whether to fake a response from the debugger
+        /// acknowledging that it has exited
+        /// </summary>
+        int DebuggerPid { get; }
     }
     public interface ISignalingTransport: ITransport
     {
--- a/src/MICore/Transports/ITransport.cs
+++ b/src/MICore/Transports/ITransport.cs
@@ -18,8 +18,11 @@ namespace MICore
         bool IsClosed { get; }
 
         /// <summary>
+        /// Process ID of the debugger process (clrdbg/lldb/gdb).
+        /// This value is only valid when using local launch options. When using non-local
+        /// options this may throw (e.g., TcpTransport) or provide bogus data (e.g., PipeTransport).
         /// It is used to know whether to fake a response from the debugger
-        /// acknowledging that it has exited
+        /// acknowledging that it has exited.
         /// </summary>
         int DebuggerPid { get; }
     }
--- a/src/MICore/Transports/LocalUnixTerminalTransport.cs
+++ b/src/MICore/Transports/LocalUnixTerminalTransport.cs
@@ -2,8 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Globalization;
 using System.IO;
 using System.Text;
+using System.Threading.Tasks;
 
 namespace MICore
 {
@@ -11,7 +13,15 @@ namespace MICore
     {
         private string _dbgStdInName;
         private string _dbgStdOutName;
-        private FileSystemWatcher _fifoWatcher;
+        private int _debuggerPid = -1;
+
+        /// <summary>
+        /// When the debugger exits, it will write into the fifo associated with this reader.
+        /// This is useful in the case of attach when the root password is required, but the
+        /// request is canceled. In that case, the debugger won't have started yet (so no messages
+        /// will be processed from it), but the trap will execute and write into this fifo.
+        /// </summary>
+        private StreamReader _exitReader;
 
         public override void InitStreams(LaunchOptions options, out StreamReader reader, out StreamWriter writer)
         {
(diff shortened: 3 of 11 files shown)


src/MICore/Transports/LocalUnixTerminalTransport.cs
```diff
@@ -33,49 +43,58 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
...
             TerminalLauncher terminal = TerminalLauncher.MakeTerminal("DebuggerTerminal", launchDebuggerCommand, localOptions.Environment);
             terminal.Launch(debuggeeDir);
 
+            using (StreamReader pidReader = new StreamReader(pidStream, Encoding.UTF8, true, UnixUtilities.StreamBufferSize))
+            {
+                _debuggerPid = int.Parse(pidReader.ReadLine(), CultureInfo.InvariantCulture);
+            }
+
+            _exitReader = new StreamReader(exitStream, Encoding.UTF8, true, UnixUtilities.StreamBufferSize);
+            Task<string> task = _exitReader.ReadLineAsync();
```
edumunoz: I wasn't sure if we want to catch any exceptions here. Given that this task will only complete at the end when the trap is executed.

src/MICore/Debugger.cs
```diff
@@ -55,6 +55,22 @@ public bool IsClosed
...
+                {
+                    if (PlatformUtilities.IsLinux() || PlatformUtilities.IsOSX())
+                    {
+                        return UnixNativeMethods.Kill(_localDebuggerPid, 0) == 0;
+                    }
+                }
+
+                return false;
+            }
+        }
```
chuckries: might want to comment that this call to kill will not actually kill the process because 0 is being passed, this through me off for a second.

src/MICore/UnixUtilities.cs
```diff
@@ -26,19 +26,26 @@ public static class UnixUtilities
...
         {
-             return string.Format(CultureInfo.InvariantCulture,
-                "cd {0}; DbgTerm=`tty`; trap 'rm {2}; rm {3}' EXIT; {1} --interpreter=mi --tty=$DbgTerm < {2} > {3};",
-                debuggeeDir,
-                debuggerCmd,
-                dbgStdInName,
-                dbgStdOutName
-                );
+            return string.Format(CultureInfo.InvariantCulture,
+               "cd {0}; DbgTerm=`tty`; trap 'rm {2}; rm {3}; rm {4}; rm {5}' EXIT; {6} --interpreter=mi --tty=$DbgTerm < {3} > {4} & pid=$! && echo $pid > {5}; wait $pid; ",
```
pieandcakes: It looks like you removed {1} and added a 4/5/6 but I don't see additional parameters being passed to String.Format().

pieandcakes: nm. I see in the final it is correct

edumunoz: That's for the first commit. I split the change into 2 commits to make reviewing easier, but I wasn't able to keep everything consistent in the intermediate commit.

The final commit has this fixed.

src/MICore/Debugger.cs
edumunoz: Fair enough. I added a comment that explains this use.

src/MICore/Transports/ITransport.cs
```diff
@@ -16,6 +16,12 @@ public interface ITransport
         void Send(string cmd);
         void Close();
         bool IsClosed { get; }
+
+        /// <summary>
+        /// It is used to know whether to fake a response from the debugger
```
gregg-miskelly: How about -- 
    Process id of the debugger (gdb/lldb-mi/clrdbg) process. It is used to know...

src/MICore/Transports/LocalUnixTerminalTransport.cs
```diff
@@ -33,49 +43,59 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
...
+                _dbgStdOutName,
+                pidFifo,
+                debuggerCmd);
 
             TerminalLauncher terminal = TerminalLauncher.MakeTerminal("DebuggerTerminal", launchDebuggerCommand, localOptions.Environment);
             terminal.Launch(debuggeeDir);
 
+            using (StreamReader pidReader = new StreamReader(pidStream, Encoding.UTF8, true, UnixUtilities.StreamBufferSize))
+            {
+                _debuggerPid = int.Parse(pidReader.ReadLine(), CultureInfo.InvariantCulture);
```
gregg-miskelly: Do we need to do something to handle the terminal process failing and therefore never running our bash script? (ex: wait on the terminal process to exit and if so cancel this read?)

andrewcrawley: The problem with waiting on the terminal process is that some terminals (e.g. gnome-terminal) work in a client-server model, where if an instance is already running, the new instance we launch will detect that, send the command to the existing instance, then exit immediately.

andrewcrawley:
LGTM.  Thanks for splitting the increasingly-complex terminal invocation into multiple lines : )

src/MICore/Transports/LocalUnixTerminalTransport.cs
gregg-miskelly: Maybe a timeout then? Ex: If the pid hasn't returned after 15 seconds something must be wrong.

---
In reply to: [62571354](https://github.com/Microsoft/MIEngine/pull/311#discussion_r62571354) [](ancestors = 62571354)

src/MICore/UnixUtilities.cs
```diff
@@ -26,19 +29,36 @@ public static class UnixUtilities
...
-                "cd {0}; DbgTerm=`tty`; trap 'rm {2}; rm {3}' EXIT; {1} --interpreter=mi --tty=$DbgTerm < {2} > {3};",
-                debuggeeDir,
-                debuggerCmd,
-                dbgStdInName,
-                dbgStdOutName
-                );
+            return string.Format(CultureInfo.InvariantCulture,
+               "cd {0}; " +
+               "DbgTerm=`tty`; " +
+               "trap 'echo {1} > {2}; rm {2}; rm {3}; rm {4}; rm {5}' EXIT; " +
```
jacdavis: Is there a race here that we may not read from pipe 2 before the rm {2} command?

src/MICore/Transports/StreamTransport.cs
```diff
@@ -207,9 +194,34 @@ public bool IsClosed
             get { return _bQuit; }
         }
 
+        public virtual int DebuggerPid
```
gregg-miskelly: Should this be abstract instead?

src/MICore/Transports/PipeTransport.cs
```diff
@@ -35,6 +36,14 @@ protected override string GetThreadName()
             return "MI.PipeTransport";
         }
 
+        public override int DebuggerPid
+        {
+            get
+            {
+                return _debuggerPid;
```
gregg-miskelly: For pipe transport we are using the pid of the pipe process and for TCP transport I think we are throwing NotImplementedException. It seems like these two should be consistent with each other. I might suggest either making this an int? and returning null in both cases, or using 0 as an invalid value and returning that in both cases.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -201,6 +201,15 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
...
                 }
+
+                // Only need to know the debugger pid on Linux and OS X local launch to detect whether
+                // the debugger is closed. If the debugger is not running anymore, the response (^exit)
+                // to the -gdb-exit command is faked to allow MIEngine to shut down.
+                _localDebuggerPid = localTransport.DebuggerPid;
+
+                // We are launching the debugger in the background, so the console will have
+                // "[job number] pid". So we clear that.
+                ResetConsole();
```
gregg-miskelly: This is currently not going to work for clrdbg - we don't support the shell command or the console exec command. Can we clear the console in the bash command instead?

src/MICore/Debugger.cs
```diff
@@ -55,6 +55,24 @@ public bool IsClosed
             }
         }
 
+        public bool IsDebuggerRunning
```
gregg-miskelly: This method is a bit weird in that it doesn't do what it says it does on Windows. That is fine for how it is being used now, but its not very obvious for the future. I would suggest -
* private instead of public
* A method instead of a property
* Something that indicates that it doesn't doing anything on Windows: IsUnixDebuggerRunning?

src/MICore/Debugger.cs
```diff
@@ -71,6 +89,7 @@ public bool IsClosed
         private LinkedList<string> _initialErrors = new LinkedList<string>();
 
         protected bool _connected;
+        protected int _localDebuggerPid = -1;
```
gregg-miskelly: nit: I know we made this mistake already in other places, but personally I don't like protected assignable fields. Suggestion: make this private and add a set method.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
jacdavis: resetting the terminal is very expensive. We had to do it that way to get rid of gdb's bogus error on launch. However, I think it might be overkill in this case. Let's try clear instead.

src/MICore/UnixUtilities.cs
```diff
@@ -26,19 +29,36 @@ public static class UnixUtilities
...
+               "{6} --interpreter=mi --tty=$DbgTerm < {3} > {4} & " +
+               "pid=$! ; " +
```
jacdavis: do we care that these won't execute if the user closes the terminal window?

edumunoz: Can you specify which commands in particular? On GitHub I can't tell which lines you refer to.

jacdavis: The ones after the gdb command, starting from pid=$!

src/MICore/Transports/PipeTransport.cs
edumunoz: `PipeTransport` is used when `useExternalConsole = false` even for local OS X and Linux, so we still need the pid in that case.

I haven't looked into when `TcpTransport` is used, but it doesn't seem like there's any access to any type of local process in that case.

src/MICore/UnixUtilities.cs
edumunoz: But they will execute because the debugger is started in the background (note the `&` at the end of that line).

src/MICore/Transports/PipeTransport.cs
gregg-miskelly: PipeTransport is also used with PipeLaunchOptions, which is the case I was thinking of. I suppose this is okay. In this case I would suggest updating the documentation for DebuggerPid to indicate that the behavior is undefined for non-local transport.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
faxue-msft: Actually sending reset string (ESC, c) isn't that expensive, it's take about a second to do a "full reset" by "echo reset"  which is much cleaner than sending (ESC, c), but we eventually didn't take that approach.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -295,7 +304,7 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
                     // This is to work around a GDB bug of warning "Failed to set controlling terminal: Operation not permitted"
                     // Reset debuggee terminal after the first module load.
                     // The clear is done by sending reset string (ESC, c) to terminal STDERR
```
faxue-msft: Move this comment to inside ResetConsole

src/MICore/UnixUtilities.cs
```diff
@@ -26,19 +29,36 @@ public static class UnixUtilities
...
-                debuggeeDir,
-                debuggerCmd,
-                dbgStdInName,
-                dbgStdOutName
-                );
+            return string.Format(CultureInfo.InvariantCulture,
+               "cd {0}; " +
+               "DbgTerm=`tty`; " +
+               "trap 'echo {1} > {2}; rm {2}; rm {3}; rm {4}; rm {5}' EXIT; " +
+               "{6} --interpreter=mi --tty=$DbgTerm < {3} > {4} & " +
```
faxue-msft: As discussed offline, debuggee might not be able to receive user input when gdb is run in background, make sure that's tested

src/MICore/UnixUtilities.cs
gregg-miskelly: It looks like if you move the starting of the debugger into a subshell by surrounding it with parenthesis bash will not print the background job notifications. Ex:

    "({6} --interpreter=mi ...
    ...
    wait $pid;)"

src/MICore/Transports/LocalUnixTerminalTransport.cs
```diff
@@ -2,16 +2,26 @@
...
         private string _dbgStdInName;
         private string _dbgStdOutName;
-        private FileSystemWatcher _fifoWatcher;
+        private int _debuggerPid = -1;
+
+        /// <summary>
+        /// When the debugger exits, it will write into the fifo associated with this reader.
+        /// This is useful in the case of attach when the root password is required, but the
+        /// request is canceled. In that case, the debugger won't have started yet (so no messages
+        /// will be processed from it), but the trap will execute and write into this fifo.
```
faxue-msft: The original bug the filewatcher trying to address is "killall -9 gdb", make sure VSCode exit debug mode when gdb is killed that way

src/MICore/UnixUtilities.cs
edumunoz: Thanks for finding this. I tried your and @jacdavis's suggestion of using `clear` instead of resetting the terminal and works fine.

We need the call to `clear` anyway (it used to be in `osxlaunchhelper.scpt`), but I moved it to the shell command.

src/MICore/Transports/StreamTransport.cs
edumunoz: I can see why. So the exception will be thrown explicitly in the subclasses if the property doesn't make sense, correct?

src/MICore/Debugger.cs
edumunoz: > A method instead of a property
Is this to explicitly indicate this involves some work and not just reading something off memory?

gregg-miskelly: Correct.

src/MICore/Transports/StreamTransport.cs
gregg-miskelly: Yes

src/MICore/Transports/LocalUnixTerminalTransport.cs
edumunoz: Verified this works both on Ubuntu and OS X.

src/MICore/UnixUtilities.cs
edumunoz: This works too, although in an upcoming PR I will probably change `wait $pid` with `fg`.

src/MICore/UnixUtilities.cs
edumunoz: As @andrewcrawley commented, as long as a process has a handle to it, the fifo will still carry the data.

src/MICore/Transports/ITransport.cs
```diff
@@ -16,6 +16,13 @@ public interface ITransport
         void Send(string cmd);
         void Close();
         bool IsClosed { get; }
+
+        /// <summary>
+        /// Process ID of the debugger process (clrdbg/lldb/gdb).
```
gregg-miskelly: I would also suggest adding -
This value is only valid when using local launch options. When using non-local options this may throw (TcpTransport) or provide bogus data (PipeTransport).

gregg-miskelly:
:shipit:

PR #312, opened 2016-05-09
Enabling detach for offroad scenario.

rajkumar42:
Note: For this to work, "detach all" from visual studio.

### DIFF
--- a/src/MICore/CommandFactories/clrdbg.cs
+++ b/src/MICore/CommandFactories/clrdbg.cs
@@ -41,11 +41,6 @@ namespace MICore
             return true;
         }
 
-        public override bool CanDetach()
-        {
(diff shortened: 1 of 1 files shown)


chuckries:
:+1:

src/MICore/CommandFactories/clrdbg.cs
```diff
@@ -43,8 +43,7 @@ public override bool AllowCommandsWhileRunning()
 
         public override bool CanDetach()
```
gregg-miskelly: You can just delete this method now. The default version returns true.

gregg-miskelly:
Otherwise looks good.

wesrupert:
:+1:

PR #313, opened 2016-05-10
Add support for cygwin on windows

jacdavis:
Cygwin maps files from windows paths to unix style
paths. It has a utility called cygpath that will
automatically do the conversion. Calling that utility
proved to be quite difficult because cygwin will fail
to initialize stacks if a 64bit app is connected to a
32 bit console. Under vscode, the 32 bit console is
inherited from vscode. This works around this by launching
that utility with process_detached to force a new console.
Note that doing so requires pinvoking to createprocess
directly since the Process object doesn't expose the
raw flags.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/CygwinFileMapper.cs
+++ b/src/MIDebugEngine/Engine.Impl/CygwinFileMapper.cs
@@ -1,0 +1,256 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
+using Microsoft.Win32.SafeHandles;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+using MICore;
+
+namespace Microsoft.MIDebugEngine
+{
+    internal class CygwinFilePathMapper
+    {
+        private DebuggedProcess _debuggedProcess;
+        private Dictionary<string, string> _cygwinToWindows;
+
+        public CygwinFilePathMapper(DebuggedProcess debuggedProcess)
+        {
+            this._debuggedProcess = debuggedProcess;
+            this._cygwinToWindows = new Dictionary<string, string>();
+        }
+
+        public string MapCygwinToWindows(string origCygwinPath)
+        {
+            if (!(_debuggedProcess.LaunchOptions is LocalLaunchOptions))
+            {
+                return origCygwinPath;
+            }
+
+            LocalLaunchOptions localLaunchOptions = (LocalLaunchOptions)_debuggedProcess.LaunchOptions;
+
+            string cygwinPath = origCygwinPath.Replace('\\', '/');
+
+            string windowsPath = cygwinPath;
+
+            lock (this._cygwinToWindows)
+            {
+                if (!this._cygwinToWindows.TryGetValue(cygwinPath, out windowsPath))
+                {
+                    if (!LaunchCygPathAndReadResult(cygwinPath, localLaunchOptions.MIDebuggerPath, out windowsPath))
+                    {
+                        return origCygwinPath;
+                    }
+
+                    this._cygwinToWindows.Add(cygwinPath, windowsPath);
+                }
+            }
+
+            return windowsPath;
+        }
+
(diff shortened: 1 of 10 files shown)


jacdavis:
@gregg-miskelly @chuckries @paulmaybee @andrewcrawley @edumunoz

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -594,6 +607,21 @@ private List<LaunchCommand> GetInitializeCommands()
                         commands.Add(new LaunchCommand("-gdb-set new-console on", ignoreFailures:true));
                     }
 
+                    // If running locally on windows, determine if gdb is running from cygwin
+                    if (localLaunchOptions != null &&
```
gregg-miskelly: Should we restrict this to GDB?

src/MIDebugEngine/Engine.Impl/CygwinFileMapper.cs
```diff
@@ -0,0 +1,256 @@
...
+using System.Runtime.InteropServices;
+using Microsoft.Win32.SafeHandles;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+using MICore;
+
+namespace Microsoft.MIDebugEngine
+{
+    class CygwinFilePathMapper
```
andrewcrawley: Accessibility modifier?

gregg-miskelly:
:shipit:

src/MIDebugEngine/Engine.Impl/CygwinFileMapper.cs
```diff
@@ -0,0 +1,256 @@
...
+                if (WaitForSingleObject(processInfo.hProcess, timeout) != 0)
+                {
+                    Debug.Fail("cygpath failed to map source file.");
+                    return false;
+                }
+
+                uint exitCode = 0;
+                if (!GetExitCodeProcess(processInfo.hProcess, out exitCode))
+                {
+                    Debug.Fail("cygpath failed to get exit code form cygpath.");
```
andrewcrawley: Typo: form -> from

src/MIDebugEngine/Engine.Impl/CygwinFileMapper.cs
```diff
@@ -0,0 +1,256 @@
...
+
+            return windowsPath;
+        }
+
+        // There is an issue launching Cygwin apps that if a process is launched using a bitness mismatched console,
+        // process launch will fail. To avoid that, launch cygpath with its own console. This requires calling CreateProcess 
+        // directly because the creation flags are not exposed in System.Diagnostics.Process. 
+        //
+        // Return true if successful. False otherwise.
+        private bool LaunchCygPathAndReadResult(string cygwinPath, string miDebuggerPath, out string windowsPath)
```
edumunoz: This could be made static.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -48,7 +50,7 @@ internal class DebuggedProcess : MICore.Debugger
         private ReadOnlyCollection<RegisterDescription> _registers;
         private ReadOnlyCollection<RegisterGroup> _registerGroups;
         private readonly EngineTelemetry _engineTelemetry = new EngineTelemetry();
-        private bool _needTerminalReset;
+        private bool _needTerminalReset;
```
edumunoz: Revert the added whitespace.

andrewcrawley:
LGTM, aside from the build break : )

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
jacdavis: I was originally thinking no because I'll eventually need a way to do this on lldb, but lldb won't support this command so I guess yes.

src/MIDebugEngine/Engine.Impl/CygwinFileMapper.cs
jacdavis: done

src/MIDebugEngine/Engine.Impl/CygwinFileMapper.cs
jacdavis: I don't typically use static as an optimization. The savings are very small and opens the code up to being used at inappropriate times. The purpose of static is to allow the method to be called from the class.

PR #314, opened 2016-05-11
Add the command used for core dump debugging in LLDB

edumunoz:
This was done by @jacdavis

There is no `-target-select core` command in lldb-mi. However, the MI Driver will translate CLI commands into MI commands if needed (when the command doesn't start with `-`).

/cc @gregg-miskelly @jacdavis


gregg-miskelly:
👍

jacdavis:
:shipit:

chuckries:
late to the party but it seems like this should be part of the command factory rather than floating in DebuggedProcess.

PR #315, opened 2016-05-11
Fix attach on cygwin which was missing my code mapping initialization

jacdavis:


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -591,6 +591,20 @@ private List<LaunchCommand> GetInitializeCommands()
                     int pid = localLaunchOptions.ProcessId;
                     commands.Add(new LaunchCommand(String.Format(CultureInfo.CurrentUICulture, "-target-attach {0}", pid), ignoreFailures: false));
 
+                    // If running locally on windows, determine if gdb is running from cygwin
+                    if (localLaunchOptions != null && PlatformUtilities.IsWindows() && this.MICommandFactory.Mode == MIMode.Gdb)
+                    {
+                        // mingw will not implement this command, but to be safe, also check if the results contains the string cygwin.
+                        LaunchCommand lc = new LaunchCommand("show configuration", null, true, null, new Action<string>((string resStr) => {
```
gregg-miskelly: Should we make this a helper method so the code can be shared between the two cases?

gregg-miskelly:
Otherwise LGTM

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
jacdavis: I thought about that. However, non of the options in these cases do that and it looked like it would require quite a few params. If it means getting a fedora'd hamster, I guess I can accomidate it.

PR #316, opened 2016-05-11
Users/jacdavis

jacdavis:
mingw enters break mode after an attach with a …  
bogus stopped record. This causes attach to fail.
Because the code paths past that point use the lack
of a frame to identify a module load, I created
a hack to generate a fake frame. No, I'm not proud
of it.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -42,6 +42,7 @@ namespace MICore
         private int _exiting;
         public ProcessState ProcessState { get; private set; }
         private MIResults _miResults;
+        public bool IsCygwin { get; protected set; }
 
         public virtual void FlushBreakStateData()
         {
@@ -1060,6 +1061,27 @@ namespace MICore
             {
                 string status = _miResults.ParseCString(cmd.Substring(8));
                 OnStateChanged("stopped", status);
+            }
+            else if (cmd.StartsWith("stopped", StringComparison.Ordinal))
+            {
+                if (PlatformUtilities.IsWindows() &&
+                    this.LaunchOptions is LocalLaunchOptions &&
+                    ((LocalLaunchOptions)this.LaunchOptions).ProcessId != 0 &&
+                    this.MICommandFactory.Mode == MIMode.Gdb &&
+                    !this.IsCygwin
+                    )
+                {
+                    // mingw enters break mode with no status during attach. All of the processing 
+                    // for *stopped expects status or it will assume the event is a module load.
+                    // to avoid regressing that code path, manufacture a fake frame. This will not
+                    // be seen by the user as it is continued during attach.
+                    string fakeMingwAttachFrame = "frame={addr=\"0x0000000000000000\",func=\"fakemodule!fakeMingwAttachFrame\"},thread-id=\"1\",stopped-threads=\"all\"";
+                    OnStateChanged("stopped", fakeMingwAttachFrame);
+                }
+                else
+                {
+                    Debug.Fail("Unknown out-of-band msg: " + cmd);
+                }
             }
             else if (cmd.StartsWith("running,", StringComparison.Ordinal))
             {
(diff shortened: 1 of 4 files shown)


jacdavis:
@gregg-miskelly @chuckries @edumunoz @andrewcrawley @paulmaybee

src/MICore/Debugger.cs
```diff
@@ -1061,6 +1062,27 @@ private void OnOutOfBand(string cmd)
                 string status = _miResults.ParseCString(cmd.Substring(8));
                 OnStateChanged("stopped", status);
             }
+            else if (cmd.StartsWith("stopped", StringComparison.Ordinal))
```
edumunoz: Isn't this condition exactly the same as the if statement right above?

src/MICore/Debugger.cs
```diff
@@ -1061,6 +1062,27 @@ private void OnOutOfBand(string cmd)
...
+            {
+                if (PlatformUtilities.IsWindows() &&
+                    this.LaunchOptions is LocalLaunchOptions &&
+                    ((LocalLaunchOptions)this.LaunchOptions).ProcessId != 0 &&
+                    this.MICommandFactory.Mode == MIMode.Gdb &&
+                    !this.IsCygwin
+                    )
+                {
+                    // mingw enters break mode with no status during attach. All of the processing
```
edumunoz: Can you clarify that `fakeMingwAttachFrame` is the actual status that is being faked?

src/MICore/Debugger.cs
```diff
@@ -1061,6 +1062,27 @@ private void OnOutOfBand(string cmd)
...
+                    this.MICommandFactory.Mode == MIMode.Gdb &&
+                    !this.IsCygwin
+                    )
+                {
+                    // mingw enters break mode with no status during attach. All of the processing 
+                    // for *stopped expects status or it will assume the event is a module load.
+                    // to avoid regressing that code path, manufacture a fake frame. This will not
+                    // be seen by the user as it is continued during attach.
+                    string fakeMingwAttachFrame = "frame={addr=\"0x0000000000000000\",func=\"fakemodule!fakeMingwAttachFrame\"},thread-id=\"1\",stopped-threads=\"all\"";
+                    OnStateChanged("stopped", fakeMingwAttachFrame);
```
gregg-miskelly: Would it be possible to either just drop this line on the ground or continue gdb? Or is there some sort of 'stopped' processing that we need to do here?

gregg-miskelly:
Otherwise signing off.

src/MICore/Debugger.cs
jacdavis: I think it might do entrypoint processing on this event. Specifically the line in DebuggedProcess.HandleBreakModeEvent if (String.IsNullOrWhiteSpace(reason) && !_bEntrypointHit).

Note setting _bEntrypointHit could be a source of bugs. Interestingly, I'm actually finding that mingw never sends entrypoint for normal cases either.

jacdavis: Yes, I've verified we need this.

src/MICore/Debugger.cs
```diff
@@ -1061,6 +1062,27 @@ private void OnOutOfBand(string cmd)
...
                 OnStateChanged("stopped", status);
             }
+            else if (cmd.StartsWith("stopped", StringComparison.Ordinal))
+            {
+                if (PlatformUtilities.IsWindows() &&
+                    this.LaunchOptions is LocalLaunchOptions &&
+                    ((LocalLaunchOptions)this.LaunchOptions).ProcessId != 0 &&
+                    this.MICommandFactory.Mode == MIMode.Gdb &&
+                    !this.IsCygwin
+                    )
```
chuckries: could this get hit in the MDD scenarios? in that case we using LocalLaunchOptions on windows.

chuckries:
signing off with comment.

src/MICore/Debugger.cs
jacdavis: If we're using gdb running under mingw and it has the same bug (incorrect format of stop command) then yes. But in that case, this is likely to behavior you want.

jacdavis: Oh, and local mdd won't have process id set

PR #317, opened 2016-05-12
Add prompt when attaching as superuser

edumunoz:
The prompt defaults to `N`. I'm open to defaulting to `y`.

/cc @andrewcrawley @jacdavis @gregg-miskelly

### DIFF
--- a/src/MICore/MICoreResources.Designer.cs
+++ b/src/MICore/MICoreResources.Designer.cs
@@ -354,5 +354,14 @@ namespace MICore {
                 return ResourceManager.GetString("Status_BreakpointPending", resourceCulture);
             }
         }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to Superuser access is required to attach to a process. Attaching as superuser can potentially harm your computer. Do you want to continue? [y/N].
+        /// </summary>
+        public static string Warn_AttachAsRootProcess {
+            get {
+                return ResourceManager.GetString("Warn_AttachAsRootProcess", resourceCulture);
+            }
+        }
     }
 }
--- a/src/MICore/MICoreResources.resx
+++ b/src/MICore/MICoreResources.resx
@@ -222,4 +222,7 @@ Error: {1}</value>
   <data name="Error_InternalFileMissing" xml:space="preserve">
     <value>Internal file "{0}" could not be found.</value>
   </data>
+  <data name="Warn_AttachAsRootProcess" xml:space="preserve">
+    <value>Superuser access is required to attach to a process. Attaching as superuser can potentially harm your computer. Do you want to continue? [y/N]</value>
+  </data>
 </root>
--- a/src/MICore/UnixUtilities.cs
+++ b/src/MICore/UnixUtilities.cs
@@ -66,6 +66,7 @@ namespace MICore
             if (PlatformUtilities.IsLinux())
             {
                 string debuggerPathCorrectElevation = localOptions.MIDebuggerPath;
+                string prompt = string.Empty;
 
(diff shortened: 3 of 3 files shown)


jacdavis:
:neckbeard:

src/MICore/Transports/LocalUnixTerminalTransport.cs
```diff
@@ -121,6 +124,8 @@ public override void Close()
             // If we are shutting down before the _exitReader has read a line it's possible that
             // there is a thread blocked doing a read() syscall. 
             ForceDisposeStreamReader(_exitReader);
+            _streamReadPidCancellationTokenSource.Cancel();
+            _streamReadPidCancellationTokenSource.Dispose();
```
gregg-miskelly: If you hit stop debugging while this is up, are we far enough along to get to Close?

gregg-miskelly:
:shipit:

src/MICore/Transports/LocalUnixTerminalTransport.cs
edumunoz: I just tried this, and no, we don't hit `Close()` or anything for that matter.

Any suggestions?

edumunoz: I was thinking to add a `SetLocalDebuggerPid` method to `ITransportCallback`, and move the `pidReadTask` to a different thread.

gregg-miskelly: In AD7, this isn't a scenario that exists (launch is sync). I think this would be my suggestion:

1. AD7Engine - make this implement a new IVSCodeEngineLaunch interface which we can define in the host assembly. This isn't the best place for it, but it seems like the best option to me. Ideally this would have a very different signature than the current LaunchSuspended method, as I think the current LaunchSuspended method is actually the wrong thing for VS Code. But for now we can just fix the obvious problem with it and add a CancellationToken to it but otherwise leave it unchanged.
2. Plumb on through the cancellation token
3. In OpenDebugAD7, add in a copy of the IVSCodeEngineLaunch cs file
4. Set the launch cancellation token if stop debugging is hit

andrewcrawley:
LGTM, +1, squirrels, etc.

edumunoz:
@Andrew-MSFT @johnkemnetz @delmyers can you review the message?

delmyers:
My only comment is to question whether the term "superuser" is generic enough to be understood by users on multiple platforms. Otherwise, it looks good.

chuckries:
I think 'superuser' is definitely common parlance in *nix world.

Andrew-MSFT:
Do we want the words "hard your computer" or is the bigger concern opening it up to security vulnerabilities?

edumunoz:
I would say it's the most generic and accurate term.

https://en.wikipedia.org/wiki/Superuser

edumunoz:
@Andrew-MSFT 

Doesn't
> **can potentially** harm your computer

imply

> opening it up to security vulnerabilities

?

That's the wording used in VS as well (included in the email I sent you).

Andrew-MSFT:
@edumunoz 

Yes, **can potentially harm** is a superset of **open up to security vulnerabilities**, my thought was wondering if it's overly scary.  I don't have a strong opposition to it, but if we think it's primarily a security concern then I'd keep the message more specific rather than be generically scary.  I'm not sure any science was put into the VS message (e.g. I don't recall us ever having a conversation about it)

edumunoz:
Followed up with @Andrew-MSFT over email. Keeping it as is.

PR #318, opened 2016-05-12
Fix bug 220654 Cannot enter breakmode with object on the stack

jacdavis:
Issue here is g++ mapped one address to two source lines for the
constructor. What I did is make the line mapping collection
handle that. Since we don't know what the "correct" line is,
it just chooses one, unless the old line is line number 0.


paulmaybee:
👍

jacdavis:
@gregg-miskelly @edumunoz @andrewcrawley @chuckries

jacdavis:
@richardstanton

PR #319, opened 2016-05-12
Query debugger for target architecture instead of relying on launch o…

paulmaybee:
…ptions


paulmaybee:
@jacdavis @gregg-miskelly

src/MICore/Debugger.cs
```diff
@@ -379,8 +379,11 @@ public void Init(ITransport transport, LaunchOptions options)
             FlushBreakStateData();
 
             _transport.Init(this, options, Logger);
+        }
```
gregg-miskelly: Did you mean to call SetTargetArch(options.TargetArchitecture) here?

gregg-miskelly:
In LanchOptions.xsd(192) TargetArchitecture is marked as required. We should make it optional.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -496,7 +496,13 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
                     }
                 }
 
-
+                var arch = await MICommandFactory.GetTargetArchitecture();
+                if (arch == TargetArchitecture.Unknown)
```
gregg-miskelly: Right now we still have target architecture in the launch options, but we are ignoring it. It seems like we should either:
1. Only ask the debugger for its target architecture if it isn't in the launch options
-or-
2. Always ask the debugger for its target architecture, but complain if it the launch options has a target architecture, and it is different from the debugger

src/MICore/CommandFactories/gdb.cs
```diff
@@ -225,5 +225,38 @@ private static string TypeBySize(uint size)
         }
 
         public override bool SupportsDataBreakpoints { get { return true; } }
+
+        public override async Task<TargetArchitecture> GetTargetArchitecture()
+        {
+            string cmd = "show architecture";
```
gregg-miskelly: Just to make sure, this gives the architecture of GDB, or the target? One potential issue - what if there was a GDB that supported multiple target architectures?

paulmaybee: This is the architecture of the target, it will change depending upon the target type.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
paulmaybee: the trouble is that since we never checked the architecture value against the target type it was often specified incorrectly. For example, in the first VSIX for Linux debugging it is hardcoded to arm. So I think 1 is not a good choice. We could do 2, but I would be opposed to halting the debugging session for this reason. Were you thinking of a popup for this, or just an error message to the output window? @gregg-miskelly

jacdavis:
:shipit:

src/MICore/Debugger.cs
paulmaybee: Actually I'd like to get rid of the launch option so that there isn't any issue with conflicts. is there some way to indicate that the use of the attribute is deprecated?

jacdavis: For vscode, from the Json I don't think there is a great way to do it, but removing it from the schema will cause a squiggly in the ui. We could also print an output message if it is set stating it is deprecated. that would work in vs and vscode.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
gregg-miskelly: Ug. How about this then - ask the debugger for its target architecture, if we understand it -- good. If not, check if it is in the launch options. If it is -- use that. If it is missing from both places, send a warning to the output window and guess x64.

paulmaybee:
@gregg-miskelly Do you want to take a look at the second commit?

gregg-miskelly:
:shipit:

PR #320, opened 2016-05-13
Remove myget.org package sources

gregg-miskelly:
This will be completed once packages have been pushed to nuget.org.


gregg-miskelly:
Verified everything worked correctly for Windows, Ubuntu, RHEL, and Debian.

PR #322, opened 2016-05-13
Monitor the shell pid and remove the exit fifo

edumunoz:
The exit fifo was causing issues in the case that MIEngine had already exited
but the shell started executing the `EXIT` trap. In that case, it would block on
echoing 'exit' to a fifo since there were no readers.

Use `getpgid` instead of `kill -0` to find if a process is running

@jacdavis @andrewcrawley @gregg-miskelly

### DIFF
--- a/src/MICore/ProcessMonitor.cs
+++ b/src/MICore/ProcessMonitor.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading;
+
+namespace MICore
+{
+    public class ProcessMonitor: IDisposable
+    {
+        private readonly TimeSpan EXIT_POLL_DELTA = TimeSpan.FromMilliseconds(200);
+        private int _processId;
+        private Timer _exitMonitorTimer;
+
+        public ProcessMonitor(int processId)
+        {
+            if (!PlatformUtilities.IsLinux() && !PlatformUtilities.IsOSX())
+            {
+                throw new NotImplementedException();
+            }
+
+            _processId = processId;
+        }
+
+        public void Start()
+        {
+            _exitMonitorTimer = new Timer(MonitorForExit, null, TimeSpan.FromMilliseconds(0), EXIT_POLL_DELTA);
+        }
+
+        public event EventHandler ProcessExited;
+
+        private bool HasExited()
+        {
+            return !UnixUtilities.IsProcessRunning(_processId);
+        }
+
+        private void MonitorForExit(object o)
+        {
+            if (HasExited())
+            {
+                _exitMonitorTimer.Change(Timeout.Infinite, Timeout.Infinite);
+                ProcessExited?.Invoke(this, null);
+            }
+        }
+
+        public void Dispose()
+        {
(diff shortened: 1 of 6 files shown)


src/MICore/ProcessMonitor.cs
```diff
@@ -0,0 +1,51 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading;
+
+namespace MICore
+{
+    public class ProcessMonitor: IDisposable
+    {
+        private readonly TimeSpan EXIT_POLL_DELTA = TimeSpan.FromMilliseconds(200);
```
edumunoz: I'm open to suggestions for a different value.

src/MICore/Transports/LocalUnixTerminalTransport.cs
```diff
@@ -64,24 +56,38 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
...
             terminal.Launch(debuggeeDir);
 
+            int shellPid = -1;
+
             using (StreamReader pidReader = new StreamReader(pidStream, Encoding.UTF8, true, UnixUtilities.StreamBufferSize))
             {
-                Task<string> readPidTask = pidReader.ReadLineAsync();
-                if (readPidTask.Wait(TimeSpan.FromSeconds(10)))
+                Task<string> readShellPidTask = pidReader.ReadLineAsync();
+                if (readShellPidTask.Wait(TimeSpan.FromSeconds(5)))
```
gregg-miskelly: Sometimes things take *much* longer than you would expect, so short timeouts can be really dangerous. I would leave this at 10.

src/MICore/ProcessMonitor.cs
```diff
@@ -0,0 +1,51 @@
...
+        private bool HasExited()
+        {
+            return !UnixUtilities.IsProcessRunning(_processId);
+        }
+
+        private void MonitorForExit(object o)
+        {
+            if (HasExited())
+            {
+                _exitMonitorTimer.Change(Timeout.Infinite, Timeout.Infinite);
```
gregg-miskelly: Would it make sense to Dispose this instead?

src/MICore/Transports/LocalUnixTerminalTransport.cs
```diff
@@ -118,9 +121,9 @@ public override void Close()
         {
             base.Close();
 
-            // If we are shutting down before the _exitReader has read a line it's possible that
-            // there is a thread blocked doing a read() syscall. 
-            ForceDisposeStreamReader(_exitReader);
+            _shellProcessMonitor.Dispose();
```
gregg-miskelly: I think this should be ?.

gregg-miskelly:
:shipit:

src/MICore/ProcessMonitor.cs
edumunoz: You are right. I had the idea in my mind that the same `ProcessMonitor` could be reused even after the process has exited. I hadn't taken into account the fact that `_processId` can't be changed and there's no point in doing so.

src/MICore/Transports/LocalUnixTerminalTransport.cs
edumunoz: Yup

PR #324, opened 2016-05-16
Interrupt before terminate

edumunoz:
Issue a break to the debuggee before terminating it.

On OS X, gdb would not be able to `kill` the debuggee as part of executing `Terminate()` if the debuggee is not interrupted.

I tested this on OS X and Ubuntu 14.04. I don't mind special casing this for OS X, but given we have the same pattern for `CmdDetach()`, I don't see how this could do any harm on other platforms.

@gregg-miskelly @jacdavis

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -517,6 +517,11 @@ namespace MICore
 
         public async Task<Results> CmdTerminate()
         {
+            if (ProcessState == ProcessState.Running)
+            {
+                await CmdBreak();
+            }
+
             await MICommandFactory.Terminate();
 
             return new Results(ResultClass.done);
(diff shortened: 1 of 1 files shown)


gregg-miskelly:
👍

src/MICore/Debugger.cs
```diff
@@ -517,6 +517,11 @@ protected bool IsCoreDump
 
         public async Task<Results> CmdTerminate()
         {
+            if (ProcessState == ProcessState.Running)
+            {
+                await CmdBreak();
+            }
```
jacdavis: This isn't a bug in your stuff, but the cmdbreak that uses signals returns before the process breaks. We really need to get that fixed at some point.

jacdavis:
🐑

jacdavis:
:shipit:

PR #325, opened 2016-05-17
Add generated design file for LaunchOptions.xsd on CoreCLR

edumunoz:
There was a file that needed updating from #319.

/cc @caslan @paulmaybee


caslan:
LGTM

PR #326, opened 2016-05-17
Use the lab builds which is what actually gets built when shipping

edumunoz:
Use the lab builds, which would've caught the build break here: #325.

/cc @caslan @andrewcrawley

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -6,7 +6,7 @@ def project = GithubProject
 def branch = GithubBranchName
 
 [true, false].each { isPR ->
-    ['Debug', 'Release', 'Desktop.Debug', 'Desktop.Release'].each { configuration ->
+    ['Lab.Debug', 'Lab.Release', 'Desktop.Debug', 'Desktop.Release'].each { configuration ->
         def newJob = job(Utilities.getFullJobName(project, configuration.toLowerCase(), isPR)) {
             steps {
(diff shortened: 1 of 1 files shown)


edumunoz:
@mmitche Can you verify this is ok?

caslan:
LGTM

netci.groovy
```diff
@@ -6,7 +6,7 @@ def project = GithubProject
 def branch = GithubBranchName
 
 [true, false].each { isPR ->
-    ['Debug', 'Release', 'Desktop.Debug', 'Desktop.Release'].each { configuration ->
+    ['Lab.Debug', 'Lab.Release', 'Desktop.Debug', 'Desktop.Release'].each { configuration ->
```
andrewcrawley: This will build the Lab version of the CoreCLR config, but the non-lab version of the Desktop config.  The Desktop configs don't have a separate "Lab" version.  You should be able to leave this line alone and specify "/p:Lab=true" in the commandline a few lines below here, which should work for both versions.

edumunoz:
As @andrewcrawley and I discussed yesterday, using Lab builds (both by explicitly specifying them or using Andrew's suggestion `/p:Lab=true`) doesn't run the unit tests.

Andrew suggested we could add a property to run unit tests, so in the end the option would be `/p:Lab=true;Tests=true`

@caslan, any objections to that?

chuckries:
@edumunoz what do we want to do with this PR?

edumunoz:
I haven't had time to look into adding an argument to run the tests.

I was hoping @caslan would give the heads up.

caslan:
Sorry, I missed the earlier communication. I don't have any issue with additional argument for tests.

PR #327, opened 2016-05-17
Enable mingw debugging to async-break. 

jacdavis:
Enable mingw debugging to async-break. This is used for breakpoint
setting in run mode and for async-break. Note that this does not
work for cygwin which has a number of issues blocking async-break.
For the first release, the user will have to manually ctrl-c the
debuggee on cygwin.

### DIFF
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -547,7 +547,38 @@ namespace MICore
 
         public virtual bool IsAsyncBreakSignal(Results results)
         {
-            return (results.TryFindString("reason") == "signal-received" && results.TryFindString("signal-name") == "SIGINT");
+            bool isAsyncBreak = false;
+            
+            if (results.TryFindString("reason") == "signal-received")
+            {
+                if (results.TryFindString("signal-name") == "SIGINT")
+                {
+                    isAsyncBreak = true;
+                }
+                else if (results.TryFindString("signal-name") == "SIGTRAP")
+                {
+                    // Mingw has no way to send the sigint and using windows console control won't work with the debuggee
+                    // redirected. This means mingw will see SIGTRAP on async-break.
+                    if (this._debugger.IsRequestingInternalAsyncBreak || this._debugger.IsRequestingRealAsyncBreak)
+                    {
+                        if (this._debugger.IsLocalGdb() && PlatformUtilities.IsWindows() && !this._debugger.IsCygwin)
+                        {
+                            ResultValue frameResult;
+                            if (results.TryFind("frame", out frameResult))
+                            {
+                                // The top frame will be in an unknown function for break injected bps since it is actually
+                                // an injected frame in ntdll doing a debug break.
+                                // NOTE: if it were possible to get the windows stack that shows the inserted break frames,
+                                // it would be better. Unfornately, gdb doesn't show them. Perhaps check if the address
+                                // lands in ntdll or kernel32?
+                                isAsyncBreak = frameResult.TryFindString("func") == "??";
+                            }
+                        }
+                    }
+                }
+            }
+
+            return isAsyncBreak;
         }
 
         /// <summary>
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -11,6 +11,7 @@ using System.Threading.Tasks;
 using System.Globalization;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.IO;
 
 namespace MICore
 {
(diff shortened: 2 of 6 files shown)


jacdavis:
@gregg-miskelly @edumunoz @andrewcrawley @paulmaybee @chuckries

src/MICore/Debugger.cs
```diff
@@ -544,15 +568,10 @@ public Task CmdBreakInternal()
         {
             this.VerifyNotDebuggingCoreDump();
 
-            // TODO May need to fix attach on windows.
-            // Note that interrupt doesn't work when attached on OS X with gdb:
+            // Note that interrupt doesn't work on OS X with gdb:
```
edumunoz: Can you reset line 571 to what it was before (specifying it's only when attached)

chuckries:
:shipit:

chuckries:
@dotnet-bot build please

src/MICore/Debugger.cs
```diff
@@ -716,6 +748,30 @@ private Task<Results> CmdUnixBreak(int debugeePid, ResultClass expectedResultCla
...
+            {
+                foreach (int pid in this._debuggeePids.Values)
+                {
+                    try
+                    {
+                        // Mingw has no kill utility. Instead, use kernel32!DebugBreakProcess. Note that this
+                        // can deadlock if the loader lock is taken by a suspended thread. This approach does
+                        // not work in cygwin
+                        Process p = Process.GetProcessById(pid);
+                        WindowsNativeMethods.DebugBreakProcess(p.Handle);
```
chuckries: this is breaking the coreclr PR build.

chuckries: It looks like SafeHandle is available in CoreCLR


---
In reply to: [63624644](https://github.com/Microsoft/MIEngine/pull/327#discussion_r63624644) [](ancestors = 63624644)

gregg-miskelly:
:shipit:

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -547,7 +547,41 @@ public virtual bool SupportsFrameFormatting
...
+                            if (results.TryFind("frame", out frameResult))
+                            {
+                                // The top frame will be in an unknown function for break injected bps since it is actually
+                                // an injected frame in ntdll doing a debug break.
+                                // NOTE: if it were possible to get the windows stack that shows the inserted break frames,
+                                // it would be better. Unfornately, gdb doesn't show them. Perhaps check if the address
+                                // lands in ntdll or kernel32?
+                                if (frameResult.TryFindString("func") == "??")
+                                {
+                                    isAsyncBreak = true;
```
edumunoz: nit:

`isAsyncBreak = frameResult.TryFindString("func") == "??"`

edumunoz:
:ship:

src/MICore/Debugger.cs
jacdavis: thanks, I'll get that fixed before checking in.

src/MICore/Debugger.cs
```diff
@@ -196,7 +205,14 @@ public Task AddInternalBreakAction(Func<Task> func)
                         CmdBreakInternal();
                         _retryCount = 0;
                         _waitingToStop = true;
-                        _breakTimer = new Timer(RetryBreak, null, BREAK_DELTA, BREAK_DELTA);
+
+                        // When using signals to stop the proces, do not kick off another break attempt. The debug break injection and
```
faxue-msft: typo

src/MICore/Debugger.cs
```diff
@@ -716,7 +748,36 @@ private Task<Results> CmdUnixBreak(int debugeePid, ResultClass expectedResultCla
...
+                {
+                    try
+                    {
+                        // Mingw has no kill utility. Instead, use kernel32!DebugBreakProcess. Note that this
+                        // can deadlock if the loader lock is taken by a suspended thread. This approach does
+                        // not work in cygwin
+                        Process p = Process.GetProcessById(pid);
+
+#if CORECLR
+                        WindowsNativeMethods.DebugBreakProcess(p.SafeHandle);
```
chuckries: SafeHandle returns a `Microsoft.Win32.SafeHandles.SafeProcessHandle`, not an `IntPtr`, so I'm pretty sure this wont' work as expected.

jacdavis: No, that doesn't exist on fx 4.0 which is what this targets.

chuckries: `SafeHandle` is also supported on desktop CLR, so I think we should just use SafeHandle.DangerousGetHandle() in all cases.

jacdavis: The SafePocessHandle does not exist on the version of corefx targeted by this code.

PR #328, opened 2016-05-18
Add a warning if the debugger is not signed and running on OS X

edumunoz:
@gregg-miskelly @jacdavis @andrewcrawley @chuckries

### DIFF
--- a/src/MICore/UnixUtilities.cs
+++ b/src/MICore/UnixUtilities.cs
@@ -20,6 +20,9 @@ namespace MICore
         // Linux specific
         private const string PtraceScopePath = "/proc/sys/kernel/yama/ptrace_scope";
 
+        // OS X specific
+        private const string CodeSignPath = "/usr/bin/codesign";
+
         /// <summary>
         /// Launch a new terminal, spin up a new bash shell, cd to the working dir, execute a tty command to get the shell tty and store it.
         /// Start the debugger in mi mode setting the tty to the terminal defined earlier and redirect stdin/stdout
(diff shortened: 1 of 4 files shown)


src/MICore/UnixUtilities.cs
```diff
@@ -160,5 +163,28 @@ internal static bool IsProcessRunning(int processId)
             // if there is no process with the ID specified.
             return UnixNativeMethods.GetPGid(processId) >= 0;
         }
+
+        public static bool IsBinarySigned(string filePath)
```
gregg-miskelly: Would it make sense to cache this to avoid needing to re-check the debugger every time?

gregg-miskelly:
Otherwise LGTM

andrewcrawley:
My only concern here is that if the user has taken the option of disabling SIP instead of signing the binary, they're going to get a warning popup at every launch anyway.  Doesn't GDB print some error message about Mach ports or something if it fails due to the signing issue?  It might be better to detect that error and tell the user how to fix it rather than warning proactively.

edumunoz:
> if the user has taken the option of disabling SIP instead of signing the binary

We will point our instructions to the gnu instructions that only specify self-signing.

> they're going to get a warning popup at every launch anyway

This is output to the Debug Window at the moment.

> Doesn't GDB print some error message about Mach ports or something if it fails due to the signing issue?

Yeah, that was my original plan. However, as we have discussed today, the errors don't get written into the pipe until a different command gets executed (gdb bug) when stdout is redirected, so it leaves as not knowing this has occurred.

src/MICore/UnixUtilities.cs
edumunoz: We are only checking once, so not sure I follow why we would cache it.

On top of that, signing is a "dynamic" property. All the user has to do is execute

`$ codesign -f -s  "<cert name>"  <debugger path>`

andrewcrawley:
Fair enough.  If the intent is to display the message in the output window, is there a reason you used MT_MESSAGEBOX instead of MT_OUTPUTSTRING?

edumunoz:
The intent is to show it prominently, ideally as a message from the top in VSCode. If people are really annoyed because they decide to turn off the security feature, then we can revisit this (but I don't see this happening).

src/MICore/UnixUtilities.cs
gregg-miskelly: I was thinking of using a static. But thinking about it more, since VS Code will recreate the process each time, that isn't going to help.

Do we know what all codesign will do to check the signature? Codesign checking on Windows is quite an expensive operation as it needs to read every byte of the file and hash it. Then it needs to go off to web services to check if the certificate has been revoked. Do we know if this will do anything like that?

jacdavis:
:shipit:

src/MICore/UnixUtilities.cs
edumunoz: I believe `codesign -v` will do the hashing. I wasn't able to check whether it verifies against revocation.

I don't think that would be an issue for self-signed binaries (like we are suggesting right now in the README).

edumunoz:
Changing the target to branch release-cpptools

PR #329, opened 2016-05-18
Child2

paulmaybee:
Child process debugging. @gregg-miskelly @jacdavis


src/MICore/Debugger.cs
```diff
@@ -569,6 +571,13 @@ public Task CmdBreakInternal()
                     return CmdUnixBreak(debuggeePid, ResultClass.done);
                 }
             }
+            else if (_transport is PipeTransport)
```
gregg-miskelly: Seems like this should be under an 'if gdb' check.

src/MICore/Debugger.cs
```diff
@@ -569,6 +571,13 @@ public Task CmdBreakInternal()
...
+            {
+                if (((PipeTransport)_transport).Interrupt(PidByInferior("i1")))
```
gregg-miskelly: Looks like this will throw if the inferior isn't found. Is that what we want?

src/MICore/Debugger.cs
```diff
@@ -1131,6 +1148,7 @@ private void OnNotificationOutput(string cmd)
             {
                 results = _miResults.ParseResultList(cmd.Substring("thread-group-exited,".Length));
                 HandleThreadGroupExited(results);
+                ThreadGroupExitedEvent(this, new ResultEventArgs(results, 0));
```
gregg-miskelly: For consistency, this should probably be under a null check.

src/MICore/Debugger.cs
gregg-miskelly: PipeTransport isn't sealed and we inherit from it for LocalTransport and ServerTransport. Should this really check the launch options to see if this is a pipe server scenario?

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -831,10 +854,14 @@ public int ExecuteOnThread(IDebugThread2 pThread)
...
         //
-        // The sample engine only supports debugging native applications and therefore only has one program per-process
         public int Stop()
         {
-            throw new NotImplementedException();
+            DebuggedProcess.WorkerThread.RunOperation(async () =>
+            {
+                await _debuggedProcess.CmdBreak();
+            });
+            // TODO: this should be returning S_ASYNC_STOP
```
gregg-miskelly: hmm... I am a little concerned that we aren't going to get to this for update 3. Well, hopefully this isn't too much of a problem.

src/MICore/LaunchOptions.xsd
```diff
@@ -300,6 +290,11 @@
               <xs:documentation>Any arguments to pass to this program.</xs:documentation>
             </xs:annotation>
           </xs:attribute>
+          <xs:attribute name="PipeCommandArguments" type="xs:string" use="optional">
```
gregg-miskelly: I had trouble understanding what this is for from the name+description.

How about:
Name: PipeUnixShellArguments
Description: Arguments passed to the pipe executable to open up a remote Unix (ex: bash or similar) shell. This is used for running 'kill' on the remote system, or other commands that the debugger may need. If not specified, the debugger will do its best without the shell. But some features, such as setting breakpoints in run mode, may not work.

src/MICore/Transports/PipeTransport.cs
```diff
@@ -24,13 +24,38 @@ public class PipeTransport : StreamTransport
...
+            proc.StartInfo.WorkingDirectory = System.IO.Path.GetDirectoryName(_pipePath);
+            proc.EnableRaisingEvents = false;
+            proc.StartInfo.RedirectStandardInput = false;
+            proc.StartInfo.RedirectStandardOutput = false;
+            proc.StartInfo.RedirectStandardError = false;
+            proc.StartInfo.UseShellExecute = false;
+            proc.StartInfo.CreateNoWindow = true;
+            proc.Start();
+            proc.WaitForExit();
+            return true;
```
gregg-miskelly: If things fail it is going to be really tricky to figure out why. Should we at least grab stderr/out and dump it to the log? Should we check to see if the pipe command succeeds?

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -135,18 +146,60 @@ internal void MarkDirty()
...
+        internal void ThreadCreatedEvent(int id, string groupId)
+        {
+            lock (_threadList)
+            {
+                var thread = _threadList.Find(t => t.Id == id);
+                if (thread == null)
+                {
+                    _stateChange = true;
+                }
+                if (!_threadGroups.ContainsKey(groupId))
```
gregg-miskelly: should this have a '!string.IsNullOrEmpty(groupId)'?

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -135,18 +146,60 @@ internal void MarkDirty()
...
+                }
             }
         }
 
+        internal void ThreadGroupExitedEvent(string groupId)
+        {
+            lock (_threadList)
+            {
+                if (_threadGroups.ContainsKey(groupId))
+                {
```
gregg-miskelly: I think remove is already fine if the provided key doesn't exist.

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -135,18 +146,60 @@ internal void MarkDirty()
...
+                    _threadGroups.Remove(groupId);
+                }
+            }
+        }
+
+        private bool IsInParent(int tid)
```
gregg-miskelly: This is going to give the wrong answer for clrdbg - we don't use thread groups. Not sure about other scenarios.

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -135,18 +146,60 @@ internal void MarkDirty()
...
+        {
+            // only those threads in the "i1" threadgroup are in the debugee, others are transient while attaching to a child process
+            return _threadGroups["i1"].Contains(tid);
```
gregg-miskelly: Could this instead work by just having the thread remember what thread group it is in and change this to:

return String.IsNullOrEmpty(_threadGroup) || _threadGroup == "i1"

src/MICore/Debugger.cs
paulmaybee: Yes, that would be a better check. I'll also check for gdb since this is only a gdb scenario (at least for now).

src/MICore/Debugger.cs
paulmaybee: Probably fine to throw since it would be a very bad situation when thread group 'i1' wasn't found as that is the debuggee processes thread group. However I'll add the check.

src/MICore/Debugger.cs
paulmaybee: done

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
paulmaybee: I am not seeing any issues in my testing. What difference in behavior is expected?

src/MICore/LaunchOptions.xsd
paulmaybee: done

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
gregg-miskelly: There are two problems with this code as-is:
1. I believe the engine will send an async break event. This can result in races where depending the UI could re-enter break mode after it has already continued. 
2. The engine is telling the UI that it is broken when it isn't yet - exec-interrupt doesn't necessarily synchronously break the target. But by returning S_OK we are telling the engine that we are broken. This may mean that we try to do a bunch of things that will not actually work since we aren't broken. Probably the worst of these things is that we could try to continue the target and then possibly wind up in a state where the UI thinks we are running while we are actually broken.

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
paulmaybee: done

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
paulmaybee: done

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
paulmaybee: fixed

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
paulmaybee: If there is no thread group then use "i1" as the default thread group. All clrdbg threads will be added to this group.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
paulmaybee: Ok, so then what behavior is expected of the engine? I can certainly return S_ASYNC_STOP here, but I assume some other behavior is also then required.

gregg-miskelly: 1. _requestingRealAsyncBreak should change to a tristate with reasons of AsyncBreak, AD7Stop, or InternalBreak.
2. In the case of stop, we should examine the MI result to determine if GDB is already stopped - in which case we should return S_OK, or if it will stop in the future, in which case we should return S_ASYNC_STOP
3. The code that is checking _requestingRealAsyncBreak should raise a stop complete event if the tristate is in 'AD7Stop' mode.

src/DebugEngineHost/HostDebugger.cs
```diff
@@ -0,0 +1,50 @@
...
+            }
+        }
+
+        private static class Internal
+        {
+            public static void LaunchDebugTarget(string filePath, string options, Guid engineId)
+            {
+                IVsDebugger4 debugger = (IVsDebugger4)Package.GetGlobalService(typeof(IVsDebugger));
+                VsDebugTargetInfo4[] debugTargets = new VsDebugTargetInfo4[1];
+                debugTargets[0].dlo = (uint)DEBUG_LAUNCH_OPERATION.DLO_CreateProcess;
```
jacdavis: Is this the correct DLO for attach?

src/MICore/CommandFactories/gdb.cs
```diff
@@ -258,5 +263,17 @@ private static string TypeBySize(uint size)
             }
             return TargetArchitecture.Unknown;
         }
+
+        public override async Task Signal(string sig)
```
jacdavis: We have other code paths that do this without an interpreter exec. Consider naming this in a way to call out the difference?

src/MICore/CommandFactories/gdb.cs
```diff
@@ -258,5 +263,17 @@ private static string TypeBySize(uint size)
...
+        {
+            string command = String.Format("-interpreter-exec console \"signal {0}\"", sig);
+            await _debugger.CmdAsync(command, ResultClass.running);
```
jacdavis: Is the result class actually running after sending a signal? I'd expect it to be break, at least for sigint and signtrap.

paulmaybee: Stopping is reported with a *stopped event. The signal command does not synchronously report the stop.

src/MICore/LaunchOptions.cs
```diff
@@ -411,7 +386,25 @@ public abstract class LaunchOptions
 
         public MIMode DebuggerMIMode { get; set; }
 
+        private Xml.LaunchOptions.BaseLaunchOptions _baseOptions;
```
jacdavis: Does base options remove any option that doesn't make sense like processid? I assume that's the motivation behind the refactoring.

src/MICore/CommandFactories/gdb.cs
paulmaybee: What other code paths? This is simply executing the debugger's 'signal' command via MI. I think any other code paths must be executing a shell command so shouldn't affect the MICommandFactory.

src/MICore/Transports/PipeTransport.cs
jacdavis: Why not just pinvoke the kill api?

paulmaybee: The kill must execute on the remote machine.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -907,7 +929,23 @@ private void Dispose()
                     {
                         code = EngineUtils.SignalMap.Instance[sigName];
                     }
-                    _callback.OnException(thread, sigName, results.Results.TryFindString("signal-meaning"), code);
+                    bool stoppedAtSIGSTOP = false;
+                    if (sigName == "SIGSTOP")
```
jacdavis: just curious, why SIGSTOP instead of SIGINT?

src/MICore/Transports/PipeTransport.cs
paulmaybee: Logged both stdout and stderr. Logged a non-zero exit code. Logged any exceptions.

src/MICore/CommandFactories/gdb.cs
jacdavis: CmdBreakUnix pinvokes a kill. The difference is that you need to do this in remote scenarios and therefore need to go through interpreter exec.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
paulmaybee: SIGSTOP/SIGCONT are used for job control. Not sure that SIGCONT would even continue a process that is not SIGSTOPed. Also not sure that you can detach from a process stopped with a SIGINT and not have it continue automatically.

src/MICore/CommandFactories/gdb.cs
paulmaybee: Ok. I can certainly rename this method, but notice that I am not using this command to cause a break, but to deliver specific signals to the process. The process may or may not break on those signals.

src/DebugEngineHost/HostDebugger.cs
paulmaybee: I believe so. The other option would be AlreadyRunning, but that seems to imply a windows process attach.

gregg-miskelly: Correct. Everything in the MIEngine goes through DLO_CreateProcess as of today.

paulmaybee:
Async stop work is in. @gregg-miskelly

src/MICore/Debugger.cs
```diff
@@ -94,6 +96,18 @@ public ResultEventArgs(Results results)
...
+        public class StoppingEventArgs: ResultEventArgs
+        {
+            public readonly BreakRequest AsyncRequest;
+            public StoppingEventArgs(Results results, uint id, BreakRequest asyncRequest = BreakRequest.None) : base(results, id)
+            {
+                AsyncRequest = asyncRequest;
+            }
+
+            public StoppingEventArgs(Results results, BreakRequest asyncRequest = BreakRequest.None) : this(results, 0, asyncRequest)
+            { }
```
gregg-miskelly: Are you missing an assignment to AsyncRequest?

src/MICore/Debugger.cs
```diff
@@ -482,10 +504,17 @@ public Task CmdStart()
...
-            _requestingRealAsyncBreak = true;
+            None,
+            Async,
+            Stop,
+            Internal
+        }
+        protected BreakRequest _requestingRealAsyncBreak = BreakRequest.None;
+        public Task CmdBreak(BreakRequest request)
+        {
+            _requestingRealAsyncBreak = request;
```
gregg-miskelly: If an internal break comes in when we already have an outstanding non-internal break, we don't want to blow away the field.

Suggestion: reorder BreakRequest from least important->most important and then you can do 'Max' of the current value and the new one.

src/MICore/Debugger.cs
```diff
@@ -505,6 +534,12 @@ private bool IsLocalGdb()
             }
         }
 
+        private bool IsRemoteGdb()
+        {
+            return this.MICommandFactory.Mode == MIMode.Gdb &&
+               this._launchOptions is PipeLaunchOptions;
```
gregg-miskelly: Instead of explicitly checking for Pipe I would suggest 'anything besides local' since we already support other remote transports (ex: TCP), and we will probably add more in the future.

gregg-miskelly: Although I am guessing in your case you really do only want to handle pipe, so maybe this should method should just be renamed.

---
In reply to: [64105587](https://github.com/Microsoft/MIEngine/pull/329#discussion_r64105587) [](ancestors = 64105587)

src/MICore/Debugger.cs
```diff
@@ -521,7 +556,7 @@ protected bool IsCoreDump
         {
             if (ProcessState == ProcessState.Running)
             {
-                await CmdBreak();
+                await CmdBreak(BreakRequest.Async);
```
gregg-miskelly: This was actually wrong before. If its trivial, we might want to fix it...

src/MICore/Debugger.cs
```diff
@@ -533,7 +568,7 @@ protected bool IsCoreDump
...
```
gregg-miskelly: Same.

src/MICore/Debugger.cs
paulmaybee: No, the second constructor invokes the first

src/MICore/Debugger.cs
paulmaybee: done

src/MICore/Debugger.cs
paulmaybee: done

src/MICore/Debugger.cs
paulmaybee: Don't think this is trivial to fix. I'll open a bug to track it.

src/MICore/Transports/PipeTransport.cs
```diff
@@ -24,13 +24,55 @@ public class PipeTransport : StreamTransport
...
+                proc.StartInfo.WorkingDirectory = System.IO.Path.GetDirectoryName(_pipePath);
+                proc.EnableRaisingEvents = false;
+                proc.StartInfo.RedirectStandardInput = false;
+                proc.StartInfo.RedirectStandardOutput = true;
+                proc.StartInfo.RedirectStandardError = true;
+                proc.StartInfo.UseShellExecute = false;
+                proc.StartInfo.CreateNoWindow = true;
+                proc.Start();
+                Task.Run(() =>
+                {
```
gregg-miskelly: There shouldn't be any need to put this in a Task.Run

src/MIDebugEngine/Engine.Impl/EngineCallback.cs
```diff
@@ -288,6 +288,12 @@ public void OnCustomDebugEvent(Guid guidVSService, Guid sourceId, int messageCod
             Send(eventObject, AD7CustomDebugEvent.IID, null);
         }
 
+        public void OnStopComplete()
+        {
+            var eventObject = new AD7StopCompleteEvent();
+            Send(eventObject, AD7StopCompleteEvent.IID, null);
```
gregg-miskelly: I _think_ you want to pass a thread

src/MIDebugEngine/AD7.Impl/AD7Events.cs
```diff
@@ -564,6 +564,11 @@ int IDebugBreakpointEvent2.EnumBreakpoints(out IEnumDebugBoundBreakpoints2 ppEnu
         #endregion
     }
 
+    internal sealed class AD7StopCompleteEvent: AD7AsynchronousEvent, IDebugStopCompleteEvent2
```
gregg-miskelly: This is a stopping event. Should be AD7StoppingEvent

src/MICore/Debugger.cs
```diff
@@ -482,10 +497,17 @@ public Task CmdStart()
...
+            None,
+            Async,
+            Stop,
+            Internal
+        }
+        protected BreakRequest _requestingRealAsyncBreak = BreakRequest.None;
+        public Task CmdBreak(BreakRequest request)
         {
-            _requestingRealAsyncBreak = true;
+            _requestingRealAsyncBreak = request;
```
gregg-miskelly: Did you forget to add your changes? This is still just overwriting the variable.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -569,6 +578,14 @@ private List<LaunchCommand> GetInitializeCommands()
                 }
             }
 
+            if (MICommandFactory.SupportsChildProcessDebugging())
+            {
+                if (_launchOptions.DebugChildProcesses)
+                {
+                    _childProcessHandler = new DebugUnixChild(this, this._launchOptions);  // TODO: let the user enable/disable this functionality
```
gregg-miskelly: Should this TODO comment be removed now?

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -40,12 +42,22 @@ internal class ThreadCache
         private DebuggedProcess _debugger;
         private List<DebuggedThread> _deadThreads;
         private List<DebuggedThread> _newThreads;
+        private Dictionary<string, List<int>> _threadGroups;
+        private static uint s_targetId;
+        private static string s_defaultGroupId = "i1";  // gdb's default group id, also used for any process without group ids
```
gregg-miskelly: Should this really be 'const'?

src/MICore/Debugger.cs
paulmaybee: Yes, I must have screwed up the add for this file.  Will fix.


_____________________________

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
paulmaybee: That comment is referring to giving the user the ability to turn off and on the catching of forked children during a debugging session. Currently the user can only catch all children or no children, there is no finer control.


_____________________________

gregg-miskelly:
Otherwise looks good to me.

PR #332, opened 2016-05-18
Generate project.json in %TEMP% rather than in install location

chuckries:
With this change we generate the project.json in %temp%\getclrdbg and
restore/publish from this directory into the final install directory. This
avoids having the project.json in the final install directory.

### DIFF
--- a/scripts/GetClrDbg.ps1
+++ b/scripts/GetClrDbg.ps1
@@ -104,13 +104,20 @@ if (-not $RuntimeID) {
 }
 Write-Host "Info: Using Runtime ID '$RuntimeID'"
 
-# create the install folder if it does not exist
-if (-not (Test-Path -Path $InstallPath -PathType Container)) {
-    New-Item -ItemType Directory -Force -Path $InstallPath
+# create the temp folder if it does not exist
+$GuidString = [System.Guid]::NewGuid()
+$TempPath = Join-Path -Path $env:TEMP -ChildPath $GuidString
+if (-not (Test-Path -Path $TempPath -PathType Container)) {
+    New-Item -ItemType Directory -Force -Path $TempPath
 }
-$InstallPath = Resolve-Path -Path $InstallPath -ErrorAction Stop
+$TempPath = Resolve-Path -Path $TempPath -ErrorAction Stop
 
-Push-Location $InstallPath -ErrorAction Stop
+# if we were given a relative path, assume its relative to the script directory and create an absolute path
+if (-not([System.IO.Path]::IsPathRooted($InstallPath))) {
+    $InstallPath = Join-Path -Path (Split-Path -Path $MyInvocation.MyCommand.Definition) -ChildPath $InstallPath
+}
+
+Push-Location $TempPath -ErrorAction Stop
 
 Write-Host "Info: Generating project.json"
 GenerateProjectJson $VersionNumber $RuntimeID
@@ -125,5 +132,7 @@ Write-Host "Info: Executing dotnet publish"
 dotnet publish -r $RuntimeID -o $InstallPath
 
 Pop-Location
+
+Remove-Item -Path $TempPath -Recurse -Force
 
 Write-Host "Successfully installed clrdbg at '$InstallPath'"


chuckries:
@gregg-miskelly @@NCarlsonMSFT @kelltrick

scripts/GetClrDbg.ps1
```diff
@@ -104,13 +104,14 @@ if (-not $RuntimeID) {
...
 Write-Host "Info: Using Runtime ID '$RuntimeID'"
 
-# create the install folder if it does not exist
-if (-not (Test-Path -Path $InstallPath -PathType Container)) {
-    New-Item -ItemType Directory -Force -Path $InstallPath
+# create the temp folder if it does not exist
+$TempPath = Join-Path -Path $env:TEMP -ChildPath "getclrdbg"
+if (-not (Test-Path -Path $TempPath -PathType Container)) {
+    New-Item -ItemType Directory -Force -Path $TempPath
 }
```
NCarlsonMSFT: Should you delete the folder if it exists? I'm a little concerned that the file gets cached somewhere I don't know about and may not realize the need to clean to get an update.

chuckries: The way project.json is generated, it will be recreated each time this script is run, nothing is cached.

scripts/GetClrDbg.ps1
```diff
@@ -104,13 +104,14 @@ if (-not $RuntimeID) {
...
 Write-Host "Info: Using Runtime ID '$RuntimeID'"
 
-# create the install folder if it does not exist
-if (-not (Test-Path -Path $InstallPath -PathType Container)) {
-    New-Item -ItemType Directory -Force -Path $InstallPath
+# create the temp folder if it does not exist
+$TempPath = Join-Path -Path $env:TEMP -ChildPath "getclrdbg"
```
rajkumar42: I'd suggest using GetTempFileName()

$TempPath = [System.IO.Path]::GetTempFileName();

scripts/GetClrDbg.ps1
NCarlsonMSFT: Got it, I see above it will override the contents regardless. Is there any other file that may be left behind that could be a problem? If not then no need to delete the folder.

gregg-miskelly:
> dotnet publish -r $RuntimeID -o $InstallPath

'dotnet publish' will create the directory if it doesn't exist?

---
Refers to: scripts/GetClrDbg.ps1:126 in 77c55ee. [](commit_id = 77c55eeb98d345999bdf14a49e6daa25ddc029b9, deletion_comment = False)

chuckries:
@gregg-miskelly yes.

scripts/GetClrDbg.ps1
chuckries: I don't believe so. 

---
In reply to: [63795897](https://github.com/Microsoft/MIEngine/pull/332#discussion_r63795897) [](ancestors = 63795897)

chuckries:
> dotnet publish -r $RuntimeID -o $InstallPath

yes.

---
In reply to: [220180314](https://github.com/Microsoft/MIEngine/pull/332#issuecomment-220180314) [](ancestors = 220180314)

---
Refers to: scripts/GetClrDbg.ps1:126 in 77c55ee. [](commit_id = 77c55eeb98d345999bdf14a49e6daa25ddc029b9, deletion_comment = False)

scripts/GetClrDbg.ps1
chuckries: I don't think this is the right way to go, but you make a good point about two things trying to run this script simultaneously. 

---
In reply to: [63795430](https://github.com/Microsoft/MIEngine/pull/332#discussion_r63795430) [](ancestors = 63795430)

scripts/GetClrDbg.ps1
```diff
@@ -104,13 +104,14 @@ if (-not $RuntimeID) {
...
+if (-not (Test-Path -Path $TempPath -PathType Container)) {
```
kelltrick: There is no point to having ```-PathType Container``` if you don't deal with having a file named 'getclrdbg'

kelltrick:
lgtm

chuckries:
Updated this to created a directory named from a new Guid under %temp% and also fixed an issue with giving the script a relative path as an install location. This should fix any caching/simultaneous use issues.

scripts/GetClrDbg.ps1
```diff
@@ -104,13 +104,20 @@ if (-not $RuntimeID) {
 }
 Write-Host "Info: Using Runtime ID '$RuntimeID'"
 
-# create the install folder if it does not exist
-if (-not (Test-Path -Path $InstallPath -PathType Container)) {
-    New-Item -ItemType Directory -Force -Path $InstallPath
+# create the temp folder if it does not exist
+$GuidString = [System.Guid]::NewGuid()
+$TempPath = Join-Path -Path $env:TEMP -ChildPath $GuidString
```
gregg-miskelly: If you are going to do it this way (which sounds reasonable to me) you should probably also delete the directory at the end of the script.

chuckries: done. 

---
In reply to: [63798580](https://github.com/Microsoft/MIEngine/pull/332#discussion_r63798580) [](ancestors = 63798580)

PR #334, opened 2016-05-18
Add a warning if the debugger is not signed and running on OS X

edumunoz:
Previously at https://github.com/Microsoft/MIEngine/pull/328

### DIFF
--- a/src/MIDebugEngine/ResourceStrings.resx
+++ b/src/MIDebugEngine/ResourceStrings.resx
@@ -207,4 +207,8 @@
   <data name="Warning_UsingDefaultArchitecture" xml:space="preserve">
     <value>Warning: Debuggee TargetArchitecture not detected, assuming x86_64.</value>
   </data>
+  <data name="Warning_DarwinDebuggerUnsigned" xml:space="preserve">
+    <value>Debugger executable '{0}' is not signed. As a result, debugging may not work properly.</value>
+    <comment>0 = debugger path</comment>
+  </data>
 </root>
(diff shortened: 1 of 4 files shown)


edumunoz:
@gregg-miskelly suggested the verification might be expensive (hashing the file, etc.).

We could go the route of dumping the signing data (`codesign -d`), which won't verify it's correct but at least it verifies it has been signed somehow.

@gregg-miskelly, do you think that may be good enough?

gregg-miskelly:
Sounds reasonable to me. But this is your call. I only wanted to bring up the issue because we have had problems with this type of thing on Windows in the past.

edumunoz:
Yeah, I'll check with the rest to see what everyone thinks.

Ping @andrewcrawley @jacdavis

edumunoz:
Ping @pieandcakes too

jacdavis:
:shipit:

andrewcrawley:
LGTM.

src/MIDebugEngine/ResourceStrings.resx
```diff
@@ -207,4 +207,8 @@
   <data name="Warning_UsingDefaultArchitecture" xml:space="preserve">
     <value>Warning: Debuggee TargetArchitecture not detected, assuming x86_64.</value>
   </data>
+  <data name="Warning_DarwinDebuggerUnsigned" xml:space="preserve">
+    <value>Debugger executable '{0}' is not signed. Debugging may not work.</value>
```
pieandcakes: "Debugging may not work properly." or "Debugging may not work as expected."

pieandcakes:
I think if -d just verifies that its signed (not that its valid) I would call that good enough instead of doing the -v full verification against valid CA and revocation lists.

PR #335, opened 2016-05-18
Throw an exception when an unrecoverable issue occurs.

edumunoz:
@jacdavis @gregg-miskelly @chuckries

### DIFF
--- a/src/MICore/MICoreResources.resx
+++ b/src/MICore/MICoreResources.resx
@@ -225,4 +225,7 @@ Error: {1}</value>
   <data name="Warn_AttachAsRootProcess" xml:space="preserve">
     <value>Superuser access is required to attach to a process. Attaching as superuser can potentially harm your computer. Do you want to continue? [y/N]</value>
   </data>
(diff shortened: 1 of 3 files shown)


src/MICore/MICoreResources.resx
```diff
@@ -225,4 +225,7 @@ Error: {1}</value>
   <data name="Warn_AttachAsRootProcess" xml:space="preserve">
     <value>Superuser access is required to attach to a process. Attaching as superuser can potentially harm your computer. Do you want to continue? [y/N]</value>
   </data>
+  <data name="Error_LocalUnixTerminalDebuggerInitializationFailed" xml:space="preserve">
+    <value>Debugger failed to initialize.</value>
```
gregg-miskelly: Might want to make your error message here a bit more unique so that at least you can search for the string. Maybe:

Failed to initialize debugger terminal.

gregg-miskelly:
Otherwise looks good to me.

src/MICore/MICoreResources.resx
edumunoz: Good point. Fixed it and merging.

PR #341, opened 2016-05-20
Add telemetry on if we're on cygwin or mingw on windows.

jacdavis:


jacdavis:
@pieandcakes

pieandcakes:
fedora squirrel

PR #346, opened 2016-05-26
Brekapoint request checksums and warnings

chuckries:


chuckries:
@gregg-miskelly @caslan @jacdavis

src/MICore/Checksum.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MICore
+{
+    public class Checksum
+    {
+#if CORECLR
```
edumunoz: Can you add a comment explaining why this is only for CoreCLR?

src/MICore/Checksum.cs
```diff
@@ -0,0 +1,123 @@
...
+            }
+
+            return builder.ToString().Trim(); ;
+        }
+#else
+        /// <summary>
+        /// Dummy concstrutor for non coreclr scnearios
+        /// </summary>
+        public Checksum()
+        {
```
edumunoz: Can we have the non-CoreCLR version throw a `NotImplementedException` if it's not going to be implemented?

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
```diff
@@ -111,7 +111,28 @@ private static MIBreakpointState StringToBreakpointState(string addr)
 
             string basename = System.IO.Path.GetFileName(documentName);     // get basename from Windows path
             basename = process.EscapePath(basename);
-            BindResult bindResults = EvalBindResult(await process.MICommandFactory.BreakInsert(basename, line, condition, ResultClass.None), pbreak);
+
+            Checksum[] checksums = null;
+#if CORECLR
```
edumunoz: Why is `#if CORECLR` needed here, given that it's checking `MICommandFactory.SupportsBreakpointChecksums()`.

src/MICore/Checksum.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            StringBuilder builder = new StringBuilder();
+            foreach (IGrouping<HashAlgorithmName, string> group in checksumGroups)
+            {
+                builder.Append(FormattableString.Invariant($"--{group.Key.Name}checksum "));
+                builder.Append(string.Join(",", group));
+                builder.Append(" ");
+            }
+
+            return builder.ToString().Trim(); ;
```
caslan: extra ;

src/MICore/Checksum.cs
```diff
@@ -0,0 +1,123 @@
+﻿using System;
```
caslan: Missing copyright notice

src/MICore/Checksum.cs
```diff
@@ -0,0 +1,123 @@
...
+                builder.Append(FormattableString.Invariant($"--{group.Key.Name}checksum "));
+                builder.Append(string.Join(",", group));
+                builder.Append(" ");
+            }
+
+            return builder.ToString().Trim(); ;
+        }
+#else
+        /// <summary>
+        /// Dummy concstrutor for non coreclr scnearios
```
caslan: "constructor"

src/MICore/Checksum.cs
```diff
@@ -0,0 +1,123 @@
...
+        public string ToMIString()
+        {
+            return FormattableString.Invariant($"--{HashAlgorithmName.Name}checksum {ToString()}");
+        }
+
+        public static string GetMIString(IEnumerable<Checksum> checksums)
+        {
+            if (checksums == null)
+            {
+                return "";
```
edumunoz: nit: `string.Empty`

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -558,6 +561,74 @@ int IDebugPendingBreakpoint2.Virtualize(int fVirtualize)
...
+        /// <returns>
+        /// Checksums calculated using the given HashAlgorithmId against the source file represented in this pending breakpoint.
+        /// If no checksums can be calculated, returns null
+        /// </returns>
+        internal Checksum[] GetChecksum(HashAlgorithmId hashAlgorithmId)
+        {
+            IDebugBreakpointChecksumRequest2 checksumRequest = _pBPRequest as IDebugBreakpointChecksumRequest2;
+            if (checksumRequest == null)
+            {
+                return null;
```
edumunoz: I would get rid of returning `null` and instead return an empty `IEnumerable`. Also, is there a reason why the return types are `Checksum[]` instead of `IEnumerable<Checksum>` (which is what is consumed in the end?)

src/MIDebugEngine/MIDebugEngine.csproj
```diff
@@ -141,6 +141,7 @@
     <Compile Include="Engine.Impl\EngineCallback.cs" />
     <Compile Include="Engine.Impl\EngineUtils.cs" />
     <Compile Include="Engine.Impl\ExceptionManager.cs" />
+    <Compile Include="AD7.Impl\HashAlgorithmId.cs" />
```
edumunoz: This does not seem to be placed properly by VS? Whenever a change is made to the csproj file, this will be moved.

src/MIDebugEngine/Engine.Impl/BreakpointManager.cs
```diff
@@ -77,8 +77,17 @@ public void BreakpointModified(object sender, EventArgs args)
             {
                 return;
             }
-            var bindList = pending.PendingBreakpoint.BindAddresses(bkpt);
-            RebindAddresses(pending, bindList);
+
+            string warning = bkpt.TryFindString("warning");
+            if (!string.IsNullOrEmpty("warning"))
```
edumunoz: I think you want `warning` here without quotes.

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
```diff
@@ -192,9 +192,16 @@ private static BindResult EvalBindResult(Results bindResult, AD7PendingBreakpoin
             Debug.Assert(bkpt.FindString("type") == "breakpoint");
 
             string number = bkpt.FindString("number");
+            string warning = bkpt.TryFindString("warning");
             string addr = bkpt.TryFindString("addr");
 
-            PendingBreakpoint bp = new PendingBreakpoint(pbreak, number, StringToBreakpointState(addr));
+            PendingBreakpoint bp;
+            if (!string.IsNullOrEmpty(warning))
```
edumunoz: Is `warning` only sent in the case of a mismatch of checksums?

Could it be the case that someone adds a warning but doesn't want the breakpoint state to go to `MIBreakpointState.Pending`?

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -146,6 +146,16 @@ public void ClearBoundBreakpoints()
             }
         }
 
+        public void SetError(AD7ErrorBreakpoint bpError, bool sendEvent = false)
```
edumunoz: Is `sendEvent` ever set to `false`?

I see lots of usages of `_BPError` in which it's either just set (e.g., in `CanBind()`) and others in which it is sent.

Maybe you want to use `SetError` across this class in the existing code?

src/MIDebugEngine/AD7.Impl/HashAlgorithmId.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
```
edumunoz: missing copyright here too

src/MICore/Checksum.cs
chuckries: the only issue is that mono doesn't have HashAlgorithmName yet. I will comment explaining this.

src/MIDebugEngine/Engine.Impl/BreakpointManager.cs
chuckries: good catch, thanks.

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
chuckries: warning is used for various things, such as symbols not loaded, checksum mismatch, condition can't be evaluated, etc. In the case of the vs debugger, we will never have a warning AND a bound breakpoint. I left the state as pending because it should be fine.

edumunoz: Cool. Consider adding a short comment regarding this?

src/MICore/Checksum.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MICore
+{
+    public class Checksum
+    {
+#if CORECLR
+        private string _checksumString = null;
+        public byte[] Bytes { get; private set; } = null;
```
gregg-miskelly: Not that this is the end of the world, but does this need to be public? Always nice to keep mutable collections private if one can...

src/MICore/Checksum.cs
```diff
@@ -0,0 +1,123 @@
...
+            }
+
+            IEnumerable<IGrouping<HashAlgorithmName, string>> checksumGroups = checksums.GroupBy(checksum => checksum.HashAlgorithmName, checksum => checksum.ToString());
+
+            StringBuilder builder = new StringBuilder();
+            foreach (IGrouping<HashAlgorithmName, string> group in checksumGroups)
+            {
+                builder.Append(FormattableString.Invariant($"--{group.Key.Name}checksum "));
+                builder.Append(string.Join(",", group));
+                builder.Append(" ");
```
gregg-miskelly: nit: you can avoid the Trim at the end if you instead remove this line, and add a line at the beginning of the foreach which is 'if (builder.Length > 0) builder.Append(' ');'

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
```diff
@@ -111,7 +111,28 @@ private static MIBreakpointState StringToBreakpointState(string addr)
...
-            BindResult bindResults = EvalBindResult(await process.MICommandFactory.BreakInsert(basename, line, condition, ResultClass.None), pbreak);
+
+            Checksum[] checksums = null;
+#if CORECLR
+            if (process.MICommandFactory.SupportsBreakpointChecksums())
+            {
+                // TODO: This will need to be configurable in the future somehow
+                // TODO: We might want the MICommandFactory to return the algorithm we should use
+                // TODO: But HashAlgorithmID is not part of MICore becuase it uses AD7Guids
+                HashAlgorithmId[] hashAlgorithmIds = { HashAlgorithmId.SHA1Normalized };
```
gregg-miskelly: 1. In this VS Case I am assuming this is NOT called on the main thread, correct? You should probably change it so that we get this stuff on the main thread along with all the other breakpoint properties. Otherwise we are adding extra thread transitions for no reason. At best that will be slow, at worst, that could hang.
2. What does the update 3 version of VSDebug do when it sees the 'SHA1Normalized' algorithm id? I am guessing we are going to need some code to either fall back to SHA1, or make it host dependent.

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
gregg-miskelly: Also please update the MI Extensions OneNote page.

---
In reply to: [64833924](https://github.com/Microsoft/MIEngine/pull/346#discussion_r64833924) [](ancestors = 64833924)

src/MICore/Checksum.cs
gregg-miskelly: I wonder if we want to just use a string. Its not like MIEngine does any actually hashing. Then we can get rid of all the ifdefs.

---
In reply to: [64833308](https://github.com/Microsoft/MIEngine/pull/346#discussion_r64833308) [](ancestors = 64833308)

src/MICore/Checksum.cs
chuckries: I will make the public property readonly

src/MIDebugEngine/Engine.Impl/Breakpoints.cs
chuckries: For 2, I have already added the following code into DocumentChecksum.h in vsdebug. We will need to merge it to update 3. 
```C++
    void SetGuidAlgorithm(REFGUID guidAlgorithm)
    {
        if (IsEqualGUID(guidAlgorithm, guidSourceHashMD5Normalized))
        {
            m_guidAlgorithm = guidSourceHashMD5;
        }
        else if (IsEqualGUID(guidAlgorithm, guidSourceHashSHA1Normalized))
        {
            m_guidAlgorithm = guidSourceHashSHA1;
        }
        else if (IsEqualGUID(guidAlgorithm, guidSourceHashSHA256Normalized))
        {
            m_guidAlgorithm = guidSourceHashSHA256;
        }
        else
        {
            m_guidAlgorithm = guidAlgorithm;
        }
    }
```

src/MIDebugEngine/MIDebugEngine.csproj
chuckries: Can you elaborate? I don't fully understand your comment. 


---
In reply to: [64830281](https://github.com/Microsoft/MIEngine/pull/346#discussion_r64830281) [](ancestors = 64830281)

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
chuckries: I had intended to do this and forgot to make the actual change, good catch. 

---
In reply to: [64832465](https://github.com/Microsoft/MIEngine/pull/346#discussion_r64832465) [](ancestors = 64832465)

src/MICore/Checksum.cs
chuckries: Went with an enum.

---
In reply to: [64843632](https://github.com/Microsoft/MIEngine/pull/346#discussion_r64843632) [](ancestors = 64843632,64833308)

src/MIDebugEngine/MIDebugEngine.csproj
edumunoz: VS organizes the entries in the `ItemGroup` alphabetically.

So the next time this csproj file is edited by VS, it will create an edit moving this line to its "correct" position (as of now, after the line for `AD7.Impl\EngineConstants.cs`).

chuckries:
@gregg-miskelly @edumunoz 

I removed the `#if CORECLR` portions and cleaned up based on feedback. Please take a second glance at this.

gregg-miskelly:
LGTM

edumunoz:
:ship:

chuckries:
Thanks all!

PR #349, opened 2016-05-27
Fix waiting for the debugger to complete in the OSX console

gregg-miskelly:
The external console code was using 'wait' to wait for the debugger process to
exit. This turns out to have problems on OSX, so this switches to using 'fg'.
there.

### DIFF
--- a/src/MICore/UnixUtilities.cs
+++ b/src/MICore/UnixUtilities.cs
@@ -42,6 +42,12 @@ namespace MICore
             string pidFifo,
             string debuggerCmd)
         {
+            // On OSX, 'wait' will return once there is a status change from the launched process rather than for it to exit, so
+            // we need to use 'fg' there. This works as our bash prompt is launched through apple script rather than 'bash -c'.
+            // On Linux, fg will fail with 'no job control' because our commands are being executed through 'bash -c', and
+            // bash doesn't support fg in this mode, so we need to use 'wait' there.
+            string waitForCompletionCommand = PlatformUtilities.IsOSX() ? "fg > /dev/null; " : "wait $pid; ";
+
             return string.Format(CultureInfo.InvariantCulture,
                 // echo the shell pid so that we can monitor it
                 "echo $$ > {3}; " +
@@ -55,12 +61,13 @@ namespace MICore
                 // we need to fake an exit by the debugger
                 "pid=$! ; " +
                 "echo $pid > {3}; " +
-                "wait $pid; ",
-                debuggeeDir,
-                dbgStdInName,
-                dbgStdOutName,
-                pidFifo,
-                debuggerCmd
+                "{5}",
+                debuggeeDir, /* 0 */
+                dbgStdInName, /* 1 */
+                dbgStdOutName, /* 2 */
+                pidFifo, /* 3 */
+                debuggerCmd, /* 4 */
+                waitForCompletionCommand /* 5 */
                 );
         }
 


gregg-miskelly:
@edumunoz please review

src/MICore/UnixUtilities.cs
```diff
@@ -42,6 +42,12 @@ public static class UnixUtilities
             string pidFifo,
             string debuggerCmd)
         {
+            // On OSX, 'wait' will return once there is a status change from the launched process rather than for it to exit, so
```
edumunoz: Just curious, do you mean something like stopping the process (via `SIGSTOP`)?

src/MICore/UnixUtilities.cs
```diff
@@ -42,6 +42,12 @@ public static class UnixUtilities
...
+            // we need to use 'fg' there. 
+            // On Linux, fg will fail with 'no job control' because our bash shell is not interactive, so we need
```
edumunoz: Do you happen to know why this doesn't happen on OS X?

I tried using `fg` when I first added this. I tried it on OS X and it worked (as you have seen since you're adding that), and I thought I had tried on Linux too, but maybe I'm misremembering.

src/MICore/UnixUtilities.cs
```diff
@@ -42,6 +42,12 @@ public static class UnixUtilities
             string pidFifo,
             string debuggerCmd)
         {
+            // On OSX, 'wait' will return once there is a status change from the launched process rather than for it to exit, so
+            // we need to use 'fg' there. 
+            // On Linux, fg will fail with 'no job control' because our bash shell is not interactive, so we need 
+            // to use 'wait' there.
+            string waitForCompletionCommand = PlatformUtilities.IsOSX() ? "fg > /dev/null; " : "wait $pid; ";
```
edumunoz: nit: You might want to move this to a `static readonly` field, but I don't really have a strong opinion on it.

src/MICore/UnixUtilities.cs
```diff
@@ -55,7 +61,7 @@ public static class UnixUtilities
                 // we need to fake an exit by the debugger
                 "pid=$! ; " +
                 "echo $pid > {3}; " +
-                "wait $pid; ",
+                waitForCompletionCommand,
```
edumunoz: nit: You might want to consider having this as an argument to `string.Format`. I don't have a strong opinion about this, but it might make it clearer since all the lines below are actual arguments to `string.Format`.

edumunoz:
Minor suggestions and questions.

:ship:

src/MICore/UnixUtilities.cs
gregg-miskelly: I _think_ any signal will cause it to return. But really I don't fully understand it - OSX has no documentation as to the behavior. It just returns immediately for OSX.

src/MICore/UnixUtilities.cs
gregg-miskelly: I can play around with it on Tuesday. I am not sure if it is a difference in the way bash gets started (AppleScript vs. Linux launches), or if it is just because OSX uses a different implementation/version of bash.

src/MICore/UnixUtilities.cs
gregg-miskelly: Putting it in a static readonly field means that the CLR must run the code before the class is executed. In this case, the check should be so trivial that it is better to repeat it each launch.

src/MICore/UnixUtilities.cs
gregg-miskelly: I don't have a strong opinion either, but sure, I can do that.

gregg-miskelly:
Made suggested changes.

PR #352, opened 2016-06-02
Fix dependencies for InstallToVSCode script

caslan:
InstallToVsCode needs to explicitly specify some dependencies that clrdbg package depends as the install script doesn't try to restore 'clrdbg' package but rather copies its binaries.
@gregg-miskelly @chuckries


chuckries:
:+1:

wesrupert:
:+1:

PR #353, opened 2016-06-06
Fix environment variables when using the external console on OSX

gregg-miskelly:
On OSX, the code was still trying to set environment variables by passing them
on to osascript, which will NOT pass them on to the child process.

This changes the code to instead pass them using 'export' statements.

### DIFF
--- a/src/MICore/TerminalLauncher.cs
+++ b/src/MICore/TerminalLauncher.cs
@@ -7,16 +7,17 @@ using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Reflection;
+using System.Text;
 
 namespace MICore
 {
     internal abstract class TerminalLauncher
     {
-        private ReadOnlyCollection<EnvironmentEntry> _environment;
         private Process _terminalProcess;
         protected string _title;
         protected string _initScript;
         protected bool _isRoot;
+        protected ReadOnlyCollection<EnvironmentEntry> _environment;
 
         public static TerminalLauncher MakeTerminal(string title, string initScript, ReadOnlyCollection<EnvironmentEntry> environment)
         {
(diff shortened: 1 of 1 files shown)


gregg-miskelly:
@edumunoz please review

src/MICore/TerminalLauncher.cs
```diff
@@ -106,7 +95,54 @@ protected override string GetProcessArgs()
...
+                    // otherwise we need to escape the single quotes
+                    foreach (char c in entry.Value)
+                    {
+                        if (c != '\'')
+                        {
+                            stringBuilder.Append(c);
+                        }
+                        else
+                        {
+                            stringBuilder.Append("'\\\''"); // end the single quoted string we are in, then add a single quote, and restart the single quotes
```
edumunoz: Can you add a comment explaining that "the single quoted string we are in" is started by appending `"='"`? I took me a while to figure that out.

I find the `'"'"'` a little easier to understand, but the one used here is fine too.

src/MICore/TerminalLauncher.cs
```diff
@@ -106,7 +95,54 @@ protected override string GetProcessArgs()
                 throw new FileNotFoundException(message);
             }
 
-            return string.Format(CultureInfo.InvariantCulture, "{0} \"{1}\" \"{2}\"", launchScript, _title, _initScript);
+            string setEnvironmentScript = GetSetEnvironmentScript();
+            return string.Format(CultureInfo.InvariantCulture, "{0} \"{1}\" \"{2}{3}\"", launchScript, _title, setEnvironmentScript, _initScript);
```
edumunoz: I know setEnvironmentScript will end with a space, but I though there was some sort of mistake because there was no space between `{2}` and `{3}`. Can you add a space there?

edumunoz:
Thanks for catching and fixing this!

:shipit:

src/MICore/TerminalLauncher.cs
gregg-miskelly: You can call me anal, but I don't want to add an extra space there since if there are no environment variables, I don't want the command line to turn into "<SPACE>..." instead of "...". I can add a comment if you think it is confusing.

src/MICore/TerminalLauncher.cs
gregg-miskelly: I updated the comment so hopefully it is more clear. I also used a '@' string which hopefully also helps.

PR #356, opened 2016-06-08
Rajkumar42/pipecwd

rajkumar42:
if PipeCwd is empty, it will be set to PipePath's directory.
Added asserts in tests to check for the default and specified scenario.

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -63,7 +63,7 @@ namespace MICore
     /// </summary>
     public sealed class PipeLaunchOptions : LaunchOptions
     {
-        public PipeLaunchOptions(string PipePath, string PipeArguments, string PipeCommandArguments)
+        public PipeLaunchOptions(string PipePath, string PipeArguments, string PipeCommandArguments, string PipeCwd)
         {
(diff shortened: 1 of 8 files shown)


src/MICore/LaunchOptions.cs
```diff
@@ -63,19 +63,27 @@ public enum LaunchCompleteCommand
...
         {
             if (string.IsNullOrEmpty(PipePath))
                 throw new ArgumentNullException("PipePath");
 
             this.PipePath = PipePath;
             this.PipeArguments = PipeArguments;
             this.PipeCommandArguments = PipeCommandArguments;
+            if (!String.IsNullOrWhiteSpace(PipeCwd))
+            {
+                this.PipeCwd = PipeCwd;
```
gregg-miskelly: Should we raise an error if this looks like a relative path or doesn't exist?

gregg-miskelly:
Otherwise LGTM

wesrupert:
:+1: After path validation

chuckries:
please squash :+1:

gregg-miskelly:
I think you also need to build the 'Debug' configuration to generate LaunchOptions.xsd.types.designer.desktop.cs

PR #357, opened 2016-06-09
Make sure a breakpoint has been reported as hit

edumunoz:
There is a problem when a breakpoint has been set on `main`, but the breakpoint actually hasn't been hit when the `EntryPoint` event is being triggered. This can happen if the user-added breakpoint has a condition that evaluates to false.

In this PR, I add a check to make sure every breakpoint reported in `FindHitBreakpoints` has actually been hit. I believe this will only work with GDB, as the other debuggers don't implement the `-break-info` command.

/cc @gregg-miskelly @jacdavis

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/BreakpointManager.cs
+++ b/src/MIDebugEngine/Engine.Impl/BreakpointManager.cs
@@ -197,9 +197,8 @@ namespace Microsoft.MIDebugEngine
         }
 
 
-        public AD7BoundBreakpoint[] FindHitBreakpoints(string bkptno, ulong addr, /*OPTIONAL*/ TupleValue frame, out bool fContinue)
+        public async Task<Tuple<AD7BoundBreakpoint[], bool>> FindHitBreakpoints(string bkptno, ulong addr, /*OPTIONAL*/ TupleValue frame)
         {
-            fContinue = false;
             List<AD7BoundBreakpoint> hitBoundBreakpoints = new List<AD7BoundBreakpoint>();
             // match based on address and bkptno since there are many
             // cases where gdb doesn't provide the breakpoint address (multple binds to the same location)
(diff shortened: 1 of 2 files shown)


src/MIDebugEngine/Engine.Impl/BreakpointManager.cs
```diff
@@ -221,13 +220,23 @@ public AD7BoundBreakpoint[] FindHitBreakpoints(string bkptno, ulong addr, /*OPTI
                     continue;
                 }
 
-                hitBoundBreakpoints.Add(currBoundBp);
-            }
+                string currBkptNo = currBoundBp.PendingBreakpoint.BreakpointId;
+                TupleValue bkpt = await _engine.DebuggedProcess.MICommandFactory.BreakInfo(currBkptNo);
+                uint? hitTimes = bkpt.TryFindUint("times");
```
gregg-miskelly: bkpt can be null, so this needs to be ?.

gregg-miskelly: Does this tell us if the breakpoint has ever been hit? Or does 'times' get reset every time we continue?

gregg-miskelly:
Is this bug actually important enough to fix? This doesn't seem like something we should bother fixing to me.

src/MIDebugEngine/Engine.Impl/BreakpointManager.cs
edumunoz: It's the total number of times the breakpoint has been hit.

edumunoz:
@gregg-miskelly I was hoping this would also fix the issue of reporting more breakpoints than are actually hit on the `AD7BreakpointEvent` that is sent in [`OnBreakpoint`](https://github.com/Microsoft/MIEngine/blob/master/src/MIDebugEngine/Engine.Impl/EngineCallback.cs#L176).

However, given that `times` doesn't get reset after each `continue`, there is no easy way to detect in gdb which breakpoints have been hit in the most recent stopped event (which is surprising to me). The stopped event only reports one breakpoint number at most. Example in which breakpoints 2 and 3 were hit:

> `*stopped,reason="breakpoint-hit",disp="keep",bkptno="2",frame=addr="0x0000000100000f29",func="test",args=],file="test.cpp",fullname="/Users/edmunoz/code/cpp2/test.cpp",line="6"},thread-id="1",stopped-threads="all"
`

I'm also not completely sure it's worth fixing, but it's not a big code change. I will update the internal bug and may re-open this PR.

delmyers:
I would like to fix this, just so that we get correct functionality. The exception to this would be if it will increase stepping/breakpoint time significantly. I would be a little bit concerned about the fact that we have to request breakpoint info frequently with this change. @gregg-miskelly, what do you think?

gregg-miskelly:
I don't know of a way to fix this completely - edmunoz's fix addresses the problem if the condition has always been false, but if the breakpoint has previously hit, I don't know of a way to resolve this. Considering this, and consider that this doesn't feel particularly important (who sets a conditional breakpoint on the first instruction of main?) I think we shouldn't fix it.

PR #358, opened 2016-06-09
Adding pipeEnv.

rajkumar42:
Adding the option to provide environment variables to the pipe program.


rajkumar42:
@chuckries @gregg-miskelly @caslan @wesrupert

src/MICore/LaunchOptions.cs
```diff
@@ -109,6 +112,11 @@ static internal PipeLaunchOptions CreateFromXml(Xml.LaunchOptions.PipeLaunchOpti
         /// [Optional] Current working directory when the pipe program is invoked.
         /// </summary>
         public string PipeCwd { get; private set; }
+        
+        /// <summary>
+        /// [Optional] Enviroment variables for the pipe program.
+        /// </summary>
+        public EnvironmentEntry[] PipeEnvironment { get; private set; }
```
gregg-miskelly: Can we make this a ReadOnlyCollection and/or IReadOnlyList. I don't like mutable public members of public types

src/MICore/Transports/PipeTransport.cs
```diff
@@ -137,6 +137,11 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
             proc.StartInfo.Arguments = pipeOptions.PipeArguments;
             proc.StartInfo.WorkingDirectory = pipeOptions.PipeCwd;
 
+            foreach (EnvironmentEntry entry in pipeOptions.PipeEnvironment)
+            {
+                proc.StartInfo.SetEnvironmentVariable(entry.Name, entry.Value);
```
gregg-miskelly: If I remember correctly (and maybe I don't) this is missing from Mono, and you want to use StartInfo.Environment.Add

gregg-miskelly:
:shipit:

wesrupert:
LGTM! :shipit:

src/MICore/Transports/PipeTransport.cs
rajkumar42: I see StartInfo.SetEnvironmentVariable is used in LocalTransport.cs as well. 
@andrewcrawley @faxue-msft  can you please confirm?

andrewcrawley: Using StartInfo.SetEnvironmentVariable is correct - it's a helper we added that does the right thing for both CoreCLR and Mono.

LGTM.

PR #360, opened 2016-06-17
Add a second check for a late HostLogger load

wesrupert:
Engine logging will currently only be recognized if the logger is set on startup. Add a second check before creating a new DebuggedProcess to accept engine logging requests sent later in the pipeline.

### DIFF
--- a/src/MICore/Logger.cs
+++ b/src/MICore/Logger.cs
@@ -41,11 +41,7 @@ namespace MICore
                 s_isInitialized = true;
                 s_initTime = DateTime.Now;
 
-                s_logger = configStore.GetLogger("EnableMIDebugLogger", "Microsoft.MIDebug.log");
-                if (s_logger != null)
-                {
-                    s_isEnabled = true;
-                }
+                LoadMIDebugLogger(configStore);
                 res.WriteLine("Initialized log at: " + s_initTime);
             }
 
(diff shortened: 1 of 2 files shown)


wesrupert:
@caslan @gregg-miskelly @chuckries, @rajkumar42, This is the third part of the logging changes. We currently only set the logger during when we load the config store. This allows us to turn on engineLogging later without having to degrade the performance of MIEngine by always prepping events just in case the logger is listening.

Unrelated, is the `cla-required` tag expected? First time I've seen it, I've had `cla-not-required` every other time...

gregg-miskelly:
@wesrupert because you are a Microsoft employee, I would have expected the 'cla-not-required' tag. Can you try opening a new PR to see if the problem happens again?

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -435,6 +435,9 @@ int IDebugEngineLaunch2.LaunchSuspended(string pszServer, IDebugPort2 port, stri
                     {
                         try
                         {
+                            // Check if the logger was enabled late.
```
gregg-miskelly: nit: might want to do this at the beginning of 'LaunchSuspended' since it should be set by then, and if we wind up adding any logging, it would be good ta capture it.

src/MICore/Logger.cs
```diff
@@ -58,6 +54,15 @@ public static Logger EnsureInitialized(HostConfigurationStore configStore)
             return res;
         }
 
+        public static void LoadMIDebugLogger(HostConfigurationStore configStore)
+        {
+            s_logger = configStore.GetLogger("EnableMIDebugLogger", "Microsoft.MIDebug.log");
```
gregg-miskelly: Should this function be in a 'if (s_logger == null)'?

gregg-miskelly:
:shipit:

rajkumar42:
:shipit:

wesrupert:
Moved to #363.

PR #361, opened 2016-06-18
Implement IDebugEngine3.SetEngineGuid

gregg-miskelly:
This checkin implements IDebugEngine3 in the MIEngine. The reason for this is
to provide support for SetEngineGuid, which allows registering the MIEngine
for an alternate engine GUID without needing to change the code.

The value of this is that, for the SSH port supplier, we can use a different
engine GUID to differentiate between LLDB/GDB/clrdbg and thus offer multiple
types of code debugging using the same gestures that the attach to process
dialog already knows (engine selection).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,9 +19,12 @@ intermediate
 *.metaproj
 *.tmp
 *.VC.opendb
+*.VC.db
 packages/
 packages.Desktop/
 .vs/
 Microsoft.VisualStudio.Glass
 project.lock.json
+src/*/*.nuget.props
+src/*/*.nuget.targets
 .DS_Store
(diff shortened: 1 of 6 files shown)


rajkumar42:
:shipit:

chuckries:
Do we need to make any changes to [this page](https://github.com/Microsoft/MIEngine/wiki/Redistribution)?

chuckries:
:+1:

wesrupert:
:+1:

gregg-miskelly:
@chuckries the change to the .cs file now is optional for folks who only distribute the engine pieces, though it doesn't hurt anything to do so.

PR #362, opened 2016-06-20
Add a second check for a late HostLogger load (attempt 2)

wesrupert:
Engine logging will currently only be recognized if the logger is set on startup. Add a second check before creating a new DebuggedProcess to accept engine logging requests sent later in the pipeline.


wesrupert:
Still showing `cla-required`, ignore this PR.

PR #369, opened 2016-06-27
Do not acquire lock when enable/disable breakpoint

faxue-msft:
I don't see a reason why we need to lock on _boundBreakpoints when enable/disable breakpoints. The problem is that this lldb-mi may send out breakpoint related messages before MIEngine receive "done" for enable/disable breakpoint, and that will cause deadlock.
I'm removing some lldb outputs to reduce the chance of deadlock, but because lldb-mi send breakpoint output on response to a lot of lldb breakpoint related Events, there is still a chance that lldb will send out breakpoint message in a bad timing and cause a deadlock.

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
@@ -432,12 +432,14 @@ namespace Microsoft.MIDebugEngine
         // Toggles the enabled state of this pending breakpoint.
         int IDebugPendingBreakpoint2.Enable(int fEnable)
         {
-            lock (_boundBreakpoints)
+            bool newValue = fEnable == 0 ? false : true;
+            if (_enabled != newValue)
             {
-                _enabled = fEnable == 0 ? false : true;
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -432,9 +432,10 @@ int IDebugPendingBreakpoint2.Delete()
         // Toggles the enabled state of this pending breakpoint.
         int IDebugPendingBreakpoint2.Enable(int fEnable)
         {
-            lock (_boundBreakpoints)
+            bool newValue = fEnable == 0 ? false : true;
+            if (_enabled != newValue)
             {
-                _enabled = fEnable == 0 ? false : true;
+                _enabled = newValue;
                 if (_bp != null)
```
gregg-miskelly: We will sometimes clear _bp, so this code should probably be:

var bp = _bp;
if (bp != null)
{
...
  bp.Enable(...

gregg-miskelly:
:shipit:

PR #374, opened 2016-06-30
Fix bug 235242 Incorrect function signatures in callstack.

jacdavis:
The newest lldb builds put the function signature in the func
response which doesn't conform the mi protocol and results
in duplicate data. This strips that off if it exists.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
@@ -226,6 +226,18 @@ namespace Microsoft.MIDebugEngine
             ulong? pc = frame.TryFindAddr("addr");
             MITextPosition textPosition = MITextPosition.TryParse(frame);
             string func = frame.TryFindString("func");
+
+            if (this._debugger.MICommandFactory.Mode == MIMode.Lldb)
+            {
+                if (func.Contains("("))
+                {
+                    // LLDB adds the types of variables to the func. This doesn't conform 
+                    // the mi protocol and results in duplicate data when the ad7 stackframe
+                    // object does the same thing. 
+                    func = func.Substring(0, func.IndexOf('('));
+                }
+            }
+
             uint level = frame.FindUint("level");
             string from = frame.TryFindString("from");
 


jacdavis:
@gregg-miskelly @chuckries @faxue-msft

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -226,6 +226,18 @@ private ThreadContext CreateContext(TupleValue frame)
...
             string func = frame.TryFindString("func");
+
+            if (this._debugger.MICommandFactory.Mode == MIMode.Lldb)
+            {
+                if (func.Contains("("))
+                {
+                    // LLDB adds the types of variables to the func. This doesn't conform 
+                    // the mi protocol and results in duplicate data when the ad7 stackframe
+                    // object does the same thing. 
+                    func = func.Substring(0, func.IndexOf('('));
```
gregg-miskelly: Is there any chance that a '(' could be part of a type name in some sort of template case?

gregg-miskelly:
Otherwise LGTM

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
jacdavis: There is certainly a risk of that but AFAIK, it shouldn't be. I don't believe '(' is a valid identifier. Do we have any cases where decorated names are shown in lldb? I haven't seen any. Arguably, the correct fix is to remove this from lldb-mi since it doesn't conform to the protocol.

gregg-miskelly: Using the Microsoft C++ compiler at least, template arguments for function types can have '(' in them. Example:

 	natcon.exe!std::_Invoker_functor::_Call<int (__cdecl*&)(int),int>(int(*)(int) & _Obj, int && <_Args_0>) Line 1375	C++
 	natcon.exe!std::invoke<int (__cdecl*&)(int),int>(int(*)(int) & _Obj, int && <_Args_0>) Line 1443	C++
 	natcon.exe!std::_Invoke_ret<int,int (__cdecl*&)(int),int>(std::_Forced<int,0> __formal, int(*)(int) & <_Vals_0>, int && <_Vals_1>) Line 1468	C++
	natcon.exe!std::_Func_impl<int (__cdecl*)(int),std::allocator<int>,int,int>::_Do_call(int && <_Args_0>) Line 212	C++

jacdavis: Probably too risky than. I'll look into fixing it on the lldb side instead.

jacdavis:
Closing due to risk

PR #375, opened 2016-06-30
Fix Kofe engine with latest MIEngine

gregg-miskelly:
The Kofe engine was still using the old constructor for MIEngine, which was
removed with commit f6bb8986a159d1ee2a24143904ab47591d8dcdac
(Implement IDebugEngine3.SetEngineGuid).

This puts it back.


chuckries:
@faxue-msft FYI

@gregg-miskelly LGTM

faxue-msft:
I guess it's not necessary to add the constructor wrapper back into DebugEngineHost.ref.cs since Kofe isn't rebuilding against the new binary anyway.
LGTM

PR #380, opened 2016-07-05
Remove async break on terminate with clrdbg

caslan:
Remove async break on terminate with clrdbg as it's not needed and causes asserts on clrdbg shutdown with chk bits.


caslan:
@gregg-miskelly @chuckries @rajkumar42

gregg-miskelly:
LGTM

chuckries:
:+1:

rajkumar42:
:+1:

PR #381, opened 2016-07-06
Fix an assert for stop debugging a running debuggee

faxue-msft:
This to fix a bug of NullReferenceException when stop debugging from run mode, Gdb will need to send a signal to debuggee to stop it and end up with a *stop event. MIEngine try to get the thread information when handling the *stop event, but can't get it since all threads has exit and the thread cache has been cleaned up.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -787,6 +787,14 @@ namespace Microsoft.MIDebugEngine
             MICommandFactory.DefineCurrentThread(tid);
 
             DebuggedThread thread = await ThreadCache.GetThread(tid);
+            if (thread == null)
+            {
+                // It's possible that the stop event happens during stop debugging and the thread cache has been cleaned up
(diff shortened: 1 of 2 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -787,6 +787,14 @@ private void Dispose()
             MICommandFactory.DefineCurrentThread(tid);
 
             DebuggedThread thread = await ThreadCache.GetThread(tid);
+            if (thread == null)
+            {
+                // It's possible that the stop event happens during stop debugging and the thread cache has been cleaned up
+                // See https://devdiv.visualstudio.com/DevDiv/VS%20Diag%20IntelliTrace/_workItems?_a=edit&id=236275&triage=true
+                // for a repro
+                return;
```
gregg-miskelly: Should we raise some sort of error if this happens when we haven't begun stop debugging?

gregg-miskelly:
Otherwise sounds good to me

gregg-miskelly:
:shipit:

PR #382, opened 2016-07-07
Fix for LLDB to allow exe arguments on launch

pieandcakes:
Reorder launch parameters to run -exec-arguments after
-file-exec-and-symbols


pieandcakes:
@richardstanton @gregg-miskelly @chuckries

pieandcakes:
Verified this did not break the gdb case on Ubuntu

richardstanton:
# 👍

gregg-miskelly:
:shipit:

chuckries:
seems like this would be worth fixing in lldb-mi as well, but whatever.

PR #383, opened 2016-07-07
Fix enable breakpoint under run mode

faxue-msft:
Before the fix, trying to enable a disabled breakpoint in run mode won't have affect, even after we enter break mode later it won't be enabled, the breakpoint will be "Pending" forever.
This fix won't affect VSCode scenario because VSCode only insert/delete, and the enable/disable code path isn't executed.

Fix bug https://devdiv.visualstudio.com/DevDiv/VS%20Diag%20IntelliTrace/_workItems/index?_a=edit&id=226898&triage=true

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
@@ -20,10 +20,20 @@ namespace Microsoft.MIDebugEngine
         private AD7Engine _engine;
         private BoundBreakpoint _bp;
 
-        private bool _enabled;
         private bool _deleted;
 
-        internal bool Enabled { get { return _enabled; } }
+        internal bool Enabled
+        {
+            get
+            {
+                return _bp.Enabled;
+            }
+            set
+            {
+                _bp.Enabled = value;
+            }
+        }
+
         internal bool Deleted { get { return _deleted; } }
         internal ulong Addr
         {
@@ -38,7 +48,6 @@ namespace Microsoft.MIDebugEngine
             _engine = engine;
             _pendingBreakpoint = pendingBreakpoint;
             _breakpointResolution = breakpointResolution;
-            _enabled = true;
             _deleted = false;
             _bp = bp;
         }
@@ -65,7 +74,7 @@ namespace Microsoft.MIDebugEngine
         // Called by the debugger UI when the user is enabling or disabling a breakpoint.
         int IDebugBoundBreakpoint2.Enable(int fEnable)
         {
-            _enabled = fEnable == 0 ? false : true;
+            Enabled = fEnable == 0 ? false : true;
             return Constants.S_OK;
         }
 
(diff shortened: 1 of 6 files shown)


faxue-msft:
@gregg-miskelly @caslan @paulmaybee

faxue-msft:
@caslan 
I'm not sure whether the "-insert-break -d" option is supported by Clrdbg MI, but if clrdbg is only executed in VSCode this won't matter, since VSCode don't disable/enable bps.

caslan:
clrdbg is not only in VS Code.

caslan:
@chuckries

chuckries:
`-d` is not currently supported in clrdbg-mi. We will parse the option, but not use it. It also looks like we do net yet implement -break-enable or -break-disable.

src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
```diff
@@ -65,7 +74,7 @@ public int Delete()
         // Called by the debugger UI when the user is enabling or disabling a breakpoint.
         int IDebugBoundBreakpoint2.Enable(int fEnable)
         {
-            _enabled = fEnable == 0 ? false : true;
+            Enabled = fEnable == 0 ? false : true;
```
gregg-miskelly: Technically this isn't quite correct, though it might not matter for the MIEngine -- this method is supposed to set the backend state, which means it should really be updating the backend if necessary, instead of just updating the property.

faxue-msft: Right, we need to disable the bound breakpiont with something like "disable 2.1", 
For this change I'm only renaming the member. I'll add a TODO here and log a bug. And fix this in a separate change

---
In reply to: [70909937](https://github.com/Microsoft/MIEngine/pull/383#discussion_r70909937) [](ancestors = 70909937)

faxue-msft:
@chuckries , did you have the chance to actually try this on clrdbg? I guess it's not totally necessary anyway since -d option is not really parsed by clrdbg.

---
In reply to: [232116025](https://github.com/Microsoft/MIEngine/pull/383#issuecomment-232116025) [](ancestors = 232116025)

chuckries:
@faxue-msft not yet, I will try this today.

src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
faxue-msft: I add an issue at https://github.com/Microsoft/MIEngine/issues/390 for not able to disabled only one of the boundbreakpoints

borgdylan:
VSCode can indeed enable/disable breakpoints (it appears so from the UI). The debugger engine misbehaves under clrdbg by rejecting breakpoints that are not in c# files but are indeed pointed to by the pdb files.

faxue-msft:
@borgdylan VsCode's enable/disable breakpoints will actually ask the underlying debugger to add/remove breakpoints, so it's not actually enable/disable from debugger's perspective

PR #384, opened 2016-07-08
Previous console work for GDB does not work with LLDB

pieandcakes:
Jackson made changes in LLDB to support -gdb-set new-console on in LLDB to
mimic the way we do console/terminal windows in Windows.

NOTE: the changes from the lldb side are also required

This also includes a change to ignore SIGHUP at the debugger level and
pass it to the debuggee. This happens when the console/terminal window is
closed.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -189,8 +189,7 @@ namespace Microsoft.MIDebugEngine
                 // For local Linux and OS X launch, use the local Unix transport which creates a new terminal and
                 // uses fifos for debugger (e.g., gdb) communication.
                 if (this.MICommandFactory.UseExternalConsoleForLocalLaunch(localLaunchOptions) &&
-                    (PlatformUtilities.IsLinux() || PlatformUtilities.IsOSX())
-                    )
+                    (PlatformUtilities.IsLinux() || (PlatformUtilities.IsOSX() && localLaunchOptions.DebuggerMIMode != MIMode.Lldb)))
                 {
                     localTransport = new LocalUnixTerminalTransport();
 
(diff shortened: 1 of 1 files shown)


pieandcakes:
@edumunoz @jacdavis @gregg-miskelly @chuckries

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -189,7 +189,7 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
                 // For local Linux and OS X launch, use the local Unix transport which creates a new terminal and
                 // uses fifos for debugger (e.g., gdb) communication.
                 if (this.MICommandFactory.UseExternalConsoleForLocalLaunch(localLaunchOptions) &&
-                    (PlatformUtilities.IsLinux() || PlatformUtilities.IsOSX())
+                    (PlatformUtilities.IsLinux() || (PlatformUtilities.IsOSX() && localLaunchOptions.DebuggerMIMode == MIMode.Gdb))
```
edumunoz: This is used for clrdbg as well. I think you want

`localLaunchOptions.DebuggerMIMode != MIMode.Lldb`

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -651,7 +651,7 @@ private List<LaunchCommand> GetInitializeCommands()
                     }
 
                     if (localLaunchOptions != null &&
-                        PlatformUtilities.IsWindows() &&
+                        (PlatformUtilities.IsWindows() || (PlatformUtilities.IsOSX() && this.MICommandFactory.Mode == MIMode.Lldb)) &&
```
edumunoz: I assume this problem is not limited to OS X. Maybe add a `TODO` so that we remember to remove the check for OS X if/when we add support for LLDB on Linux?

edumunoz:
:ship:

chuckries:
:+1:

PR #385, opened 2016-07-08
Fix for Stack Frame parameters on LLDB

pieandcakes:
LLDB already appends parameter types to function names in its MI stack
info calls.

Also fixed some spelling errors.

@wiktork @gregg-miskelly @chuckries

### DIFF
--- a/src/MICore/CommandFactories/lldb.cs
+++ b/src/MICore/CommandFactories/lldb.cs
@@ -36,6 +36,16 @@ namespace MICore
             return false;
         }
 
+        public override bool SupportsFrameFormatting
+        {
+            get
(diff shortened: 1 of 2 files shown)


wiktork:
LGTM

faxue-msft:
this fixed is also needed by MDD IOS scenario, LGTM

chuckries:
Does this have anything to do with PR #374 that jackson abandoned?

pieandcakes:
Yes, This is a workaround for the time being to prevent us from adding the parameters, thus avoiding duplication.

chuckries:
:+1:

PR #386, opened 2016-07-11
Allow lldb-mi's conditional breakpoint -break-insert to work with -f

pieandcakes:
-break-insert's -f on lldb-mi requires a parameter.

@faxue-msft @chuckries

### DIFF
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -409,7 +409,7 @@ namespace MICore
 
         #region Breakpoints
 
-        private StringBuilder BuildBreakInsert(string condition)
+        protected virtual StringBuilder BuildBreakInsert(string condition)
         {
             StringBuilder cmd = new StringBuilder("-break-insert -f ");
             if (condition != null)
--- a/src/MICore/CommandFactories/lldb.cs
+++ b/src/MICore/CommandFactories/lldb.cs
@@ -46,6 +46,25 @@ namespace MICore
             }
         }
 
+        protected override StringBuilder BuildBreakInsert(string condition)
+        {
+            // LLDB's use of the pending flag requires an optional parameter or else it fails.
+            // We will use "on" for now. 
+            // TODO: Fix this on LLDB-MI's side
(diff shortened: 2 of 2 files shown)


src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -411,7 +411,17 @@ public virtual Task<bool> SetStepFiltering(bool enabled)
 
         private StringBuilder BuildBreakInsert(string condition)
         {
-            StringBuilder cmd = new StringBuilder("-break-insert -f ");
+            string pendingFlag = "-f ";
+            if(this.Mode == MIMode.Lldb)
```
andrewcrawley: Formatting nitpick: missing a space between the "if" and the open-paren.

edumunoz: Why not make this method `protected` and override it in `LlldbMICommandFactory`?

pieandcakes: @edumunoz Forgot that was in an overloaded class. Took your suggestion

faxue-msft:
:shipit:

src/MICore/CommandFactories/lldb.cs
```diff
@@ -46,6 +46,25 @@ public override bool SupportsFrameFormatting
             }
         }
 
+        protected override StringBuilder BuildBreakInsert(string condition)
+        {
+            // LLDB's use of the pending flag requires an optional parameter or else it fails.
+            // We will use "on" for now. 
+            // TODO: Fix this on LLDB-MI's side
+            string pendingFlag = "-f on ";
```
edumunoz: Move `pendingFlag` as part of ` cmd` like it's done in `MICommandFactory`?

edumunoz:
:ship: it

src/MICore/CommandFactories/lldb.cs
pieandcakes: I wanted to call out specifically what was different and why it was different. I'll keep it separate for now until we fix LLDB side

chuckries:
:+1:

PR #387, opened 2016-07-12
Change signing error message to not be shown for lldb-mi

pieandcakes:
The executable that is signed is not lldb-mi but debugserver

@edumunoz


pieandcakes:
I thought about making this GDB only. let me know if you think that's a better idea.

edumunoz:
:+1:

PR #391, opened 2016-07-21
Users/jacdavis

jacdavis:
Fix working directory for lldb launch.

    Normally, we set the working dir using the -environment-cd
    mi command. lldb does support this command, but sadly, it doesn't
    work, at least on mac. Instead, I'm setting the working dir
    of lldb at launch which will carry over to the debuggee.

### DIFF
--- a/src/MICore/Transports/LocalTransport.cs
+++ b/src/MICore/Transports/LocalTransport.cs
@@ -27,7 +27,10 @@ namespace MICore
             Process proc = new Process();
             proc.StartInfo.FileName = localOptions.MIDebuggerPath;
             proc.StartInfo.Arguments = "--interpreter=mi";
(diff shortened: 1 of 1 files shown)


jacdavis:
@gregg-miskelly @pieandcakes @edumunoz

pieandcakes:
:shipit:

edumunoz:
Does this affect the ability of `lldb-mi` to find other binaries in its current directory (e.g., `debugserver`)?

jacdavis:
Doesn’t seem to. Testing isn’t finding any issue with that.

edumunoz:
`lldb-mi` will use the system provided `debugserver` if it can't find the one beside it.
I would double check it's not using that one.

PR #392, opened 2016-07-25
InstallToVSCode: remove System.Diagnostics.StackTrace dependency

gregg-miskelly:
InstallToVSCode declared a dependency on System.Diagnostics.StackTrace. As
far as I can tell, this is not needed, and having it breaks Dev15 versions
of clrdbg as it pulls down an older version of System.Reflection.Metadata.
This removes the dependency.


caslan:
LGTM

gregg-miskelly:
@rajkumar42 can you take a quick look?

rajkumar42:
:+1:

PR #396, opened 2016-07-27
Jacdavisrelease

jacdavis:
@gregg-miskelly @pieandcakes @richardstanton @paulmaybee @chuckries

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -517,7 +517,7 @@ namespace Microsoft.MIDebugEngine
                 }
 
                 // Must wait to get the target architecture until the first break event.
-				EnqueueInternalBreakAction(ObtainTargetArchitecture);
+                EnqueueInternalBreakAction(ObtainTargetArchitecture);
 
                 success = true;
             }
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -516,20 +516,8 @@ namespace Microsoft.MIDebugEngine
                     }
                 }
 
-                var arch = await MICommandFactory.GetTargetArchitecture();
-                if (arch == TargetArchitecture.Unknown)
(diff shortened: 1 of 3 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -516,20 +516,8 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
...
-                    }
-                    else
-                    {
-                        WriteOutput(ResourceStrings.Warning_UsingDefaultArchitecture);
-                        arch = TargetArchitecture.X64;  // use as default
-                    }
-                }
-                SetTargetArch(arch);
+                // Must wait to get the target architecture until the first break event.
+				EnqueueInternalBreakAction(ObtainTargetArchitecture);
```
gregg-miskelly: >ObtainTargetArchitecture [](start = 31, length = 24)

This winds up setting the 'Is64BitArch' property which is used before the first time we enter break state (ex: AD7Module.cs(88)). I wonder if we need a different way of detecting this.

jacdavis: shoot. I didn't see that one. This is very unfortunate because this bug is completely breaking our last release so we wanted to ship this today.

jacdavis: I wonder if I can execute this code sequence before the target is started to ensure break mode. Let me try that.

gregg-miskelly: Maybe that doesn't matter for VS Code - you could run some experiments and find out. Assuming it doesn't, you could checkin your changes to release-cpptools now, and fix lldb-mi for the next release.

jacdavis: It wouldn't matter for vscode, but I'd rather see if I can get the correct fix sooner rather than later.

jacdavis:
Gregg found a fatal flaw in this change. While it is possible to ship vscode with this, I'm going to try a fix that will work everywhere.

PR #398, opened 2016-07-28
Fix for racecondition on startup with target architecture fix.

jacdavis:
The target architecture fix introduced a race condition between
the -exec-run or -target-attach and the module loads finishing.
This was causing many debuggees to fail during attach or startup
due to the process state not being able to send the console command
to obtain the architecture. This moves the acquisition of that state
to between the -file-exec-and-symbols and the actual running.

NOTE: needed to set the binary on attach so that the debuggers
know the architecture. Apparently, the miengine wasn't doing that
before.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -661,7 +661,7 @@ namespace MICore
             PostCommand("-gdb-exit");
         }
 
-        private string Escape(string str)
+        protected string Escape(string str)
         {
             StringBuilder outStr = new StringBuilder();
             for (int i = 0; i < str.Length; ++i)
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -661,7 +661,7 @@ namespace MICore
             PostCommand("-gdb-exit");
         }
 
-        protected string Escape(string str)
+        private string Escape(string str)
         {
             StringBuilder outStr = new StringBuilder();
             for (int i = 0; i < str.Length; ++i)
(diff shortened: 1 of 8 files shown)


jacdavis:
@gregg-miskelly @paulmaybee @pieandcakes

gregg-miskelly:
>             }

Call AddGetTargetArchitectureCommand here unless the TargetArchitecture is already set in the launch options?

---
Refers to: src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs:589 in f4cb620. [](commit_id = f4cb6202ff3616e92c3fb8d111c860e29beda6c4, deletion_comment = False)

gregg-miskelly:
Otherwise LGTM

src/MICore/Debugger.cs
```diff
@@ -661,7 +661,7 @@ public void CmdExitAsync()
             PostCommand("-gdb-exit");
         }
 
-        private string Escape(string str)
+        protected string Escape(string str)
```
pieandcakes: Did you mean to change this? I don't see this used anywhere.

pieandcakes:
otherwise :shipit:

src/MICore/Debugger.cs
jacdavis: I did because I was originally using it. I must have pulled that out though.

gregg-miskelly:
LGTM

pieandcakes:
🚢

jacdavis:
Wound up putting this into master as we decided not to submit this for the current release. Abandaning the pull request, but this is submitted.

PR #401, opened 2016-08-01
Address some of the PR feedback

caslan:
- Made sure the new code is built only for the appropriate configuration and the solution continues to build for every configuration (I was getting build breaks with Desktop configuration)
- Moved unit test project over to MIEngine repo
- Some couple other minor changes per code review feedback


caslan:
@chuckries

PR #402, opened 2016-08-01
Build Break fix

pieandcakes:
@jander-msft @jacdavis 

Apparently using the RESX generator in VS breaks things. 

This is fixing the build break.

### DIFF
--- a/src/MIDebugEngine/ResourceStrings.Designer.cs
+++ b/src/MIDebugEngine/ResourceStrings.Designer.cs
@@ -10,6 +10,7 @@
 
 namespace Microsoft.MIDebugEngine {
     using System;
+	using System.Reflection;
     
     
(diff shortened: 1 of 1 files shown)


chuckries:
I can't find the PR that this came in with. Did the PR build not catch this or did we merge it without making sure the PR build passed?

chuckries:
:+1:

pieandcakes:
@chuckries it was in in https://github.com/Microsoft/MIEngine/commit/1da32dacab1da4f15c36c9f244fd55574d5defc2 . Apparently if you edit the resx with VS it removes these two lines which then causes the build break. @jander-msft  pointed it out and this is the fix for it.

jacdavis:
Squirrels and all that jazz

PR #407, opened 2016-08-09
Add noDebug flag for clrdbg

wesrupert:
Add `--noDebug` flag to clrdbg's `-exec-run` command to enable Ctrl-F5 support for VS Code.

### DIFF
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -242,7 +242,7 @@ namespace MICore
         /// <summary>
         /// Tells GDB to spawn a target process previous setup with -file-exec-and-symbols or similar
         /// </summary>
-        public async Task ExecRun()
+        public virtual async Task ExecRun()
         {
             string command = "-exec-run";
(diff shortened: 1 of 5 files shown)


wesrupert:
@caslan @gregg-miskelly @chuckries

caslan:
LGTM

gregg-miskelly:
Do you have the full end-to-end working? This approach feels slightly dirty to me -- just the fact that so  little of the MIEngine is aware of the fact that it isn't debugging. That said, if you have gotten everything to work and you haven't hit any snags with this approach I don't have a problem with it.

rajkumar42:
Would you be able to attach to the process that you launched with -nodebug? We should give a meaningful error message.

rajkumar42:
:shipit:

wesrupert:
End-to-end works. Since this change is to enable VS Code, which wants to track running state on the process, there is no further opportunity to attach from that instance of VS Code. Attaching to it via another instance or from VS should not be impacted, as the process is merely launched then monitored. Clrdbg should not have any other locks on the process or its debug info.

chuckries:
:+1:

PR #409, opened 2016-08-10
Switch to new credentials dialog from VC Linux extension

caslan:
- Removed previous credentials dialog code
- Added code to invoke new credentials dialog from VC Linux extension
- Fit and finish fixes around stale passwords, cancelling connecting to remote system etc.


caslan:
@gregg-miskelly @rajkumar42

src/SSHDebugPS/AD7Port.cs
```diff
@@ -49,10 +55,25 @@ public void EnsureConnected()
...
+
         public int EnumProcesses(out IEnumDebugProcesses2 processEnum)
         {
             IEnumDebugProcesses2 result = null;
             var connection = GetConnection(ConnectionReason.Deferred);
+
+            if (connection == null)
+            {
+                processEnum = null;
+                return HR.E_REMOTE_CONNECT_USER_CANCELED;
```
caslan: SDM currently doesn't suppress showing an error dialog for this value in this call as it does in other code paths. Added a work item for it.

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -20,54 +21,102 @@ internal static Connection GetInstance(string name, ConnectionReason reason)
...
+                        break;
+                    }
+                    catch (RemoteAuthenticationException)
+                    {
+                        IVsConnectionManager connectionManager = (IVsConnectionManager)ServiceProvider.GlobalProvider.GetService(typeof(IVsConnectionManager));
+                        IConnectionManagerResult result = connectionManager.ShowDialog(StringResources.AuthenticationFailureHeader, StringResources.AuthenticationFailureDescription, connectionInfo);
+
+                        if (result.DialogResult == ConnectionManagerDialogResult.Succeeded)
+                        {
+                            // Update the credentials in the store
```
caslan: This should probably be automatically handled by the dialog. Will change after dialog makes that change.

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -20,54 +21,102 @@ internal static Connection GetInstance(string name, ConnectionReason reason)
...
+                PasswordConnectionInfo newConnectionInfo = new PasswordConnectionInfo(hostName, userName, new System.Security.SecureString());
+
+                IConnectionManagerResult result = connectionManager.ShowDialog(newConnectionInfo);
+
+                if (result.DialogResult == ConnectionManagerDialogResult.Succeeded)
                 {
-                    return new Connection(remoteSystem);
+                    // Retrieve the newly added connection
+                    store.Load();
+                    connectionInfo = store.Connections.First(info => info.Id == result.StoredConnectionId);
```
rajkumar42: >info.Id == result.StoredConnectionId [](start = 69, length = 36)

Ignore case?

rajkumar42:
:shipit:

src/SSHDebugPS/ConnectionManager.cs
caslan: It's an "int", not a string

gregg-miskelly:
Is the dll a reference dll? Can we checkin an IL file instead? If we checkin the binary, it will be there forever.

gregg-miskelly:
:shipit:

PR #410, opened 2016-08-12
Updating GetClrDbg.sh script to accept switch arguments.

rajkumar42:
Added functionality to remove existing installation, skip downloads and launch clrdbg after installation.
And some refactoring.

### DIFF
--- a/scripts/GetClrDbg.sh
+++ b/scripts/GetClrDbg.sh
@@ -30,8 +30,6 @@ get_script_directory()
     popd > /dev/null 2>&1
 }
 
-get_script_directory
-
 print_help()
 {
     echo 'GetClrDbg.sh [-usdh] -v V [-l L]'
@@ -307,6 +305,8 @@ check_latest()
         echo "Info: Previous installation at "$__InstallLocation" not found"
     fi
 }
+
+get_script_directory
 
 if [ -z "$1" ]; then
     print_help
--- a/scripts/GetClrDbg.sh
+++ b/scripts/GetClrDbg.sh
@@ -310,6 +310,7 @@ check_latest()
 
 if [ -z "$1" ]; then
     print_help
+    echo "Error: Missing arguments for GetClrDbg.sh"
     exit 1
 else
     parse_and_get_arguments $@
@@ -360,14 +361,14 @@ else
     echo 'Info: Executing dotnet restore'
     dotnet restore > dotnet_restore.log 2>&1
     if [ $? -ne 0 ]; then
-        echo "dotnet restore failed"
+        echo "Error: dotnet restore failed"
         exit 1
     fi
 
     echo 'Info: Executing dotnet publish'
     dotnet publish -o . > dotnet_publish.log 2>&1
     if [ $? -ne 0 ]; then
-        echo "dotnet publish failed"
+        echo "Error: dotnet publish failed"
         exit 1
     fi
 
--- a/scripts/GetClrDbg.sh
+++ b/scripts/GetClrDbg.sh
@@ -16,7 +16,7 @@ __SkipDownloads=false
 __LaunchClrDbg=false
 
 # Removes existing installation of ClrDbg in the Install Location.
-__RemoveExisting=false
+__RemoveExistingOnUpgrade=false
 
 # Internal, fully specified version of the ClrDbg. Computed when the meta version is used.
 __ClrDbgVersion=
@@ -34,10 +34,10 @@ get_script_directory
 
 print_help()
 {
-    echo 'GetClrDbg.sh [-rsdh] -v V [-l L]'
+    echo 'GetClrDbg.sh [-usdh] -v V [-l L]'
     echo ''
     echo 'This script downloads and configures clrdbg, the Cross Platform .NET Debugger'
-    echo '-r    Removes the existing installation before installing the specified version. Can be used for a clean install of the debugger.'
+    echo '-u    Deletes the existing installation directory of the debugger before installing the current version.'
     echo '-s    Skips any steps which requires downloading from the internet.'
     echo '-d    Launches debugger after the script completion.'
     echo '-h    Prints usage information.'
@@ -118,7 +118,7 @@ generate_nuget_config()
 # Parses and populates the arguments
 parse_and_get_arguments()
 {
-    while getopts "v:l:srhd" opt; do
+    while getopts "v:l:suhd" opt; do
         case $opt in
             v)
                 __ClrDbgMetaVersion=$OPTARG;
@@ -126,8 +126,8 @@ parse_and_get_arguments()
             l)
                 __InstallLocation=$OPTARG
                 ;;
-            r)
-                __RemoveExisting=true
+            u)
+                __RemoveExistingOnUpgrade=true
                 ;;
             s)
                 __SkipDownloads=true
@@ -142,8 +142,8 @@ parse_and_get_arguments()
             \?)
                 echo "Error: Invalid Option: -$OPTARG"
                 print_help
-                exit 1
-                ;;
+                exit 1;
+				;;
             :)
                 echo "Error: Option expected for -$OPTARG"
                 print_help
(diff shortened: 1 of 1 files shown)


rajkumar42:
@chuckries  @gregg-miskelly  @wesrupert 
Tested with Linux. Will test it on mac.

scripts/GetClrDbg.sh
```diff
@@ -76,97 +103,238 @@ generate_nuget_config()
...
-    print_help
-    exit 1
-elif [ "$1" == "-h" ]; then
-    print_help
-    exit 1
-fi
+# Parses and populates the arguments
+parse_and_get_arguments()
+{
+    echo "Parsing Arguments..."
```
gregg-miskelly: Seems like the script shouldn't output this level of information except maybe in some sort of verbose mode.

scripts/GetClrDbg.sh
```diff
@@ -76,97 +103,238 @@ generate_nuget_config()
...
-            echo "Error: '$1' does not look like a valid version number."
-            exit 1
-        fi
-        __ClrDbgVersion=$1
-        ;;
-esac
+# Parses and populates the arguments for the legacy commandline.
+get_legacy_arguments()
+{
+    echo "Parsing Legacy Arguments... consider moving to using switches."
```
gregg-miskelly: I don't know that we want this output...

scripts/GetClrDbg.sh
```diff
@@ -76,97 +103,238 @@ generate_nuget_config()
...
+    fi
+}
+
+# Converts relative location of the installation directory to absolute location.
+convert_install_path_to_absolute()
+{
+    if [ -z $__InstallLocation ]; then
+        __InstallLocation=$pwd
+    else
+        __InstallLocation=$script_dir/$__InstallLocation
```
gregg-miskelly: Isn't this wrong if $__InstalationLocation is a full path?

scripts/GetClrDbg.sh
```diff
@@ -76,97 +103,238 @@ generate_nuget_config()
...
-pushd $__InstallPath > /dev/null 2>&1
-if [ "$?" -ne 0 ]; then
-    echo "Error: Unable to cd to install directory '$__InstallPath'"
+if [ -z "$1" ]; then
+    print_help
     exit 1
+else
+    parse_and_get_arguments $@
+
+    if [ -z $__ClrDbgMetaVersion ]; then
```
gregg-miskelly: Maybe I am missing something, but this seems like the wrong check to me. It seems like we really want something like -- the first argument that doesn't start with '-' is the version, the second is the install dir.

scripts/GetClrDbg.sh
```diff
@@ -1,15 +1,42 @@
...
 print_help()
 {
-    echo 'GetClrDbg.sh <version> [<install path>]'
+    echo 'GetClrDbg.sh [-rsdh] -v V [-l L]'
     echo ''
     echo 'This script downloads and configures clrdbg, the Cross Platform .NET Debugger'
-    echo 'If <install path> is not specified, clrdbg will be installed to the directory'
-    echo 'from which this script was executed.' 
-    echo '<version> can be "latest" or a version number such as 14.0.25109-preview-2865786'
+    echo '-r    Deletes the existing installation of the debugger at the installation location. Can be used for a clean install of the debugger.'
```
gregg-miskelly: The '-r' argument doesn't seem to act quite like what I was thinking. I could definitely see this being useful for some scenarios, but the one I think we need is 'clean the old directory on upgrade'.

scripts/GetClrDbg.sh
```diff
@@ -1,15 +1,42 @@
 #! /bin/bash
 
-script_dir=`dirname $0`
+script_dir="$( cd $( dirname "$0" ); pwd )"
```
wesrupert: `pushd`/`popd` to avoid state modification in an assignment

scripts/GetClrDbg.sh
rajkumar42: Thanks for pointing out! Forgot the remove the logging. I don't think there is a need for verbose yet.

scripts/GetClrDbg.sh
rajkumar42: got it!

scripts/GetClrDbg.sh
rajkumar42: its pretty complex to determine if a path is rooted or not in *nix. To make it simpler I have limited the location to be relative to script directory. Any suggestion?

scripts/GetClrDbg.sh
rajkumar42: getopts breaks the while loop when it finds a non-hypen token. So it will not populate __ClrDbgMetaVersion.

scripts/GetClrDbg.sh
rajkumar42: This is a documentation error. "-r" will remove the directory. The install script will continue to download and install. Combining with "-s" will remove the directory and not download.

 Not specifying "-r" will result in install or upgrade.

scripts/GetClrDbg.sh
rajkumar42: Can you provide the one-liner?

For me using pushd/popd puts the pushd stack on the variable.

scripts/GetClrDbg.sh
gregg-miskelly: I think I would suggest first ensuring the directory exists, and then pushd'ing to it. Then you can get the full directory name via pwd.

scripts/GetClrDbg.sh
gregg-miskelly: I see. That works.

scripts/GetClrDbg.sh
rajkumar42: Fixed it using the method you suggested. Thanks!

scripts/GetClrDbg.sh
rajkumar42: Added a sanity check.

scripts/GetClrDbg.sh
rajkumar42: Fixed.

scripts/GetClrDbg.sh
```diff
@@ -76,97 +115,271 @@ generate_nuget_config()
...
-echo 'Info: Executing dotnet publish'
-dotnet publish -o . > dotnet_publish.log 2>&1
-if [ $? -ne 0 ]; then
-    echo "dotnet publish failed"
-    exit 1
+
+if [ "$__SkipDownloads" = true ]; then
+    echo "Info: Skipping downloads"
+else
+    prepare_install_location
```
gregg-miskelly: >prepare_install_location [](start = 4, length = 24)

Do we still need this?

scripts/GetClrDbg.sh
```diff
@@ -76,97 +115,271 @@ generate_nuget_config()
...
+    pushd $__InstallLocation > /dev/null 2>&1
+    if [ "$?" -ne 0 ]; then
+        echo "Error: Unable to cd to install directory '$__InstallLocation'"
+        exit 1
+    fi
```
gregg-miskelly: >fi [](start = 4, length = 2)

In upgrade scenarios, do we want to blow away all the old files before we attempt to download the new ones?

scripts/GetClrDbg.sh
rajkumar42: Yep, if -r is specified, we will remove that folder.

gregg-miskelly:
LGTM

scripts/GetClrDbg.sh
```diff
@@ -1,15 +1,54 @@
...
+
+# Gets the script directory
+get_script_directory()
+{
+    pushd $(dirname "$0") > /dev/null 2>&1
+    __ScriptDirectory=$(pwd)
+    popd > /dev/null 2>&1
+}
+
+get_script_directory
```
chuckries: [nit] place all function definitions before executing script execution.

chuckries:
:+1:

wesrupert:
LGTM! :shipit:

PR #411, opened 2016-08-16
Fix DebuggedProcess.HandleBreakModeEvent for deleted breakpoints

gregg-miskelly:
The change to support async stop events had a bug that caused us to leave the debuggee in stop state but not send a stopping event to the UI if a breakpoint was deleted in run mode by the user, but the breakpoint was hit before it made it to the command line debugger.

This fixes it.


gregg-miskelly:
@paulmaybee please review

paulmaybee:
👍

PR #413, opened 2016-08-16
Deploying ClrDbg as part of debugger initialization for SSH transport.

rajkumar42:
- Downloading and invoking GetClrDbg.sh on the remote machine.
- Displaying text on the progress dialog while the clrdbg is getting installed.
- Displaying error message when the installation fails.


rajkumar42:
abandoning pr.

PR #414, opened 2016-08-16
Deploying ClrDbg.

rajkumar42:
- Downloading and invoking GetClrDbg.sh on the remote machine.
- Displaying text on the progress dialog while the clrdbg is getting installed.
- Displaying error message when the installation fails.

### DIFF
--- a/src/DebugEngineHost/VSImpl/VsWaitLoop.cs
+++ b/src/DebugEngineHost/VSImpl/VsWaitLoop.cs
@@ -120,5 +120,10 @@ namespace Microsoft.DebugEngineHost.VSImpl
         {
             _messagePump.SetProgressInfo(totalSteps, currentStep, progressText);
         }
+
+        public void SetText(string text)
+        {
+            _messagePump.SetWaitText(text);
+        }
     }
 }
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -405,6 +405,13 @@ namespace MICore
         public string StartRemoteDebuggerCommand { get; private set; }
         public Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.IDebugUnixShellPort UnixPort { get; private set; }
 
+        public static bool LastDebuggerLaunchSuccessful = false;
+
+        private static string GetClrDbgUrl = "https://raw.githubusercontent.com/Microsoft/MIEngine/dev/rajkumar42/updatescript/scripts/GetClrDbg.sh";
+        private static string DebugerInstallationDirectory = "~/.vs-debugger";
+        private static string DebuggerFirstLaunchCommand = "mkdir -p {0} && wget -q {1} -O {0}/GetClrDbg.sh && chmod +x {0}/GetClrDbg.sh && {0}/GetClrDbg.sh -v latest -l {0}/latest -d -u";
+        private static string DebuggerSubsequentLaunchCommand = "{0}/GetClrDbg.sh -v latest -l {0}/latest -d";
+
         public UnixShellPortLaunchOptions(string startRemoteDebuggerCommand, Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.IDebugUnixShellPort unixPort, MIMode miMode, BaseLaunchOptions baseLaunchOptions)
         {
             this.UnixPort = unixPort;
@@ -422,8 +429,14 @@ namespace MICore
                         startRemoteDebuggerCommand = "lldb-mi --interpreter=mi";
                         break;
                     case MIMode.Clrdbg:
-                        // TODO: Replace with a set of commands for downloading clrdbg from the web if it isn't already present
-                        startRemoteDebuggerCommand = "~/clrdbg/out/Linux/bin/x64.Debug/clrdbg/clrdbg --interpreter=mi";
+                        if (!LastDebuggerLaunchSuccessful)
+                        {
+                            startRemoteDebuggerCommand = string.Format(CultureInfo.InvariantCulture, DebuggerFirstLaunchCommand, DebugerInstallationDirectory, GetClrDbgUrl);
+                        }
+                        else
+                        {
+                            startRemoteDebuggerCommand = string.Format(CultureInfo.InvariantCulture, DebuggerSubsequentLaunchCommand, DebugerInstallationDirectory);
+                        }
                         break;
 
                     default:
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -18,6 +18,7 @@ using System.Diagnostics;
 using Microsoft.DebugEngineHost;
 using MICore.Xml.LaunchOptions;
 using System.Reflection;
+using Microsoft.VisualStudio.Debugger.Interop;
 
 namespace MICore
 {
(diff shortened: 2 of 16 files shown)


rajkumar42:
@caslan @gregg-miskelly @chuckries @wesrupert

src/MICore/LaunchOptions.cs
```diff
@@ -405,6 +405,13 @@ public sealed class UnixShellPortLaunchOptions : LaunchOptions
         public string StartRemoteDebuggerCommand { get; private set; }
         public Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.IDebugUnixShellPort UnixPort { get; private set; }
 
+        public static bool LastDebuggerLaunchSuccessful = false;
+
+        private static string GetClrDbgUrl = "https://raw.githubusercontent.com/Microsoft/MIEngine/dev/rajkumar42/updatescript/scripts/GetClrDbg.sh";
+        private static string DebugerInstallationDirectory = "~/.vs-debugger";
```
caslan: typo, "debuger"

src/MICore/LaunchOptions.cs
```diff
@@ -405,6 +405,13 @@ public sealed class UnixShellPortLaunchOptions : LaunchOptions
...
+        private static string DebuggerFirstLaunchCommand = "mkdir -p {0} && wget -q {1} -O {0}/GetClrDbg.sh && chmod +x {0}/GetClrDbg.sh && {0}/GetClrDbg.sh -v latest -l {0}/latest -d -u";
```
caslan: I would add comment explaining the difference between first and second launch commands

caslan: in terms of what they actually do

src/MICore/LaunchOptions.cs
```diff
@@ -405,6 +405,13 @@ public sealed class UnixShellPortLaunchOptions : LaunchOptions
         public string StartRemoteDebuggerCommand { get; private set; }
         public Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.IDebugUnixShellPort UnixPort { get; private set; }
 
+        public static bool LastDebuggerLaunchSuccessful = false;
+
+        private static string GetClrDbgUrl = "https://raw.githubusercontent.com/Microsoft/MIEngine/dev/rajkumar42/updatescript/scripts/GetClrDbg.sh";
```
caslan: As discussed, we can pull these into engine metrics

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -616,7 +618,10 @@ private List<LaunchCommand> GetInitializeCommands()
 
                     CheckCygwin(commands, localLaunchOptions);
 
-                    this.AddExecutablePathCommand(commands);
+                    if (this.MICommandFactory.Mode != MIMode.Clrdbg)
```
caslan: You might want to add a comment why this is not applicable for clrdbg

src/MICore/Transports/UnixShellPortTransport.cs
```diff
@@ -17,17 +18,24 @@ public class UnixShellPortTransport : ITransport, IDebugUnixShellCommandCallback
         private string _startRemoteDebuggerCommand;
         private IDebugUnixShellAsyncCommand _asyncCommand;
         private bool _bQuit;
+        private bool _debuggerLaunched = false;
+        private HostWaitLoop _waitLoop;
+
+        private static string ErrorPrefix = "Error:";
```
wesrupert: Do we need to worry about localization with this prefix?

src/DebugEngineHost/VSImpl/VsWaitLoop.cs
```diff
@@ -120,5 +120,10 @@ public void SetProgress(int totalSteps, int currentStep, string progressText)
         {
             _messagePump.SetProgressInfo(totalSteps, currentStep, progressText);
         }
+
+        public void SetText(string text)
```
gregg-miskelly: >public [](start = 8, length = 6)

FYI, we should implement this in OpenDebugAD7 also

src/MICore/LaunchOptions.cs
```diff
@@ -405,6 +405,13 @@ public sealed class UnixShellPortLaunchOptions : LaunchOptions
         public string StartRemoteDebuggerCommand { get; private set; }
         public Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.IDebugUnixShellPort UnixPort { get; private set; }
 
+        public static bool LastDebuggerLaunchSuccessful = false;
```
gregg-miskelly: >LastDebuggerLaunchSuccessful [](start = 27, length = 28)

This needs to be a HashSet of port name instead of just a bool.

src/MICore/LaunchOptions.cs
gregg-miskelly: >latest [](start = 160, length = 6)

We shouldn't be using 'latest'.

gregg-miskelly: >wget [](start = 76, length = 4)

One problem with this command: If the user already has a version of clrdbg, and wget fails (ex: no internet connectivity), we will not continue on and start clrdbg. I think you want something like:

( wget -q {1} -O {0}/GetClrDbg.download && mv {0}/GetClrDbg.download {0}/GetClrDbg.sh ) || [ -f {0}/GetClrDbg.sh ]

src/MICore/Transports/UnixShellPortTransport.cs
gregg-miskelly: No

---
In reply to: [75037013](https://github.com/Microsoft/MIEngine/pull/414#discussion_r75037013) [](ancestors = 75037013)

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -61,6 +62,7 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
             Engine = engine;
             _libraryLoaded = new List<string>();
             _loadOrder = 0;
+            _waitLoop = waitLoop;
```
gregg-miskelly: >_waitLoop [](start = 12, length = 9)

I wonder if we should avoid storing this in a member variable when it shouldn't be used after launch. If we do want it in a member variable, I think we want to make sure we clear it once Initialize returns.

gregg-miskelly:
:shipit:

src/MICore/LaunchOptions.cs
gregg-miskelly: If you use '-q' with wget it will not output an error if there is no internet connectivity.

rajkumar42: Agreed! I've incorporated the changes.

src/MICore/LaunchOptions.cs
rajkumar42: Done!

src/MICore/LaunchOptions.cs
rajkumar42: Nice catch!

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
rajkumar42: Yep it makes sense.

gregg-miskelly:
LGTM

PR #418, opened 2016-08-20
Not deploying the contracts assembly. 

rajkumar42:
Not deploying the contracts assembly. 
And wrapping the presence of liblinux check with try/catch.

### DIFF
--- a/src/MIDebugPackage/SuppressFromVsix.targets
+++ b/src/MIDebugPackage/SuppressFromVsix.targets
@@ -14,5 +14,6 @@
     <SuppressFromVsix Include="System.Runtime.Serialization.dll"/>
     <SuppressFromVsix Include="Newtonsoft.Json.dll"/>
     <SuppressFromVsix Include="liblinux.dll"/>
+    <SuppressFromVsix Include="Microsoft.VisualStudio.Linux.ConnectionManager.Contracts.dll"/>
   </ItemGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


rajkumar42:
@gregg-miskelly

caslan:
Per one of the work items in our list, we should change this code to just querying for the service. If a version of VC++ extension with IVsConnectionManager is not installed, QueryService call will fail. We already do QueryService in ConnectionManager class

caslan:
Just realized the other checkin already does this. I'm not sure why this change is needed.

gregg-miskelly:
It is needed because IVsConnectionManager is defined in an interface which may not be installed.

LGTM

gregg-miskelly:
One other thought: might want to add a comment describing while the catch is needed.

rajkumar42:
@gregg-miskelly i did add a comment.. did i miss something?

caslan:
We could just query for guid as well, but if this works, ok.

gregg-miskelly:
@rajkumar42 Ah... I was looking for it on the `catch` rather than the `try`. That works.

gregg-miskelly:
@caslan `IVsConnectionManager` didn't have a GuidAttribute defined. So if we want to query by Guid we would need to change the interface.

rajkumar42:
Thanks @gregg-miskelly @caslan, I'm merging this now. If there is any actionable feedback later, I will send a different PR to address that.

PR #422, opened 2016-08-23
Enabling SSH Transport to work with GDB

rajkumar42:
Fixing: gdb prompt sent by the gdb has a trailing space.
Setting -file-exec-and-symbols for gdb.
try/catch eat DebuggerDisposedException on detach, which is expected.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -379,7 +379,7 @@ namespace MICore
             {
                 if (_isClosed)
                 {
-                    source.SetException(new ObjectDisposedException("Debugger"));
+                    source.SetException(new DebuggerDisposedException());
                 }
                 else
                 {
@@ -493,7 +493,7 @@ namespace MICore
             {
                 if (_internalBreakActionCompletionSource != null)
                 {
-                    _internalBreakActionCompletionSource.SetException(new ObjectDisposedException("Debugger"));
+                    _internalBreakActionCompletionSource.SetException(new DebuggerDisposedException());
                 }
                 _internalBreakActions.Clear();
             }
@@ -689,7 +689,7 @@ namespace MICore
             {
                 if (this.ProcessState == MICore.ProcessState.Exited)
                 {
-                    throw new ObjectDisposedException("Debugger");
+                    throw new DebuggerDisposedException();
                 }
                 else
                 {
@@ -704,7 +704,7 @@ namespace MICore
                 {
                     if (this.ProcessState == MICore.ProcessState.Exited)
                     {
-                        throw new ObjectDisposedException("Debugger");
+                        throw new DebuggerDisposedException();
                     }
                     else
                     {
@@ -761,7 +761,7 @@ namespace MICore
             {
                 if (_isClosed)
                 {
-                    throw new ObjectDisposedException("Debugger");
+                    throw new DebuggerDisposedException();
                 }
 
                 id = ++_lastCommandId;
@@ -971,11 +971,26 @@ namespace MICore
 
             internal void Abort()
             {
-                _completionSource.SetException(new ObjectDisposedException("Debugger"));
+                _completionSource.SetException(new DebuggerDisposedException(Command));
             }
         }
 
         private readonly Dictionary<uint, WaitingOperationDescriptor> _waitingOperations = new Dictionary<uint, WaitingOperationDescriptor>();
+
+        /// <summary>
+        /// Returns true it is a (gdb) prompt.
+        /// </summary>
+        /// <param name="prompt">The prompt from the debugger</param>
+        /// <returns>True if (gdb) prompt, false otherwise</returns>
+        /// <remarks>For SSH transport connecting to gdb, the (gdb) prompt string as a following space following it.</remarks>
+        private static bool IsGdbPrompt(string prompt)
+        {
+            const string gdbPrompt = "(gdb)";
+
+            return 
+                prompt.StartsWith(gdbPrompt, StringComparison.Ordinal) && 
+                prompt.Trim().Length == gdbPrompt.Length;
+        }
 
         public void ProcessStdOutLine(string line)
         {
@@ -983,7 +998,7 @@ namespace MICore
             {
                 return;
             }
-            else if (line == "(gdb)")
+            else if (IsGdbPrompt(line))
             {
                 if (_consoleDebuggerInitializeCompletionSource != null)
                 {
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -982,7 +982,7 @@ namespace MICore
         /// </summary>
         /// <param name="prompt">The prompt from the debugger</param>
         /// <returns>True if (gdb) prompt, false otherwise</returns>
-        /// <remarks>For SSH transport connecting to gdb, the (gdb) prompt string as a following space following it.</remarks>
+        /// <remarks>For SSH transport connecting to gdb, it has a trailing space following the prompt like "(gdb) ".</remarks>
         private static bool IsGdbPrompt(string prompt)
         {
             const string gdbPrompt = "(gdb)";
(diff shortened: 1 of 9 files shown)


rajkumar42:
@gregg-miskelly @jacdavis @caslan @wesrupert @chuckries

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -616,10 +616,19 @@ private List<LaunchCommand> GetInitializeCommands()
 
                     CheckCygwin(commands, localLaunchOptions);
 
-                    // ClrDbg doesn't need it.
-                    if (this.MICommandFactory.Mode != MIMode.Clrdbg)
+                    // ClrDbg doesn't need -file-exec-and-symbols set.
+                    if (this.MICommandFactory.Mode == MIMode.Gdb)
                     {
-                        this.AddExecutablePathCommand(commands);
+                        if (_launchOptions is UnixShellPortLaunchOptions)
```
gregg-miskelly: >_launchOptions is UnixShellPortLaunchOptions [](start = 28, length = 44)

Since your code doesn't actually depend on any particular transport, would it make sense to instead check if the executable path is set or not defined?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -744,6 +753,34 @@ private void AddExecutablePathCommand(IList<LaunchCommand> commands)
             commands.Add(new LaunchCommand("-file-exec-and-symbols " + exe, description, ignoreFailures: false, failureHandler: failureHandler));
         }
 
+        private void DetermineAndAddExecutablePathCommand(IList<LaunchCommand> commands)
+        {
+            // Runs a shell command to get the full path of the exe.
+            string absoluteExePath = string.Format(CultureInfo.InvariantCulture, @"shell readlink -f /proc/{0}/exe", _launchOptions.ProcessId);
```
gregg-miskelly: >" [](start = 82, length = 1)

This isn't going to work if the target OS is OSX. Might want to change the bash command to fail gracefully if it isn't Linux.

src/MICore/Debugger.cs
```diff
@@ -971,19 +971,34 @@ internal void OnComplete(Results results, MICommandFactory commandFactory)
...
         }
 
         private readonly Dictionary<uint, WaitingOperationDescriptor> _waitingOperations = new Dictionary<uint, WaitingOperationDescriptor>();
 
+        /// <summary>
+        /// Returns true it is a (gdb) prompt.
+        /// </summary>
+        /// <param name="prompt">The prompt from the debugger</param>
+        /// <returns>True if (gdb) prompt, false otherwise</returns>
+        /// <remarks>For SSH transport connecting to gdb, the (gdb) prompt string as a following space following it.</remarks>
```
caslan: You might want to fix this sentence for grammar. (From what I can tell you are saying (gdb) prompt has a space following it)

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
gregg-miskelly: Something like:
shell [ $(uname -s) == "Linux" ] && { readlink -f /proc/{0}/exe; true; } || echo "Error: Unsupported operating system family '$(uname -s)'"

---
In reply to: [75935390](https://github.com/Microsoft/MIEngine/pull/422#discussion_r75935390) [](ancestors = 75935390)

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -744,6 +753,34 @@ private void AddExecutablePathCommand(IList<LaunchCommand> commands)
...
+
+            Action<string> failureHandler = (string miError) =>
+            {
+                string message = string.Format(CultureInfo.CurrentUICulture, ResourceStrings.Error_FailedToGetExePath, miError);
```
gregg-miskelly: >miError [](start = 119, length = 7)

Do we get a reasonable error back if the process isn't one the user has rights to debug?

gregg-miskelly:
:shipit:

src/MICore/DebuggerDisposedException.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// </summary>
+        public DebuggerDisposedException(string message, Exception innerException, string abortedCommand = null) : base(message, innerException)
+        {
+            AbortedCommand = abortedCommand;
+        }
+
+        /// <summary>
+        /// Constructor for the DebuggerDisposedException which takes message and aborted command.
+        /// </summary>
+        public DebuggerDisposedException(string message, string abortedCommand = null) : base(Debugger, message)
```
wesrupert: How is this overload compatible with the existence of the constructor on line 21? Isn't the call `new DebuggerDisposedException("some string")` ambiguous?

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -771,7 +771,19 @@ public int Detach()
...
 
-            _pollThread.RunOperation(() => _debuggedProcess.CmdDetach());
+            try
+            {
+                _pollThread.RunOperation(() => _debuggedProcess.CmdDetach());
+            }
+            catch (DebuggerDisposedException e)
+            {
+                // Detach command could cause DebuggerDisposedException and we ignore that.
+                if (e.AbortedCommand != "-target-detach")
```
wesrupert: As of C# 6 you can use exception filters like so:

```csharp
catch (DebuggerDisposedException e) if (e.AbortedCommand != "-target-detach") 
{
    // Detach command could cause DebuggerDisposedException and we ignore that.
}
```

Does mono support this? More curious than a necessary change.

wesrupert: From the [msdn release article](https://blogs.msdn.microsoft.com/csharpfaq/2014/11/20/new-features-in-c-6/):

> If the parenthesized expression after ‘if’ evaluates to true, the catch block is run, otherwise the exception keeps going. Exception filters are preferable to catching and rethrowing because they leave the stack unharmed. If the exception later causes the stack to be dumped, you can see where it originally came from, rather than just the last place it was rethrown.

It's also more compact and looks nicer in my opinion.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
rajkumar42: I wanted it to be specific to *nix and GDB. I'll write a comment for now.

---
In reply to: [75935168](https://github.com/Microsoft/MIEngine/pull/422#discussion_r75935168) [](ancestors = 75935168)

src/MICore/DebuggerDisposedException.cs
rajkumar42: Good question! I guess [https://msdn.microsoft.com/en-us/library/dd264739.aspx](https://msdn.microsoft.com/en-us/library/dd264739.aspx) explains at "Overload Resolution" 

---
In reply to: [75942444](https://github.com/Microsoft/MIEngine/pull/422#discussion_r75942444) [](ancestors = 75942444)

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
rajkumar42: I've added a todo. 

---
In reply to: [75937575](https://github.com/Microsoft/MIEngine/pull/422#discussion_r75937575) [](ancestors = 75937575)

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
rajkumar42: I've added a todo. Will need some testing with liblinux to see if it is more robust.

---
In reply to: [75936997](https://github.com/Microsoft/MIEngine/pull/422#discussion_r75936997) [](ancestors = 75936997,75935390)

PR #425, opened 2016-08-24
Escape quotes in expression

edumunoz:
This is needed to evaluate expressions that contain quotes and spaces: e.g.,
`message = "hello world"`

This is needed for both GDB and LLDB, but not CLRDBG.

This was tested on Ubuntu (GDB), OS X (LLDB) and Windows (GDB from MinGW).

@pieandcakes @jacdavis


edumunoz:
More detailed description of the bug at https://devdiv.visualstudio.com/DevDiv/_workitems?id=256458

jacdavis:
:shipit:

PR #429, opened 2016-08-29
Fix adding disabled breakpoint for lldb-mi

Dmitri-Botcharnikov:
The disabled breakpoint inserted by MI command  `-break-insert -f -d hello3.cs:8` which is wrong sequence of flags (the filename:line should follow -f flags). LLDB-MI responds with the following message:
```
^error,msg="Command 'break-insert'. Command Args. Validation failed. Args missing additional information: f"
```
This PR places -f flag just before filename:line specification.
cc @chunseoklee @seanshpark @ayuckhulk

### DIFF
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -411,7 +411,7 @@ namespace MICore
 
         protected virtual StringBuilder BuildBreakInsert(string condition, bool enabled)
         {
-            StringBuilder cmd = new StringBuilder("-break-insert -f ");
+            StringBuilder cmd = new StringBuilder("-break-insert ");
             if (condition != null)
             {
                 cmd.Append("-c \"");
@@ -435,6 +435,7 @@ namespace MICore
                 cmd.Append(" ");
             }
 
+            cmd.Append("-f ");
             cmd.Append(filename);
             cmd.Append(":");
             cmd.Append(line.ToString());


gregg-miskelly:
@jacdavis @pieandcakes can you review?

chuckries:
I believe @pieandcakes fixed this for lldb-mi. Their implementation was incorrect as '-f' is just a simple flag and should not be required to be followed by a location.

jacdavis:
If there is a difference between lldb and gdb here, we should probably split this into their respective command factories. @pieandcakes, do you recall a change you made here?

pieandcakes:
In our version of lldb-mi, we have fixed it so that the -f flag does not require a parameter. This is in keeping with the gdb/mi spec as it should be simple flag. I'm in the process of pushing this change back to lldb.llvm.org. 

If we want a "fix" for this temporarily to support other lldb-mi versions, we'd want to change this specifically for only the lldb instance within src/MICore/CommandFactories for lldb

Dmitri-Botcharnikov:
Ok, so this PR is wrong and I close it.
@pieandcakes could you please share the patch for lldb-mi?

pieandcakes:
@Dmitri-Botcharnikov sorry, I misspoke. We have the temporary fix in mainline MIEngine:

https://github.com/Microsoft/MIEngine/commit/b50bc9a05ceeb0aa738060eab1dc4a418485de00

you need to specify an MIMode of lldb in order for it to work. 

I have a request in to lldb to fix it on lldb's side and will share the patch when it gets approved.

Dmitri-Botcharnikov:
@pieandcakes: thank you, it works.

Dmitri-Botcharnikov:
@pieandcakes BTW which is the best way to propose patches to lldb? We have couple of them regarding lldb-mi implementation among others.

pieandcakes:
@Dmitri-Botcharnikov You can either file a bug and propose the patch inside of it or sign up to contribute. I found the information here to be helpful: http://llvm.org/docs/DeveloperPolicy.html

PR #431, opened 2016-08-29
update file version for VS vNext

faxue-msft:
@gregg-miskelly @chuckries

### DIFF
--- a/build/version.settings.targets
+++ b/build/version.settings.targets
@@ -2,7 +2,7 @@
   <PropertyGroup>    
     <!--SxS: These three properties should be changed at the start of a new version, VersionZeroYear should be the year
     before the start of the project. When updating the version, also update MIEngine\metadata.json.-->
-    <MajorVersion>14</MajorVersion>
+    <MajorVersion>15</MajorVersion>
     <MinorVersion>0</MinorVersion>
     <VersionZeroYear>2013</VersionZeroYear>
 
--- a/metadata.json
+++ b/metadata.json
@@ -1,5 +1,5 @@
 {
   "correlationId": "595D6107-E618-4BD9-85A0-C323EFC400D1",
   "owner":"miengine",
-  "version": "14.0"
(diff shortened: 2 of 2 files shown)


build/version.settings.targets
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>    
     <!--SxS: These three properties should be changed at the start of a new version, VersionZeroYear should be the year
     before the start of the project. When updating the version, also update MIEngine\metadata.json.-->
-    <MajorVersion>14</MajorVersion>
+    <MajorVersion>15</MajorVersion>
     <MinorVersion>0</MinorVersion>
     <VersionZeroYear>2013</VersionZeroYear>
```
gregg-miskelly: I would change 'VersionZeroYear' to 2015 when we do this.

gregg-miskelly:
I would hold off on this for another month. We are planning to ship one more dev14 version, and it might be confusing that the version number in Dev15.

gregg-miskelly:
@caslan what do you think?

build/version.settings.targets
faxue-msft: I thought it's the start time for miengine project, isn't it?

faxue-msft:
@gregg-miskelly
Just to be clear, do you mean shipping MIEngine as dev14 update 4 for coreclr debugging?

caslan:
As an OOB. We are planning to release by the end of September for SSH support for VC++ Linux extension.

build/version.settings.targets
gregg-miskelly: It says what real year should have a '1' in the year part of the version number (MIEngine uses build numbers of YMMDD). But the max number of years which this scheme can accept is 6, so when we rev the major year, we also want to update the VersionZeroYear.

faxue-msft:
Per our discussion, we'll stick with the version number 14

PR #434, opened 2016-09-06
When debugging through SSH as root, stdin is printed in stdout stream.

rajkumar42:
Fix: ignoring the input string, if we see it when processing the stdout.

### DIFF
--- a/src/SSHDebugPS/AD7UnixAsyncCommand.cs
+++ b/src/SSHDebugPS/AD7UnixAsyncCommand.cs
@@ -23,6 +23,8 @@ namespace Microsoft.SSHDebugPS
         private bool _beginReceived;
         private bool _isClosed;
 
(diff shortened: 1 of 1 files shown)


rajkumar42:

wesrupert:
LGTM! :shipit:

src/SSHDebugPS/AD7UnixAsyncCommand.cs
```diff
@@ -23,6 +23,8 @@ internal class AD7UnixAsyncCommand : IDebugUnixShellAsyncCommand
         private bool _beginReceived;
         private bool _isClosed;
 
+        private string startCommand;
```
gregg-miskelly: Use '_' prefix

gregg-miskelly:
Aside from that nit, look good to me.

PR #437, opened 2016-09-07
Computing RID based on /etc/os-release instead of dotnet --info

rajkumar42:
Tested it with docker containers!

### DIFF
--- a/scripts/GetClrDbg.sh
+++ b/scripts/GetClrDbg.sh
@@ -21,7 +21,6 @@ __RemoveExistingOnUpgrade=false
 # Internal, fully specified version of the ClrDbg. Computed when the meta version is used.
 __ClrDbgVersion=
 
-
 # Gets the script directory
 get_script_directory()
 {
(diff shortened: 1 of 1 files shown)


rajkumar42:
@gregg-miskelly @caslan @chuckries @wesrupert

chuckries:
:+1: please squash.

scripts/GetClrDbg.sh
```diff
@@ -49,32 +48,63 @@ print_help()
...
+            __RuntimeID=rhel.7-x64
+            ;;
+        debian)
+            __RuntimeID=debian.8-x64
+            ;;
+        ol)
+            __RuntimeID=centos.7-x64
+            ;;
+        elementaryOS)
+            if [[ "$PlatformVersionID" == 0.3* ]]; then
```
gregg-miskelly: >. [](start = 43, length = 1)

Does the '.' need to be escaped?

gregg-miskelly: Looks like it doesn't need to be.

---
In reply to: [77927758](https://github.com/Microsoft/MIEngine/pull/437#discussion_r77927758) [](ancestors = 77927758)

gregg-miskelly:
:shipit:

PR #441, opened 2016-09-13
Set environment variables through the appropriate command

edumunoz:
Previously, in order to set the desired environment variables intended for the
debuggee, MIEngine was setting them for the debugger process, given that
subprocesses (debuggee) inherit the environment. This is not the case by default
with LLDB, and is cleaner to only set the environment variables in the debuggee
directly. This is done by using GDB/LLDB commands that only affect the debuggee.
CLRDBG doesn't support these commands, so for now the old behavior is preserved.

@pieandcakes @jacdavis @gregg-miskelly


gregg-miskelly:
:shipit:

PR #443, opened 2016-09-15
Remove workaround for updating old connection info

caslan:
Removed the workaround for updating connection info as the credentials dialog now does this automatically. Also, fixed an issue with comparing the enum dialogresult.


caslan:
@gregg-miskelly @rajkumar42

rajkumar42:
:shipit:

PR #444, opened 2016-09-15
Allow lldb-mi to assign expressions to variables

edumunoz:
lldb-mi doesn't evaluate expressions passed in -var-assign <expression>
This workaround evaluates the expression as a different step. I will create a bug in the lldb database to see if this is something they're interested in fixing.

### DIFF
--- a/src/MICore/CommandFactories/lldb.cs
+++ b/src/MICore/CommandFactories/lldb.cs
@@ -160,5 +160,26 @@ namespace MICore
         {
             throw new NotImplementedException("lldb catch command");
         }
+
+        /// <summary>
+        /// Assigns the value of an expression to a variable.
+        /// Since LLDB only accepts assigning values to variables, the expression may need to be evaluated.
+        /// However, since the result of evaluating an expression in LLDB can return some extra information:
+        /// e.g., 'a' --> 97 'a'. We don't want to assign the value "97 'a'". Instead, we first try
+        /// assigning what the user passed, only falling back to evaluation if the first assignment fails.
+        /// </summary>
+        public async override Task<string> VarAssign(string variableName, string expression, int threadId, uint frameLevel)
+        {
+            try
+            {
+                return await base.VarAssign(variableName, expression, threadId, frameLevel);
+            }
+            catch (UnexpectedMIResultException)
+            {
+                Results results = await VarCreate(expression, threadId, frameLevel, 0, ResultClass.done);
+                string value = results.FindString("value");
+                return await base.VarAssign(variableName, value, threadId, frameLevel);
+            }
+        }
     }
 }
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -705,8 +705,11 @@ namespace Microsoft.MIDebugEngine
 
             _engine.DebuggedProcess.WorkerThread.RunOperation(async () =>
             {
+                int threadId = Client.GetDebuggedThread().Id;
+                uint frameLevel = _ctx.Level;
+
                 _engine.DebuggedProcess.FlushBreakStateData();
-                Value = await _engine.DebuggedProcess.MICommandFactory.VarAssign(_internalName, expression);
+                Value = await _engine.DebuggedProcess.MICommandFactory.VarAssign(_internalName, expression, threadId, frameLevel);
             });
         }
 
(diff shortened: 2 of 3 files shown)


edumunoz:
@gregg-miskelly @jacdavis @pieandcakes

src/MICore/CommandFactories/lldb.cs
```diff
@@ -160,5 +160,12 @@ public override Task Catch(string name, bool onlyOnce = false, ResultClass resul
         {
             throw new NotImplementedException("lldb catch command");
         }
+
+        public async override Task<string> VarAssign(string variableName, string expression, int threadId, uint frameLevel)
+        {
+            Results results = await VarCreate(expression, threadId, frameLevel, 0, ResultClass.done);
```
edumunoz: The argument being passed as `0` is the `enum_EVALFLAGS` passed to evaluate.

I see it's not actually being used. Should I instead construct some other flags more descriptive of the type of evaluations that we want to allow here (even if it's not enforced), or rather take this as a paramter for `VarAssign` and pass some flags from the client of this method (in `Variables.cs`).

gregg-miskelly: The EVALFLAGS get used in the clrdbg code path. Since this is the lldb command factory, you know that the eval flags will never be used, so it seems okay to just pass zero.

jacdavis: I personally am fine with what you're doing here.

src/MICore/CommandFactories/lldb.cs
```diff
@@ -160,5 +160,12 @@ public override Task Catch(string name, bool onlyOnce = false, ResultClass resul
...
+            string value = results.FindString("value");
```
jacdavis: what happens if the evaluation fails?

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -702,11 +702,13 @@ public bool IsReadOnly
         public void Assign(string expression)
         {
             this.VerifyNotDisposed();
+            int threadId = Client.GetDebuggedThread().Id;
+            uint frameLevel = _ctx.Level;
```
jacdavis: be careful here. Since this is all async, I think it is possible for the debuggee to get continued before this point. In which case, the frame could be gone or completely blow up.

src/MICore/CommandFactories/lldb.cs
gregg-miskelly: Does `value` have any extra stuff on it that LLDB wouldn't want on an assignment? Ex: if you have a `char*` variable and assign it `(char*)0x12345` (where 12345 is the address of a valid string) will this wind up adding anything extra? Might be interesting to indicate what types we have tested.

edumunoz: It will return an error as expected since the expected result class is `done`.

edumunoz: That's a good point.

For primitive types in C++, this doesn't work for `bool` or `char` because lldb doesn't treat `true`, `false`, `'a'` as valid values. The hope here was that those values would evaluate to the corresponding integer value (e.g., `'a'` --> `97`). I'll be creating a bug in the lldb database to see how we can address this.

For `char*`, it's being displayed as a tree structure in VSCode, so it won't allow to set the value.

Is there a way you recommend we can get the actual integer value for `char` or `bool` directly from lldb?

In order to avoid regressions, do you think `lldb::VarAssign` should first try the vanilla `VarAssign` (not trying to evaluate the expression), and if that fails, then try to evaluate the expression and set the result with another `VarAssign`?

src/MIDebugEngine/Engine.Impl/Variables.cs
edumunoz: Should these two lines be moved within the lambda passed to `RunOperation` to avoid race conditions that change the `threadId` or `frameLevel`?

I think I see why a `continue` could also cause a race condition that messes up this operation. Not sure if a continue would automatically dispose these `VariableInformation` object (I couldn't find where).

However, I expect the same problem to occur for `VariableInformation::Eval`, which for some reason doesn't use the `_engine.DebuggedProcess.WorkerThread.RunOperation pattern`. I think the worst that can happen is that `threadId` and `frameLevel` are not valid anymore and the resulting expression evaluation will fail.

Do you agree?

jacdavis: I agree that failure is fine, as long as it fails cleanly and doesn't result in a crash or anything like that.

PR #445, opened 2016-09-15
Fix for displaying char* and char** values

pieandcakes:
changing to var-list-children to use --all-values. lldb treats any object
with children as not simple, and therefore will not show values when
passing --simple-values as the parameter.


pieandcakes:
@edumunoz @jacdavis @gregg-miskelly

PR #446, opened 2016-09-15
Implementing ExecuteSyncCommand.

rajkumar42:
Refactoring kill and kill child processes to UnixUtilities.

### DIFF
--- a/src/MICore/Transports/ClientServerTransport.cs
+++ b/src/MICore/Transports/ClientServerTransport.cs
@@ -69,12 +69,12 @@ namespace MICore
 
         public int ExecuteSyncCommand(string commandDescription, string commandText, int timeout, out string output, out string error)
         {
-            throw new NotImplementedException();
+            return _clientTransport.ExecuteSyncCommand(commandDescription, commandText, timeout, out output, out error);
         }
 
         public bool CanExecuteCommand()
         {
-            return false;
+            return _clientTransport.CanExecuteCommand();
         }
     }
 }
--- a/src/MICore/Transports/ClientServerTransport.cs
+++ b/src/MICore/Transports/ClientServerTransport.cs
@@ -66,5 +66,15 @@ namespace MICore
         {
             _clientTransport.Send(cmd);
         }
+
+        public int ExecuteSyncCommand(string commandDescription, string commandText, int timeout, out string output, out string error)
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool CanExecuteCommand()
+        {
+            return false;
+        }
     }
 }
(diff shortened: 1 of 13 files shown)


rajkumar42:
@gregg-miskelly @caslan @chuckries @wesrupert

gregg-miskelly:
>                 if (PlatformUtilities.IsLinux() || PlatformUtilities.IsOSX())

This should probably change to 'CanExecuteCommand'

---
Refers to: src/MICore/Debugger.cs:622 in ab47616. [](commit_id = ab47616c64d1a437880f78008e033c488a3411ee, deletion_comment = False)

gregg-miskelly:
>             {

And eliminate this

---
Refers to: src/MICore/Debugger.cs:610 in ab47616. [](commit_id = ab47616c64d1a437880f78008e033c488a3411ee, deletion_comment = False)

src/MICore/Transports/ClientServerTransport.cs
```diff
@@ -66,5 +66,15 @@ public void Send(string cmd)
         {
             _clientTransport.Send(cmd);
         }
+
+        public int ExecuteSyncCommand(string commandDescription, string commandText, int timeout, out string output, out string error)
+        {
+            throw new NotImplementedException();
```
gregg-miskelly: >throw new NotImplementedException(); [](start = 12, length = 36)

these should delegate to _clientTransport

src/MICore/Transports/PipeTransport.cs
```diff
@@ -361,5 +300,66 @@ private void OnProcessExit(object sender, EventArgs e)
...
+                proc.StartInfo.RedirectStandardInput = false;
+                proc.StartInfo.RedirectStandardOutput = true;
+                proc.StartInfo.RedirectStandardError = true;
+                proc.StartInfo.UseShellExecute = false;
+                proc.StartInfo.CreateNoWindow = true;
+                proc.Start();
+                proc.WaitForExit(timeout);
+                exitCode = proc.ExitCode;
+            }
+            finally
```
gregg-miskelly: >finally [](start = 12, length = 7)

It doesn't seem like this finally block should cover all this. Like if say the CreateNoWindow property setter threw for some crazy reason, we don't want to call this.

Is there any of the methods that you want to call this even if it throws?

gregg-miskelly:
:shipit:

src/MICore/Transports/PipeTransport.cs
rajkumar42: Did some tests, streams are empty when the process did not launch. So there is no reason to read it on finally. 

---
In reply to: [79071150](https://github.com/Microsoft/MIEngine/pull/446#discussion_r79071150) [](ancestors = 79071150)

rajkumar42:
Tested C# and gdb scenarios with VSCode on Linux.

PR #447, opened 2016-09-15
Fix a crashing issue when stop debugging

faxue-msft:
https://devdiv.visualstudio.com/DevDiv/VS%20Diag%20IntelliTrace/_workItems/index?_a=edit&id=264224&triage=true

The issue is MI try to read stack or thread information after debuggee has
been killed, and eventually cause unhandled exception and crash VS. Fix by
checking terminating status before querying debuggee information.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -365,7 +365,7 @@ namespace Microsoft.MIDebugEngine
                 }
                 catch (Exception e) when (ExceptionHelper.BeforeCatch(e, Logger, reportOnlyCorrupting: true))
                 {
-                    if (this.ProcessState == MICore.ProcessState.Exited)
+                    if (_terminating)
                     {
                         return; // ignore exceptions after the process has exited
                     }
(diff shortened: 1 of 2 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -365,7 +365,7 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
                 }
                 catch (Exception e) when (ExceptionHelper.BeforeCatch(e, Logger, reportOnlyCorrupting: true))
                 {
-                    if (this.ProcessState == MICore.ProcessState.Exited)
+                    if (_terminating)
```
gregg-miskelly: You might want `_terminating || ...` but you certainly don't want to **only** do terminate here as the same could happen if the debuggee ran to completion, or detach cases. It might want sense to instead of using '_terminating' use a new '_exitting'.

faxue-msft: It safer that way. 
It worth noticing that for all the scenarios I tested, this code can only be reached when stop debugging and debugger send "kill" command, GDB will have output of "signal-received" and debugger in turn tries to handle that as a break event. When debuggee run to exit we don't execute this code path.

PR #449, opened 2016-09-20
Fixing dotnet restore error issues

rajkumar42:
printing dotnet_restore.log to console
catching exception when the command times out

### DIFF
--- a/src/MICore/MICoreResources.resx
+++ b/src/MICore/MICoreResources.resx
@@ -131,7 +131,9 @@
   <data name="Error_DebuggerInitializeFailed_StdErr" xml:space="preserve">
     <value>Unable to establish a connection to {0}. The following message was written to stderr:
 
-{1}</value>
+{1}
+
+See output window for details.</value>
   </data>
   <data name="Error_DebugServerInitializationFailed" xml:space="preserve">
     <value>Debug server process failed to initialize.</value>
--- a/src/MICore/MICoreResources.resx
+++ b/src/MICore/MICoreResources.resx
@@ -133,7 +133,7 @@
 
 {1}
 
-See output window for details.</value>
+See Output Window for details.</value>
   </data>
   <data name="Error_DebugServerInitializationFailed" xml:space="preserve">
     <value>Debug server process failed to initialize.</value>
--- a/src/SSHDebugPS/AD7UnixAsyncCommand.cs
+++ b/src/SSHDebugPS/AD7UnixAsyncCommand.cs
@@ -151,16 +151,16 @@ namespace Microsoft.SSHDebugPS
                         _streamingShell.Close();
                     }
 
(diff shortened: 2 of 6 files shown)


rajkumar42:
@gregg-miskelly @caslan @wesrupert @chuckries

src/MICore/MICoreResources.resx
```diff
@@ -131,7 +131,9 @@
   <data name="Error_DebuggerInitializeFailed_StdErr" xml:space="preserve">
     <value>Unable to establish a connection to {0}. The following message was written to stderr:
 
-{1}</value>
+{1}
+
+See output window for details.</value>
```
caslan: I think this needs to be "Output Window"

src/SSHDebugPS/AD7UnixAsyncCommand.cs
```diff
@@ -143,9 +143,25 @@ private void Close()
...
 
-            _streamingShell.Close();
+                try
+                {
+                    if (_streamingShell != null && _streamingShell.IsOpen)
+                    {
+                        _streamingShell.Close();
+                    }
+
+                    if (_streamingShell != null)
```
wesrupert:
```csharp
_streamingShell?.Dispose();
_streamingShell = null;
```

src/SSHDebugPS/AD7UnixAsyncCommand.cs
```diff
@@ -143,9 +143,25 @@ private void Close()
                     return;
 
                 _isClosed = true;
-            }
 
-            _streamingShell.Close();
+                try
+                {
+                    if (_streamingShell != null && _streamingShell.IsOpen)
```
wesrupert: I think this line can be simplified to `if (_streamingShell?.IsOpen)`.

src/SSHDebugPS/AD7UnixAsyncCommand.cs
```diff
@@ -143,9 +143,25 @@ private void Close()
...
+
+                    if (_streamingShell != null)
+                    {
+                        _streamingShell.Dispose();
+                        _streamingShell = null;
+                    }
+                }
+                catch (ThreadInterruptedException)
+                {
+                    // This can happen if the command failed on a timeout. Say dotnet restore failed because of insufficient permissions etc.
```
wesrupert: This comment can be improved, it doesn't currently indicate that it's okay that we don't try to dispose again.

gregg-miskelly:
:shipit:

src/SSHDebugPS/AD7UnixAsyncCommand.cs
rajkumar42: Primitives return values gets converted into nullable when using ?.  
Keeping it as is.

PR #453, opened 2016-09-23
Fixing empty PipeCwd issue.

rajkumar42:
PipeCwd will be null if the provided PipeCwd is not absolute.

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -80,6 +80,7 @@ namespace MICore
             this.PipePath = pipePath;
             this.PipeArguments = pipeArguments;
             this.PipeCommandArguments = pipeCommandArguments;
+            this.PipeCwd = pipeCwd;
 
             this.PipeEnvironment = (pipeEnvironment != null) ? pipeEnvironment.Select(e => new EnvironmentEntry(e)).ToArray() : new EnvironmentEntry[] { };
         }
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -81,15 +81,6 @@ namespace MICore
             this.PipeArguments = pipeArguments;
             this.PipeCommandArguments = pipeCommandArguments;
 
-            if (!String.IsNullOrWhiteSpace(pipeCwd))
-            {
-                this.PipeCwd = pipeCwd;
-            }
-            else
-            {
-                this.PipeCwd = Path.GetDirectoryName(pipePath);
-            }
-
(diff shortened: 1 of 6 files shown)


rajkumar42:
@gregg-miskelly

src/MICore/LaunchOptions.cs
```diff
@@ -64,6 +64,14 @@ public enum LaunchCompleteCommand
     /// </summary>
     public sealed class PipeLaunchOptions : LaunchOptions
     {
+        /// <summary>
+        /// Creates an instance of PipeLaunchOptions
+        /// </summary>
+        /// <param name="pipePath">Path of the pipe program</param>
```
gregg-miskelly: Might want to add something here, and to the xsd file (line 292) something like --

This should either be a full path to an executable or the name of a program on the PATH.

src/MICore/Transports/PipeTransport.cs
```diff
@@ -93,9 +93,11 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
         {
             PipeLaunchOptions pipeOptions = (PipeLaunchOptions)options;
 
+            string workingDirectory = pipeOptions.PipeCwd;
             if (!LocalLaunchOptions.CheckDirectoryPath(pipeOptions.PipeCwd))
```
gregg-miskelly: The one problem with this is that if the user provides a 'PipeCmd' attribute which is bogus, we will no longer throw. We might want to instead remove LaunchOptions.cs(82) and change this line to be:

```csharp
    if (!string.IsNullOrEmpty(workingDirectory))
    {
       if (!CheckDirectoryPath()..) throw;
    }
   else
   {
      workingDirectory = Path.GetDirectoryName(pipePath);
      if (!CheckDirectoryPath()..)
      {
        your new code
      }
```

gregg-miskelly:
>             }

I think you want to remove this too

---
Refers to: src/MICore/LaunchOptions.cs:91 in 40c92ab. [](commit_id = 40c92ab6a3a0c6b652d94b90cc3b9d47086c5be4, deletion_comment = False)

gregg-miskelly:
:shipit:

rajkumar42:
>             }

yep! thanks

---
In reply to: [249264922](https://github.com/Microsoft/MIEngine/pull/453#issuecomment-249264922) [](ancestors = 249264922)

---
Refers to: src/MICore/LaunchOptions.cs:91 in 40c92ab. [](commit_id = 40c92ab6a3a0c6b652d94b90cc3b9d47086c5be4, deletion_comment = False)

PR #456, opened 2016-09-25
Removing file exists check for PipePath.

rajkumar42:
Previously added checks for existence of PipeProgram expects the absolute path. That breaks docker tools scenario where the absolute path is not provided.


rajkumar42:
@gregg-miskelly @caslan

PR #463, opened 2016-10-04
Do not check NDK_HOME for Java debugging

faxue-msft:
Skip the check for NDK_HOME when it's Java debugging.


src/AndroidDebugLauncher/InstallPaths.cs
```diff
@@ -11,6 +11,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using System.Globalization;
+using MICore;
 
 namespace AndroidDebugLauncher
```
gregg-miskelly: The comments on GDBPath, and GDBServerPath is wrong with your change (they are now optional). Please fix.

PR #465, opened 2016-10-05
Add extention point for debugger launchers

paulmaybee:
Looks up Xml launch options name in the registry in order to discover the clsid for the correct launcher to handle the options. 
Does not touch current IOS and Android launcher code paths. They can be updated to use the same extension mechanism but I don't have the time before next release to perform the appropriate testing on those scenarios.

### DIFF
--- a/src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
+++ b/src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
@@ -187,12 +187,22 @@ namespace Microsoft.DebugEngineHost
         {
             throw new NotImplementedException();
         }
-    }
 
-    /// <summary>
-    /// The host logger returned from HostConfigurationStore.GetLogger.
-    /// </summary>
-    public sealed class HostLogger
+        /// <summary>
+        /// Lookup the clsid of the launcher that understands these options
+        /// </summary>
+        /// <param name="launchOptionsName">launch options type name</param>
+        /// <returns></returns>
+        public Guid GetCustomLauncherClsid(string launchOptionsName)
+        {
+            throw new NotImplementedException();
+    }
+}
+
+/// <summary>
+/// The host logger returned from HostConfigurationStore.GetLogger.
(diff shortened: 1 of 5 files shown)


src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
```diff
@@ -187,12 +187,22 @@ public T GetDebuggerConfigurationSetting<T>(string settingName, T defaultValue)
         {
             throw new NotImplementedException();
         }
+
+        /// <summary>
+        /// Lookup the clsid of the launcher that understands these options
+        /// </summary>
+        /// <param name="launchOptionsName">launch options type name</param>
+        /// <returns></returns>
+        public Guid GetCustomLauncherClsid(string launchOptionsName)
```
gregg-miskelly: Since we need to define a new interface anyway, I would suggest we add 'GetCustomLauncher' (no Clsid) and return the object directly. Since Clsid aren't exsactly a natural concept on non-windows.

src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
```diff
@@ -187,12 +187,22 @@ public T GetDebuggerConfigurationSetting<T>(string settingName, T defaultValue)
...
+        {
+            throw new NotImplementedException();
     }
```
gregg-miskelly: indenting is weird here

paulmaybee:
@gregg-miskelly @faxue-msft

src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
paulmaybee: fixed

src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
paulmaybee: fixed

PR #467, opened 2016-10-06
Allow natvis files to load with:

jacdavis:
0) solution file
1) full path
2) local launch options exe relative path
3) global visualizer relative path.

Note that the priority is that order.

@gregg-miskelly @edumunoz @paulmaybee @chuckries

### DIFF
--- a/src/MIDebugEngine/Natvis.Impl/Natvis.cs
+++ b/src/MIDebugEngine/Natvis.Impl/Natvis.cs
@@ -201,22 +201,26 @@ namespace Microsoft.MIDebugEngine.Natvis
                     globalNatVisPath = Path.Combine(globalVisualizersDirectory, fileName);
                 }
 
-                // For local launch, try and load natvis next to the target exe if it exists.
-                // If the file doesn't exist, and also doesn't exist in the global folder fail.
+                // For local launch, try and load natvis next to the target exe if it exists and if 
+                // the exe is rooted. If the file doesn't exist, and also doesn't exist in the global folder fail.
                 if (_process.LaunchOptions is LocalLaunchOptions)
                 {
-                    string localNatvisPath = Path.Combine(Path.GetDirectoryName((_process.LaunchOptions as LocalLaunchOptions).ExePath), fileName);
+                    string exePath = (_process.LaunchOptions as LocalLaunchOptions).ExePath;
+                    if (Path.IsPathRooted(exePath))
+                    {
+                        string localNatvisPath = Path.Combine(Path.GetDirectoryName(exePath), fileName);
 
-                    if (File.Exists(localNatvisPath))
-                    {
-                        LoadFile(localNatvisPath);
-                        return;
-                    }
-                    else if (!File.Exists(globalNatVisPath))
-                    {
-                        // Neither local or global path exists, report an error.
-                        _process.WriteOutput(String.Format(CultureInfo.CurrentCulture, ResourceStrings.FileNotFound, localNatvisPath));
-                        return;
+                        if (File.Exists(localNatvisPath))
+                        {
+                            LoadFile(localNatvisPath);
+                            return;
+                        }
+                        else if (globalNatVisPath == null || !File.Exists(globalNatVisPath))
+                        {
+                            // Neither local or global path exists, report an error.
+                            _process.WriteOutput(String.Format(CultureInfo.CurrentCulture, ResourceStrings.FileNotFound, localNatvisPath));
+                            return;
+                        }
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -191,16 +191,44 @@ public void Initialize(string fileName)
...
+                if (_process.LaunchOptions is LocalLaunchOptions)
+                {
+                    string natvisPath = Path.Combine(Path.GetDirectoryName((_process.LaunchOptions as LocalLaunchOptions).ExePath), fileName);
+
+                    if (File.Exists(natvisPath))
+                    {
+                        LoadFile(natvisPath);
+                        return;
+                    }
+                    else if (!File.Exists(globalNatVisPath))
```
edumunoz: I don't have the full context on how natvis files are loaded, but it seems a bit strange that:

```
If the launch options are local {
- We try the local one if it exists and return
- Check if the global one exists and show error if it doesn't
}

- Load the global one if it exists

```
It would be more clear:

```
fileToLoad = globalNatVisPath
If the launch options are local {
- Set fileToLoad to natvisPath if natvisPath exists (natvisPath probably should be renamed to localNatVisPath)
}

- Load fileToLoad if it exists, show error if it doesn't

```

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -191,16 +191,44 @@ public void Initialize(string fileName)
             {
                 // failed to find the VS Service
             }
-            string natvisPath = null;
-            string directory = _process.Engine.GetMetric("GlobalVisualizersDirectory") as string;
-            if (!string.IsNullOrEmpty(directory) && !string.IsNullOrEmpty(fileName))
+
+            if (string.IsNullOrEmpty(Path.GetDirectoryName(fileName)))
```
edumunoz: Nit: I would revert the order of this if and else statements (so check whether it's not null or empty). I think it makes following the loading logic easier.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -191,16 +191,44 @@ public void Initialize(string fileName)
...
+                if (!string.IsNullOrEmpty(globalVisualizersDirectory) && !string.IsNullOrEmpty(fileName))
+                {
+                    globalNatVisPath = Path.Combine(globalVisualizersDirectory, fileName);
+                }
 
-            if (!string.IsNullOrEmpty(natvisPath))
+                // For local launch, try and load natvis next to the target exe if it exists.
+                if (_process.LaunchOptions is LocalLaunchOptions)
+                {
+                    string natvisPath = Path.Combine(Path.GetDirectoryName((_process.LaunchOptions as LocalLaunchOptions).ExePath), fileName);
```
gregg-miskelly: I think you also need o check that 'ExePath' is rooted.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
gregg-miskelly: Don't we also need a check that `globalNatVisPath != null`?

src/MIDebugEngine/Natvis.Impl/Natvis.cs
gregg-miskelly: This will return true if `fileName` is something like `folder\name.ext`. Seems like this should be Path.IsRooted instead.

jacdavis: good point

src/MIDebugEngine/Natvis.Impl/Natvis.cs
jacdavis: Unfortunately, the scenario has been a bit bastardized by the existing launch options. In the vs case, we expect the natvis file to come from solution or from the global path. For vscode, we expect it to come from either a fullpath, a workspace relative path (if possible in the local case) or the global path. I'd love to revisit this, but the cost seems too high now (changes across too many products). 

I agree the code isn't all that readable, but it seems to be the best I can do to capture the behavior I described above. Perhaps I can make it a bit more readable.

jacdavis: The behavior of File.Exists isn't clear to me if this is necessary (given that vb treats null as an empty string), but I'll add that check just in case.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
jacdavis: Unrooted exe shouldn't be possible, but this may be called before that is validated. I added a check just to be safe.

PR #475, opened 2016-10-18
Fix crash in SSH

caslan:
This PR fixes a crash in SSH debugging when tearing down the SSH session. The root issue is StreamingShell.Dispose method tries to interrupt the thread responsible for processing output events and then waits for it to finish. The following sequence of events cause an exception to be raised and go unhandled. 

1 - PollThread calls transport.close due to an attach error
2 - PollThread calls StreamingShell.Dispose
3 - StreamingShell.Dispose calls Interrupt on the output processing thread
4 - Output processing thread either runs into lock statement in OnOutputReceived event or in Close() methods
5 - This results in ThreadInterruptedException going unhandled

Moved the unnecessary locks around so that output processing thread doesn't need to lock. Also made sure Close() is called once.
@gregg-miskelly @rajkumar42


rajkumar42:
:shipit:

PR #479, opened 2016-10-22
Invalid condition in RetryBreak timer guard clause

dosvidos:
As it is stated in comment - SIGINT delivery is reliable, so no need to re-send it, but Cygwin can't currently async-break reliably, so use retries there. However, the actual condition "!IsCygwin" in the guard clause seems to be wrong and should be inverted.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -223,7 +223,7 @@ namespace MICore
                         // When using signals to stop the proces, do not kick off another break attempt. The debug break injection and
                         // signal based models are reliable so no retries are needed. Cygwin can't currently async-break reliably, so
                         // use retries there.
-                        if (!IsLocalGdb() && !this.IsCygwin)
+                        if (!IsLocalGdb() && this.IsCygwin)
                         {
                             _breakTimer = new Timer(RetryBreak, null, BREAK_DELTA, BREAK_DELTA);
                         }


jacdavis:
Thanks for pointing this out. However, I'm not convinced the condition is correct in either case. I believe the reason this code was like this was that signal based stops won't need retries. However, on cygwin we cannot stop the process either way. It doesn't support the break command and doesn't respond to signals properly. I believe mingw suffers from the same problem. Currently, we have to rely on the user to ctrl-c the process to break or to add breakpoints in run mode. I did have a very complex workaround for cygwin which we decided not to checkin. However, I was never able to work around the issue in mingw. 

I believe what we want to capture is anytime signals are not being used. However, as mentioned above, cygwin and mingw are futile. So, I believe the correct condition would be to !iscygwin && !ismingw although I don't recall if we have state for !ismingw.

chuckries:
@dotnet-bot build this please

chuckries:
@dotnet-bot test this please

haneefdm:
> `!iscygwin && !ismingw`

@jacdavis I think this is the right condition, but not willing take a bet. Neither of them respond properly (or at all). I tried so many ways and got nothing using multiple Win32 APIs, using /usr/bin/kill on cygwin, using cygwin PIDs rather than windows PIDs, using it on the debugger rather than the debuggee and nothing worked. _Sending a signal more than once is futile and just not right_. Not the same as hitting Ctrl-C in the console twice for command-line gdb which has a special meaning to gdb; uses the first one to try to interrupt and the second one asking the user if they want to kill the inferior because the interrupt failed

Only thing reliable I found is to hit Ctrl-C in the console allocated to the program -- as documented.

I also did not get why the whole retry was needed 'cause it either works or it doesn't -- using a signal or a `-exec-interrupt`. A timeout was required for sure. But I don't have a test case where retries can help. I am sure there was a reason at one point in time.

And, be careful as both cygwin and mingw respond properly if gdb is not controlling the target directly and is configured to use a gdbserver for controlling the target execution.  And there are multiple versions of mingw gdbs for embedded systems -- so `!cygwin != mingw` strictly speaking

I have an ongoing related PR #855 which is why I am commenting here.

WardenGnaw:
Closing old PRs

PR #481, opened 2016-10-26
check for empty disassembly instructions

faxue-msft:
For iOS debugging, sometimes lldb may return empty result when the starting address is invalid. This might also happen for gdb debugging.


faxue-msft:
@paulmaybee @gregg-miskelly

paulmaybee:
👍

PR #482, opened 2016-10-27
Two natvis fixes in order to support gnu stl:

paulmaybee:
1. allow decimal constants to have trailing 'u' and 'l'
2. The TreeItems ValueNode element only accepted a simple field name, now allows any natvis expression value


paulmaybee:
@jacdavis @gregg-miskelly @faxue-msft

jacdavis:
:shipit:

PR #483, opened 2016-10-27
Correctly generate full names for C# arrays

gregg-miskelly:
This is a partial fix for full names with clrdbg. A full fix with come later with the new architecture so that that clrdbg will directly report full names. With this checkin, the MIEngine will generate a full name of '(myArray)[0]' instead of '(myArray).[0]'.


gregg-miskelly:
@paulmaybee can you review?

paulmaybee:
From this fix I assume that the clrdbg returns "[index]" rather then just a numeric index for an array element. If so lgtm.
👍

gregg-miskelly:
Correct

PR #485, opened 2016-10-28
Add Initialize to HostTelemetry

pieandcakes:
This is to preload IVsTelemetryService. We ran into an issue where, when
launching the debugger in VS that if we try and send telemetry, it
deadlocks since the main thread is waiting on the debugger launch which is
in turn trying to send telemetry and is stuck waiting to get on the main
thread to get the IVsTelemetryService. 

@faxue-msft @gregg-miskelly @chuckries @caslan @jacdavis

### DIFF
--- a/src/DebugEngineHost/Host.cs
+++ b/src/DebugEngineHost/Host.cs
@@ -6,6 +6,8 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.Internal.VisualStudio.Shell;
+using Microsoft.Internal.VisualStudio.Shell.Interop;
 
 namespace Microsoft.DebugEngineHost
 {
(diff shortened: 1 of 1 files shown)


faxue-msft:
:shipit:

src/DebugEngineHost/HostTelemetry.cs
```diff
@@ -50,6 +50,18 @@ public static void SendEvent(string eventName, params KeyValuePair<string, objec
...
         /// <summary>
+        /// Initialize Host Telemetry
+        /// </summary>
+        [Conditional("LAB")]
+        public static void Initialize()
+        {
+            // Force the IVsTelemetryService to complete its lazy loading. There is a hang caused by trying to
+            // send telemetry while Visual Studio is launching the debugger if the telemetry helper also needs 
+            // to load the telemetry service on the main thread.
+            var telemetryService = TelemetryHelper.TelemetryService;
```
gregg-miskelly: You are going to need try/catches around this and probably to move this to another class. Otherwise you will break glass scenarios. See other uses of Telemetry for an example.

src/MIDebugEngine/Engine.Impl/EngineTelemetry.cs
```diff
@@ -13,6 +13,7 @@ namespace Microsoft.MIDebugEngine
 {
     internal class EngineTelemetry
     {
+        private const string Event_LoadHostTelemetry = @"VS/Diagnostics/Debugger/MIEngine/LoadHostTelemetryEvent";
```
gregg-miskelly: Maybe I am being blind, but I don't see where this is used.

src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
```diff
@@ -485,6 +485,15 @@ public void SetProgress(int totalSteps, int currentStep, string progressText)
     public static class HostTelemetry
     {
         /// <summary>
+        /// Initialize Host Telemetry 
+        /// </summary>
+        [Conditional("LAB")]
+        public static void Initialize()
```
gregg-miskelly: I don't have a problem with the change, but if you want a slightly simpler fix, you could eliminate the API change and instead have the existing method which initializes the main thread do both tasks.

pieandcakes:
@chuckries Can you take another look? I think this is what you suggested and I've verified it still solves my problem.

Issue #489, opened 2016-11-04
DebugEngineHost for private distrutibution through VSIX


seanshpark:
When I build and debug run MIEngine inside VS2015, it works fine.
With the guide from [Redistribution](https://github.com/Microsoft/MIEngine/wiki/Redistribution) when I make a release version of VSIX and test, I get this error:
```
The debugger is not properly installed.  Cannot debug the requested type of code.  Run setup to install or repair the debugger.
```
I've found that inside of VSIX in Release folder, DebugEngineHost.dll is result from the DebugEngineHost.stub and it's a Contract dll.
How do I let the VS2015 not use this from stub but the one from DebugEngineHost project?
Should I unpack VSIX and replace the DLL and Zip it again?

seanshpark:
CC @manu-silicon , @gregg-miskelly , @Dmitri-Botcharnikov

gregg-miskelly:
@seanshpark my guess would be that one of the following two things is happening:

1. Some project is referencing the stub project without `<Private>false</Private>` ([example](https://github.com/Microsoft/MIEngine/blob/master/src/MICore/MICore.csproj#L115)).
2. The '[BuiltProjectOutputGroupDependencies](https://github.com/Microsoft/MIEngine/blob/master/src/MIDebugPackage/MIDebugPackage.csproj#L243)' is mistakenly including the stub assembly.

I would probably suggest checking the projects for 1, and if that doesn't work, comb through the verbose build logs to see what is dragging it in.

seanshpark:
@gregg-miskelly , Thanks!. Found that `SSHDebugPS.csproj` did not have `<Private>false</Private>`. I'll prepare a patch.

PR #491, opened 2016-11-09
Fix DebugEngineHost stub reference for SSHDebugPS

seanshpark:
- Related issue: #489
- SSHDebugPS needs to add <Private>false</Private> for DEbugEngineHost.stub and also DesignTime
- Additionally remove extra ">" in csproj


gregg-miskelly:
@seanshpark thanks for submitting this! Looks good to me.

PR #492, opened 2016-11-09
Updating version of clrdbg to get fixes.

rajkumar42:


rajkumar42:

chuckries:
👍 Do we have any plans to add a new moniker at some point?

gregg-miskelly:
@chuckries yes, when there is a breaking change

gregg-miskelly:
@rajkumar42 Are you also going to update the script that we download in the SSH attach to process?

rajkumar42:
@gregg-miskelly Yes I will update the script which is used in the SSH scenario when the clrdbg zips are available for all the distros.

PR #493, opened 2016-11-10
Fixing locking of zip files

rajkumar42:
Fixing locking of zip files by moving the downloading and extraction into a separate method.


rajkumar42:

PR #494, opened 2016-11-10
Fix addition of extra breakpoints.

Dmitri-Botcharnikov:
Enabling/disabling breakpoint in VS breakpoint pane adds lots of unnecessary subordinate breakpoints to pending one. This patch fix this.
before
after

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
@@ -351,7 +351,7 @@ namespace Microsoft.MIDebugEngine
         {
             lock (_boundBreakpoints)
             {
-                if (!IsDataBreakpoint && _boundBreakpoints.Find((b) => b.Addr == bp.Addr) != null)
+                if (bp.Addr == 0 || (!IsDataBreakpoint && _boundBreakpoints.Find((b) => b.Addr == bp.Addr) != null))
                 {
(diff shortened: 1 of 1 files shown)


gregg-miskelly:
@Dmitri-Botcharnikov thanks for taking the time to submit a PR!
@chuckries @paulmaybee can you help review this?

There are a couple of scenarios where I think we want to allow breakpoints without an address:

1. I think data breakpoints don't have an address
2. I don't think clrdbg breakpoints have an address. You can detect that with: `_engine.DebuggedProcess.MICommandFactory.Mode == MIMode.Clrdbg`
3. There is some handling of 'address == 0' breakpoints for when we get the address later, so I _think_ maybe we need to allow 1 address == 0 break point for a pending.

Paul/Chuck do you remember case 3?
@Dmitri-Botcharnikov what was the scenario where you were seeing this problem? Do you know if the extra breakpoints were coming in during the initial bind or via a breakpoint-modified event?

Dmitri-Botcharnikov:
@gregg-miskelly: we use MIEngine  for debugging C# apps on CoreCLR / Linux through lldb-mi. I set breakpoint and start remote debugging (Windows -> Linux). The breakpoint is created as pending with zero address of course. After method is jitted the notification form debugger is sent and breakpoint in VS updated with real native address. After bp hits, I try to check/uncheck enabling box in breakpoints pane and it's time when all extra breakpoints are created: in response to breakpoint modified event.

PR #495, opened 2016-11-12
Windows defender locks zip file after downloading and resulting in failure to delete.

rajkumar42:
Fixing it by directly extracting it through an open stream.


rajkumar42:
@gregg-miskelly @chuckries

PR #499, opened 2016-11-16
Add support for MIDebugLog command:

paulmaybee:
Debug.MIDebugLog /On[:path] [/Output] - start logging to one or both of "path" and the Output Window
  Debug.MIDebugLog /Off - stop logging

### DIFF
--- a/src/DebugEngineHost/HostConfigurationStore.cs
+++ b/src/DebugEngineHost/HostConfigurationStore.cs
@@ -88,14 +88,13 @@ namespace Microsoft.DebugEngineHost
         /// </summary>
         /// <param name="enableLoggingSettingName">[Optional] In VS, the name of the settings key to check if logging is enabled. If not specified, this will check 'Logging' in the AD7 Metrics.</param>
         /// <param name="logFileName">[Required] name of the log file to open if logging is enabled.</param>
-        /// <returns>[Optional] If logging is enabled, the logging object.</returns>
+        /// <returns>If no error then logging object. If file cannot be openened then throw an exception. Otherwise return an empty logger - the user can explictly reconfigure it later</returns>
         public HostLogger GetLogger(string enableLoggingSettingName, string logFileName)
         {
             if (string.IsNullOrEmpty(logFileName))
             {
-                throw new ArgumentNullException("logFileName");
+                return new HostLogger();    // empty logger
             }
-
             object enableLoggingValue;
             if (!string.IsNullOrEmpty(enableLoggingSettingName))
             {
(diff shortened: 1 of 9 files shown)


src/DebugEngineHost/HostLogger.cs
```diff
@@ -12,27 +12,89 @@ namespace Microsoft.DebugEngineHost
...
+            if (Path.IsPathRooted(logFileName) || (!string.IsNullOrEmpty(tempDirectory) && Directory.Exists(tempDirectory)))
+            {
+                string filePath = Path.Combine(tempDirectory, logFileName);
+
+                try
+                {
+                    FileStream stream = File.Open(filePath, FileMode.Create, FileAccess.Write, FileShare.Read);
+                    writer = new StreamWriter(stream);
+                }
+                catch (IOException)
```
gregg-miskelly: In the new case, I wonder if we want to let the user know this failed?

src/MIDebugPackage/MIDebugPackagePackage.cs
```diff
@@ -223,6 +227,81 @@ private int MIDebugExec(uint nCmdExecOpt, IntPtr pvaIn, IntPtr pvaOut)
...
 
+
+        /// <summary>
+        /// The syntax for the MIDebugLog command. Notes:
+        /// ':' : The switch takes a value
+        /// '!' : The value is required
+        /// '(' ... ')' : Auto complete list for the switch (I don't know what is valid here except for 'd')
+        /// d : A path
+        /// </summary>
+        private const string LogMIDebugCommandSyntax = "O,On:(d) Output Off";
```
paulmaybee: Done

src/MIDebugPackage/MIDebugPackagePackage.cs
```diff
@@ -291,6 +370,19 @@ private void LaunchDebugTarget(string filePath, string options)
             debugger.LaunchDebugTargets4(1, debugTargets, processInfo);
         }
 
+        private void EnableLogging(bool output, string logFile)
```
paulmaybee: Done

src/MICore/Logger.cs
```diff
@@ -27,6 +27,15 @@ public class Logger
...
         private static int s_count;
         private int _id;
+        public class LogInfo
+        {
+            public string logFile;
+            public HostLogger.OutputCallback logToOutput;
+            public bool enabled;
+        };
+        private static LogInfo s_cmdLogInfo = new LogInfo();
+        public static LogInfo CmdLogInfo { get { return s_cmdLogInfo; } set { s_cmdLogInfo = value; } }
```
paulmaybee: Done

src/DebugEngineHost/HostConfigurationStore.cs
```diff
@@ -88,14 +88,13 @@ public void GetExceptionCategorySettings(Guid categoryId, out HostConfigurationS
...
         /// <param name="enableLoggingSettingName">[Optional] In VS, the name of the settings key to check if logging is enabled. If not specified, this will check 'Logging' in the AD7 Metrics.</param>
         /// <param name="logFileName">[Required] name of the log file to open if logging is enabled.</param>
-        /// <returns>[Optional] If logging is enabled, the logging object.</returns>
+        /// <returns>If no error then logging object. If file cannot be openened then throw an exception. Otherwise return an empty logger - the user can explictly reconfigure it later</returns>
         public HostLogger GetLogger(string enableLoggingSettingName, string logFileName)
         {
             if (string.IsNullOrEmpty(logFileName))
             {
-                throw new ArgumentNullException("logFileName");
+                return new HostLogger();    // empty logger
```
paulmaybee: This (and below) were to support a scenario where the old-style config caused a logger to already exist and sometime later it was being replaced by command. Only one HostLogger would ever exist, and would (almost) always exist but might not do anything. However I could change it so that the Logger closes the old one and switches to a new HostLogger object when a change is made.

paulmaybee: If we are going to switch loggers then additional locking is necessary in the Logger to prevent message losses on switching logs. Or maybe that just isn't an issue we care about?

gregg-miskelly: Can't we just swap the pointers (Interlocked.Exchange) and Flush the old one if we had one?

paulmaybee:
Yeah, but I think we still may lose a message:
Thread 1: gets pointer to logger
Thread 2: swaps pointers and Closes old one
Thread 1: attempts to write to closed logger, message is lost

src/DebugEngineHost/HostConfigurationStore.cs
paulmaybee: done

src/DebugEngineHost/HostConfigurationStore.cs
```diff
@@ -110,26 +109,22 @@ public HostLogger GetLogger(string enableLoggingSettingName, string logFileName)
                 !(enableLoggingValue is int) ||
                 ((int)enableLoggingValue) == 0)
             {
-                return null;
+                return new HostLogger();    // empty logger
```
paulmaybee: No HostConfigurationStore is available on which to call GetLoggerFromCmd. I will move this functionality directly into the HostLogger.

PR #503, opened 2016-11-18
Make the MIEngine project buildable in Dev15

gregg-miskelly:
This checkin makes the changes to the MIEngine projects so that they are buildable/launchable in Dev15.

Testing:
- Verified that the project can be built from the IDE and from build.cmd in both Dev14 and Dev15
- Verified that, when building in Dev14, we get exactly the same IL as before
- Verified that we can launch the project in Dev15

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -12,7 +12,7 @@ if not defined VisualStudioVersion (
         goto :EnvSet
     )
 
-    echo Error: build.cmd requires Visual Studio 2015.
+    echo Error: build.cmd requires Visual Studio 2015 or to be run from a developer command prompt.
     exit /b 1
 )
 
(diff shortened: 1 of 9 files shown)


faxue-msft:
Build.cmd still depends on VS140COMNTOOLS to work properly, how will it work when Dev15 is installed?

faxue-msft:
otherwise LGTM

gregg-miskelly:
Build.cmd will work now as long as it is run from a Dev command prompt.

Dev15 doesn't add any global environment variables, so I can't make the equivalent code work unless I sniff the Dev15 install directory. I am tempted to think we should just run it from a dev command prompt unless that is hard to do.

gregg-miskelly:
I will update the error message though.

PR #505, opened 2016-11-22
Ensure logger in enabled when originally initialized via command.

paulmaybee:
Engine logger was only being enabled if the registry was previously updated to enable logging. If MIEngineLog command was called on a clean registry the log would never become enabled. This fixes the problem by enabling the log under the command initialization case too.


gregg-miskelly:
:shipit:

PR #508, opened 2016-11-24
Move InstallToVSCode to use 1.1 .NET Core

gregg-miskelly:
Testing: Verified that basic debugging still works on Windows, Ubuntu 14 and OSX.


gregg-miskelly:
@chuckries please review

PR #509, opened 2016-11-25
Fixed mdd android c++ miengine glass tests

richardzeng1985:
Verify this fix works in both VS2015 U3 (micro update 3.5) and VS2017 RC:

1) Fixed the build error in android C++ debuggee projects, we might need to revisit this if we release new version of android application later.
2) Fixed test script to retrieve the correct %errorlevel% code when failed to build projects. Currently the test script marked all of glass tests as successfully even the projects failed to build due to missing the redirect "stderr" code.
3) Fixed the load exceptions issues in glass2 ": Unable to load one or more of the requested types"
4) Fixed the "COM API failures" issues due to fail to load one binary in glass2 runtime.

### DIFF
--- a/test/Android/androidtest.cmd
+++ b/test/Android/androidtest.cmd
@@ -35,6 +35,12 @@ set _GlassDir=%_ProjectRoot%%_GlassPackageName%\
 if NOT exist "%_GlassDir%glass2.exe" echo Getting Glass from NuGet.& call "%_ProjectRoot%tools\NuGet\nuget.exe" install %_GlassPackageName% -Version %_GlassPackageVersion% -ExcludeVersion -OutputDirectory %_ProjectRoot%
 if NOT "%ERRORLEVEL%"=="0" echo ERROR: Failed to get Glass from NuGet.& exit /b -1
 
+:: Copy binary to folder "Microsoft.VisualStudio.Glass" which required by glass2.exe in runtime for Visual Studio 2017
+if defined VS150COMNTOOLS (
+xcopy /Y /D "%VSINSTALLDIR%Common7\IDE\Remote Debugger\x86\Microsoft.VisualStudio.OLE.Interop.dll" "%_GlassDir%"
+if not "%ERRORLEVEL%"=="0" echo ERROR: Unable to copy the binaries from Visual Studio installation.& exit /b -1
+)
+
 :: Ensure the project has been built
 if NOT exist "%_GlassDir%Microsoft.MIDebugEngine.dll" echo The project has not been built. Building now with default settings.& call %_ProjectRoot%build.cmd
 if NOT "%ERRORLEVEL%"=="0" echo ERROR: Failed to build MIEngine project.& exit /b -1
@@ -192,7 +198,7 @@ exit /b -1
     set LastTestSucceeded=false
     
     ::Build the app
-    call msbuild /p:Platform=%_Platform%;VS_NDKRoot="%_NdkRoot%";VS_SDKRoot="%_SdkRoot%";PackageDebugSymbols=true > build.log
+    call msbuild /p:Platform=%_Platform%;VS_NDKRoot="%_NdkRoot%";VS_SDKRoot="%_SdkRoot%";PackageDebugSymbols=true > build.log 2>&1
     if NOT "%ERRORLEVEL%"=="0" echo ERROR: Failed to build %~1. See build.log for more information.& set FAILED_TESTS="%~1" "%FAILED_TESTS%"& goto RunSingleTestDone
     
     ::Deploy the app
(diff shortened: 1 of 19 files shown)


test/Android/androidtest.cmd
```diff
@@ -34,6 +35,14 @@ set _GlassDir=%_ProjectRoot%%_GlassPackageName%\
 if NOT exist "%_GlassDir%glass2.exe" echo Getting Glass from NuGet.& call "%_ProjectRoot%tools\NuGet\nuget.exe" install %_GlassPackageName% -Version %_GlassPackageVersion% -ExcludeVersion -OutputDirectory %_ProjectRoot%
 if NOT "%ERRORLEVEL%"=="0" echo ERROR: Failed to get Glass from NuGet.& exit /b -1
 
+:: Copy binaries to folder "Microsoft.VisualStudio.Glass" which required by glass2.exe in runtime for Visual Studio 2017
+if defined VS150COMNTOOLS (
+xcopy /Y /D "%VSINSTALLDIR%Common7\IDE\Remote Debugger\x86\Microsoft.VisualStudio.Debugger.Interop.15.0.dll" "%_GlassDir%"
+xcopy /Y /D "%VSINSTALLDIR%Common7\IDE\Remote Debugger\x86\Microsoft.VisualStudio.OLE.Interop.dll" "%_GlassDir%"
+xcopy /Y /D "%VSINSTALLDIR%Common7\IDE\PrivateAssemblies\Microsoft.VisualStudio.Shell.14.0.dll" "%_GlassDir%"
```
gregg-miskelly: The shell dll shouldn't be needed (if it is, we have a bug)

test/Android/androidtest.cmd
```diff
@@ -34,6 +35,14 @@ set _GlassDir=%_ProjectRoot%%_GlassPackageName%\
 if NOT exist "%_GlassDir%glass2.exe" echo Getting Glass from NuGet.& call "%_ProjectRoot%tools\NuGet\nuget.exe" install %_GlassPackageName% -Version %_GlassPackageVersion% -ExcludeVersion -OutputDirectory %_ProjectRoot%
 if NOT "%ERRORLEVEL%"=="0" echo ERROR: Failed to get Glass from NuGet.& exit /b -1
 
+:: Copy binaries to folder "Microsoft.VisualStudio.Glass" which required by glass2.exe in runtime for Visual Studio 2017
+if defined VS150COMNTOOLS (
+xcopy /Y /D "%VSINSTALLDIR%Common7\IDE\Remote Debugger\x86\Microsoft.VisualStudio.Debugger.Interop.15.0.dll" "%_GlassDir%"
```
gregg-miskelly: Actually, we don't need the 15 interop assembly either

test/Android/androidtest.cmd
richardzeng1985: Following code "Host.EnsureMainThreadInitialized();" rely on the shell dll, is it expected ? 

        public AD7Engine()
        {
            Host.EnsureMainThreadInitialized();

            _breakpointManager = new BreakpointManager(this);
        }

richardzeng1985: Following code "Host.EnsureMainThreadInitialized();" rely on the shell dll, is it expected ? 

        public AD7Engine()
        {
            Host.EnsureMainThreadInitialized();

            _breakpointManager = new BreakpointManager(this);
        }

test/Android/androidtest.cmd
richardzeng1985: glass2 will throw exception "Unable to load one or more of the requested types."  in following code when initialized:

            Type t = typeof(GlassCommand);
            Type[] rgtypes = t.Module.GetTypes();

By checking the $exception.LoaderExceptions, you can see it relied on "Microsoft.VisualStudio.Debugger.Interop.15.0" in Dev15, is it expected ?

test/Android/androidtest.cmd
gregg-miskelly: This is a bug. I just submitted a PR to fix this: https://github.com/Microsoft/MIEngine/pull/510

test/Android/androidtest.cmd
gregg-miskelly: Are you running the glass tests from an admin command prompt? Maybe try deleting HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Glass (note: this is a **32** bit registry key, so if you are using 64-bit regedit, look under HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node).

The version of glass that AndroidTest.cmd pulls is Dev14-based. So it definitely shouldn't need the 15 interop assembly.

test/Android/androidtest.cmd
richardzeng1985: Seems we should handle the exception in src/MIDebugEngine/AD7.Impl/AD7Engine.cs  instead, otherwise, glass will still throw the exception.

test/Android/androidtest.cmd
richardzeng1985: Yes, it works after I deleted the Glass register key.

richardzeng1985:
I was not able to create a new pull request, so try to close this one first.

PR #510, opened 2016-11-28
VS 2015 Glass fixes

gregg-miskelly:
- Host was mistakenly taking a hard dependency on the existance of a shell dll. This moves the dependency to be a soft dependency.
- On the machine I was running things at least, there was a different NDK, and the .apk files where dropping to a different place than when AndroidTest.cmd expected. This made AndroidTest.cmd a bit more forgiving.


gregg-miskelly:
@chuckries please review

PR #511, opened 2016-11-29
Fixed mdd android C++ miengine based glass tests with new updates

richardzeng1985:
Fixed following items which can be running glass tests on machine with VS2017 RC installed:
1.	Fixed the build error in android C++ debuggee projects, we might need to revisit this if we release new version of android application later.
2.	Fixed test script to retrieve the correct %errorlevel% code when failed to build projects. 
3.	Fixed the load exceptions issues in glass2 ": Unable to load one or more of the requested types"
4.	Fixed the "COM API failures" issues due to missing vs types when invoke miengine code.

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7Engine.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Engine.cs
@@ -70,7 +70,14 @@ namespace Microsoft.MIDebugEngine
 
         public AD7Engine()
         {
-            Host.EnsureMainThreadInitialized();
+            try
+            {
+                Host.EnsureMainThreadInitialized();
+            }
+            catch
+            {
+                // In miengine based glass tests, VS types will be missing and throw exception. Ignore the exceptions.
+            }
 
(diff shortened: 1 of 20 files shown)


src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -70,7 +70,14 @@ static AD7Engine()
 
         public AD7Engine()
         {
-            Host.EnsureMainThreadInitialized();
+            try
+            {
+                Host.EnsureMainThreadInitialized();
+            }
+            catch
```
gregg-miskelly: This should really be inside of the implementation of Host.EnsureMainThreadInitialized. I tried fixing this again: https://github.com/Microsoft/MIEngine/pull/512. Can you give that a try?

richardzeng1985: Yeah, it works correctly after apply for changes in #512

gregg-miskelly:
Looks good @richardzeng1985. Thanks!

PR #514, opened 2016-12-08
Implement IDebugDocumentContext150

chuckries:
This allows the MIEngine to opt out of Visual Studio's default source
search directories. We were seeing an issue where linux's crt source
files names might match source files in Microsoft's crt, which is on the
Default Source Search Path. This was causing VS to show incorrect source
files.

This change required us to create our own definition of
IDebugDocumentContext150, which was previously defined in
Microsoft.VisualStudio.Debugger.Interop.15.0.dll. Unfortunately we
cannot take a dependency on this dll because:
1. It is compatible with portable clr code, so this would not work for
x-plat scenarios
2. We need MIEngine to continue to run in dev14, where
Debugger.Interop.15.0.dll is not available.

### DIFF
--- a/src/MIDebugEngine/AD7.Definitions/IDebugDocumentContext150.cs
+++ b/src/MIDebugEngine/AD7.Definitions/IDebugDocumentContext150.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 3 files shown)


chuckries:
@gregg-miskelly

gregg-miskelly:
:shipit:

PR #515, opened 2016-12-08
Update references/NuGet.exe to bring build/deploy to dev15

chuckries:
I found that the project was failing to build/deploy into the
experimental instance on dev15. This updates the shell.immutable
references to be fully qualified, ensuring that we pick up the correct
versions from VSSDK. It looks like shell.immutable does some weird
binding redirect stuff in dev15, which was causing us to fail without
this fix.

This also updates nuget.exe to a version which undresntands msbuild v15,
so that we no longer see a spurious nuget error on build.


chuckries:
I tested that build/deploy from VS works as expected in both dev14 and dev15.

chuckries:
@gregg-miskelly

PR #522, opened 2016-12-16
Fixing crash bug when attaching through SSH.

rajkumar42:
Exceptions were not caught because it was a void async method, which crashed the application. Fixed it to return a Task.
Added actionable error message for the user in situations when this can happen.

### DIFF
--- a/src/MIDebugEngine/ResourceStrings.resx
+++ b/src/MIDebugEngine/ResourceStrings.resx
@@ -215,7 +215,7 @@
 
 {1} failed with message: {2}
 
-Updates can cause binaries replaced on disk and resulting in this failure. Try re-launching the application or restarting the machine.</value>
+This may occur if the process's executable was changed after the process was started, such as when installing an update. Try re-launching the application or restarting the machine.</value>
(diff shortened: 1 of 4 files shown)


rajkumar42:
@gregg-miskelly @chuckries

src/MIDebugEngine/ResourceStrings.resx
```diff
@@ -213,7 +213,9 @@
   <data name="Error_ExePathInvalid" xml:space="preserve">
     <value>Program path '{0}' is missing or invalid.
 
-{1} failed with message: {2}</value>
+{1} failed with message: {2}
+
+Updates can cause binaries replaced on disk and resulting in this failure. Try re-launching the application or restarting the machine.</value>
```
gregg-miskelly: nit: The first sentence seems a little awkward to me. How about:

This may occur if the process's executable was changed after the process was started, such as when installing an update.

rajkumar42: Thanks for the text!

PR #523, opened 2016-12-17
Fixing SSH attach to work with other shells.

rajkumar42:
Invoking bash shell before any other command does the trick.


rajkumar42:
@gregg-miskelly @chuckries

rajkumar42:
Tested with zsh.

PR #524, opened 2016-12-19
Merging master to release_mdd

rajkumar42:
The merge contains 2 bug fixes for SSH needed for RC3
- Fixed: VS crashes when binaries are updated.
- Fixed: SSH does not work on non-bash default shells.


rajkumar42:
@gregg-miskelly @chuckries

Issue #520, opened 2016-12-15
User sees an error when detaching from a currently running process


paulmaybee:
When debugging using gdbserver on a remote host MIEngine does not wait for the interrupt to finish before issuing the -target-detach. I have seen the same behavior on "Stop Debugging".

17: (8886412) <-1014-exec-interrupt
17: (8886420) ->1014^done
17: (8886420) ->(gdb)
17: (8886420) 1014: elapsed time 9
17: (8886421) <-1015-target-detach
17: (8886431) ->1015^error,msg="Cannot execute this command while the target is running.\nUse the \"interrupt\" command to stop the target\nand then try again."
17: (8886431) ->(gdb)
17: (8886431) 1015: elapsed time 10
17: (8886724) ->=thread-created,id="2",group-id="i1"
17: (8886725) ->~"[New Thread 706]\n"
17: (8886725) ->~"\nProgram received signal "
[New Thread 706]

Program received signal 17: (8886725) ->~"SIGINT, Interrupt.\n"
SIGINT, Interrupt.
17: (8886770) ->~"0xb6f06250 in ?? ()\n"
0xb6f06250 in ?? ()
17: (8886770) ->*stopped,reason="signal-received",signal-name="SIGINT",signal-meaning="Interrupt",frame={addr="0xb6f06250",func="??",args=[]},thread-id="1",stopped-threads="all",core="0"
17: (8886770) <-1016-thread-info
17: (8887826) ->1016^done,threads=[{id="2",target-id="Thread 706",frame={level="0",addr="0xb6f03998",func="??",args=[]},state="stopped",core="0"},{id="1",target-id="Thread 699",frame={level="0",addr="0xb6f06250",func="??",args=[]},state="stopped",core="0"}],current-thread-id="1"
17: (8887826) ->(gdb)

PR #529, opened 2017-01-04
Wait for process to stop before terminating or detaching. Also fixes …

paulmaybee:
…process reference cleanup on abort event.
Issue #520  
@gregg-miskelly @jacdavis @faxue-msft

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -603,10 +603,13 @@ namespace MICore
                 _terminating = true;
                 if (ProcessState == ProcessState.Running && this.MICommandFactory.Mode != MIMode.Clrdbg)
                 {
+                    await AddInternalBreakAction(() => MICommandFactory.Terminate());
                     await CmdBreak(BreakRequest.Async);
                 }
-
-                await MICommandFactory.Terminate();
+                else
+                {
+                    await MICommandFactory.Terminate();
+                }
             }
 
             return new Results(ResultClass.done);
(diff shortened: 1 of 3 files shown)


src/MICore/Debugger.cs
```diff
@@ -617,10 +620,13 @@ protected bool IsCoreDump
             _detaching = true;
             if (ProcessState == ProcessState.Running)
```
gregg-miskelly: You also want: `&& this.MICommandFactory.Mode != MIMode.Clrdbg`

src/MICore/Debugger.cs
```diff
@@ -603,10 +603,13 @@ protected bool IsCoreDump
                 _terminating = true;
                 if (ProcessState == ProcessState.Running && this.MICommandFactory.Mode != MIMode.Clrdbg)
                 {
+                    await AddInternalBreakAction(() => MICommandFactory.Terminate());
                     await CmdBreak(BreakRequest.Async);
```
gregg-miskelly: AddInternalBreakAction already does a CmdBreak, so can't we remove this?

paulmaybee: yes, done

src/MICore/Debugger.cs
paulmaybee: done

gregg-miskelly:
LGTM. Merge when ready...

paulmaybee:
Currently not planning on going to ask. If early access users start complaining about hitting it all the time we may change our minds.

PR #533, opened 2017-01-21
Fixed improper encapsulation of 'pwd' 

ptoonen:
When no installation path was provided, the string was set by an empty one due to incorrect encapsulation of the pwd command.


gregg-miskelly:
@rajkumar42 FYI

rajkumar42:
Thanks @ptoonen !

PR #535, opened 2017-01-24
Merging master to getclrdbg-release for docker 

rajkumar42:
Integrating fixes made in master to getclrdbg-release so that docker scenarios are also benefited.


rajkumar42:
@gregg-miskelly @caslan

Issue #536, opened 2017-01-25
getclrdbgsh is broken


Spongman:
all it does is create a directory ~/clrdbg containing a single (ironically-named) file 'success.txt'.

````
~$ curl -sSL https://aka.ms/getclrdbgsh | bash /dev/stdin vs2015u2 ~/clrdbg
Info: Creating install directory
Using arguments
    Version                    : 'vs2015u2'
    Location                   : '/home/user/clrdbg'
    SkipDownloads              : 'false'
    LaunchClrDbgAfter          : 'false'
    RemoveExistingOnUpgrade    : 'false'
Info: Using clrdbg version '15.0.26022.0'
Info: Previous installation at /home/user/clrdbg not found
Info: Determining Runtime ID
Info: Using Runtime ID 'debian.8-x64'
Downloading https://vsdebugger.azureedge.net/clrdbg-15-0-26022-0/clrdbg-debian.8-x64.zip
chmod: cannot access â./clrdbgâ: No such file or directory
Info: Successfully installed clrdbg at '/home/user/clrdbg'

~$ ls -lR ~
.:
total 4
drwxr-xr-x 2 user user 4096 Jan 24 16:58 clrdbg

./clrdbg:
total 4
-rw-r--r-- 1 user user 13 Jan 24 16:58 success.txt

~$ cat ./clrdbg/success.txt
15.0.26022.0
````

rajkumar42:
Thanks @Spongman, it should be fixed now.

PR #537, opened 2017-01-25
remove '-f' flag from unzip

Spongman:
unzip -f doesn't do what you think it does.


Spongman:
fixes #536

gregg-miskelly:
@Spongman Thanks for fixing this!

Issue #147, opened 2016-01-05
Multiple source files with the same name can cause breakpoint isues.


chuckries:
From VS Feedback item:

1) Make two shared objects by having 2 projects in a Solution
    (Denote these by project/library 'A' and 'B')
2) In two separate projects, create a C++ source file in each with the same name 
    (in my solution/workflow, "OnLoad.cpp" appears in multiple projects (which are made into multiple .so files)
3) Put a breakpoint in one of the "OnLoad.cpp" files 
    (in my case, I put a breakpoint in the last library loaded by my application)
    (App loads project 'A' and then project 'B' libraries -- which means my breakpoint was in project 'B')
4) Start Debugging using GDB. If the breakpoint is in said file contained in project 'B', the debugger interprets project 'A's file with the same name and will use the breakpoints in project 'A's library -- not project 'B' as you desired.

erwanh1112:
Here are somewhat simplified repro steps on VS2015 U3 + KB3165756 (version 14.0. 25431.01)

1. Create a new solution and project Console Application (Linux) named A
2. Add to the solution a new project  Console Application (Linux) named B
3. In the Solution Explorer, right-click on the solution and select Project Dependencies... Add a dependency on project A to project B, such that B builds first
4. Ensure that A is the active project
5. _In project B_, Set a breakpoint (F9) on main.cpp@6 -- i.e. on the printf() line
6. Press F5. Enter the information to connect to the remote Linux box

The debugger breaks on A.exe!main.cpp@6. The debugger should not break at all since the breakpoint is actually set on project B, which is not part of the debug session. In Visual Studio, when debugging starts, the breakpoint visually moves from B.exe!main.cpp@6 to A.exe!main.cpp@6 . When debugging stops, the breakpoint moves back to B.exe!main.cpp@6

When looking at the debug output in VS, we can see the following:

-file-exec-and-symbols .../bin/x64/Debug/A.exe
-break-insert main
-break-insert -f main.cpp:6
-symbol-list-lines main.cpp
-exec-run

It appears that the MIEngine notifies GDB to set a breakpoint on the executable being debugged (i.e. A.exe) on main.cpp@6, event though the breakpoint is actually set on a different project B. Note that the issue reproduces with either GDB or GDBSERVER

gregg-miskelly:
[...]

PR #541, opened 2017-01-26
Support breakpoints in sourcefiles with same base name.

paulmaybee:
- projects can send in source file mapping data
 - users can supply solution-specific additional source map file
@gregg-miskelly @pieandcakes

### DIFF
--- a/src/DebugEngineHost/HostNatvisProject.cs
+++ b/src/DebugEngineHost/HostNatvisProject.cs
@@ -37,6 +37,19 @@ namespace Microsoft.DebugEngineHost
             paths.ForEach((s) => loader(s));
         }
 
+        public static string FindSolutionRoot()
+        {
+            string path = null;
+            try
+            {
+                ThreadHelper.Generic.Invoke(() => { path = Internal.FindSolutionRootImpl(); });
+            }
+            catch (Exception)
+            {
+            }
+            return path;
+        }
+
         private static class Internal
         {
             // from vsshell.idl
@@ -87,6 +100,20 @@ namespace Microsoft.DebugEngineHost
                 {
                     LoadNatvisFromProject(proj[0], paths, solutionLevel: true);
                 }
+            }
+
+            public static string FindSolutionRootImpl()
+            {
+                string root = null;
+                string slnFile;
+                string slnUserFile;
+                var solution = (IVsSolution)Package.GetGlobalService(typeof(SVsSolution));
+                if (solution == null)
+                {
+                    return null; // failed to find a solution
+                }
+                solution.GetSolutionInfo(out root, out slnFile, out slnUserFile);
+                return root;
             }
 
             private static void LoadNatvisFromProject(IVsHierarchy hier, List<string> paths, bool solutionLevel)
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -436,9 +436,11 @@ namespace MICore
                 cmd.Append(" ");
             }
 
+            cmd.Append("\"");
             cmd.Append(filename);
             cmd.Append(":");
             cmd.Append(line.ToString());
+            cmd.Append("\"");
 
             return await _debugger.CmdAsync(cmd.ToString(), resultClass);
         }
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -426,7 +426,24 @@ namespace MICore
             return cmd;
         }
 
-        public virtual async Task<Results> BreakInsert(string filename, uint line, string condition, bool enabled, IEnumerable<Checksum> checksums = null, ResultClass resultClass = ResultClass.done)
+        internal bool PreparePath(string path, bool useUnixFormat, out string pathMI)
+        {
+            bool requiresQuotes = false;
+            path = path.Trim();
+            if (useUnixFormat)  // convert directory separators
+            {
+                path = path.Replace('\\', '/');
+            }
+            if (path.IndexOf(' ') != -1)    // path contains spaces. Convert to c-string format
+            {
+                path = path.Replace(@"\", @"\\");   // escape any backslashes in the path
+                requiresQuotes = true;              // parameter containing the name will need to be quoted
+            }
+            pathMI = path;
+            return requiresQuotes;
+        }
+
+        public virtual async Task<Results> BreakInsert(string filename, bool useUnixFormat, uint line, string condition, bool enabled, IEnumerable<Checksum> checksums = null, ResultClass resultClass = ResultClass.done)
         {
             StringBuilder cmd = BuildBreakInsert(condition, enabled);
 
@@ -436,11 +453,19 @@ namespace MICore
                 cmd.Append(" ");
             }
 
-            cmd.Append("\"");
-            cmd.Append(filename);
+            string filenameMI;
+            bool quotes = PreparePath(filename, useUnixFormat, out filenameMI);
+            if (quotes)
+            {
+                cmd.Append("\"");
+            }
+            cmd.Append(filenameMI);
             cmd.Append(":");
             cmd.Append(line.ToString());
-            cmd.Append("\"");
+            if (quotes)
+            {
+                cmd.Append("\"");
+            }
 
             return await _debugger.CmdAsync(cmd.ToString(), resultClass);
         }
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -172,6 +172,39 @@ namespace MICore
         public string Value { get; private set; }
     }
 
+    public sealed class PathMapEntry
+    {
+        public PathMapEntry()
+        {
+        }
+
+        public PathMapEntry(Xml.LaunchOptions.PathMapEntry xmlEntry)
+        {
+            if (string.IsNullOrEmpty(xmlEntry.EditorPath))
+            {
+                throw new ArgumentNullException("EditorPath");
+            }
+            this.EditorPath = xmlEntry.EditorPath;
+            this.CompilerPath = xmlEntry.CompilerPath == null ? string.Empty : xmlEntry.CompilerPath;
+            this.UseForBreakpoints = xmlEntry.UseForBreakpoints;
+        }
+
+        public string EditorPath { get; set; }
+        public string CompilerPath { get; set; }
+        public bool UseForBreakpoints { get; set; }
+
+        public static ReadOnlyCollection<PathMapEntry> CreateCollectionFromXml(Xml.LaunchOptions.PathMapEntry[] source)
+        {
+            PathMapEntry[] pathArray = source?.Select(x => new PathMapEntry(x)).ToArray();
+            if (pathArray == null)
+            {
+                pathArray = new PathMapEntry[0];
+            }
+
+            return new ReadOnlyCollection<PathMapEntry>(pathArray);
+        }
+    }
+
     /// <summary>
     /// Launch options class when VS should launch an instance of an MI Debugger to connect to an MI Debugger server
     /// </summary>
@@ -888,6 +921,18 @@ namespace MICore
             }
         }
 
+        private ReadOnlyCollection<PathMapEntry> _sourceMap;
+
+        public ReadOnlyCollection<PathMapEntry> SourceMap
+        {
+            get { return _sourceMap; }
+            set
+            {
+                VerifyCanModifyProperty("SourceMap");
+                _sourceMap = value;
+            }
+        }
+
         public string GetOptionsString()
         {
             try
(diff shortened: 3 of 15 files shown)


paulmaybee:
#541

paulmaybee:
#147

src/MICore/LaunchOptions.cs
```diff
@@ -888,6 +921,18 @@ protected set
             }
         }
 
+        private ReadOnlyCollection<PathMapEntry> _sourceMap;
```
pieandcakes: nit: if you are creating a sourceMap can we just call it a SourceMapEntry?

src/MICore/LaunchOptions.cs
```diff
@@ -172,6 +172,39 @@ public EnvironmentEntry(Xml.LaunchOptions.EnvironmentEntry xmlEntry)
...
+            if (string.IsNullOrEmpty(xmlEntry.EditorPath))
+            {
+                throw new ArgumentNullException("EditorPath");
+            }
+            this.EditorPath = xmlEntry.EditorPath;
+            this.CompilerPath = xmlEntry.CompilerPath == null ? string.Empty : xmlEntry.CompilerPath;
+            this.UseForBreakpoints = xmlEntry.UseForBreakpoints;
+        }
+
+        public string EditorPath { get; set; }
```
pieandcakes: CurrentPath and CompileTimePath ? Compiler path sounds like the path to the compiler

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1916,5 +1916,65 @@ private Task<string> ResetConsole()
...
             return ConsoleCmdAsync(@"shell echo -e \\033c 1>&2");
         }
+
+        public bool MapCurrentSrcToCompilerSrc(string currentSrc, out string compilerSrc)
+        {
+            StringComparison comp = PlatformUtilities.IsWindows() ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
+            foreach (var e in _launchOptions.SourceMap)
+            {
+                if (e.UseForBreakpoints && currentSrc.StartsWith(e.EditorPath, comp))
+                {
```
pieandcakes: Should you replace the directory separators for both so that they match? I know we update some of the directory separators in some instances but I don't remember if we do it for breakpoints.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1916,5 +1916,65 @@ private Task<string> ResetConsole()
...
+                    var file = currentSrc.Substring(e.EditorPath.Length);
+                    if (string.IsNullOrEmpty(file)) // matched the whole directory string
+                    {
+                        break;  // use default
+                    }
+                    if (file[0] == Path.DirectorySeparatorChar || file[0] == Path.AltDirectorySeparatorChar)
+                    {
+                        file = file.Substring(1);   // Trim the directory separator
+                    }
+                    compilerSrc = Path.Combine(e.CompilerPath,file);    // map to the compiled location
```
pieandcakes: missing space: (e.CompilerPath, file)

pieandcakes: Will this work properly with the directory separator characters for the target OS? If you go from windows to Linux, will doing Path.Combine give you the Linux directory separators?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1916,5 +1916,65 @@ private Task<string> ResetConsole()
...
+        }
+
+        public bool MapCompilerSrcToCurrentSrc(string compilerSrc, out string currentName)
+        {
+            StringComparison comp = _launchOptions.UseUnixSymbolPaths ? StringComparison.Ordinal : StringComparison.OrdinalIgnoreCase;
+            foreach (var e in _launchOptions.SourceMap)
+            {
+                if (string.IsNullOrEmpty(e.CompilerPath))
+                {
+                    continue;   // don't try to map back if path has an emty compiler src tree
```
pieandcakes: nit: spelling in comment

src/DebugEngineHost/HostNatvisProject.cs
```diff
@@ -89,6 +102,20 @@ public static void FindNatvisInSolutionImpl(List<string> paths)
...
+                string root = null;
+                string slnFile;
+                string slnUserFile;
+                var solution = (IVsSolution)Package.GetGlobalService(typeof(SVsSolution));
+                if (solution == null)
+                {
+                    return null; // failed to find a solution
+                }
+                solution.GetSolutionInfo(out root, out slnFile, out slnUserFile);
+                return root;
```
gregg-miskelly: Does this correctly handle the 'open folder' case too?

src/MICore/LaunchOptions.xsd
```diff
@@ -476,4 +495,33 @@
...
+          </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="CompilerPath" type="xs:string">
+          <xs:annotation>
+            <xs:documentation>The path to the source tree at compile time.</xs:documentation>
+          </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="UseForBreakpoints" type="xs:boolean">
+          <xs:annotation>
+            <xs:documentation>False if entry is only used for stack frame location mapping. True if this entry should also be used when specifying breakpoint locations.</xs:documentation>
```
gregg-miskelly: >. [](start = 167, length = 1)

Might want to also indicate what the default value is.

src/MICore/LaunchOptions.cs
gregg-miskelly: I like Pierson's idea of 'CompileTimePath'. I personally think 'CurrentPath' could be confusing in remote debugging cases. Should I ask around the team room and see if people can understand?

---
In reply to: [98105505](https://github.com/Microsoft/MIEngine/pull/541#discussion_r98105505) [](ancestors = 98105505)

src/MICore/LaunchOptions.cs
gregg-miskelly: If we are going to call it that, I am assuming the more important place to fix it is in the XSD. Correct?

---
In reply to: [98104661](https://github.com/Microsoft/MIEngine/pull/541#discussion_r98104661) [](ancestors = 98104661)

src/MICore/LaunchOptions.cs
```diff
@@ -1055,10 +1100,51 @@ public static LaunchOptions GetInstance(HostConfigurationStore configStore, stri
...
+            }
+
             launchOptions.SetInitializationComplete();
             return launchOptions;
         }
 
+        private void Merge(SupplementalLaunchOptions suppOptions)
+        {
+            // merge the source mapping lists
+            List<PathMapEntry> map = new List<PathMapEntry>(SourceMap);
```
gregg-miskelly: >SourceMap [](start = 60, length = 9)

I feel like the new values should go at the front of the list so they take precedence over what was supplied by the project system.

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -436,9 +436,11 @@ protected virtual StringBuilder BuildBreakInsert(string condition, bool enabled)
                 cmd.Append(" ");
             }
 
+            cmd.Append("\"");
             cmd.Append(filename);
             cmd.Append(":");
             cmd.Append(line.ToString());
+            cmd.Append("\"");
```
gregg-miskelly: >\" [](start = 24, length = 2)

Unfortunately clrdbg doesn't like these quotes. Is LLDB happy with them?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1916,5 +1916,65 @@ private Task<string> ResetConsole()
...
+        {
+            StringComparison comp = PlatformUtilities.IsWindows() ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
+            foreach (var e in _launchOptions.SourceMap)
+            {
+                if (e.UseForBreakpoints && currentSrc.StartsWith(e.EditorPath, comp))
+                {
+                    var file = currentSrc.Substring(e.EditorPath.Length);
+                    if (string.IsNullOrEmpty(file)) // matched the whole directory string
+                    {
+                        break;  // use default
```
gregg-miskelly: >break;  // use default [](start = 24, length = 22)

Shouldn't we allow full matches and return e.EditorPath?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1916,5 +1916,65 @@ private Task<string> ResetConsole()
...
+                    }
+                    if (file[0] == Path.DirectorySeparatorChar || file[0] == Path.AltDirectorySeparatorChar)
+                    {
+                        file = file.Substring(1);   // Trim the directory separator
+                    }
```
gregg-miskelly: >} [](start = 20, length = 1)

In the else case don't we need to verify that e.EditorPath ended with a directory separator? Otherwise we will say that c:\MyDirectory_Alt\Foo.cpp is under a path map entry for 'c:\MyDirectory'.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1916,5 +1916,65 @@ private Task<string> ResetConsole()
...
+                {
+                    var file = compilerSrc.Substring(e.CompilerPath.Length);
+                    if (string.IsNullOrEmpty(file)) // matched the whole directory string
+                    {
+                        break;  // use default
+                    }
+                    if (file[0] == Path.DirectorySeparatorChar || file[0] == Path.AltDirectorySeparatorChar)
+                    {
+                        file = file.Substring(1);   // Trim the directory separator
+                    }
```
gregg-miskelly: >} [](start = 20, length = 1)

Same comment as above

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1916,5 +1916,65 @@ private Task<string> ResetConsole()
...
+            }
+            currentName = compilerSrc;
+            return false;
+        }
+
+        public string GetMappedFileFromTuple(TupleValue tuple)
+        {
+            string file = tuple.Contains("fullname") ? tuple.FindString("fullname") : tuple.TryFindString("file");
+            string currentName;
+            MapCompilerSrcToCurrentSrc(file, out currentName);
```
gregg-miskelly: >MapCompilerSrcToCurrentSrc [](start = 12, length = 26)

I think you need an 'if (file != null) around thsi.

gregg-miskelly:
@paulmaybee I will be on vacation tomorrow. But @caslan can review if you are ready. Or I will be back Monday if you don't mind waiting.

src/MICore/CommandFactories/MICommandFactory.cs
paulmaybee: This is according to the spec for MI. You can't quote the file name alone, you must quote the entire parameter value. So if it doesn't work on lldb it is a bug on their side. However I don't have an apple available to check it. Can one of you do this?  It should be enough to start lldb in MI mode and enter: -break-insert "main.c:10".  But I will leave them off if there are no spaces in the filename and that should prevent any previously working scenario from failing. @pieandcakes

src/DebugEngineHost/HostNatvisProject.cs
paulmaybee: What is the 'open folder' case?

paulmaybee: Ah, you mean Open Folder, followed by a remote attach. I'll doubt that case works. I'll verify.

paulmaybee: So getting the "solution" does work in the open folder case, however it was not being invoked in the attach code path. Fixed.

src/MICore/LaunchOptions.cs
paulmaybee: If the two cases are CompileTime and Current I think that these are obvious. I could make it even more explicit, e.g. "CurrentLocal", if you thought it was better. Sure, ask.

src/MICore/LaunchOptions.cs
paulmaybee: correct

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
paulmaybee: we do do it for breakpoints

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
paulmaybee: Path combine will give you the separator for the current host OS. We adjust this in the caller to Unix separators if necessary.

src/MICore/CommandFactories/MICommandFactory.cs
gregg-miskelly: @pieandcakes can you check lldb-mi?

src/MICore/LaunchOptions.cs
```diff
@@ -172,6 +172,63 @@ public EnvironmentEntry(Xml.LaunchOptions.EnvironmentEntry xmlEntry)
...
+            {
+                return _editorPath;
+            }
+            set
+            {
+                if (value == null)
+                {
+                    throw new ArgumentNullException("EditorPath");
+                }
+                this._editorPath = value;
```
gregg-miskelly: We might want to reject the empty string also, as currently this will cause us to crash (and clearly it shouldn't be used anyway).

src/MICore/CommandFactories/MICommandFactory.cs
pieandcakes: @gregg-miskelly @paulmaybee lldb does not like the quotes.

pieandcakes:
I'll give it a shot with our tests once it is checked in.

PR #545, opened 2017-02-01
Package the UnixPortSupplier interop assembly

andrewcrawley:
Ensure that the MS.VS.Debugger.Interop.UnixPortSupplier.DesignTime
assembly is packaged into the MDD NuGet package so it can be referenced
from other projects.

### DIFF
--- a/MIEngine.mdd.nuspec
+++ b/MIEngine.mdd.nuspec
@@ -14,6 +14,7 @@
   <files>
     <file src="drop\Release\ReferenceAssemblies\*.xml" target="ref\dotnet" />
     <file src="drop\Release\ReferenceAssemblies\Microsoft.DebugEngineHost.dll" target="ref\dotnet" />
+    <file src="drop\Release\ReferenceAssemblies\Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.DesignTime.dll" target="ref\dotnet" />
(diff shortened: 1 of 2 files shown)


MIEngine.mdd.nuspec
```diff
@@ -14,6 +14,7 @@
   <files>
     <file src="drop\Release\ReferenceAssemblies\*.xml" target="ref\dotnet" />
     <file src="drop\Release\ReferenceAssemblies\Microsoft.DebugEngineHost.dll" target="ref\dotnet" />
+    <file src="drop\Release\ReferenceAssemblies\Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.DesignTime.dll" target="ref\dotnet" />
```
gregg-miskelly: Would it be difficult to make a separate nuget package instead? This guy makes sense to be in its own package so that other things can easily reference it.

MIEngine.mdd.nuspec
```diff
@@ -14,6 +14,7 @@
...
     <file src="drop\Release\Microsoft.MICore.dll" target="ref\dotnet" />
```
gregg-miskelly: We should probably include Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.DesignTime.xml also.

PR #554, opened 2017-02-14
Add non-shell commands to SSH port supplier

andrewcrawley:
Extends the SSH port supplier to allow commands to be run without starting
a shell and allocating a PTY on the remote host.

### DIFF
--- a/src/SSHDebugPS/AD7UnixAsyncCommand.cs
+++ b/src/SSHDebugPS/AD7UnixAsyncCommand.cs
@@ -53,7 +53,7 @@ namespace Microsoft.SSHDebugPS
             {
                 if (_command != null)
                 {
-                    _command.Write(text + "\r");
+                    _command.Write(text + "\n");
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


gregg-miskelly:
>     [Guid("808CC6CA-45B1-47D5-9779-62BAA597BA50")]

We should probably change the GUID

---
Refers to: src/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.cs:18 in e67acae. [](commit_id = e67acaedc5733872e836d8f9491c74b8722fd0c9, deletion_comment = False)

src/SSHDebugPS/AD7UnixAsyncCommand.cs
```diff
@@ -8,180 +8,71 @@
...
-                    return;
-                }
-
-                if (line.EndsWith(_startCommand, StringComparison.Ordinal))
+                if (_command != null)
                 {
-                    // When logged in as root, shell sends a copy of stdin to stdout.
-                    // This ignores the shell command that was used to launch the debugger.
-                    continue;
+                    _command.Write(text + "\r");
```
gregg-miskelly: Is there a reason for `\r` instead of `\n`?

PR #555, opened 2017-02-14
Fix Natvis for DisplayString conditionals in C

pieandcakes:
Issue: gdb returns "1" and "0" instead of "true" and "false" when
debugging C and a condition is passed for evaluation.

Changed code to check for either "true" or "1"

https://github.com/Microsoft/vscode-cpptools/issues/476

@rajkumar42 @gregg-miskelly @chuckries

### DIFF
--- a/src/MIDebugEngine/Natvis.Impl/Natvis.cs
+++ b/src/MIDebugEngine/Natvis.Impl/Natvis.cs
@@ -846,7 +846,11 @@ namespace Microsoft.MIDebugEngine.Natvis
             if (!String.IsNullOrWhiteSpace(condition))
             {
                 string exprValue = GetExpressionValue(condition, variable, scopedNames);
-                res = !String.IsNullOrEmpty(exprValue) && (exprValue.Equals("true", StringComparison.OrdinalIgnoreCase) || exprValue.Equals("1", StringComparison.OrdinalIgnoreCase));
+
+                bool exprBool = false;
+                int exprInt = 0;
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -846,7 +846,7 @@ private bool EvalCondition(string condition, IVariableInformation variable, IDic
             if (!String.IsNullOrWhiteSpace(condition))
             {
                 string exprValue = GetExpressionValue(condition, variable, scopedNames);
-                res = !String.IsNullOrEmpty(exprValue) && (exprValue == "true");
+                res = !String.IsNullOrEmpty(exprValue) && (exprValue.Equals("true", StringComparison.OrdinalIgnoreCase) || exprValue.Equals("1", StringComparison.OrdinalIgnoreCase));
```
paulmaybee: WIll it always be 0 or 1? Maybe the test should be !="0"?

paulmaybee:
👍

src/MIDebugEngine/Natvis.Impl/Natvis.cs
paulmaybee: You would have to check for numeric field contents though.

pieandcakes:
@gregg-miskelly  @paulmaybee Does the rewrite work for you? I changed it based on Paul's suggestion

Issue #144, opened 2016-01-05
Gdb returns stack level of -1 in some cases


Andrew-MSFT:
I suggest we open a bug and do the fix. Treat the -1 as a zero I suppose; maybe just ignore the thread? Will have to try with the sample project to see if we can repro the behavior though.  

(8240) ->1009^done,threads=[{id="10",target-id="Thread 936",frame={level="0",addr="0x40036fc0",func="__ioctl",args=[],from="libc.so"},state="stopped",core="0"},{id="9",target-id="Thread 935",frame={level="0",addr="0x40036fc0",func="__ioctl",args=[],from="libc.so"},state="stopped",core="0"},{id="8",target-id="Thread 931",frame={level="0",addr="0x400380e0",func="__futex_syscall3",args=[],from="libc.so"},state="stopped",core="0"},{id="7",target-id="Thread 930",frame={level="0",addr="0x400380e0",func="__futex_syscall3",args=[],from="libc.so"},state="stopped",core="0"},{id="6",target-id="Thread 929",frame={level="0",addr="0x400380e0",func="__futex_syscall3",args=[],from="libc.so"},state="stopped",core="0"},{id="5",target-id="Thread 928",frame={level="0",addr="0x400380e0",func="__futex_syscall3",args=[],from="libc.so"},state="stopped",core="0"},{id="4",target-id="Thread 927",frame={level="0",addr="0x40037ba8",func="recvmsg",args=[],from="libc.so"},state="stopped",core="0"},{id="3",target-id="Thread 926",frame={level="0",addr="0x400379a0",func="__rt_sigtimedwait",args=[],from="libc.so"},state="stopped",core="0"},{id="2",target-id="Thread 924",frame={level="0",addr="0x400380e0",func="__futex_syscall3",args=[],from="libc.so"},state="stopped",core="0"},{id="1",target-id="Thread 921",frame={level="-1",addr="0x40757610",func="??",args=[]},state="stopped",core="0"}],current-thread-id="1"

suprstefan:
I'm pretty sure I hit this bug when trying to debug an application on a raspberry pi. It has to communicate with the camera and therefore calls `bcm_host_init()` which spawns four threads (according to the output tab) and if I then hit a breakpoint I get an error saying:

> Error while trying to enter break state. Debugging will now stop. Unrecognized format of field "level" in result: {level=-1,addr=0xb6cc902c,func=??,args=[]}

[...]

PR #556, opened 2017-02-16
Fix for Frame's level=-1

pieandcakes:
* When frame's level=-1, make it 0 since it is represented by uint
* Fix spelling error

Fix for #144 

@paulmaybee @gregg-miskelly

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -551,18 +551,18 @@ namespace Microsoft.MIDebugEngine
 
             // On Windows ';' appears to correctly works as a path seperator and from the documentation, it is ':' on unix
             string pathEntrySeperator = _launchOptions.UseUnixSymbolPaths ? ":" : ";";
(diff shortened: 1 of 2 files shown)


gregg-miskelly:
This feels like the wrong fix to me -- we use the frame id to control what frame we use in subsequent evaluations. So if we handle it this way we will wind up switching back to the root frame and issuing evaluations there. Seems like we should instead change this to a 'uint?' and implicitly fail any call that requires frame affinity.

paulmaybee:
Can you reproduce this bug?  I have only seen it occur when the architecture of gdb was not matched with the architecture of gdbserver (for example trying to use an x86 gdb to debug remotely to an arm device).

pieandcakes:
@paulmaybee yes, that is the exact instance I kept running into it.

paulmaybee:
Then this in not the right fix. The right fix is to use the arm gdb.

paulmaybee:
And by that I mean one compiled for x86 but targeting arm.

pieandcakes:
We should probably also behave better by looking at Gregg's suggestion about `uint?` as the option. I will abandon this for now and might take a look again when I have time.

PR #557, opened 2017-02-17
Fix for StopAtEntry when using gdbserver

pieandcakes:
We get a stopped Event when gdb<->gdbserver is trying to setup but then we mark the EntrypointHit so stopAtEntry doesn't occur. 

@paulmaybee @rajkumar42 @gregg-miskelly

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -551,18 +551,18 @@ namespace Microsoft.MIDebugEngine
 
             // On Windows ';' appears to correctly works as a path seperator and from the documentation, it is ':' on unix
             string pathEntrySeperator = _launchOptions.UseUnixSymbolPaths ? ":" : ";";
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -989,7 +989,24 @@ private void Dispose()
...
-                this.EntrypointHit = true;
+                
+                // MinGW sends a stopped event on attach. gdb<->gdbserver also sends a stopped event when first attached.
+                // If this is a gdb<->gdbserver connection, ignore this as the entryPoint
+                if (this._launchOptions is LocalLaunchOptions &&
+                    !String.IsNullOrWhiteSpace(((LocalLaunchOptions)this._launchOptions).MIDebuggerServerAddress))
+                {
+                    // If the stopped event occurs on gdbserver, ignore it unless it contains a filename.
+                    TupleValue frame = results.Results.TryFind<TupleValue>("frame");
+                    if (frame.Contains("file"))
```
gregg-miskelly: I am a little confused by this part since aren't we going to continue from this event no matter what on line 993?

gregg-miskelly:
Oh, and please verify that you can still F5 an Android project and hit a breakpoint.

pieandcakes:
@gregg-miskelly It works!

PR #558, opened 2017-02-23
Fixes for thread-created and terminate/detach

pieandcakes:
1. Fix for Pause on OS X with lldb. LLDB does not break on attach so
threads were not being sent to the UI. Fix was to send back messages when
threads are created directly

2. fix for detach/terminate. With the change to do an internal break, it
was sending -exec-interrupt, -exec-abort and then -exec-continue which was
causing errors. added check in DoInternalBreakActions to only continue if
we are not terminating or detaching.

3. Fix for ErrorEvent. on LLDB, the message format is ^error,message=""
but in gdb it is ^error,msg="". Added a special case for it.

@gregg-miskelly @paulmaybee @rajkumar42

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -384,9 +384,15 @@ namespace Microsoft.MIDebugEngine
             ErrorEvent += delegate (object o, EventArgs args)
             {
                 // NOTE: Exceptions leaked from this method may cause VS to crash, be careful
-
                 ResultEventArgs result = (ResultEventArgs)args;
-                _callback.OnError(result.Results.FindString("msg"));
+                // In lldb, the format is ^error,message=""
+                // In gdb/vsdbg it is ^error,msg=""
+                string message = result.Results.TryFindString("msg");
+                if(String.IsNullOrWhiteSpace(message))
+                {
+                    message = result.Results.TryFindString("message");
+                }
+                _callback.OnError(message);
             };
 
             ThreadCreatedEvent += delegate (object o, EventArgs args)
--- a/src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
@@ -1,12 +1,12 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using MICore;
 using System;
 using System.Collections.Generic;
-using System.Text;
-using System.Threading.Tasks;
-using MICore;
 using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
 
 namespace Microsoft.MIDebugEngine
 {
@@ -45,6 +45,30 @@ namespace Microsoft.MIDebugEngine
         private Dictionary<string, List<int>> _threadGroups;
         private static uint s_targetId;
         private const string c_defaultGroupId = "i1";  // gdb's default group id, also used for any process without group ids
+
+        private List<DebuggedThread> DeadThreads
+        {
+            get
+            {
+                if (_deadThreads == null)
+                {
+                    _deadThreads = new List<DebuggedThread>();
+                }
+                return _deadThreads;
+            }
+        }
+
+        private List<DebuggedThread> NewThreads
+        {
+            get
+            {
+                if (_newThreads == null)
+                {
+                    _newThreads = new List<DebuggedThread>();
+                }
+                return _newThreads;
+            }
+        }
 
         static ThreadCache()
         {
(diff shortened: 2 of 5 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -384,9 +384,15 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
             ErrorEvent += delegate (object o, EventArgs args)
             {
                 // NOTE: Exceptions leaked from this method may cause VS to crash, be careful
-
                 ResultEventArgs result = (ResultEventArgs)args;
-                _callback.OnError(result.Results.FindString("msg"));
+                string msgString = "msg";
+                // In lldb, the format is ^error,message=""
+                // In gdb/vsdbg it is ^error,msg=""
+                if (_launchOptions.DebuggerMIMode == MIMode.Lldb)
```
gregg-miskelly: Since this is a bug in LLDB that they might eventually fix, I might suggest instead Trying to fix 'msg', and if that property doesn't exist look for 'message'

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -151,9 +151,11 @@ internal void ThreadCreatedEvent(int id, string groupId)
         {
             lock (_threadList)
             {
-                var thread = _threadList.Find(t => t.Id == id);
-                if (thread == null)
+                bool bNew;
+                var thread = FindThread(id, out bNew);
+                if (bNew)
                 {
+                    _callback.OnThreadStart(thread);
```
gregg-miskelly: The one problem I can see with this is that in CollectThreadsInfo we will not just have the id, but we will also have the 'TargetId'. It looks like the 'TargetId' is what we expose to AD7. Here, we only have the Id. AD7 is not going to be happy if a thread's ID changes.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -384,9 +384,15 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
             ErrorEvent += delegate (object o, EventArgs args)
             {
                 // NOTE: Exceptions leaked from this method may cause VS to crash, be careful
-
                 ResultEventArgs result = (ResultEventArgs)args;
-                _callback.OnError(result.Results.FindString("msg"));
+                string msgString = "msg";
```
paulmaybee: I suggest TryFind "msg" first and then for "message" (if LLDB) so that this doesn't fail in case they ever "fix" the issue on the LLDB side.

pieandcakes: I'll do that

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
pieandcakes: I'll do that

pieandcakes: I'll do that

pieandcakes:
@gregg-miskelly @paulmaybee I refactored the thread info block and am now reporting thread created and thread exited events. when we get them.

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -147,15 +171,35 @@ internal void MarkDirty()
             }
         }
 
-        internal void ThreadCreatedEvent(int id, string groupId)
+        internal async void ThreadCreatedEvent(int id, string groupId)
         {
-            lock (_threadList)
+            ResultValue resVal = null;
```
gregg-miskelly: You need a try/catch around this method now, as async void methods can never let exceptions leave

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -46,6 +46,30 @@ internal class ThreadCache
         private static uint s_targetId;
         private const string c_defaultGroupId = "i1";  // gdb's default group id, also used for any process without group ids
 
+        private List<DebuggedThread> deadThreads
```
gregg-miskelly: >d [](start = 37, length = 1)

nit: the convention in this code base is to capitalize the first letter of properties.

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -231,6 +301,72 @@ private ThreadContext CreateContext(TupleValue frame)
...
+        }
+
+        private DebuggedThread SetThreadInfoFromResultValue(ResultValue resVal, out bool newThread)
+        {
+            newThread = false;
+            int threadId = resVal.FindInt("id");
+            string targetId = resVal.TryFindString("target-id");
+
+            DebuggedThread thread = FindThread(threadId, out newThread);
+            thread.Alive = true;
```
gregg-miskelly: >thread.Alive = true; [](start = 12, length = 20)

Should this really be in CollectThreadInfo?

pieandcakes: @gregg-miskelly This was there already. I just moved it. I think its there because if thread info lists it, it is still alive.

pieandcakes: @gregg-miskelly This was there already. I just moved it. I think its there because if thread info lists it, it is still alive.

pieandcakes:
@gregg-miskelly Verified on android that everything still works.

PR #561, opened 2017-03-01
Fix for threading issue

pieandcakes:
* Issue was timing in that the new thread was being populated in MIEngine
* without the id being added to the threadgroup. Moved the relevant block.
* Also blocked it from requesting thread-info from clrdbg since this
* doesn't work while running in clrdbg

@gregg-miskelly @rajkumar42 @chuckries @caslan

### DIFF
--- a/src/MICore/CommandFactories/gdb.cs
+++ b/src/MICore/CommandFactories/gdb.cs
@@ -155,6 +155,7 @@ namespace MICore
                 _currentFrameLevel = frameLevel;
             }
         }
+
         public override async Task<Results> ThreadInfo(uint? threadId = null)
         {
             Results results = await base.ThreadInfo(threadId);
@@ -164,6 +165,7 @@ namespace MICore
             }
             return results;
         }
+
         public override async Task<List<ulong>> StartAddressesForLine(string file, uint line)
         {
             string cmd = "info line " + file + ":" + line;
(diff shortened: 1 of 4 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -388,17 +386,17 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
...
+                if (String.IsNullOrWhiteSpace(message))
                 {
                     message = result.Results.TryFindString("message");
                 }
                 _callback.OnError(message);
             };
 
-            ThreadCreatedEvent += delegate (object o, EventArgs args)
+            ThreadCreatedEvent += async delegate (object o, EventArgs args)
             {
```
gregg-miskelly: >{ [](start = 12, length = 1)

This method is now async void, so it needs a try/catch

src/MICore/CommandFactories/gdb.cs
```diff
@@ -155,6 +155,7 @@ public override bool UseExternalConsoleForLocalLaunch(LocalLaunchOptions localLa
                 _currentFrameLevel = frameLevel;
             }
         }
+
```
gregg-miskelly: Might want to undo your changes to the files with whitespace-only diffs

pieandcakes: @gregg-miskelly Was fixing missing newlines. Do you want me to remove them?

PR #571, opened 2017-03-28
2015 linux extension: merge with master

paulmaybee:
Add one additional commit which causes this build to target dev14 - liblinux.IL will reference version 14 now rather than version 15.


gregg-miskelly:
I am a little late to the party, but looks reasonable to me.

Issue #574, opened 2017-04-05
ExclusiveCmdAsync stuck on "info sharedlibrary"


menting311:
@gregg-miskelly 
When attaching to a running process through VS 17 it would always hang at some point in the process. So I  cloned this project and but some breakpoints in to see where it is hanging. Turns out that the command "info sharedlibrary" is getting hung up, if I but some code in Debugger.cs to return "" when this command is going to run it no longer hangs.

I am also able to duplicate this randomly if I run the same command on the command line when attaching to a pid.
This is what I run:

 gdb --interpreter=mi
 Then from (gdb) prompt (I pulled these from the output of the MIEngine debug session):
1001-gdb-set target-async on
1002-gdb-set stop-on-solib-events 1
1003-interpreter-exec console "shell readlink -f /proc/1716/exe”
1004-file-exec-and-symbols /code/tmenting/dev/wms/src/my_pg
1005-interpreter-exec console "show architecture"
1006-target-attach 1716
1007-thread-info 1
1008-break-insert -f hubmgrmenu.cpp:290
1009-symbol-list-lines /code/tmenting/dev/wms/src/my_pg.cpp
1010-break-insert -f my_pgm.cpp:143
1011-interpreter-exec console "info sharedlibrary"

Sometimes I get the full list and that returns me to the (gdb) prompt. However, the problem is that sometimes I get part of the output displayed to screen, and it waits for me to press enter to display the rest. It is almost like it is being piped to "less" and I have to enter to continue

How can we work around this? I am evaluating VS17 to see if it works for our environment. I would like to get this figured out soon before my trial periods start running out.

I have more data I could send privately if you are willing. Thanks!

paulmaybee:
I think you can turn off the pagination and that should keep gdb from asking for the RET.

From the command window enter
                Debug.MIDebugExec set pagination off

Or you can add the same text to your project properties
                Properties->Debugger->Additional Debugger Commands = set pagination off

We should be starting gdb with the –batch option which makes this command unnecessary. I will see that it gets fixed.

paulmaybee:
Disregard the “-batch” comment. That’s not the correct fix on our side.

From: Paul Maybee
Sent: Thursday, April 6, 2017 3:20 PM
To: 'Microsoft/MIEngine' <reply@reply.github.com>; Microsoft/MIEngine <MIEngine@noreply.github.com>
Cc: Subscribed <subscribed@noreply.github.com>
[...]

PR #577, opened 2017-04-07
Turn off pagination (Issue #574)

paulmaybee:
Long output from commands (more that a screen full) causes gdb to pause waiting for user input. This change turns off that behavior. @gregg-miskelly @rajkumar42

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -553,7 +553,10 @@ namespace Microsoft.MIDebugEngine
 
             commands.AddRange(_launchOptions.SetupCommands);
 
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -553,6 +553,8 @@ private List<LaunchCommand> GetInitializeCommands()
 
             commands.AddRange(_launchOptions.SetupCommands);
 
+            commands.Add(new LaunchCommand("-interpreter-exec console \"set pagination off\""));
```
gregg-miskelly: Should this be GDB-only?

gregg-miskelly: Should we ignore failures?

rajkumar42:
👍

PR #579, opened 2017-04-10
Don't apply natvis visualizations to pretty-printed ("dynamic") results

paulmaybee:
When both pretty-printing and natvis are enabled (the new default for Linux projects) then the combination of features seems to give best results. In particular, there is no pretty printing for std::list, but nativs will format a list just fine. With this fix a list\<string\> will be formatted optimally, using natvis to decode the list and allowing the string to be pretty-printed.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -39,6 +39,7 @@ namespace Microsoft.MIDebugEngine
         bool IsVisualized { get; }
         bool IsReadOnly { get; }
         enum_DEBUGPROP_INFO_FLAGS PropertyInfoFlags { get; set; }
+        bool IsPreformatted { get; set; }
     }
 
     internal class SimpleVariableInformation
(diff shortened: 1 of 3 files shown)


src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -675,6 +683,7 @@ private bool IsMapType()
             return DisplayHint == "map";
         }
 
+        [DebuggerHidden()]
```
gregg-miskelly: Is this just to keep the debugger from func-evaling the property? DebuggerHidden is a pretty big hammer. This code will no longer be debuggable, even with JMC off. I would suggest changing this to be a method instead of a property, as it looks like it really shouldn't be a property. If that isn't possible, I would suggest using Debugger.NotifyOfCrossThreadDependancy in the 'not-yet-fetched' case.

paulmaybee: changed IsReadonly into a method

PR #582, opened 2017-04-19
fix: Debugger was not stopping on unkown breakpoints. Manually entere…

paulmaybee:
…d breaks/catches were not working.


paulmaybee:
User reported the problem with "catch throw". See https://vsfeedback/comment/853185

gregg-miskelly:
Sorry Paul, I noticed I messed this.

Issue #583, opened 2017-05-02
Parsing of gdb output is inefficient


paulmaybee:
User reports that placing a breakpoint in a file with 20K lines is taking about 2 minutes. Analysis of the MIEngine log points to the parsing of the large output string being returned by gdb from a request for the line number mapping from that file.

paulmaybee:
Fixed in PR #584

PR #584, opened 2017-05-02
Speed up parsing of gdb output strings by using indexes into the orig…

paulmaybee:
…inal string rather than creating new strings at each step.

This results in a huge speedup. In the example reported (Issue #583) the time to parse when from 106 seconds to .1 second.

### DIFF
--- a/src/MICore/MIResults.cs
+++ b/src/MICore/MIResults.cs
@@ -740,7 +740,6 @@ namespace MICore
 
         public string ParseCString(string input)
         {
-            input = input.Trim();
             if (input == null)
             {
                 throw new ArgumentNullException("input");
--- a/src/MICore/MIResults.cs
+++ b/src/MICore/MIResults.cs
@@ -659,9 +659,9 @@ namespace MICore
             public int IndexOf(string theString, char c)
             {
                 int i = theString.IndexOf(c, Start);
-                if (i < 0)
+                if (i < 0 || i >= Extent)
                 {
-                    return i;
+                    return -1;
                 }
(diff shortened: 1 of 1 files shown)


gregg-miskelly:
@paulmaybee wow!

src/MICore/MIResults.cs
```diff
@@ -596,6 +596,93 @@ public override string ToString()
...
+                    if (theString[Start + i] != pattern[i])
+                    {
+                        return false;
+                    }
+                }
+                return true;
+            }
+        }
+
+        private string theResult;
```
gregg-miskelly: nit: MIEngine uses '_' for private members. How about '_resultString'?

src/MICore/MIResults.cs
```diff
@@ -596,6 +596,93 @@ public override string ToString()
...
+                return theString.Substring(Start, Length);
+            }
+            public int IndexOf(string theString, char c)
+            {
+                int i = theString.IndexOf(c, Start);
+                if (i < 0)
+                {
+                    return i;
+                }
+                return i - Start;   // Span relative offset
```
gregg-miskelly: Not sure if this will actually matter based on how you are using this class, but there should really should also be:

```csharp
if (i >= this.Extent)
{
    return -1;
}
```

paulmaybee: fixed

src/MICore/MIResults.cs
paulmaybee: fixed

PR #585, opened 2017-05-03
Update package references

gregg-miskelly:
The MIEngine was still compiling against incredibly ancient packages. This switches it over to something recent - the packages that came out at the same time as .NET Core 1.0 RTW.

The primary IL level difference of this change is to increment the reference assembly version of the following in the non-'Desktop' configurations.

    System.Xml.XmlSerializer
    System.IO.FileSystem
    System.Net.Http
    System.IO.FileSystem.Primitives

**Additional changes**:
* Fixes the 'Desktop' configurations to stop dropping a bunch of facade assemblies that shouldn't really be needed.
* Stops having the IOS launcher need to partially depend on facade assemblies. To do this I added an MIEngine definition of RemoteCertificateValidationCallback.

**Testing**: Verified Android debugging in VS 2015 and VS 2017. Both F5'ing the debug configuration in the experimental instance and installing the lab configuration over top our shipping bits.

### DIFF
--- a/src/MICore/project.json
+++ b/src/MICore/project.json
@@ -1,25 +1,42 @@
 {
-  "supports": {
-    "net46.app": {},
-    "dnxcore50.app": {}
-  },
   "dependencies": {
-    "MicroBuild.Core": "0.2.0",
-    "Microsoft.NETCore": "5.0.0",
-    "Microsoft.NETCore.Portable.Compatibility": "1.0.0",
-    "System.Diagnostics.Process": "4.0.0-beta-23225",
-    "System.Net.Security": "4.0.0-beta-23225",
-    "System.Net.Sockets": "4.1.0-beta-23225",
-    "System.Reflection.TypeExtensions": "4.0.0",
-    "System.Security.Cryptography.X509Certificates": "4.0.0-beta-23225",
-    "System.Threading.Thread": "4.0.0-beta-23225",
-    "System.Xml.XmlSerializer": "4.0.11-beta-23225",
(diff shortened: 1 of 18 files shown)


src/MICore/project.json
```diff
@@ -1,25 +1,42 @@
...
-    "System.Diagnostics.Process": "4.0.0-beta-23225",
-    "System.Net.Security": "4.0.0-beta-23225",
-    "System.Net.Sockets": "4.1.0-beta-23225",
-    "System.Reflection.TypeExtensions": "4.0.0",
-    "System.Security.Cryptography.X509Certificates": "4.0.0-beta-23225",
-    "System.Threading.Thread": "4.0.0-beta-23225",
-    "System.Xml.XmlSerializer": "4.0.11-beta-23225",
-    "System.Runtime.InteropServices.RuntimeInformation": "4.0.0-beta-23225",
-    "System.IO.FileSystem.Watcher": "4.0.0-beta-23225"
+    "Microsoft.VisualStudio.Debugger.Interop.Portable": "1.0.1",
```
pieandcakes: was this needed? I don't see this in the old file?

pieandcakes: Or did you just make most of the files the same so that it was easier to maintain?

gregg-miskelly: I made the files mostly the same to make them easier to maintain. Though I can try taking that one out.

---
In reply to: [114674680](https://github.com/Microsoft/MIEngine/pull/585#discussion_r114674680) [](ancestors = 114674680)

gregg-miskelly: We actually seem to need this package in MICore, so I am not sure why it worked before without it.

PR #587, opened 2017-05-05
Add ability to parse json as well as XML for launchoptions

pieandcakes:


src/MICore/JsonLaunchOptions.cs
```diff
@@ -0,0 +1,392 @@
+﻿// © Microsoft. All rights reserved.
+
+using System;
```
pieandcakes: This is a generated file that I had to hand tweak to get it to work right (removing some items, adding others, etc.

src/MICore/JsonLaunchOptions.cs
```diff
@@ -0,0 +1,392 @@
...
+        /// </summary>
+        [JsonProperty("filterStdout", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public bool? FilterStdout { get; set; }
+
+        /// <summary>
+        /// Search stderr stream for server-started pattern and log stderr to debug output. Defaults to false.
+        /// </summary>
+        [JsonProperty("filterStderr", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public bool? FilterStderr { get; set; }
+
```
paulmaybee: The filterStdout and filterStderr options are related to LaunchOptions.serverStarted and should be placed there, not in BaseOptions

pieandcakes: @paulmaybee Are you saying this is only an option on Launch and not Attach? In our base VSCode json we have the filter in both.

gregg-miskelly:
@pieandcakes in case you didn't notice - the build failed

src/MICore/JsonLaunchOptions.cs
paulmaybee: I am saying that these options are ALWAYS used in conjunction with serverStarted (and also debugServerPath and debugServerArgs) and don't make sense in any other context. I am not sure whether there are any attach scenarios that would use all of these, but I would not rule it out.

paulmaybee: You could improve the whole experience and group them together in a new ServerOptions class.

src/MICore/project.json
```diff
@@ -32,7 +34,8 @@
     "System.Threading.Thread": "4.0.0",
     "System.Threading.Timer": "4.0.1",
     "System.Xml.ReaderWriter": "4.0.11",
-    "System.Xml.XmlSerializer": "4.0.11"
+    "System.Xml.XmlSerializer": "4.0.11",
+    "Newtonsoft.Json": "9.0.1"
```
gregg-miskelly: >Newtonsoft.Json [](start = 5, length = 15)

The version of Newtonsoft.Json.dll in devenv has an assembly version of 8.0.0.0 for VS 2017 and 6.0.0.0 for VS 2015. So you need to either compile again an older version of this, or update Install.cmd + targets now, and SetupPackages when you checkin to ship yet-another-copy of this dll.

gregg-miskelly:
> // © Microsoft. All rights reserved.

You need:


---
Refers to: src/MICore/JsonLaunchOptions.cs:1 in 2b19c08. [](commit_id = 2b19c082aeb7f7fecc572e2a6b30e901834169e4, deletion_comment = False)

src/MICore/JsonLaunchOptions.cs
```diff
@@ -0,0 +1,392 @@
...
+        [JsonProperty("pipeTransport", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public PipeTransport PipeTransport { get; set; }
+    }
+
+    public partial class AttachOptions : BaseOptions
+    {
+        #region Public Properties for Serialization
+        
+        [JsonProperty("processId")]
+        public Dictionary<string, object> ProcessId { get; private set; }
```
gregg-miskelly: >Dictionary [](start = 15, length = 10)

Why is this a Dictionary instead of an object? or an int?

pieandcakes: Because it was generated by the tool that way. I can change it

src/MICore/project.json
pieandcakes: @gregg-miskelly Are you saying I should just use 6.0.0.0?

src/MICore/JsonLaunchOptions.cs
pieandcakes: @paulmaybee I moved it to LaunchOptions. I don't want to change the schema at this point since the current schema is what we have in the VSCode experience and I think our goal right now is to enable a similar scenario. 

When is the debugserver used? this isn't for the gdbserver connection as I think that one uses the miDebugServerAddress option.

chuckries:
:shipit:

src/MICore/JsonLaunchOptions.cs
paulmaybee: It is used to start a local process that "acts" as a gdbserver when the target host is not capable of being debugged via ssh and gdbserver. The typical case is when using OpenOCD to debug an IoT device.

src/MICore/project.json
chuckries: Does this mean we want to compile against 8.0.3?

gregg-miskelly: I don't know which one you should pick, but here are the options:
1. Compile against a 6.0.x version. Looks like the latest is 6.0.8. You would likely want to compile against the 'portable-net45+wp80+win8+wpa81+aspnetcore50' version in this case. In this case you don't need any more work assuming this version actually has everything you need.
2. Compile against a 8.0.x version. Looks like the latest is 8.0.3. In this case you don't need to do anything for VS 2017, but you should fix up Install.cmd the same way we do for System.IO.FileSystem.Primitives.dll.
3. Keep compiling against this version, but update the various setup things for the new dependency.

If it was me, I would start with 1. If that works - great! If not, keep going till I found something that actually works.

gregg-miskelly: Looking at https://github.com/JamesNK/Newtonsoft.Json/releases I didn't see anything that blocks us from compiling against the older version. But I could be missing something.

src/MICore/MICore.csproj
```diff
@@ -102,10 +102,14 @@
   </ItemGroup>
   <ItemGroup Condition="'$(IsCoreClr)' == 'false'">
     <Reference Include="System" />
-    <Reference Include="System.Xml" />
     <Reference Include="System.Net.Http" />
+    <Reference Include="System.Runtime.Serialization" />
+    <Reference Include="System.Xml" />
     <Reference Include="Microsoft.VisualStudio.Debugger.InteropA">
-      <HintPath>$(NuGetPackagesDirectory)/Microsoft.VisualStudio.Debugger.Interop.Portable.1.0.1/lib/portable-net45+net46+dnxcore50/Microsoft.VisualStudio.Debugger.InteropA.dll</HintPath>
+      <HintPath>Microsoft.VisualStudio.Debugger.Interop.Portable.1.0.1/lib/portable-net45+net46+dnxcore50/Microsoft.VisualStudio.Debugger.InteropA.dll</HintPath>
```
gregg-miskelly: >Microsoft [](start = 16, length = 9)

Did you mean to remove `$(NuGetPackagesDirectory)/`?

src/MICore/LaunchOptions.cs
```diff
@@ -1032,94 +1190,108 @@ public static LaunchOptions GetInstance(HostConfigurationStore configStore, stri
             object launcher = null;
             object launcherXmlOptions = null;
 
-            try
+            if (options.Trim()[0] == '{')
             {
-                XmlSerializer serializer;
-                using (XmlReader reader = OpenXml(options))
+                launchOptions = LocalLaunchOptions.CreateFromJson(options);
```
gregg-miskelly: >launchOptions [](start = 16, length = 13)

Do you need a try/catch here like we have in the XML case?

src/MICore/LaunchOptions.cs
```diff
@@ -1032,94 +1190,108 @@ public static LaunchOptions GetInstance(HostConfigurationStore configStore, stri
             object launcher = null;
             object launcherXmlOptions = null;
 
-            try
+            if (options.Trim()[0] == '{')
```
gregg-miskelly: >(options.Trim() [](start = 15, length = 15)

Might want to move this up so you don't Trim it twice. Ex: change line 1183 to `options = options?.Trim(); if (string.IsNullOrEmpty(options))`

src/MICore/JsonLaunchOptions.cs
```diff
@@ -0,0 +1,392 @@
...
+        /// Full path to program executable.
+        /// </summary>
+        [JsonProperty("program")]
+        public string Program { get; set; }
+
+        /// <summary>
+        /// The type of the engine. Must be "cppdbg".
+        /// </summary>
+        [JsonProperty("type", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public string Type { get; set; }
```
gregg-miskelly: You might want to remove this since it doesn't look like it is actually used.

src/MICore/LaunchOptions.cs
```diff
@@ -310,54 +341,85 @@ public bool IsValidMiDebuggerPath()
             return File.Exists(MIDebuggerPath);
         }
 
-        static internal LocalLaunchOptions CreateFromXml(Xml.LaunchOptions.LocalLaunchOptions source)
+        static internal LocalLaunchOptions CreateFromJson(string json)
         {
```
gregg-miskelly: >{ [](start = 8, length = 1)

You don't need to do it now (though it might make it easier to test, so maybe you want to), but I would suggest adding some unit tests for this stuff. It has been helpful in the XML case.

src/MICore/LaunchOptions.cs
```diff
@@ -1350,6 +1522,28 @@ private IEnumerable<string> GetSOLibSearchPathCandidates()
...
+            }
+
+            this.VisualizerFile = options.VisualizerFile;
+            this.ShowDisplayString = options.ShowDisplayString.GetValueOrDefault(false);
+
+            this.AdditionalSOLibSearchPath = String.IsNullOrEmpty(this.AdditionalSOLibSearchPath) ?
+                options.AdditionalSOLibSearchPath :
+                String.Concat(this.AdditionalSOLibSearchPath, ";", options.AdditionalSOLibSearchPath);
+
+            // This was set in OpenDebugAD7 for !CORECLR
```
gregg-miskelly: >This [](start = 15, length = 4)

Might be worth adding a TODO to have a launch option for this. It didn't make sense in VS Code since we don't have any UI in VS Code. But it does make sense in VS.

src/MICore/MICore.csproj
pieandcakes: No i'll add it back

src/MICore/JsonLaunchOptions.cs
pieandcakes: @gregg-miskelly I added validation for this as we're going to require cppdbg to be the type.

src/MICore/project.json
pieandcakes: I went ahead and changed it to 6.0.8

src/MICore/LaunchOptions.cs
pieandcakes: @gregg-miskelly added a line in the task table

pieandcakes:
Verified that its loading the correct Newtonsoft.Json against VS 2015.

PR #592, opened 2017-05-10
Sign files from drop dir

gregg-miskelly:
This is an attempt to fix our signing issue by signing files from the drop directory instead of signing before DropFiles runs.


pieandcakes:
:shipit:

PR #597, opened 2017-05-12
Fix where Glass is copying Newtonsoft.json from

pieandcakes:


src/MICore/MICore.csproj
```diff
@@ -199,7 +200,8 @@
   <ItemGroup>
     <GlassDirCopy Include="$(OutDir)$(AssemblyName)$(TargetExt)" />
     <GlassDirCopy Include="$(OutDir)$(AssemblyName).XmlSerializers.dll" Condition="'$(Configuration)'!='Debug-PortablePDB'" />
-    <GlassDirCopy Include="$(OutDir)Newtonsoft.Json.dll" />
+	<GlassDirCopy Include="$(NuGetPackagesDirectory)\Newtonsoft.Json\6.0.8\lib\portable-net45+wp80+win8+wpa81+aspnetcore50\Newtonsoft.Json.dll" Condition="'$(IsCoreClr)' == 'true'" />
```
rajkumar42: Do you need `Condition="'$(IsCoreClr)' == 'xyz'"`? 
Shouldn't this just copy the binary if
`<GlassDirCopy Include="$(NuGetPackagesDirectory)\Newtonsoft.Json\6.0.8\lib\portable-net45+wp80+win8+wpa81+aspnetcore50\Newtonsoft.Json.dll" /> ` and remove the duplicate.

gregg-miskelly: There is a one character difference between the two lines -- `\` vs `.` separating the version from the package name.

rajkumar42: Got it!!!

PR #600, opened 2017-05-17
ClrDbg is obselete.

rajkumar42:


rajkumar42:
@gregg-miskelly @caslan please review.

PR #601, opened 2017-05-18
Fix dragging cursor when changing the next line to execute. This was …

paulmaybee:
…broken when line mapping was added.

See https://devdiv.visualstudio.com/defaultcollection/DevDiv/_workitems?id=427193&triage=true&_a=edit

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -1906,10 +1906,20 @@ namespace Microsoft.MIDebugEngine
         public async Task<List<ulong>> StartAddressesForLine(string file, uint line)
         {
             List<ulong> addresses = new List<ulong>();
-            SourceLineMap srcLines = await SourceLineCache.GetLinesForFile(file);
-            if (srcLines == null || srcLines.Count == 0)
+            string compFile;
+            SourceLineMap srcLines;
+            if (MapCurrentSrcToCompileTimeSrc(file, out compFile))  // found a remote mapping for this source file
             {
-                srcLines = await SourceLineCache.GetLinesForFile(System.IO.Path.GetFileName(file));
+                file = compFile;
+                srcLines = await SourceLineCache.GetLinesForFile(file);
+            }
+            else
+            {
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1967,13 +1977,14 @@ public bool MapCurrentSrcToCompileTimeSrc(string currentSrc, out string compiler
...
                         {
-                            file = file.Substring(1);   // Trim the directory separator
+                            file = file.Trim(new char[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar });   // Trim the directory separator(s)
                         }
                         else if (lastDirectoryChar != Path.DirectorySeparatorChar && lastDirectoryChar != Path.AltDirectorySeparatorChar)
                         {
                             continue;   // match didn't end at a directory separator, not actually a match
                         }
                         compilerSrc = Path.Combine(e.CompileTimePath, file);    // map to the compiled location
+                        compilerSrc = compilerSrc.Replace('\\', '/'); // use Unix notation for the compiled path
```
gregg-miskelly: Do we still want Unix slashes in Android or MinGW scenarios?

paulmaybee: Neither of those are remote compile scenarios. But I suppose someone could use this for moved source too. How about I just look for '\\' in the compileTimePath and don't do this step if found?

gregg-miskelly: In other places in MIEngine we check `_launchOptions.UseUnixSymbolPaths`. But your solution sounds reasonable to me also.

paulmaybee: UseUnixSymbolPaths is set to false when using local launch options on windows. When using gdbserver it will not indicate that we still need Unix style paths because we remotely compiled.

gregg-miskelly: Ah right. Makes sense then.

PR #602, opened 2017-05-19
Fix disassembly near start/end of valid address spaces

paulmaybee:
Result of this bug would be 
1. Goto disassembly leaves cursor at top of disassembly page with no way to see previous instructions.
2. Yellow arrow would sometimes not show up in disassembly when near edge of address space

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
@@ -134,10 +134,16 @@ namespace Microsoft.MIDebugEngine
             if (iInstructions != 0)
             {
                 IEnumerable<DisasmInstruction> instructions = null;
-                _engine.DebuggedProcess.WorkerThread.RunOperation(async () =>
+                // When failing to read all instructions try looking for smaller and smaller numbers of instructions
+                // so that fetching disassembly works near the start and end of valid address spaces
+                while (instructions == null && (iInstructions > 1 || -iInstructions > 1)) 
                 {
-                    instructions = await _engine.DebuggedProcess.Disassembly.FetchInstructions(_addr, (int)iInstructions);
-                });
+                    _engine.DebuggedProcess.WorkerThread.RunOperation(async () =>
+                    {
+                        instructions = await _engine.DebuggedProcess.Disassembly.FetchInstructions(_addr, (int)iInstructions);
+                    });
+                    iInstructions = iInstructions/2;
+                }
                 if (instructions == null)
                 {
                     return Constants.E_FAIL;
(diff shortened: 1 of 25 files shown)


src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -134,10 +134,16 @@ public int Seek(enum_SEEK_START dwSeekStart, IDebugCodeContext2 pCodeContext, ul
             if (iInstructions != 0)
             {
                 IEnumerable<DisasmInstruction> instructions = null;
-                _engine.DebuggedProcess.WorkerThread.RunOperation(async () =>
+                // When failing to read all instructions try looking for smaller and smaller numbers of instructions
+                // so that fetching disassembly works near the start and end of valid address spaces
+                while (instructions == null && (iInstructions > 1 || -iInstructions > 1))
```
gregg-miskelly: nit: might be easier to read if the check was `iInstructions != 0` instead of `(iInstructions > 1 || -iInstructions > 1)` unless there is a difference between those two checks that I am not seeing.

gregg-miskelly: This would be a bit more efficient if the while loop was in the delegate instead of here. Otherwise it will need to thread transition back and forth.

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -134,10 +134,16 @@ public int Seek(enum_SEEK_START dwSeekStart, IDebugCodeContext2 pCodeContext, ul
...
+                // so that fetching disassembly works near the start and end of valid address spaces
+                while (instructions == null && (iInstructions > 1 || -iInstructions > 1)) 
                 {
-                    instructions = await _engine.DebuggedProcess.Disassembly.FetchInstructions(_addr, (int)iInstructions);
-                });
+                    _engine.DebuggedProcess.WorkerThread.RunOperation(async () =>
+                    {
+                        instructions = await _engine.DebuggedProcess.Disassembly.FetchInstructions(_addr, (int)iInstructions);
+                    });
+                    iInstructions = iInstructions/2;
```
gregg-miskelly: Two questions: 
1. Suppose there are 19 valid instructions before the page offset and we ask for 20 initially. Now we will only wind up only asking for 10. Isn't it a problem that we didn't get those extra 9 valid instructions?
2. For the invalid instructions, don't we want to fill the disassembly window with '??' lines so that it will still scroll. If so, shouldn't we take the difference between what was asked for and the number of valid instructions and seek 1-byte for each invalid?

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
paulmaybee: Well, != 0 or 1, but just 0 will work too.

PR #604, opened 2017-05-25
Fixing Nuget packages path

pieandcakes:
This fixes ordering so that $(NuGetPackagesDirectory) is specified and we can use it to find the MicroBuild props file. This is an issue because ..\packages\ dir exists on CoreCLR builds but not the Desktop.* builds. For Desktop.* builds we put the NuGet files in packages.Desktop. This was causing our release build to fail.

I have verified that this change by queuing up a release and a nightly run of builds against this branch and they both succeeded.

### DIFF
--- a/BuildAndTest.proj
+++ b/BuildAndTest.proj
@@ -32,14 +32,14 @@
 
   <Target Name="Build">
     <MSBuild Targets="Build"
-	         Projects="src\prebuild\prebuild.csproj" />
+             Projects="src\prebuild\prebuild.csproj" />
     <MSBuild Targets="Build"
              Projects="@(Solution)" />
   </Target>
 
   <Target Name="Rebuild">
     <MSBuild Targets="Rebuild"
-	         Projects="src\prebuild\prebuild.csproj" />
+             Projects="src\prebuild\prebuild.csproj" />
     <MSBuild Targets="Rebuild"
              Projects="@(Solution)" />
   </Target>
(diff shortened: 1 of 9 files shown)


pieandcakes:
@gregg-miskelly @chuckries @rajkumar42 I got it to succeed both the release and the nightly builds now. Please review.

BuildAndTest.proj
```diff
@@ -32,11 +32,15 @@
 
   <Target Name="Build">
     <MSBuild Targets="Build"
+	         Projects="src\prebuild\prebuild.csproj" />
```
gregg-miskelly: nit: tab instead of spaces

src/DebugEngineHost.Stub/DebugEngineHost.Stub.csproj
```diff
@@ -5,9 +5,10 @@
     NOTE: Ths must be set BEFORE improting miengine.settings.targets-->
     <AssemblyVersion>1.0.0</AssemblyVersion>
   </PropertyGroup>
-  <Import Project="..\packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.props" Condition="Exists('..\packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.props')" />
-  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+    <!-- miengine.settings.targets must be imported first as it sets $(NuGetPackagesDirectory) -->
```
gregg-miskelly: nit: this should be two space indented instead of four to match with the rest of the file.

PR #608, opened 2017-05-31
copy loc files to drop location

pieandcakes:
Copy the loc files into an LCID folder in the drop location.

### DIFF
--- a/src/MIDebugPackage/MIDebugPackage.csproj
+++ b/src/MIDebugPackage/MIDebugPackage.csproj
@@ -215,6 +215,59 @@
       <TranslationFile>$(MIEngineRoot)loc\LCL\{Lang}\lib\OpenFolderSchema.json.lcl</TranslationFile>
     </FilesToLocalize>
   </ItemGroup>
+  <Target Name="DropLocalizeDir" AfterTargets="Localize;DropFiles" Condition="'$(LocalizationEnabled)'=='true'">
+    <PropertyGroup>
+      <LocalizedDropDir>$(DropRootDir)\loc\</LocalizedDropDir>
+      <LocalizeBinDir>$(OutDir)localize\</LocalizeBinDir>
+    </PropertyGroup>
+    <ItemGroup>
+      <Locale Include="$(LocalizeBinDir)ENU\*.json">
+        <LCID>1033</LCID>
+      </Locale>
+      <Locale Include="$(LocalizeBinDir)CSY\*.json">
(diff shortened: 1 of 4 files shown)


src/MIDebugPackage/MIDebugPackage.csproj
```diff
@@ -215,6 +215,59 @@
...
+        <LCID>1055</LCID>
+      </Locale>
+      <Locale Include="$(LocalizeBinDir)CHS\*.json">
+        <LCID>2052</LCID>
+      </Locale>
+      <Locale Include="$(LocalizeBinDir)CHT\*.json">
+        <LCID>1028</LCID>
+      </Locale>
+    </ItemGroup>
+    <MakeDir Condition="!Exists($(LocalizedDropDir))" Directories="$(LocalizedDropDir)" />
```
gregg-miskelly: Do we need to create the various subdirectories also? Or will Copy do that for us? If Copy will do it, can we remove this step entirely?

src/MIDebugPackage/MIDebugPackage.csproj
```diff
@@ -215,6 +215,59 @@
       <TranslationFile>$(MIEngineRoot)loc\LCL\{Lang}\lib\OpenFolderSchema.json.lcl</TranslationFile>
     </FilesToLocalize>
   </ItemGroup>
+  <Target Name="DropLocalizeDir" AfterTargets="Localize;DropFiles" Condition="'$(LocalizationEnabled)'=='true'">
+    <PropertyGroup>
+      <LocalizedDropDir>$(DropRootDir)\loc\</LocalizedDropDir>
+      <LocalizeBinDir>$(OutDir)localize\</LocalizeBinDir>
+    </PropertyGroup>
+    <ItemGroup>
+      <Locale Include="$(LocalizeBinDir)ENU\*.json">
```
gregg-miskelly: nit: Since each item represents a file to copy rather than a language, we might want to give this a better name. `LocalizedFileToDrop` maybe?

src/MIDebugPackage/MIDebugPackage.csproj
pieandcakes: @gregg-miskelly looks like copy will create it. I removed it.

src/MIDebugPackage/MIDebugPackage.csproj
pieandcakes: Done

PR #609, opened 2017-05-31
Update to copy out the lcg file to the drop

pieandcakes:
per @danyeh we need to copy the lcg file out to the drop for them to localize. Making the copy for the drop copy the file.

### DIFF
--- a/src/MIDebugPackage/MIDebugPackage.csproj
+++ b/src/MIDebugPackage/MIDebugPackage.csproj
@@ -221,7 +221,8 @@
       <LocalizeBinDir>$(OutDir)localize\</LocalizeBinDir>
     </PropertyGroup>
     <ItemGroup>
-      <LocalizedFileToDrop Include="$(LocalizeBinDir)ENU\*.json">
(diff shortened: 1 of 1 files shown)


src/MIDebugPackage/MIDebugPackage.csproj
```diff
@@ -221,7 +221,8 @@
       <LocalizeBinDir>$(OutDir)localize\</LocalizeBinDir>
     </PropertyGroup>
     <ItemGroup>
-      <LocalizedFileToDrop Include="$(LocalizeBinDir)ENU\*.json">
+      <!-- to copy out lcg file -->
+      <LocalizedFileToDrop Include="$(LocalizeBinDir)ENU\*">
```
chuckries: We don't need this for other languages as well?

pieandcakes: @chuckries No. This is a hack to get the file out. I have it checked in but their tool expects it to be in the drop.

PR #611, opened 2017-06-02
Pipe Transport

pieandcakes:
Adding PipeTransport for the AnyCode scenario

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -74,6 +74,10 @@ namespace MICore
         /// <param name="pipeCwd">Current working directory of pipe program. If empty directory of the pipePath is set as the cwd.</param>
         /// <param name="pipeEnvironment">Environment variables set before invoking the pipe program</param>
         public PipeLaunchOptions(string pipePath, string pipeArguments, string pipeCommandArguments, string pipeCwd, MICore.Xml.LaunchOptions.EnvironmentEntry[] pipeEnvironment)
+            : this(pipePath, pipeArguments, pipeCommandArguments, pipeCwd, (pipeEnvironment != null) ? pipeEnvironment.Select(e => new EnvironmentEntry(e)).ToArray() : new EnvironmentEntry[] { })
+        { }
+
+        public PipeLaunchOptions(string pipePath, string pipeArguments, string pipeCommandArguments, string pipeCwd, IList<EnvironmentEntry> pipeEnvironment)
         {
             if (string.IsNullOrEmpty(pipePath))
                 throw new ArgumentNullException("PipePath");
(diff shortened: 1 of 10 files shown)


src/MICore/LaunchOptions.cs
```diff
@@ -83,7 +87,106 @@ public PipeLaunchOptions(string pipePath, string pipeArguments, string pipeComma
             this.PipeCommandArguments = pipeCommandArguments;
             this.PipeCwd = pipeCwd;
 
-            this.PipeEnvironment = (pipeEnvironment != null) ? pipeEnvironment.Select(e => new EnvironmentEntry(e)).ToArray() : new EnvironmentEntry[] { };
+            this.PipeEnvironment = new ReadOnlyCollection<EnvironmentEntry>(pipeEnvironment ?? new List<EnvironmentEntry>(0));
+        }
+
+        private static string gdbPath = @"/usr/bin/gdb";
```
paulmaybee: gdbPathDefault

src/MICore/LaunchOptions.cs
```diff
@@ -248,7 +351,7 @@ public static ReadOnlyCollection<SourceMapEntry> CreateCollection(Xml.LaunchOpti
 
         public static ReadOnlyCollection<SourceMapEntry> CreateCollection(Dictionary<string, string> source)
         {
-            IList<SourceMapEntry> sourceMaps = source?.Select(x => new SourceMapEntry() { EditorPath = x.Value, CompileTimePath = x.Key }).ToList();
+            IList<SourceMapEntry> sourceMaps = source?.Select(x => new SourceMapEntry() { EditorPath = x.Key, CompileTimePath = x.Value}).ToList();
```
paulmaybee: Looks like you are never setting UseForBreakpoints in the SourceMapEntry. Breakpoints are the primary usage. You should either change the default to true or add the field to the json specification.

src/MICore/LaunchOptions.cs
pieandcakes: Fixed

src/MICore/LaunchOptions.cs
pieandcakes: @paulmaybee I'll add the default to true.

src/MICore/LaunchOptions.cs
```diff
@@ -83,7 +87,106 @@ public PipeLaunchOptions(string pipePath, string pipeArguments, string pipeComma
...
+
+            Json.LaunchOptions.PipeTransport pipeTransport = parsedOptions["pipeTransport"].ToObject<Json.LaunchOptions.PipeTransport>();
+
+            // PipeProgram must be specified
+            if (String.IsNullOrWhiteSpace(pipeTransport.PipeProgram))
+            {
+                throw new InvalidLaunchOptionsException(String.Format(CultureInfo.InvariantCulture, MICoreResources.Error_EmptyPipePath));
+            }
+
+            string pipeProgram = EnsurePipeProgramExists(pipeTransport);
```
gregg-miskelly: This function doesn't handle the case that pipeProgram is a command on the PATH. You might want to omit this and just deal with Process.Start failing, which seems to be what we do for XML.

src/MICore/LaunchOptions.cs
```diff
@@ -248,7 +351,7 @@ public static ReadOnlyCollection<SourceMapEntry> CreateCollection(Xml.LaunchOpti
 
         public static ReadOnlyCollection<SourceMapEntry> CreateCollection(Dictionary<string, string> source)
         {
-            IList<SourceMapEntry> sourceMaps = source?.Select(x => new SourceMapEntry() { EditorPath = x.Value, CompileTimePath = x.Key }).ToList();
+            IList<SourceMapEntry> sourceMaps = source?.Select(x => new SourceMapEntry() { EditorPath = x.Key, CompileTimePath = x.Value, UseForBreakpoints = true }).ToList();
```
gregg-miskelly: >UseForBreakpoints [](start = 137, length = 17)

I would suggest adding a // TODO comment and some sort of work item to support setting this to false as there are some cases where one wouldn't want that on.

src/MICore/LaunchOptions.cs
```diff
@@ -1849,6 +1909,39 @@ public static MIMode ConvertMIModeAttribute(Xml.LaunchOptions.MIMode source)
             Debug.Assert((uint)MIMode.Clrdbg == (uint)Xml.LaunchOptions.MIMode.clrdbg, "Enum values don't line up!");
             return (MIMode)source;
         }
+
+        protected static string ParseArguments(IEnumerable<string> arguments, bool alwaysQuote = false)
```
gregg-miskelly: This is a different meaning for `quoteArgs` then we use in the C# extension, and I don't think this is what you want.

This is what I think you want:
  If true: arguments containing spaces get quoted
  If false: arguments are never quoted, even if they have spaces.

Here is the code that I think you want:

```C#
        /// <summary>Enclose the given string in quotes if it contains space or tab characters.</summary>
        public static string Quote(string arg)
        {
            // Check to see if it has already been quoted
            if (arg[0] == '"')
            {
                return arg;
            }

            if (arg.IndexOfAny(ArgumentSeparators) >= 0)
            {
                return '"' + arg + '"';
            }
            return arg;
        }

        /// <summary>Join the arguments list, quoting arguments when necessary.</summary>
        public static string JoinArguments(IEnumerable<string> args, bool withQuote)
        {
            StringBuilder stringBuilder = new StringBuilder();
            if (args != null)
            {
                foreach (string arg in args)
                {
                    if (stringBuilder.Length != 0)
                        stringBuilder.Append(' ');

                    if (withQuote)
                    {
                        stringBuilder.Append(Quote(arg));
                    }
                    else
                    {
                        stringBuilder.Append(arg);
                    }
                }
            }
            return stringBuilder.ToString();
        }
```

src/MICore/LaunchOptions.cs
```diff
@@ -83,7 +87,106 @@ public PipeLaunchOptions(string pipePath, string pipeArguments, string pipeComma
...
+        {
+            // Debugger path. Assume /usr/bin/gdb unless specified
+            string dbgPath = String.IsNullOrWhiteSpace(debuggerPath) ? debuggerPathDefault : debuggerPath;
+
+            // debugger command: /usr/bin/gdb --interpreter=mi
+            string dbgCmdArguments = String.Format(CultureInfo.InvariantCulture, "{0} {1}", dbgPath, "--interpreter=mi");
+
+            string userArguments = ParseArguments(pipeArgs, quoteArgs);
+
+            if (userArguments.Contains("${debuggerCommand}"))
```
gregg-miskelly: >if [](start = 12, length = 2)

We should probably extract this if/else block into a helper method that we can use both here and also at:

```
PipeTransport.cs(319):     string fullCommand = string.Format(CultureInfo.InvariantCulture, "{0} {1} {2}", _pipePath, _cmdArgs, commandText);
```

(Or just fix up that line also)

src/MICore/LaunchOptions.cs
pieandcakes: added comment

src/MICore/LaunchOptions.cs
pieandcakes: Removed

src/MICore/LaunchOptions.cs
pieandcakes: made it slightly different but that works.

pieandcakes:
@gregg-miskelly Can you take a look at the latest set of changes?

pieandcakes:
Got Email from Gregg about signoff.

PR #613, opened 2017-06-06
Dev/paulmay/disassembly fix

paulmaybee:
Disassembly behaved oddly when scrolling backwards and when encountering unmapped memory ranges.
Now return '??' for out-of-range memory. Enable scrolling forwards and backwards consistently.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/Disassembly.cs
+++ b/src/MIDebugEngine/Engine.Impl/Disassembly.cs
@@ -54,7 +54,7 @@ namespace Microsoft.MIDebugEngine
             return 0 <= i && i <= _instructions.Length;
         }
 
-        public bool TryFetch(ulong addr, int cnt, out IEnumerable<DisasmInstruction> instructions)
+        public bool TryFetch(ulong addr, int cnt, out ICollection<DisasmInstruction> instructions)
         {
             instructions = null;
             if (!Contains(addr, cnt))
@@ -66,6 +66,19 @@ namespace Microsoft.MIDebugEngine
                 cnt = -cnt;
             }
             instructions = new ArraySegment<DisasmInstruction>(_instructions, i, cnt);
+            return true;
+        }
+
+        public bool TryFetch(ulong startAddr, ulong endAddr, out ICollection<DisasmInstruction> instructions)
+        {
+            instructions = null;
+            if (!Contains(startAddr, 1))
+                return false;
+            if (!Contains(endAddr, 1))
+                return false;
+            int i = FindIndex(startAddr);
+            int j = FindIndex(endAddr);
+            instructions = new ArraySegment<DisasmInstruction>(_instructions, i, j-i);
             return true;
         }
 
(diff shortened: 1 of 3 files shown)


pieandcakes:
@paulmaybee I don't know enough about memory window but it LGTM. @gregg-miskelly probably should give it a look too.

src/MIDebugEngine/Engine.Impl/Disassembly.cs
```diff
@@ -166,10 +143,123 @@ public Disassembly(DebuggedProcess process)
...
+        {
+            ICollection<DisasmInstruction> ret = null;
+            ulong defaultAddr = address >= (ulong)nInstructions ? address - (ulong)nInstructions : 0;
+
+            lock (_disassemlyCache)
+            {
+                // check the cache
+                var kv = _disassemlyCache.FirstOrDefault((p) => p.Value.TryFetch(address, -nInstructions, out ret));
+                if (kv.Value != null)
+                    return ret.First().Addr;
```
gregg-miskelly: This will seek back nInstruction bytes. But if the count of instructions is less than nInstructions, don't we need to keep looking?

src/MIDebugEngine/Engine.Impl/Disassembly.cs
```diff
@@ -166,10 +143,123 @@ public Disassembly(DebuggedProcess process)
...
+            }
+            ulong endAddress;
+            ulong startAddress;
+            var range = await _process.FindValidMemoryRange(address, (uint)(_process.MaxInstructionSize * nInstructions)+1, (int)(_process.MaxInstructionSize * -nInstructions));
```
gregg-miskelly: Since we only went back a single byte per instruction on line 158, shouldn't this value just be the starting value of `address`?

src/MIDebugEngine/Engine.Impl/Disassembly.cs
```diff
@@ -166,10 +143,123 @@ public Disassembly(DebuggedProcess process)
...
+        /// <returns> address - n on failure, else the address of an instruction n back from the target address</returns>
+        public async Task<ulong> SeekBack(ulong address, int nInstructions)
+        {
+            ICollection<DisasmInstruction> ret = null;
+            ulong defaultAddr = address >= (ulong)nInstructions ? address - (ulong)nInstructions : 0;
+
+            lock (_disassemlyCache)
+            {
+                // check the cache
+                var kv = _disassemlyCache.FirstOrDefault((p) => p.Value.TryFetch(address, -nInstructions, out ret));
```
paulmaybee: done

src/MIDebugEngine/Engine.Impl/Disassembly.cs
paulmaybee: We provide a range large enough to contain all the instructions we are looking for (3rd parameter is offset from address). Gdb will return the  valid addresses in that range.

src/MIDebugEngine/Engine.Impl/Disassembly.cs
```diff
@@ -166,10 +143,123 @@ public Disassembly(DebuggedProcess process)
...
+                return defaultAddr;
+            }
+            lock (_disassemlyCache)
+            {
+                // check the cache with the adjusted range
+                var kv = _disassemlyCache.FirstOrDefault((p) => p.Value.TryFetch(startAddress, address < endAddress ? address : endAddress, out ret));
+            }
+            if (ret == null)
+            {
+                DisasmInstruction[] instructions = await Disassemble(_process, startAddress, endAddress);
```
paulmaybee: Gdb seems to do a good job of finding the correct instructions. That said, we always call VerifyDisassembly for any newly disassembled address range. If we fail to find the "right" address as the start of an instruction - because it is not really the start of an instruction - then we just use whatever gdb first returned as disassembly for the address range.

paulmaybee: Ah, if you meant this comment for below in "FetchInstruction" then yes. We need VerifyDisassembly there too.

Issue #612, opened 2017-06-03
In Json parsing for Source File Map, add method for users to provide "useBreakpoints" boolean


pieandcakes:

pieandcakes:
Closed with Pull Request #614

PR #614, opened 2017-06-07
Change SourceFileMapping to allow string or object as the value

pieandcakes:
Change for sourcefilemapping to allow string or an object with string+bool  as the value. 

Also fixed some schema issues and changed description

### DIFF
--- a/src/MIDebugPackage/OpenFolderSchema.json
+++ b/src/MIDebugPackage/OpenFolderSchema.json
@@ -12,7 +12,11 @@
           "templateId": "launch",
           "displayName": "C/C++ Launch for MinGW/Cygwin (gdb)",
           "description": "Debug launch configuration for C/C++ hosted in MinGW or Cygwin.",
-          "fileExtensions": [ "*.exe", "makefile" ],
+          "fileExtensions": [
+            "*.exe",
+            "*.out",
+            "makefile"
+          ],
           "initialConfiguration": {
             "cwd": "${workspaceRoot}",
             "program": "${debugInfo.target}",
@@ -25,7 +29,11 @@
           "templateId": "attach",
           "displayName": "C/C++ Attach for MinGW/Cygwin (gdb)",
           "description": "Debug attach configuration for C/C++ hosted in MinGW or Cygwin.",
-          "fileExtensions": [ "*.exe", "makefile" ],
+          "fileExtensions": [
+            "*.exe",
+            "*.out",
+            "makefile"
+          ],
           "initialConfiguration": {
             "cwd": "${workspaceRoot}",
             "program": "${debugInfo.target}",
@@ -41,7 +49,6 @@
           "cpp_schema": {
             "type": "object",
             "properties": {
-
               "program": {
                 "type": "string",
                 "description": "Full path to program executable."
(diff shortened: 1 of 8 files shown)


pieandcakes:
This is for issue #612

src/MIDebugPackage/OpenFolderSchema.json
```diff
@@ -207,18 +219,35 @@
...
+              "sourceFileMapOptions": {
+                "type": "object",
+                "properties": {
+                  "editorPath": {
+                    "type": "string",
+                    "description": "The location of the source code for the editor to locate."
+                  },
+                  "useForBreakpoints": {
+                    "type": "boolean",
+                    "description": "Use this source mapping for breakpoint binding? Default is true."
```
gregg-miskelly: >for breakpoint binding [](start = 60, length = 22)

I wonder if people will understand what 'for breakpoint binding' means?

How about:
Use this source mapping when setting breakpoints? Default is 'true'. Set to 'false' if files under 'editorPath' were not always originally compiled from the compiler path. Note that if if a source file doesn't match any entry in the map, the breakpoint will be set using just the file name.

pieandcakes: @gregg-miskelly Took another stab at it. How about this wording

pieandcakes:
@gregg-miskelly @chuckries Also added newlines in the description text for some of the descriptions so that the default and examples are on a new line.

src/RequiredNetFX46FacadeAssemblies.targets
```diff
@@ -13,5 +13,6 @@
     <RequiredNetFX46FacadeAssemblies Include="$(NuGetPackagesDirectory)\System.Net.Sockets\4.1.0\lib\net46\System.Net.Sockets.dll"/>
     <RequiredNetFX46FacadeAssemblies Include="$(NuGetPackagesDirectory)\System.Security.Cryptography.X509Certificates\4.1.0\lib\net46\System.Security.Cryptography.X509Certificates.dll"/>
     <RequiredNetFX46FacadeAssemblies Include="$(NuGetPackagesDirectory)\System.Runtime.InteropServices.RuntimeInformation\4.0.0\lib\net45\System.Runtime.InteropServices.RuntimeInformation.dll"/>
+    <RequiredNetFX46FacadeAssemblies Include="$(NuGetPackagesDirectory)\System.Runtime.Serialization.Primitives\4.1.1\lib\net46\System.Runtime.Serialization.Primitives.dll"/>
```
gregg-miskelly: >RequiredNetFX46FacadeAssemblies [](start = 5, length = 31)

Update src\MIDebugPackage\Install.cmd also

src/MIDebugPackage/OpenFolderSchema.json
```diff
@@ -194,31 +208,48 @@
                   "debuggerPath": {
                     "type": "string",
                     "description": "The full path to the debugger on the target machine, for example /usr/bin/gdb."
-                  },
-                  "pipeEnv": {
+                  },                  "pipeEnv": {
```
gregg-miskelly: >                   [](start = 20, length = 18)

nit: your spacing looks like it got messed up here

src/MIDebugPackage/OpenFolderSchema.json
```diff
@@ -194,31 +208,48 @@
...
+              "sourceFileMapOptions": {
+                "type": "object",
+                "properties": {
+                  "editorPath": {
+                    "type": "string",
+                    "description": "The location of the source code for the editor to locate."
+                  },
+                  "useForBreakpoints": {
+                    "type": "boolean",
+                    "description": "Use this source mapping when setting breakpoints. If 'false', only the filename and line number is sent. If 'true', only when this mapping applies will the entire path to the file and line number be sent, otherwise just filename and line number is sent. \nDefault is 'true'."
```
gregg-miskelly: >If 'false', only the filename and line number is sent [](start = 86, length = 53)

Should there be a question mark instead of a period at the end of the first sentence?

RE: "sent" - instead of "only the filename and line number is sent" how about "breakpoints are set using just file name and line number (no path)" since it might not be obvious to people where we are sending this information. Similar comment in the next sentences too.

Last: Do we need a '\n' somewhere in there?

src/MIDebugPackage/OpenFolderSchema.json
pieandcakes: Fixed

src/RequiredNetFX46FacadeAssemblies.targets
pieandcakes: Fixed

src/MIDebugPackage/OpenFolderSchema.json
pieandcakes: @gregg-miskelly  I'll take another stab at it. I was trying to avoid the first sentence sounding like a question. 
I can add a '\n' after the first sentence.

src/MIDebugPackage/OpenFolderSchema.json
```diff
@@ -204,21 +218,38 @@
...
+              "sourceFileMapOptions": {
+                "type": "object",
+                "properties": {
+                  "editorPath": {
+                    "type": "string",
+                    "description": "The location of the source code for the editor to locate."
+                  },
+                  "useForBreakpoints": {
+                    "type": "boolean",
+                    "description": "When setting breakpoints, this source mapping should be used. \nIf 'false', only the filename and line number is used for setting breakpoints. \nIf 'true', breakpoints will be set with the full path to the file and line number only when this source mapping is used. Otherwise just filename and line number will be used when setting breakpoints. \nDefault is 'true'."
```
pieandcakes: @gregg-miskelly does this sound better:
>"When setting breakpoints, this source mapping should be used. \nIf 'false', only the filename and line number is used for setting breakpoints. \nIf 'true', breakpoints will be set with the full path to the file and line number only when this source mapping is used. Otherwise just filename and line number will be used when setting breakpoints. \nDefault is 'true'."

pieandcakes:
@gregg-miskelly Removed the changes for fixing the missing DLL issue. We're going to make a different change and I didn't want to leave it in this bug.

gregg-miskelly:
Okay, merge at will...

PR #619, opened 2017-06-09
Updated Json parsing with custom enum converter

pieandcakes:
* Remove reliance on System.runtime.serialization.primitives.dll
* fixed error message strings. There were some instances where Error_InvalidLaunchOptions string showed up twice because InvalidLaunchOptionsException already prepends the message.

This is to fix a missing assembly issue in VS2017 where System.Runtime.Serialization.Primitives version 4.1.1 was not installed

### DIFF
--- a/src/MICore/JsonLaunchOptions.cs
+++ b/src/MICore/JsonLaunchOptions.cs
@@ -3,11 +3,10 @@
 
 using System;
 using System.Collections.Generic;
-using System.Runtime.Serialization;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Converters;
-using Newtonsoft.Json.Linq;
+using System.Diagnostics;
 using System.Globalization;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace MICore.Json.LaunchOptions
 {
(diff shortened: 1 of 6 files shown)


src/MICore/LaunchOptions.cs
```diff
@@ -457,7 +455,7 @@ static internal LocalLaunchOptions CreateFromJson(JObject parsedOptions)
 
             if (launchOptions == null)
             {
-                throw new InvalidLaunchOptionsException(MICoreResources.Error_InvalidLaunchOptions);
+                throw new InvalidLaunchOptionsException(MICoreResources.Error_LaunchOptionsNull);
```
gregg-miskelly: >Error_LaunchOptionsNull [](start = 72, length = 23)

Is this just an internal error? Or can the user actually do something to get here? If it is just an internal error, we might not want to add a new localizable string. If it is not just an internal error, should we better describe what is wrong with the launch options?

src/MICore/JsonLaunchOptions.cs
```diff
@@ -185,17 +184,14 @@ public partial class LaunchOptions : BaseOptions
...
         /// The command to execute after the debugger is fully setup in order to cause the target process to run. Allowed values are "exec-run", "exec-continue", "None". The default value is "exec-run".
         /// </summary>
-        [JsonProperty("launchCompleteCommand", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        [JsonProperty("launchCompleteCommand", DefaultValueHandling = DefaultValueHandling.Ignore),
+        JsonConverter(typeof(LaunchCompleteCommandConverter))]
         public LaunchCompleteCommandValue? LaunchCompleteCommand { get; set; }
-
-        [JsonConverter(typeof(StringEnumConverter))]
+        
         public enum LaunchCompleteCommandValue
```
gregg-miskelly: >LaunchCompleteCommandValue [](start = 20, length = 26)

Now that you don't need this to make the serializer happy, I might suggest using the existing 'LaunchCompleteCommand' type.

pieandcakes: @gregg-miskelly I added that change

src/MICore/LaunchOptions.cs
pieandcakes: @gregg-miskelly I changed it to the message that says we weren't able to serialize it. It happens when we can't serialize it as either launch or attach options.

PR #620, opened 2017-06-12
Recent changes in MIEngine broke an existing Launcher extension. Add…

paulmaybee:
… test to verify that api surface is intact.


paulmaybee:
The change that broke us was related to the LocalLaunchOptions constructor; third parameter changed from an array to an IList.

paulmaybee:
Would like to change the Launcher so that rather than a fully constructed LauchOptions object it passes Xml back after processing its custom parameters. This would significantly reduce the api surface we have to worry about.

Issue #616, opened 2017-06-08
MIEngine hangs on _reader.ReadLineAsync() when gdb debugs wrong type of executable


pieandcakes:
Tested with MSYS2's x64 version of gdb against (accidently) a WSL compiled app. Happens when waiting for a response from -exec-run

Expecting the gdb error that the executable could not be run. I can see the errors in redirected stdout but not with MIEngine.

this happens both with VS and with VSCode's C++ extension.

pieandcakes:
Also happens when you step out of the main frame because gdb returns an error: 

> ->1023^error,msg="\"finish\" not meaningful in the outermost frame."

pieandcakes:
After investigation, this seems to be an issue with gdb on a custom Linux subsystem running on Windows. The problem seems to be that the `^error` messages aren't being flushed. A workaround is to add a new line after every command sent to ensure that it is being flushed.

PR #624, opened 2017-06-14
Fix for hang with VS on gdb error message (#616)

pieandcakes:
Fix for Issue #616 
* Hang was because gdb on MinGW/Cygwin would send ^error but not flush.
Fix is to send additional newline at the end of every command if we are
running local launch on gdb on windows

* Fix for error messages during continue or step showing useless message
in VS. Fix is to catch the exceptions instead of letting it go through,
return the actual exception's message and send E_ABORT

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -49,6 +49,8 @@ namespace MICore
 
         public bool IsCygwin { get; protected set; }
 
+        public bool IsLocalGdbOnWindows { get; protected set; }
+
         public virtual void FlushBreakStateData()
         {
         }
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -49,7 +49,7 @@ namespace MICore
 
         public bool IsCygwin { get; protected set; }
 
-        public bool IsLocalGdbOnWindows { get; protected set; }
+        public bool SendNewLineAfterCmd { get; protected set; }
 
         public virtual void FlushBreakStateData()
         {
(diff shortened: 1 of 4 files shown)


src/MICore/Debugger.cs
```diff
@@ -49,6 +49,8 @@ public class Debugger : ITransportCallback
 
         public bool IsCygwin { get; protected set; }
 
+        public bool IsLocalGdbOnWindows { get; protected set; }
```
gregg-miskelly: There are some cases where we can be using a local GDB on Windows where this will not be true. I am guessing this flag matches the cases where you want to send the extra new line, but doesn't quite match when we are using local GDB. So I would suggest changing the name of this flag to 'SendExtraNewLine'.

src/MICore/Transports/StreamTransport.cs
```diff
@@ -147,6 +147,12 @@ protected void Echo(string cmd)
             lock (_locker)
             {
                 _writer?.WriteLine(cmd);
+                // https://github.com/Microsoft/MIEngine/issues/616 : If it is local gdb (MinGW/Cygwin) on Windows, we need to send an extra line after commands 
+                // so that if it errors, the error will come through. 
+                if((_callback is Debugger) && ((Debugger)_callback).IsLocalGdbOnWindows)
```
gregg-miskelly: Checking if '_callback' is Debugger feels kind of fragile. Could we instead fix it in Debugger.cs line 1450 with 'if (IsLocalGdbOnWindows) _transport.Send(string.Empty);

src/MICore/Debugger.cs
pieandcakes: @gregg-miskelly What cases would that be? I'd be ok with renaming it. Bash on Windows uses pipe so it isn't going through this path and I can't think of another "local" launch that uses gdb that wouldn't want this.

src/MICore/Transports/StreamTransport.cs
pieandcakes: @gregg-miskelly I can do that but the Logger will log all the newlines too making the logfile much longer.

PR #625, opened 2017-06-14
Change check for isTerminating

pieandcakes:
During test runs, because it is so fast, we could be terminating and fall through to the second case. In this instance, the Android MDD Sanity test was running so fast that it _terminating was true so it was falling through to the second case. What we want to do is that if it is _terminating, to do neither block of work.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -286,18 +286,21 @@ namespace MICore
             this.ProcessState = ProcessState.Stopped;
             FlushBreakStateData();
 
-            if (!results.Contains("frame") && !_terminating)
+            if (!_terminating)
             {
-                if (ModuleLoadEvent != null)
+                if (!results.Contains("frame"))
                 {
-                    ModuleLoadEvent(this, new ResultEventArgs(results));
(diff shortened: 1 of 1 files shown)


pieandcakes:
Bug: https://devdiv.visualstudio.com/web/wi.aspx?pcguid=011b8bdf-6d56-4f87-be0d-0092136884d9&id=434110

pieandcakes:
Glass run with fix:

Run on device: x.x.x.x:5555 KitKat
Running 'Attach'
Running 'Breakpoint'
Running 'BreakPointEnableAndDisable'
Running 'BreakPointWhenTrueCondition'
Running 'CallStack'
Running 'Eval'
Running 'EvalComplexExpression'
Running 'EvalCompoundDataTypes'
Running 'EvalErrors'
Running 'EvalPrimitiveTypes'
Running 'Exceptions'
Running 'FunctionBreakPoints'
Running 'Locals'
Running 'MultiThreadingCallStack'
Running 'Sanity'
Running 'Stepping'
Running 'SwitchFramesInCallStack'
----------------------------------------
All tests completed succesfully.

PR #627, opened 2017-06-15
Remove Breakpoint that we set at main after entrypoint is hit

pieandcakes:
Looks like we never remove the breakpoint we set at main and if the first line of code is in the while loop, then it keeps getting hit because it isn't being removed which comes back as an exception in VS.
@gregg-miskelly @chuckries 

fix for Bug
https://devdiv.visualstudio.com/web/wi.aspx?pcguid=011b8bdf-6d56-4f87-be0d-0092136884d9&id=451217


pieandcakes:
@paulmaybee

Issue #629, opened 2017-06-16
Using miDebuggerServerAddress on Windows sends wrong cwd path to -environment-cd


pieandcakes:
Seems like we are sending -environment-cd with windows path separators when miDebuggerServerAddress is specified on a Windows machine.

pieandcakes:
Fixed with: https://github.com/Microsoft/MIEngine/pull/631

PR #630, opened 2017-06-16
Fix for Issue #629.

pieandcakes:
Unix path separators should be used when miDebuggerServerAddress is specified

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -341,20 +341,24 @@ namespace MICore
             {
                 if (item.Value is String)
                 {
-                    sourceMaps.Add(new SourceMapEntry() {
+                    sourceMaps.Add(new SourceMapEntry()
+                    {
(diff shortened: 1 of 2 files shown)


gregg-miskelly:
@pieandcakes I don't think your fix is correct. The problem is that 'UseUnixSymbolPaths' gets used for deciding what to do for symbol paths (as the name implies), and we want to use Windows symbols paths when running on Windows with gdbserver as the symbols are loaded on Windows.

Sounds like we either need another property, or to change this to a method for dealing with the working directory.

pieandcakes:
@gregg-miskelly I'll create a separate property then and update EscapePath. The EscapePath call seems to want to use this to determine `/` versus `\\`.

gregg-miskelly:
@pieandcakes You probably want to rename `EscapePath` to `EscapeSymbolSourcePath` and add a new `EscapePath` which take a `bool useUnixPaths`.

PR #631, opened 2017-06-17
Refactor Fix for Issue#629

pieandcakes:
Was pointed out that symbol paths and regular path escaping needs to be
different. made a change to support that.

@wardenGnaw

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -567,7 +567,7 @@ namespace Microsoft.MIDebugEngine
 
             // On Windows ';' appears to correctly works as a path seperator and from the documentation, it is ':' on unix
             string pathEntrySeperator = _launchOptions.UseUnixSymbolPaths ? ":" : ";";
-            string escapedSearchPath = string.Join(pathEntrySeperator, _launchOptions.GetSOLibSearchPath().Select(path => EscapePath(path, ignoreSpaces: true)));
+            string escapedSearchPath = string.Join(pathEntrySeperator, _launchOptions.GetSOLibSearchPath().Select(path => EscapeSymbolPath(path, ignoreSpaces: true)));
             if (!string.IsNullOrWhiteSpace(escapedSearchPath))
             {
                 if (_launchOptions.DebuggerMIMode == MIMode.Gdb)
@@ -624,7 +624,7 @@ namespace Microsoft.MIDebugEngine
                     this.AddGetTargetArchitectureCommand(commands);
 
                     // Add core dump information (linux/mac does not support quotes around this path but spaces in the path do work)
-                    string coreDump = _launchOptions.UseUnixSymbolPaths ? _launchOptions.CoreDumpPath : EscapePath(_launchOptions.CoreDumpPath);
+                    string coreDump = this.UseUnixPathSeparators ? _launchOptions.CoreDumpPath : this.EscapePath(_launchOptions.CoreDumpPath);
                     string coreDumpCommand = _launchOptions.DebuggerMIMode == MIMode.Lldb ? String.Concat("target create --core ", coreDump) : String.Concat("-target-select core ", coreDump);
                     string coreDumpDescription = String.Format(CultureInfo.CurrentCulture, ResourceStrings.LoadingCoreDumpMessage, _launchOptions.CoreDumpPath);
                     commands.Add(new LaunchCommand(coreDumpCommand, coreDumpDescription, ignoreFailures: false));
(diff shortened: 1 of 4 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1249,7 +1249,22 @@ internal WorkerThread WorkerThread
             get { return _worker; }
         }
 
-        internal string EscapePath(string path, bool ignoreSpaces = false)
+        internal string EscapePath(string path)
```
gregg-miskelly: >internal [](start = 8, length = 8)

I might suggest adding a comment to try and explain when to use this method and when to use EscapeSymbolPath. Might be worth renaming it to 'EscapeTargetMachinePath' also

```
/// <summary>
/// This method should be used to prepare file paths that are used are the target machine. For GDB Server scenarios, this means wherever gdbserver is being run.
/// </summary>
```
And on EscapeSymbolPath, maybe:
```
/// <summary>
/// This method should be used to escape paths that are used by GDB (and NOT gdbserver). Any path that gdbserver would use in remote server scenarios should use EscapeTargetMachinePath instead.
/// </summary>
```

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1268,6 +1283,24 @@ internal string EscapePath(string path, bool ignoreSpaces = false)
             return path;
         }
 
+        internal bool UseUnixPathSeparators
+        {
+            get
+            {
+                if (PlatformUtilities.IsWindows())
```
WardenGnaw: nit: combine to single if

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -777,7 +777,7 @@ private void CheckCygwin(List<LaunchCommand> commands, LocalLaunchOptions localL
 
         private void AddExecutablePathCommand(IList<LaunchCommand> commands)
         {
-            string exe = EscapePath(_launchOptions.ExePath);
+            string exe = this.EscapePath(_launchOptions.ExePath);
```
gregg-miskelly: >EscapePath [](start = 30, length = 10)

hmm... as I think about it more, I think this isn't complicated enough :). In Android gdb server scenarios at least, I believe this and the working directory and both local (Windows) paths.

Is this how we tell GDB the exe path in GDB Server launch scenarios?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
gregg-miskelly: Personally I like it better this way (I hate super complicated if statements)...

---
In reply to: [122555606](https://github.com/Microsoft/MIEngine/pull/631#discussion_r122555606) [](ancestors = 122555606)

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
pieandcakes: This is how we specify the executable to GDB.

pieandcakes: Aren't Android LaunchOptions not LocalLaunchOptions? I thought they were AndroidLaunchOptions

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
pieandcakes: I combined them for Andrew. I didn't care either way and as long as the Windows check is first, then I think it will be clearer?

WardenGnaw: Ah. I was basing it on what it was previously:
```
if (this is LocalLaunchOptions && PlatformUtilities.IsWindows()) 
```
either way is ok.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
pieandcakes: @gregg-miskelly

gregg-miskelly:
@pieandcakes Android launch options come in as Android-specific, but the Android launcher creates a LocalLaunchOptions object and passes it back to MIEngine. Regardless, I don't *think* Android is special in this regard: at least in GDB server attach scenarios, I think the value passed to '-file-exec-and-symbols' is a local path. But I have never tried to start a new process under gdbserver, so I am not sure how that is supposed to work.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -2012,7 +2059,7 @@ public bool MapCurrentSrcToCompileTimeSrc(string currentSrc, out string compiler
                             continue;   // match didn't end at a directory separator, not actually a match
                         }
                         compilerSrc = Path.Combine(e.CompileTimePath, file);    // map to the compiled location
-                        if (compilerSrc.IndexOf('\\') < 0)
+                        if (compilerSrc.IndexOf('\\') > 0)
```
pieandcakes: @paulmaybee I think this was a logic error, but I wanted to check to see if there was a case I wasn't considering that this was trying to catch.

paulmaybee: Yeah. Looks like an error. As is the statement is a no op.

PR #632, opened 2017-06-17
Refactor removing entrypoint breakpoint.

pieandcakes:
Found a condition where we would hit the clause of bkpt != null before
hitting the entrypoint breakpoint. Refactored how we handle hitting the
entrypoint and removing the fake bp.

@wardengnaw

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -49,6 +49,7 @@ namespace Microsoft.MIDebugEngine
         private bool _needTerminalReset;
         private HashSet<Tuple<string, string>> _fileTimestampWarnings;
         private ProcessSequence _childProcessHandler;
+        private bool _deleteEntryPointBreakpoint;
 
         public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngineCallback callback, WorkerThread worker, BreakpointManager bpman, AD7Engine engine, HostConfigurationStore configStore, HostWaitLoop waitLoop = null) : base(launchOptions, engine.Logger)
         {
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -724,6 +726,7 @@ private List<LaunchCommand> GetInitializeCommands()
                     }
 
                     commands.Add(new LaunchCommand("-break-insert main", ignoreFailures: true));
+                    this._deleteEntryPointBreakpoint = true;
```
gregg-miskelly: I wonder if we should instead install a command handler here so we can find the bpno if the '-bread-insert' succeeds. There is a slight risk of deleting some other random thing that we don't understand otherwise.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -722,9 +725,31 @@ private List<LaunchCommand> GetInitializeCommands()
...
+                        if (index > 0)
+                        {
+                            string trimmedInnerText = bkptResult.Substring(index).Trim('\r', '\n', '{', '}');
+                            Dictionary<string, string> dict = trimmedInnerText
+                                .Split(',')
+                                .Select(value => value.Split('='))
+                                .Where(x => x.Length == 2)
+                                .ToDictionary(x => x.First(), x => x.Last());
+
+                            if (dict.Keys.Contains("number"))
```
WardenGnaw: Will this ever be false? There is already a check for indexOf "number=".

pieandcakes: in case of bad split? I'm worried about trying to access a key that doesn't exist and it throwing an exception. This seemed safer.

WardenGnaw: ok.

pieandcakes:
This apparently breaks Android Tests. I'm going to hang on to this until I figure out what is wrong.

pieandcakes:
Looks like I was missing a rebase and the actual breaking change was when the paths changed for Linux path separators that was fixed last week. reran after rebase:

Running 'Attach'
Running 'Breakpoint'
Running 'BreakPointEnableAndDisable'
Running 'BreakPointWhenTrueCondition'
Running 'CallStack'
Running 'Eval'
Running 'EvalComplexExpression'
Running 'EvalCompoundDataTypes'
Running 'EvalErrors'
Running 'EvalPrimitiveTypes'
Running 'Exceptions'
Running 'FunctionBreakPoints'
Running 'Locals'
Running 'MultiThreadingCallStack'
Running 'Sanity'
Running 'Stepping'
Running 'SwitchFramesInCallStack'
----------------------------------------
All tests completed succesfully.

PR #636, opened 2017-07-01
Fix termination on MinGW/Cygwin by calling Windows API to terminate.

pieandcakes:
MinGW and Cygwin do not allow Async break to occur without user
intervention. The problem is with -exec-abort (stop debugging) while in
run mode requires us to do an async break so that we can abort. Since we
can't do that without additional user intervention, we will terminate
the child process instead.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -602,14 +602,6 @@ namespace MICore
             }
         }
 
-        [DllImport("kernel32.dll", SetLastError = true)]
-        private static extern IntPtr OpenProcess(int dwDesiredAccess, [MarshalAs(UnmanagedType.Bool)] bool bInheritHandle, int dwProcessId);
-        [DllImport("kernel32.dll", SetLastError = true)]
-        [return: MarshalAs(UnmanagedType.Bool)]
-        private static extern bool TerminateProcess(IntPtr hProcess, uint uExitCode);
-        [DllImport("kernel32.dll", SetLastError = true)]
-        private static extern bool CloseHandle(IntPtr hHandle);
-
         public async Task<Results> CmdTerminate()
         {
             if (!_terminating)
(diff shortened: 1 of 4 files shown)


src/MICore/Debugger.cs
```diff
@@ -604,14 +607,49 @@ protected bool IsCoreDump
...
+                if (ProcessState == ProcessState.Running &&
+                    this.MICommandFactory.Mode != MIMode.Clrdbg)
+                {
+                    // MinGW and Cygwin on Windows don't support async break. Because of this,
+                    // the normal path of sending an internal async break so we can exit doesn't work.
+                    // Therefore, we will call TerminateProcess on the debuggee with the exit code of 0
+                    // to terminate debugging. 
+                    if ((this.IsCygwin || this.IsMinGW) && _debuggeePids.Count > 0)
+                    {
+                        int debuggeePid = _debuggeePids.First().Value;
```
gregg-miskelly: >_debuggeePids [](start = 42, length = 13)

Is there a reason you only pick the first?

src/MICore/Debugger.cs
```diff
@@ -604,14 +607,49 @@ protected bool IsCoreDump
...
+                            {
+                                // OperationThread's _runningOpCompleteEvent is doing WaitOne(). Calling MICommandFactory.Terminate() will Set() it, unblocking the UI. 
+                                await MICommandFactory.Terminate();
+                                return new Results(ResultClass.done);
+                            }
+                        }
+                        finally
+                        {
+                            if (handle != IntPtr.Zero)
+                                Marshal.FreeCoTaskMem(handle);
```
gregg-miskelly: >FreeCoTaskMem [](start = 40, length = 13)

You need to P/Invoke to CloseHandle

src/MICore/Debugger.cs
```diff
@@ -604,14 +607,49 @@ protected bool IsCoreDump
...
-                {
-                    await AddInternalBreakAction(() => MICommandFactory.Terminate());
+                if (ProcessState == ProcessState.Running &&
+                    this.MICommandFactory.Mode != MIMode.Clrdbg)
+                {
+                    // MinGW and Cygwin on Windows don't support async break. Because of this,
+                    // the normal path of sending an internal async break so we can exit doesn't work.
+                    // Therefore, we will call TerminateProcess on the debuggee with the exit code of 0
+                    // to terminate debugging. 
+                    if ((this.IsCygwin || this.IsMinGW) && _debuggeePids.Count > 0)
```
gregg-miskelly: >) [](start = 82, length = 1)

Exclude gdbserver cases?

src/MICore/Debugger.cs
```diff
@@ -604,14 +607,49 @@ protected bool IsCoreDump
             }
         }
 
+        [DllImport("kernel32.dll", SetLastError = true)]
+        public static extern IntPtr OpenProcess(int dwDesiredAccess, [MarshalAs(UnmanagedType.Bool)] bool bInheritHandle, int dwProcessId);
+        [DllImport("kernel32.dll", SetLastError = true)]
+        [return: MarshalAs(UnmanagedType.Bool)]
+        static extern bool TerminateProcess(IntPtr hProcess, uint uExitCode);
```
rajkumar42: nit: private?

src/MICore/Debugger.cs
pieandcakes: @gregg-miskelly I took the code from CmdBreakInternal() (line 697) that breaks on the first thread. Would it be better to iterate through the list and terminate each individual item?

src/MICore/Debugger.cs
pieandcakes: Fixed

src/MICore/Debugger.cs
pieandcakes: Fixed

src/MICore/Debugger.cs
gregg-miskelly: That is what I would have thought, but @paulmaybee would be more familiar with the child process code. Maybe there is a reason that isn't needed.

paulmaybee: Doesn't hurt to terminate all pids. Normally having more than one is a transient state. However there is a non-trivial amount of work done by gdb in between reporting that a new process is available and detaching so that vs can reattach separate miengine instance. If something goes wrong/hangs then killing all children is most likely the right way to recover.

Issue #645, opened 2017-08-30
[VS] SSH Attach in VS with a space in path causes attach to fail


pieandcakes:

pieandcakes:
Customer Comment: Debugging Linux SSH is attached to the process does not support paths with spaces.

PR #646, opened 2017-08-30
Fix handling of paths with spaces when trying to use SSH Attach from VS

pieandcakes:
https://github.com/microsoft/miengine/issues/645

@chuckries @rajkumar42 @gregg-miskelly

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -848,6 +848,12 @@ namespace Microsoft.MIDebugEngine
                 string trimmedExePath = exePath.Trim();
                 try
                 {
+                    // If the folder contains a space, we need to quote the path.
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -848,6 +848,12 @@ private void DetermineAndAddExecutablePathCommand(IList<LaunchCommand> commands,
                 string trimmedExePath = exePath.Trim();
                 try
                 {
+                    // If the folder contains a space, we need to quote the path.
+                    if (trimmedExePath.Contains(' '))
+                    {
+                        trimmedExePath = "\"" + trimmedExePath + "\"";
```
gregg-miskelly: Did you test LLDB?

gregg-miskelly: Never mind, looks like this code is only used for GDB

Issue #350, opened 2016-05-31
AD7PendingBreakpoint.Bind() code path makes several AD7 COM calls off of the main thread. 


chuckries:
We aren't seeing specific issues from this yet, but AD7PendingBreakpoint.Bind() ends up calling several AD7 COM api calls off of the main thread, which is a bad. In VS IDE, this will invoke vsdebug directly from background threads.

Most of these calls happen in the BindAsync function or lower in MIEngine. BindAsync() is either executed directly on the worker thread if we are stopped, or indirectly on the event thread when we stop due to internal break.

Some of the calls are HostMarshal calls, which become Marhsal.GetObjectForIntPtr calls in VS IDE, or calls on AD7 interface objects, such as IDebugDocumentPosition2.GetFileName

Example: 

``` C#
case enum_BP_LOCATION_TYPE.BPLT_CODE_FILE_LINE:
    {
        IDebugDocumentPosition2 docPosition = HostMarshal.GetDocumentPositionForIntPtr(_bpRequestInfo.bpLocation.unionmember2);

        // Get the name of the document that the breakpoint was put in
        EngineUtils.CheckOk(docPosition.GetFileName(out documentName));

        // Get the location in the document that the breakpoint is in.
        EngineUtils.CheckOk(docPosition.GetRange(startPosition, endPosition));

        // Get the document checksum
        if (_engine.DebuggedProcess.MICommandFactory.SupportsBreakpointChecksums())
        {
            // TODO: This will need to be configurable in the future somehow
            // TODO: We might want the MICommandFactory to return the algorithm we should use
            // TODO: But HashAlgorithmID is not part of MICore becuase it uses AD7Guids
            HashAlgorithmId[] hashAlgorithmIds = { HashAlgorithmId.SHA1Normalized };

            try
            {
                checksums = GetChecksums(hashAlgorithmIds);
            }
            catch (Exception)
            {
                // If we fail to get a checksum there's nothing else we can do
            }
        }

        break;
    }
```

This whole code path needs to be cleaned up so that AD7 interface calls are only made from the main thread.

pieandcakes:
Leave a comment

PR #649, opened 2017-09-05
Move calls to bpLocation's unionmembers to main thread.

pieandcakes:
This should fix part of #350 's issues in BindAsync().
@gregg-miskelly @chuckries 

VSTS bug 451112

### DIFF
--- a/src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
+++ b/src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
@@ -64,15 +64,6 @@ namespace Microsoft.DebugEngineHost
         {
             throw new NotImplementedException();
         }
-
-        /// <summary>
-        /// Returns true if on the main thread.
-        /// </summary>
-        /// <returns></returns>
-        public static bool OnMainThread()
-        {
-            throw new NotImplementedException();
-        }
     }
 
     /// <summary>
@@ -364,6 +355,16 @@ public sealed class HostLogger
         /// <returns>In VS, a thread-safe wrapper on top of the underlying SDM event callback which allows
         /// sending events on any thread. In VS Code, this just returns the provided ad7Callback. </returns>
         public static IDebugEventCallback2 GetThreadSafeEventCallback(IDebugEventCallback2 ad7Callback)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// In VS, calls Marshal.Release on IUnknown to derefcount when we are done with object.
+        /// </summary>
+        /// <param name="unk">In VS, the IUnknown</param>
+        /// <returns>In VS, the result from calling Marshal.Release</returns>
+        public static int Release(IntPtr unk)
         {
             throw new NotImplementedException();
         }
--- a/src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
+++ b/src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
@@ -64,6 +64,15 @@ namespace Microsoft.DebugEngineHost
         {
             throw new NotImplementedException();
         }
+
+        /// <summary>
+        /// Returns true if on the main thread.
+        /// </summary>
+        /// <returns></returns>
+        public static bool OnMainThread()
+        {
+            throw new NotImplementedException();
+        }
     }
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -248,29 +271,25 @@ int IDebugPendingBreakpoint2.Bind()
...
+                                    codeAddress = ((AD7MemoryAddress)_codePosition).Address;
                                     break;
                                 }
                             case enum_BP_LOCATION_TYPE.BPLT_CODE_FILE_LINE:
                                 {
-                                    IDebugDocumentPosition2 docPosition = HostMarshal.GetDocumentPositionForIntPtr(_bpRequestInfo.bpLocation.unionmember2);
-
                                     // Get the name of the document that the breakpoint was put in
-                                    EngineUtils.CheckOk(docPosition.GetFileName(out documentName));
+                                    EngineUtils.CheckOk(_docPosition.GetFileName(out documentName));
```
pieandcakes: @gregg-miskelly I guess my question is in these calls where its GetFileName or GetRange, I should be caching that in the bind call also?

pieandcakes:
@dotnet-bot test please

pieandcakes:
@dotnet-bot test this please

pieandcakes:
@paulmaybee I don't know if changing the DebugAdapterHost API is going to affect you, but I want to make sure you know about it.

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -116,14 +130,13 @@ public AD7DocumentContext GetDocumentContext(ulong address, string functionName)
         {
             if ((enum_BP_LOCATION_TYPE)_bpRequestInfo.bpLocation.bpLocationType == enum_BP_LOCATION_TYPE.BPLT_CODE_FILE_LINE)
             {
-                IDebugDocumentPosition2 docPosition = HostMarshal.GetDocumentPositionForIntPtr(_bpRequestInfo.bpLocation.unionmember2);
                 string documentName;
-                EngineUtils.CheckOk(docPosition.GetFileName(out documentName));
+                EngineUtils.CheckOk(_docPosition.GetFileName(out documentName));
```
gregg-miskelly: Can we just use `_documentName`/`_startPosition`?

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -34,7 +34,21 @@ internal class AD7PendingBreakpoint : IDebugPendingBreakpoint2
         private bool _deleted;
         private bool _pendingDelete;
 
-        public DebuggedProcess DebuggedProcess {  get { return _engine.DebuggedProcess;  } }
+        private IDebugFunctionPosition2 _functionPosition;
```
gregg-miskelly: Can we avoid storing the AD7 objects (`_functionPosition`, `_codePosition`, `_docPosition`) as member variables if we don't use them in GetDocumentContext?

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
gregg-miskelly: I think we should use the cached info. Really this whole thing is a bit wrong in that we should really be asking GDB where the break point bound. But if we aren't going to do that, it is better to use the info we cached then to re-ask VS for it.

src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
```diff
@@ -64,6 +64,15 @@ public static HostUIIdentifier GetHostUIIdentifier()
         {
             throw new NotImplementedException();
         }
+
+        /// <summary>
```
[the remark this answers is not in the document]
pieandcakes: @gregg-miskelly Are you saying move the method to the HostMarshal class itself or make the calls to this new API from all the methods in the VS implementation of the HostMarshal class?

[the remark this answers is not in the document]
paulmaybee: Agree with Gregg here. Move the OnMainThead checks into the VS implementation of HostMarshal.Xxx. Then no need to worry about breaking any other host implementations.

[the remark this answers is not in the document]
pieandcakes: @gregg-miskelly @paulmaybee these should be fixed now. Thanks

PR #650, opened 2017-09-06
Fix framelevel when break all occurs in Visual Studio

pieandcakes:
VS keeps state of framelevel when break all occurs. gdb assumes when you
break all that the framelevel of current thread is 0. reset the tracking
framelevelid in MIEngine to 0 so that we will call -stack-select-frame
every time we break all if we are expecting a non-zero framelevel.

VSTS bug 410289


pieandcakes:
@dotnet-bot test this please

PR #655, opened 2017-09-27
Add detection for unsupported GDB versions

benmcmorran:
I'm trying to improve the error messaging in Visual Studio when GDB versions 7.12-1, 7.12-2, 7.12-3, and 7.12.1-1 are used for debugging. These specific versions of GDB have a bug that causes them to terminate early in MinGW when invoked with the `--interpreter=mi` flag. The current Visual Studio error message in this case is somewhat cryptic, and this PR adds more detail. The new parts are in bold.

> Unable to start debugging. **GDB version 7.12.1 is not supported. Consider upgrading.**
> 
> Unable to establish a connection to GDB. The following message was written to stderr:
> 
> This application has requested the Runtime to terminate it in an unusual way.
> Please contact the application's support team for more information.
> 
> See Output Window for details.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -12,6 +12,7 @@ using System.Linq;
 using Microsoft.Win32.SafeHandles;
 using Microsoft.DebugEngineHost;
 using System.Runtime.InteropServices;
+using System.Text.RegularExpressions;
 
 namespace MICore
 {
@@ -25,6 +26,9 @@ namespace MICore
 
     public class Debugger : ITransportCallback
     {
+        private const string Event_UnsupportedWindowsGdb = "VS/Diagnostics/Debugger/MIEngine/UnsupportedWindowsGdb";
+        private const string Property_GdbVersion = "VS.Diagnostics.Debugger.MIEngine.GdbVersion";
+
         public event EventHandler BreakModeEvent;
         public event EventHandler RunModeEvent;
         public event EventHandler ProcessExitEvent;
@@ -943,7 +947,25 @@ namespace MICore
                     {
                         if (_consoleDebuggerInitializeCompletionSource != null)
                         {
-                            MIDebuggerInitializeFailedException exception = new MIDebuggerInitializeFailedException(this.MICommandFactory.Name, _initialErrors.ToList().AsReadOnly(), _initializationLog.ToList().AsReadOnly());
+                            MIDebuggerInitializeFailedException exception;
+                            string version = GdbVersionFromLog();
+
+                            // We can't use IsMinGW or IsCygwin because we never connected to the debugger
+                            bool isMinGWOrCygwin = _launchOptions is LocalLaunchOptions &&
+                                    PlatformUtilities.IsWindows() &&
+                                    this.MICommandFactory.Mode == MIMode.Gdb;
+                            if (isMinGWOrCygwin && version != null && IsUnsupportedWindowsGdbVersion(version))
+                            {
+                                exception = new MIDebuggerInitializeFailedUnsupportedGdbException(
+                                    this.MICommandFactory.Name, _initialErrors.ToList().AsReadOnly(), _initializationLog.ToList().AsReadOnly(), version);
+                                SendUnsupportedWindowsGdbEvent(version);
+                            }
+                            else
+                            {
+                                exception = new MIDebuggerInitializeFailedException(
+                                    this.MICommandFactory.Name, _initialErrors.ToList().AsReadOnly(), _initializationLog.ToList().AsReadOnly());
+                            }
+
                             _initialErrors = null;
                             _initializationLog = null;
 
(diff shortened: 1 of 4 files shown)


src/MICore/MICoreResources.resx
```diff
@@ -288,4 +288,10 @@ Error: {1}</value>
     <value>Unknown launchCompleteCommand value '{0}'. Expected values are 'exec-run', 'exec-continue' and 'None'.</value>
     <comment>{0} value</comment>
   </data>
+  <data name="Error_UnsupportedGdb" xml:space="preserve">
+    <value>GDB version {0} is not supported. Consider upgrading.
```
paulmaybee: AFAIK we have only seen this issue on Cygwin. Have you verified this bug occurs on other platforms (Ubuntu)? I don't think we want to put out this message for other platforms unless we know its a problem.

gregg-miskelly: I think you probably want to grab the 'Cygwin' part of the text also, and then change the unsupported version info to include that.

benmcmorran: I just verified that MI mode works on GDB 7.12 Ubuntu, so as far as I know it's only a MinGW/Cygwin issue. Added a check for MinGW/Cygwin.

src/MICore/Debugger.cs
```diff
@@ -943,7 +944,24 @@ public void OnDebuggerProcessExit(/*OPTIONAL*/ string exitCode)
                     {
                         if (_consoleDebuggerInitializeCompletionSource != null)
                         {
-                            MIDebuggerInitializeFailedException exception = new MIDebuggerInitializeFailedException(this.MICommandFactory.Name, _initialErrors.ToList().AsReadOnly(), _initializationLog.ToList().AsReadOnly());
+                            MIDebuggerInitializeFailedException exception;
+                            string version = GdbVersionFromLog();
+
+                            // We can't use IsMinGW or IsCygwin because we never connected to the debugger
+                            bool isMinGWOrCygwin = _launchOptions is LocalLaunchOptions &&
```
gregg-miskelly: I don't think you really want/need this minGW/Cygwin check. It isn't actually that accurate (there are other gdbs that can run on Windows) and you are already explicitly checking for 'Cygwin' in the version info.

gregg-miskelly:
@pieandcakes do you want to take a look also?

src/MICore/Debugger.cs
benmcmorran: We don't actually do a Cygwin check -- we just use the Cygwin version number if it's available. The same problem shows up in MinGW, but the GDB version line has less information in this case.

>GNU gdb (GDB) 8.0.1

I could search through the initialization log some more to look for MinGW specific strings, like `x86_64-w64-mingw32`. Thoughts?

gregg-miskelly: Oh got it. I missed the `?` at the end of the Cygwin section of the RegEx.

I think your code is as good as we can do then. The one suggestion I might make is to enhance the comment in `GdbVersionFromLog` to also list the MinGW example.

benmcmorran: Updated the comment. Also, I don't have write access for this repo, so someone on the team will need to merge once the PR looks good.

pieandcakes: You can replace this with `this.IsMinGW() || this.IsCygwin()`

src/MICore/Debugger.cs
```diff
@@ -978,6 +996,28 @@ public void OnDebuggerProcessExit(/*OPTIONAL*/ string exitCode)
...
+                if (match.Success)
+                {
+                    return match.Groups[1].Success ? match.Groups[1].Value : match.Groups[2].Value;
+                }
+            }
+
+            return null;
+        }
+
+        bool IsUnsupportedGdbVersion(string version)
```
pieandcakes: can we change this to be IsUnsupportedWindowsGdbVersion or add a comment that this is only the unsupported versions running on a Windows subsystem?

src/MICore/MICoreResources.resx
pieandcakes: I agree with Gregg. either add the type of gdb or add something about not being supported on Windows ?

src/MICore/Debugger.cs
benmcmorran: When I was testing `this.IsMinGW` was always false. I think that's because it's set in `CheckCygwin()`, but that function is never called when the debugger fails to connect.

pieandcakes: So what you are saying is that in the instance of this failure we never get to the point where we are issuing a `show configuration` call. That's fair then. 

Do we want to add any additional telemetry to track the versions we are seeing when that exception is thrown? I don't think we log the exception message because it may contain PII.

benmcmorran:
@pieandcakes Thanks. I think this is ready to merge.

pieandcakes:
Talked to @benmcmorran offline, he is going to follow up on two more small items and will let me know.

benmcmorran:
@pieandcakes Telemetry added. I think this is ready for final review and merge.

PR #660, opened 2017-11-17
Localize SSH provider project.

richardstanton:
Update the SSH provider project to allow microbuild localization.

 @pieandcakes @gregg-miskelly @jacdavis @chuckries @paulmaybee

### DIFF
--- a/build/miengine.targets
+++ b/build/miengine.targets
@@ -26,6 +26,27 @@
     </Task>
   </UsingTask>
 
+    <!-- Localize the output assembly if requested -->
+  <ItemGroup Condition=" '$(LocalizeOutputAssembly)'=='true' ">
+    <FilesToLocalize Include="$(TargetPath)">
+      <TranslationFile>$(RepositoryRoot)LOC\LCL\{Lang}\$(TargetFileName).lcl</TranslationFile>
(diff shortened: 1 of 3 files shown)


build/miengine.targets
```diff
@@ -26,6 +26,27 @@
     </Task>
   </UsingTask>
 
+    <!-- Localize the output assembly if requested -->
+  <ItemGroup Condition=" '$(LocalizeOutputAssembly)'=='true' ">
+    <FilesToLocalize Include="$(TargetPath)">
+      <TranslationFile>$(RepositoryRoot)LOC\LCL\{Lang}\$(TargetFileName).lcl</TranslationFile>
```
pieandcakes: If we care about cross compat if we need to be able to build on Linux in the future, these should match the case of the folders in the repo

pieandcakes:
@richardstanton left you one comment, but otherwise LGTM

:shipit:

gregg-miskelly:
@richardstanton I believe you also need to update a .nuspec to include the localized dlls in MIEngine's internal nuspec.

pieandcakes:
@richardstanton Did you try and build the nuget package locally? After looking at the MIEngine.mdd.nuspec it looks like its pulling in stuff from drop\<configuration>\loc so as long as your file is there it should be in the package.

PR #661, opened 2017-11-20
Fix Go To Module / EnumModules

gregg-miskelly:
The MIEngine had a faulty implementation of EnumModules where it would create new module objects instead of returning the same module objects that it used for stack frames / module events. This fixes it.

This also removes some dead code the MIEngine had for sending IDebugSymbolSearchEvent2's. This code also incorrectly created new modules, but that didn't matter since the code was never called anyway.

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7Module.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Module.cs
@@ -20,8 +20,8 @@ namespace Microsoft.MIDebugEngine
 
         public AD7Module(DebuggedModule debuggedModule, DebuggedProcess process)
         {
-            Debug.Assert(debuggedModule.Client == null, "Why is the client of the module already set?");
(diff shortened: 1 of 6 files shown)


src/MIDebugEngine/AD7.Impl/AD7Module.cs
```diff
@@ -20,6 +20,7 @@ internal class AD7Module : IDebugModule2, IDebugModule3
 
         public AD7Module(DebuggedModule debuggedModule, DebuggedProcess process)
         {
+            Debug.Assert(debuggedModule.Client == null, "Why is the client of the module already set?");
```
pieandcakes: i feel like this should be below the null check below it.

pieandcakes:
@gregg-miskelly Approved with a small nit comment.

paulmaybee:
:thumbsup:

PR #664, opened 2017-11-20
Copy localized SSH provider assemblies to drop folder.

richardstanton:
This copies satellite assemblies into the drop folder using the same subfolder name from the bin folder.

@pieandcakes @gregg-miskelly @jacdavis @chuckries @paulmaybee

### DIFF
--- a/build/miengine.targets
+++ b/build/miengine.targets
@@ -26,43 +26,6 @@
     </Task>
   </UsingTask>
 
-    <!-- Localize the output assembly if requested -->
-  <ItemGroup Condition=" '$(LocalizeOutputAssembly)'=='true' ">
-    <FilesToLocalize Include="$(TargetPath)">
-      <TranslationFile>$(RepositoryRoot)loc\lcl\{Lang}\$(TargetFileName).lcl</TranslationFile>
-      <LciCommentFile>$(RepositoryRoot)loc\lci\$(TargetFileName).lci</LciCommentFile>
-    </FilesToLocalize>
-    <!-- Ensure new localized assemblies are signed -->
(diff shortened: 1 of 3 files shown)


build/miengine.targets
```diff
@@ -47,6 +47,22 @@
     </ItemGroup>
   </Target>
 
+  <!-- Custom target that will be invoked after signing and loc to copy signed satellite assemblies
```
pieandcakes: I think this would be better either in its own targets file or in DropFiles.targets. This file doesn't have the MicroBuild props that you depend on.

build/miengine.targets
```diff
@@ -47,6 +47,22 @@
...
+       to the drop location.
+       This target is conditioned to only run if localization is currently turned on. -->
+  <Target Name="CopySatellitesToDrop" AfterTargets="SignFiles" Condition="'$(LocalizationEnabled)'=='true'">
```
pieandcakes: You used LocalizedOutputs in the other target but you use CopySatellites here. Should this be CopySignedLocalizedOutputsToDrop?

Also, I feel like this should be "AfterTargets='GatherLocalizedOutputsForSigning'"

richardstanton: Maybe, I thought something like copy signed localized outputs was getting a bit lengthy, so I tried to find a more concise name that was still clear. 
I think after GatherLocalizedOutputsForSigning is too early because the signing hasn't happened yet.

pieandcakes: fair enough. the naming didn't seem to link the two together which is why I commented.

build/miengine.targets
richardstanton: Sounds reasonable.

PR #673, opened 2017-12-04
Change -stack-list-arguments call to be more performant in some cases

pieandcakes:
Made change to call -stack-list-arguments with --no-values when the
value flag is not specified. This improves perf for large objects in stack walks but it does add an
extra call per variable to -var-create and -var-delete to get the type.
The calls to -var-create/-var-delete are faster overall than the
-stack-list-arguments with --simple-values.

If both value and types are requested, we fall back to requesting
-stack-list-arguments with simple values which will return types and
values.

Testing: Tested against 200 deep stack passing either a list of 6 ints per stack, or 3 vectors of size 5000 + 3 ints. The new solution looks to be more performant in both cases.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -725,7 +725,7 @@ namespace Microsoft.MIDebugEngine
                     {
                         commands.Add(new LaunchCommand("-exec-arguments " + _launchOptions.ExeArguments));
                     }
-                    
+
                     Func<string, Task> breakMainSuccessHandler = (string bkptResult) =>
                     {
                         int index = bkptResult.IndexOf("number=", StringComparison.Ordinal);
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1771,13 +1772,31 @@ internal void OnPostedOperationError(object sender, Exception e)
                         string[] names = ((ResultListValue)argList).FindAllStrings("name");
                         foreach (var n in names)
                         {
-                            args.Add(new SimpleVariableInformation(n, /*isParam*/ true, null, null));
+                            // If the types of the arguments are requested, get that from a call to -var-create
+                            if (types)
```
chuckries: Doesn't this do *both* costly things now if I specify values && types? If we asked for values, don't we already have the type data here?

pieandcakes: @chuckries @gregg-miskelly  No, because you would not be in this block if you had both. Line 1762 would be true as if you went through the "simple values" your argList would be a ValueTypeList and it would run through that if block and not this else block.

pieandcakes: If it helps, without the value, its a NamedResultValue so it falls into the else block.

PR #674, opened 2017-12-06
Change to calling step commands with frame id.

pieandcakes:
Test failure caused by var-create changing frames, which revealed a bug where step commands would operate based on selected Frame on the thread instead of the top frame. Changed to call with frame.

### DIFF
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -217,31 +217,31 @@ namespace MICore
         public async Task ExecStep(int threadId, ResultClass resultClass = ResultClass.running)
         {
             string command = "-exec-step";
-            await ThreadCmdAsync(command, resultClass, threadId);
+            await ThreadFrameCmdAsync(command, resultClass, threadId, 0);
         }
 
         public async Task ExecNext(int threadId, ResultClass resultClass = ResultClass.running)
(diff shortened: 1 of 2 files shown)


src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -217,31 +217,31 @@ public virtual void DefineCurrentThread(int threadId)
         public async Task ExecStep(int threadId, ResultClass resultClass = ResultClass.running)
         {
             string command = "-exec-step";
-            await ThreadCmdAsync(command, resultClass, threadId);
+            await ThreadFrameCmdAsync(command, resultClass, threadId, 0);
```
gregg-miskelly: Did you see if this works in LLDB? I think we will now pass `--frame`. Not sure if lldb-mi will complain.

pieandcakes:
@dotnet-bot test please

pieandcakes:
@dotnet-bot  please test

pieandcakes:
@dotnet-bot test this please

src/MICore/CommandFactories/MICommandFactory.cs
pieandcakes: Looks like it is fine, but now it has the same behavior of stepping based on the top frame of the thread

PR #678, opened 2017-12-12
handle var-create failure case in stackwalk

pieandcakes:


pieandcakes:
This is to address https://github.com/microsoft/vscode-cpptools/issues/1339

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1772,26 +1772,27 @@ internal void OnPostedOperationError(object sender, Exception e)
...
-                                Results results = await MICommandFactory.VarCreate(n, thread.Id, (uint)level, 0);
-
-                                string type = results.FindString("type");
-                                args.Add(new SimpleVariableInformation(n, /*isParam*/ true, null, String.IsNullOrWhiteSpace(type) ? null : type));
-
-                                string varName = results.TryFindString("name");
-                                if (!String.IsNullOrWhiteSpace(varName))
+                                Results results = await MICommandFactory.VarCreate(n, thread.Id, (uint)level, 0, ResultClass.None);
+                                // Only get the type if the result wasn't an error
+                                if (results.ResultClass != ResultClass.error)
```
gregg-miskelly: Rather than `!= error` I think you want to check for the result class that you actually expect.

PR #680, opened 2018-01-02
Fix for entrypoint bp delete and watson

pieandcakes:
* change 'async void' to 'async Task' and add await
* change to Expect possible failure on deleting the entrypoint bp. We
will try and delete it once now and if it fails, move on.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -1126,7 +1126,11 @@ namespace MICore
                     {
                         miError = results.FindString("msg");
                     }
-
+                    else
+                    {
+                        miError = results.ResultClass.ToString();
+                    }
+                
                     _completionSource.SetException(new UnexpectedMIResultException(commandFactory.Name, this.Command, miError));
                 }
                 else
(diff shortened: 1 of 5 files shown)


src/MICore/Debugger.cs
```diff
@@ -1126,7 +1126,11 @@ internal void OnComplete(Results results, MICommandFactory commandFactory)
                     {
                         miError = results.FindString("msg");
                     }
-
+                    else
+                    {
+                        miError = results.ResultClass.ToString();
```
gregg-miskelly: >results.ResultClass.ToString(); [](start = 34, length = 31)

Shouldn't this be some sort of new localized string like "Unexpected ResultClass from MI Debugger. Expected '{0}' but received '{1}'."

gregg-miskelly:
>                 await ConsoleCmdAsync("process handle --pass true --stop false --notify false SIGHUP", true);

If you haven't already double checked this, we should probably make sure this still works.

---
Refers to: src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs:1233 in 3579e19. [](commit_id = 3579e1939f18192bd5fe0dd19e1ac7625f2d1943, deletion_comment = False)

PR #686, opened 2018-01-31
Support launchOptions on remote attach operations

paulmaybee:
During an attach operation look for the file "Microsoft.MIEngine.Attach.Options.xml" in the root of the solution/open folder. Look in the file for SupplementalAttachOptions. Use these to customize the normal attach processing. If the ServerOptions element is present then use gdbserver to perform the attach rather than gdb.

Depends upon a new field in liblinux's GdbServerStartInfo class.

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -1396,7 +1396,7 @@ namespace MICore
             LaunchOptions options;
             if (isServerMode)
             {
-                string addr = unixPort.AttachToProcess((uint)processId, attachOptions.ServerOptions.PreAttachCommand);
+                string addr = unixPort.AttachToProcess(processId, attachOptions.ServerOptions.PreAttachCommand);
                 options = new LocalLaunchOptions(attachOptions.ServerOptions.MIDebuggerPath, addr, null);
                 options._miMode = miMode;
                 options.ExePath = attachOptions.ServerOptions.ExePath;
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -830,31 +830,6 @@ namespace MICore
             }
         }
 
-        public static UnixShellPortLaunchOptions CreateForAttachRequest(Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.IDebugUnixShellPort unixPort,
-            int processId,
-            MIMode miMode,
-            string getClrDbgUrl,
-            string remoteDebuggingDirectory,
-            string remoteDebuggingSubDirectory,
-            string debuggerVersion)
-        {
-            var @this = new UnixShellPortLaunchOptions(startRemoteDebuggerCommand: null,
-                                                       unixPort: unixPort,
-                                                       miMode: miMode,
-                                                       baseLaunchOptions: null,
-                                                       getClrDbgUrl: getClrDbgUrl,
-                                                       remoteDebuggerInstallationDirectory: remoteDebuggingDirectory,
-                                                       remoteDebuggerInstallationSubDirectory: remoteDebuggingSubDirectory,
-                                                       clrdbgVersion: debuggerVersion);
-
-            @this.ProcessId = processId;
-            @this.SetupCommands = new ReadOnlyCollection<LaunchCommand>(new LaunchCommand[] { });
-            @this.LoadSupplementalOptions(null);
-            @this.SetInitializationComplete();
-
-            return @this;
-        }
-
         /// <summary>
         /// Records for a specific portname, the remote launch was successful.
         /// </summary>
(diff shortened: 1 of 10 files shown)


src/MICore/LaunchOptions.cs
```diff
@@ -1399,17 +1374,64 @@ public static LaunchOptions GetInstance(HostConfigurationStore configStore, stri
...
         {
-            if (SourceMap == null)
+            var suppOptions = GetOptionsFromFile<Xml.LaunchOptions.SupplementalAttachOptions>("Microsoft.MIEngine.Attach.Options.xml", logger);
+            string connection;
+            ((IDebugPort2)unixPort).GetPortName(out connection);
+            AttachOptionsForConnection attachOptions = null;
+            if (suppOptions != null)
             {
-                SourceMap = new ReadOnlyCollection<SourceMapEntry>(new List<SourceMapEntry>());
+                attachOptions = suppOptions.OptionsForConnection.FirstOrDefault((o) => o.ConnectionName == connection || o.ConnectionName == "*" || o.ConnectionName == null);
```
pieandcakes: >o.ConnectionName == null [](start = 148, length = 24)

IsNullOrWhitespace? is just whitespace here a concern?

gregg-miskelly:
Is there a reason to create a new XML options file rather than extending the current one?

src/MICore/LaunchOptions.xsd
```diff
@@ -371,6 +371,102 @@
...
+      </xs:attribute>
+      <xs:attribute name="ShowDisplayString" type="xs:boolean" use="optional">
+        <xs:annotation>
+          <xs:documentation>
+            If true, instead of showing Natvis-DisplayString value as a child of a dummy element, it is shown immediately.
+            Should only be enabled if debugger is fast enough providing the value.
+          </xs:documentation>
+        </xs:annotation>
+      </xs:attribute>
+    <xs:attribute name="ConnectionName" type="xs:string" use="optional">
```
gregg-miskelly: >xs [](start = 5, length = 2)

nit: The indenting seems a bit off in parts of this. Might want to reformat it all to make it easier to read.

src/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.cs
```diff
@@ -62,11 +62,24 @@ public interface IDebugUnixShellPort
         /// <returns>Home directory of the user.</returns>
         string GetUserHomeDirectory();
 
+        /// <summary>
+        /// Attaches gdbserver to a process.
+        /// </summary>
+        /// <param name="processId">Id of the process.</param>
+        /// <param name="preAttachCommand">Command to run before starting gdbserver.</param>
+        /// <returns>Communications addr:port</returns>
+        string AttachToProcess(int processId, string preAttachCommand);
```
gregg-miskelly: >AttachToProcess [](start = 15, length = 15)

This is a COM interface, so if we are going to add new things it should be in a new interface.

src/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.cs
```diff
@@ -62,11 +62,24 @@ public interface IDebugUnixShellPort
         /// <returns>Home directory of the user.</returns>
         string GetUserHomeDirectory();
 
+        /// <summary>
+        /// Attaches gdbserver to a process.
```
gregg-miskelly: >Attaches [](start = 12, length = 8)

This feels a bit wrong to be in the port supplier to me. Is there a reason we have this here instead of the engine? (And expose out whatever port supplier functionality the engine might need to make it work).

gregg-miskelly: Oh I see, it is because liblinux has GDP server-specific stuff. Make sense.

---
In reply to: [165417036](https://github.com/Microsoft/MIEngine/pull/686#discussion_r165417036) [](ancestors = 165417036)

src/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.cs
gregg-miskelly: >AttachToProcess [](start = 15, length = 15)

I might suggest calling this 'GDBServerAttachProcess' since it isn't a generic mechanism.

src/SSHDebugPS/AD7Port.cs
```diff
@@ -216,5 +221,11 @@ public bool IsLinux()
         {
             return GetConnection(ConnectionReason.Deferred).IsLinux();
         }
+
+        public void Clean()
+        {
+            GetConnection(ConnectionReason.Deferred).Clean();
```
gregg-miskelly: >GetConnection(ConnectionReason.Deferred) [](start = 12, length = 40)

If `_connection` is null, should this actually be a nop?

src/SSHDebugPS/Connection.cs
```diff
@@ -126,6 +127,19 @@ internal string GetUserHomeDirectory()
             return _remoteSystem.FileSystem.GetDirectory(liblinux.IO.SpecialDirectory.Home).FullPath;
         }
 
+        public string AttachToProcess(int pid, string preAttachCommand)
+        {
+            var gdbStart = new liblinux.Services.GdbServerStartInfo();
+            gdbStart.ProcessId = pid;   // indicates an attach operation
+            gdbStart.PreLaunchCommand = preAttachCommand;
+            _gdbserver = _remoteSystem.Services.GdbServer.Start(gdbStart);
```
gregg-miskelly: >Start [](start = 58, length = 5)

Do you know what the error reporting is like if gdbserver fails to attach?

gregg-miskelly: >_gdbserver [](start = 12, length = 10)

Should we really be throwing if this comes back as null?

src/MICore/LaunchOptions.xsd
paulmaybee: fixed

src/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.cs
paulmaybee: fixed

paulmaybee: fixed

src/SSHDebugPS/AD7Port.cs
paulmaybee: fixed

src/SSHDebugPS/Connection.cs
paulmaybee: Start throws on failure with an explicit message.

paulmaybee: Start does this, so removed the null test and doc'ed the behavior.

PR #687, opened 2018-02-02
Fixing watson crash where Regex is throwing ArgumentNullException

pieandcakes:
Fix for Watson crash because _startPattern can be null. 

may want to guard the overall TransportLoop better too to prevent VS from crashing.

### DIFF
--- a/src/MICore/Transports/ServerTransport.cs
+++ b/src/MICore/Transports/ServerTransport.cs
@@ -36,7 +36,7 @@ namespace MICore
             proc.StartInfo.FileName = localOptions.DebugServer;
             proc.StartInfo.Arguments = localOptions.DebugServerArgs;
             proc.StartInfo.WorkingDirectory = miDebuggerDir;
-            _startPattern = localOptions.ServerStarted;
+            _startPattern = localOptions.ServerStarted; // Can be null
             _messagePrefix = Path.GetFileNameWithoutExtension(localOptions.DebugServer);
 
             InitProcess(proc, out reader, out writer);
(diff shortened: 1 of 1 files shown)


src/MICore/Transports/ServerTransport.cs
```diff
@@ -36,15 +36,26 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
...
             _messagePrefix = Path.GetFileNameWithoutExtension(localOptions.DebugServer);
 
             InitProcess(proc, out reader, out writer);
         }
 
         protected override string FilterLine(string line)
         {
-            if (!_started && Regex.IsMatch(line, _startPattern, RegexOptions.None, new TimeSpan(0, 0, 0, 0, 10) /* 10 ms */))
+            Debug.Assert(_startPattern != null, "Why is _startPattern null");
+            if (!_started && _startPattern != null && Regex.IsMatch(line, _startPattern, RegexOptions.None, new TimeSpan(0, 0, 0, 0, 10) /* 10 ms */))
```
paulmaybee: I think you want to allow startpattern to be null and treat it as a match when it is, so
(!_started && (_startPattern == null || Regex...))

pieandcakes: @paulmaybee Should I treat _startPattern the same if it is an Empty string?

paulmaybee: yes

src/MICore/Transports/ServerTransport.cs
```diff
@@ -36,15 +36,26 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
...
+            
             _messagePrefix = Path.GetFileNameWithoutExtension(localOptions.DebugServer);
 
             InitProcess(proc, out reader, out writer);
         }
 
         protected override string FilterLine(string line)
         {
-            if (!_started && Regex.IsMatch(line, _startPattern, RegexOptions.None, new TimeSpan(0, 0, 0, 0, 10) /* 10 ms */))
+            Debug.Assert(_startPattern != null, "Why is _startPattern null");
```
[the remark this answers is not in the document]
pieandcakes: I removed the first fix and added a comment that the value can be null.

PR #693, opened 2018-02-24
Fixing GDB error if ' exists in path

WardenGnaw:
Addresses issue for the C++ Extension

https://github.com/Microsoft/vscode-cpptools/issues/1554

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -815,7 +815,7 @@ namespace Microsoft.MIDebugEngine
                 throw new LaunchErrorException(message);
             };
 
-            commands.Add(new LaunchCommand("-file-exec-and-symbols \"" + exe + "\"", description, ignoreFailures: false, failureHandler: failureHandler));
+            commands.Add(new LaunchCommand("-file-exec-and-symbols " + exe, description, ignoreFailures: false, failureHandler: failureHandler));
         }
 
         private void DetermineAndAddExecutablePathCommand(IList<LaunchCommand> commands, UnixShellPortLaunchOptions launchOptions)
@@ -1321,7 +1321,7 @@ namespace Microsoft.MIDebugEngine
                 path = path.Replace(@"\", @"\\");
             }
 
-            if (path.IndexOf(' ') != -1)
+            if (path.IndexOf(' ') != -1 || path.IndexOf('\'') != -1)
             {
                 path = '"' + path + '"';
             }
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -815,7 +815,7 @@ private void AddExecutablePathCommand(IList<LaunchCommand> commands)
                 throw new LaunchErrorException(message);
             };
 
-            commands.Add(new LaunchCommand("-file-exec-and-symbols " + exe, description, ignoreFailures: false, failureHandler: failureHandler));
+            commands.Add(new LaunchCommand("-file-exec-and-symbols \"" + exe + "\"", description, ignoreFailures: false, failureHandler: failureHandler));
```
pieandcakes: Please add a comment for why we're adding quotes.

pieandcakes:
Otherwise LGTM. :shipit:

gregg-miskelly:
>             if (path.IndexOf(' ') != -1)

Your fix should be here instead

---
Refers to: src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs:1324 in b673208. [](commit_id = b673208fee7bf2d263f0216098ffcbc32890e600, deletion_comment = False)

WardenGnaw:
Would I also need to add it in https://github.com/Microsoft/MIEngine/blob/d7e8da83d1fa1e719c949bf1f192147e09752160/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs#L1335

It currently works in VsCode:
```
1: (1453) <-1012-symbol-list-lines E:\\weirdfolder'\\main.cpp
1: (1454) ->1012^done,lines=[{pc="0x00401460",line="4"},{pc="0x00401471",line="4"},{pc="0x00401476",line="5"},{pc="0x0040149b",line="6"},{pc="0x004014a0",line="7"},{pc="0x004014a8",line="0"},{pc="0x004014ba",line="7"},{pc="0x004014c0",line="7"},{pc="0x004014cf",line="0"},{pc="0x004014e5",line="7"},{pc="0x004014e8",line="7"},{pc="0x004014ee",line="7"},{pc="0x00401504",line="0"}]
```

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1321,7 +1321,7 @@ internal string EnsureProperPathSeparators(string path)
                 path = path.Replace(@"\", @"\\");
             }
 
-            if (path.IndexOf(' ') != -1)
+            if (path.IndexOf(' ') != -1 || path.IndexOf('\'') != -1)
```
gregg-miskelly: nit: use `IndexOfAny`. Also, is it worth checking other special characters? Or do we just want to fix these when people complain about them?

pieandcakes: I think we can take the tact of updating it when there is reports of other characters that it doesn't like. This is the first time I've seen this come up.

PR #695, opened 2018-03-07
Set state to stopped when continue command is aborted.

paulmaybee:
When using openocd setting up breakpoints can fail because the number of hardware supported breakpoints has been exceeded. On continue gdb will set the process as "running" but then issue an error with "Command aborted" leaving the process stopped (see below). It never sends a "stopped" even though it had issued a "running" so the debugger still thinks the process is executing. 
This change will look for the "Command aborted" string and then issue a "stopped" so that the debugger can return to break mode. It has the side effect of causing an "unhandled exception" popup. I have placed a string into the popup saying that the debugger was unable to continue the process (see below). Ideally the user wouldn't see the popup at all since there was no event. I could modify the exception handling looking for some "magic" exception that should be eaten if you think it is important enough to do.

Relevant gdb log:
1: (83623) <-1048-exec-next-instruction
1: (83635) ->1048^running
openocd: Info : no hardware breakpoint available
1: (83636) ->*running,thread-id="all"
openocd: Error: can't add breakpoint: resource not available
1: (83636) 1048: elapsed time 13
openocd: Info : no hardware breakpoint available
openocd: Error: can't add breakpoint: resource not available
1: (83638) ->&"Warning:\n"
1: (83639) ->&"Cannot insert breakpoint 7.\n"
1: (83640) ->&"Cannot access memory at address 0xbf80b5b8\n"
1: (83641) ->&"Cannot insert breakpoint 7.\n"
1: (83643) ->&"Cannot access memory at address 0xbf869f50\n"
1: (83644) ->&"\n"
1: (83644) ->1048^error,msg="Command aborted."
1: (83646) ->(gdb)
1: (83648) ->&"\n"
1: (83649) ->^done
1: (83650) ->(gdb)

Popup:
image

### DIFF
--- a/src/MICore/MICoreResources.resx
+++ b/src/MICore/MICoreResources.resx
@@ -301,4 +301,7 @@ Error: {1}</value>
   <data name="Error_CommandAborted" xml:space="preserve">
     <value>Command aborted. See the output window for additional details.</value>
   </data>
+  <data name="Info_UnableToContinue" xml:space="preserve">
+    <value>Debugger was unable to set the process running.</value>
+  </data>
 </root>
--- a/src/MICore/MICoreResources.resx
+++ b/src/MICore/MICoreResources.resx
@@ -298,4 +298,7 @@ Error: {1}</value>
     <value>Unexpected ResultClass from MI Debugger. Expected '{0}' but received '{1}'.</value>
     <comment>{0} expected value, {1} received value</comment>
   </data>
+  <data name="Error_CommandAborted" xml:space="preserve">
+    <value>Command aborted. See the output window for additional details.</value>
+  </data>
 </root>
--- a/src/MICore/MICoreResources.resx
+++ b/src/MICore/MICoreResources.resx
@@ -302,6 +302,6 @@ Error: {1}</value>
     <value>Command aborted. See the output window for additional details.</value>
   </data>
   <data name="Info_UnableToContinue" xml:space="preserve">
-    <value>Debugger was unable to set the process running.</value>
+    <value>Debugger was unable to continue the process.</value>
   </data>
 </root>
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -393,6 +393,17 @@ namespace Microsoft.MIDebugEngine
                 {
                     message = result.Results.TryFindString("message");
                 }
+                // if the command was abort (usually because of breakpoints failing to bind) then gdb writes messages into the output
+                if (this.MICommandFactory.Mode == MIMode.Gdb && message == "Command aborted.")
+                {
+                    message = MICoreResources.Error_CommandAborted;
+                    if (ProcessState == ProcessState.Running)
+                    {
+                        // assume that it was a continue command that got aborted and return to stopped state:
+                        // this occurs when using openocd to debug embedded devices and it runs out of hardware breakpoints.
+                        ScheduleStdOutProcessing(@"*stopped,reason=""exception-received"",signal-name=""SIGINT"",thread-id=""1"",exception=""""");
+                    }
+                }
                 _callback.OnError(message);
             };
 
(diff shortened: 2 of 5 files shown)


src/MICore/MICoreResources.resx
```diff
@@ -298,4 +298,10 @@ Error: {1}</value>
     <value>Unexpected ResultClass from MI Debugger. Expected '{0}' but received '{1}'.</value>
     <comment>{0} expected value, {1} received value</comment>
   </data>
+  <data name="Error_CommandAborted" xml:space="preserve">
+    <value>Command aborted. See the output window for additional details.</value>
+  </data>
+  <data name="Info_UnableToContinue" xml:space="preserve">
+    <value>Debugger was unable to set the process running.</value>
```
pieandcakes: "Debugger was unable to continue the process." or "Debugger was unable to set the process to 'run'.?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -393,6 +393,17 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
                 {
                     message = result.Results.TryFindString("message");
                 }
+                // if the command was abort (usually because of breakpoints failing to bind) then gdb writes messages into the output
+                if (this.MICommandFactory.Mode == MIMode.Gdb && message == "Command aborted.")
+                {
+                    message = MICoreResources.Error_CommandAborted;
+                    if (ProcessState == ProcessState.Running)
```
pieandcakes: Should this be ProcessState == stopped or should we add an exception handling to the response for continue instead of doing it here? I think we run the risk of getting any command abort in this state that can indicate something else is wrong with gdb. Maybe add another check to see if we know we are in this specific scenario?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -393,6 +393,17 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
...
+                    {
+                        // assume that it was a continue command that got aborted and return to stopped state:
+                        // this occurs when using openocd to debug embedded devices and it runs out of hardware breakpoints.
+                        ScheduleStdOutProcessing(string.Format(CultureInfo.CurrentCulture, @"*stopped,reason=""exception-received"",signal-name=""SIGINT"",thread-id=""1"",exception=""{0}""", MICoreResources.Info_UnableToContinue));
```
gregg-miskelly: I am not sure if you want to do more, or if you are happy enough with this solution, but I **think** the debugger package would be okay receiving an error event that was a stopping event. So if you want, you could add a new 'OnStoppingError' callback that sent the message event with the stopping attribute.

gregg-miskelly: Don't you need to send the real thread id instead of always using `1`?

paulmaybee: VS doesn't seem to want to accept a either a message event or error event as a stopping event (sending the event results in a bad arg error).

paulmaybee: Switched to using gdb's currentThread.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
paulmaybee: == Running is correct. The problem is that the "^error" arrives AFTER the command has completed with a "^running" result (see log in description above), so we can't handle this as a response to the continue command itself. We could have some kind of "last command" variable and check for a continue op, but that is a pretty ugly solution too.  I have only ever seen this behavior happen in continue/step operations. Also I have looked at the gdb sources and only see this string generated when stepping or continuing. So I think it is pretty safe to make this assumption. But happy to consider alternatives.

src/MICore/MICoreResources.resx
paulmaybee: done

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
gregg-miskelly: I think you missed changing this file.

paulmaybee: yup. Fixed.

PR #699, opened 2018-03-22
Redirect terminal output and error

WardenGnaw:
Unread stdout and stderr messages would be returned to VsCode then VsCode
would try to JSON parse the message and fail with an unexpected token.

This fix will redirect the error and output it to the logger if enabled.

### DIFF
--- a/src/MICore/TerminalLauncher.cs
+++ b/src/MICore/TerminalLauncher.cs
@@ -51,7 +51,7 @@ namespace MICore
             _isRoot = UnixNativeMethods.GetEUid() == 0;
         }
 
-        public void Launch(string workingDirectory)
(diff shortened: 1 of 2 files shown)


src/MICore/TerminalLauncher.cs
```diff
@@ -61,12 +61,26 @@ public void Launch(string workingDirectory)
                     UseShellExecute = false,
                     WorkingDirectory = workingDirectory,
                     FileName = GetProcessExecutable(),
-                    Arguments = GetProcessArgs()
+                    Arguments = GetProcessArgs(),
+                    RedirectStandardOutput = true,
```
andrewcrawley: Might be nice to have a comment explaining why this is necessary.

PR #703, opened 2018-03-26
Adding redirect for check BinarySign process

WardenGnaw:


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -178,7 +178,7 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
                 if (PlatformUtilities.IsOSX() &&
                     localLaunchOptions.DebuggerMIMode != MIMode.Clrdbg &&
                     localLaunchOptions.DebuggerMIMode != MIMode.Lldb &&
-                    !UnixUtilities.IsBinarySigned(localLaunchOptions.MIDebuggerPath))
+                    !UnixUtilities.IsBinarySigned(localLaunchOptions.MIDebuggerPath, _callback.OnOutputString, _callback.OnError))
```
gregg-miskelly: You probably don't want to use 'OnError' here. OnError is supposed to put up a message box. Granted VS Code doesn't support that currently, so it doesn't. But still, this is probably more severe than you want. You probably should want to use OnOutputString for both.

PR #705, opened 2018-03-29
UseShellExecute must be false when redirecting IO

WardenGnaw:
With the redirection, need to set UseShellExecute to false. The default
value of it is true.

Addresses https://github.com/Microsoft/vscode-cpptools/issues/1756

### DIFF
--- a/src/MICore/UnixUtilities.cs
+++ b/src/MICore/UnixUtilities.cs
@@ -184,7 +184,7 @@ namespace MICore
                 StartInfo =
                 {
                     CreateNoWindow = true,
-                    UseShellExecute = true,
+                    UseShellExecute = false,
                     FileName = CodeSignPath,
                     Arguments = "--display " + filePath,
                     RedirectStandardOutput = true,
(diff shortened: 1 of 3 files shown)


src/MICore/UnixUtilities.cs
```diff
@@ -194,12 +194,12 @@ public static bool IsBinarySigned(string filePath, Logger logger)
...
             p.OutputDataReceived += (sender, e) =>
             {
-                OutputNonEmptyString(e.Data, "stdout: ", logger);
+                OutputNonEmptyString(e.Data, "binSign-stdout: ", logger);
             };
 
             p.ErrorDataReceived += (sender, e) =>
             {
-                OutputNonEmptyString(e.Data, "stderr: ", logger);
+                OutputNonEmptyString(e.Data, "binSign-stderr: ", logger);
```
pieandcakes: Please either call this "verifySign" or "codeSign". "binSign" seems a little weird.

src/MICore/UnixUtilities.cs
```diff
@@ -208,6 +209,14 @@ public static bool IsBinarySigned(string filePath, Action<string> outputCallback
             return p.ExitCode == 0;
         }
 
+        public static void OutputNonEmptyString(string str, string prefix, Logger logger)
```
pieandcakes: Why does this need to be public? Can you make it internal ?

Issue #710, opened 2018-04-15
deadlock in WorkerThread::TrySetOperationInternal


RYDB3RG:
there seems to be a deadlock in `WorkerThread.TrySetOperationInternal`: 

![image](https://user-images.githubusercontent.com/7002848/38779238-f371bac2-40c5-11e8-96c8-6ad6c6230d81.png)

while it holds `_eventLock `, it might re-enter via `_runningOpCompleteEvent.WaitOne()` itself, thus leading to a deadlock.

repro is easy: run into breakpoint, have a non-empty watch window, toggle "Hexadecimal display"

PR #711, opened 2018-04-16
Make checking the radix async. Eliminates a deadlock in accessing the…

paulmaybee:
… WorkerThread.
Reported in #710

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
@@ -245,11 +245,14 @@ namespace Microsoft.MIDebugEngine
 
         private void EnsureLocalsAndParameters()
         {
-            uint radix = _radix;
-            Engine.DebuggedProcess.WorkerThread.RunOperation(async () =>
+            uint radix = Engine.CurrentRadix();
+            if (radix != Engine.DebuggedProcess.MICommandFactory.Radix)
             {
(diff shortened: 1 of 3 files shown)


src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
```diff
@@ -245,7 +245,11 @@ public void SetFrameInfo(enum_FRAMEINFO_FLAGS dwFieldSpec, out FRAMEINFO frameIn
 
         private void EnsureLocalsAndParameters()
         {
-            uint radix = Engine.CurrentRadix();
+            uint radix = _radix;
+            Engine.DebuggedProcess.WorkerThread.RunOperation(async () =>
```
gregg-miskelly: nit: This is kind of inefficient - we will switch to the worker thread and then immediately switch back to the main thread to obtain the display radix.

Suggestion: Split `CurrentRadix` into one function to obtain the Radix from AD7, and another to Set the radix. Something like:

Variable.cs: `await _engine.UpdateRadix(_engine.CurrentRadix());`
AD7StackFrame.cs:
```
    uint radix = Engine.CurrentRadix();
    if (radix != _debuggedProcess.MICommandFactory.Radix)
    {
       Engine.DebuggedProcess.WorkerThread.RunOperation(() => Engine.UpdateRadix(radix));
    });
```

PR #712, opened 2018-04-18
Add deploy directory to the MEF probe path

paulmaybee:
Running into MEF cache building errors after some new reference to the Linux workload were added.

### DIFF
--- a/src/MIDebugPackage/MIDebugPackagePackage.cs
+++ b/src/MIDebugPackage/MIDebugPackagePackage.cs
@@ -37,6 +37,7 @@ namespace Microsoft.MIDebugPackage
     // This attribute is needed to let the shell know that this package exposes some menus.
     [ProvideMenuResource("Menus.ctmenu", 1)]
     [Guid(GuidList.guidMIDebugPackagePkgString)]
+    [ProvideBindingPath]
     public sealed class MIDebugPackagePackage : Package, IOleCommandTarget
     {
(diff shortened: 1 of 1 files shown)


src/MIDebugPackage/MIDebugPackagePackage.cs
```diff
@@ -37,6 +37,7 @@ namespace Microsoft.MIDebugPackage
     // This attribute is needed to let the shell know that this package exposes some menus.
     [ProvideMenuResource("Menus.ctmenu", 1)]
     [Guid(GuidList.guidMIDebugPackagePkgString)]
+    [ProvideBindingPath]
```
gregg-miskelly: To make sure I understand, this is needed because some Linux workload dll is referencing some MIEngine dll?

I think this needs to be under `#if LAB`, and we should probably have a comment explaining why we have it.

gregg-miskelly:
@paulmaybee In case you don't know - Pierson is currently out. His wife just had a baby.

src/MIDebugPackage/MIDebugPackagePackage.cs
paulmaybee: Modified to:
#if LAB
    // Adding this dll location to VS probe path. Custom launcher types are referenced by Linux/Azure Sphere workloads. 
    [ProvideBindingPath]
#endif

PR #713, opened 2018-04-18
Support json initialization of a custom launcher

paulmaybee:


src/MICore/LaunchOptions.cs
```diff
@@ -2079,6 +2106,25 @@ public interface IPlatformAppLauncherSerializer : IDisposable
...
+    {
+        /// <summary>
+        /// Initializes the launcher from the json launch settings
+        /// </summary>
+        /// <param name="exePath">[Required] Path to the executable provided in the VsDebugTargetInfo by the project system. Some launchers may ignore this.</param>
+        /// <param name="args">[Optional] Arguments to the executable provided in the VsDebugTargetInfo by the project system. Some launchers may ignore this.</param>
+        /// <param name="dir">[Optional] Working directory of the executable provided in the VsDebugTargetInfo by the project system. Some launchers may ignore this.</param>
+        /// <param name="launcherJsonOptions">[Required] JObject options structure</param>
+        /// <param name="targetEngine">Indicates the type of debugging being done.</param>
+        void SetJsonLaunchOptions(string exePath, string args, string dir, object launcherJsonOptions, TargetEngine targetEngine);
```
gregg-miskelly: >object [](start = 75, length = 6)

This is always going to be JObject, correct? Should we strongly type it?

paulmaybee: Yes. Done.

gregg-miskelly: >SetJsonLaunchOptions [](start = 13, length = 20)

We are now passing Newtonsoft.Json objects across assemblies, which means we will need to make sure that both assemblies are bound to the same Newtonsoft.Json module. I would suggest adding a comment like this:

```
<!-- NOTE: MIEngine's usage of Newtonsoft.Json's flows through public interfaces exposed by MICore (IPlatformJsonAppLauncher). So this version needs to be within the binding redirect range used by devenv.exe (currently 4.5.0.0-9.0.0.0) -->
src\MICore\MICore.csproj:111:    <Reference Include="Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
```

paulmaybee:
@gregg-miskelly I have one issue with this solution. Because of the new type introduced in MICore it doesn't allow an extension that supports json launch options to also work against older versions of the MIEngine. How would you feel about reusing the existing SetLaunchOptions api and passing the JObject via the "object" parameters instead of introducing a new type? It would then be up to the extension to check the type of the parameter and throw an exception if it doesn't support json. This would allows an extension to work against both old and new versions of MIEngine. It does subtly change the requirements of a launcher implementation but I believe that the only non-MIEngine user of it today is the Azure Sphere extension. Downside is that it hides the JObject usage so we would want to leave a similar comment in place in the csproj file.

PR #716, opened 2018-05-03
Pick up correct arch folder for NDK x64

IanMatthewHuff:
VS had some blocking changes in deploying to x64 Android devices or emulators. After clearing up those issues an issue was then hit in mi-engine where it could not find the NDK GDB directory. For android it uses x86_64 for x64bit devices and ToNDKArchitectureName was returning the wrong value for this. Tested deploying to x64, x86, ARM, and ARM64 emulators.


chuckries:
LGTM

PR #719, opened 2018-05-07
CHeck architecture of target even if the user has specified a value

paulmaybee:
Customer is reporting a crash in the VS memory tool when they specified target ARM but the device was ARM64. Memory tool crashed trying to format a long address into a small buffer.
Change the target architecture determination to let the target arch check override the users specification.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -909,27 +909,25 @@ namespace Microsoft.MIDebugEngine
             // 1. if the command factory can discover the target architecture then use that
             // 2. else if the user specified an architecture then use that
             // 3. otherwise default to x64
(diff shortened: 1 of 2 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -888,51 +888,48 @@ private void DetermineAndAddExecutablePathCommand(IList<LaunchCommand> commands,
...
-                else
-                {
-                    SetTargetArch(MICommandFactory.ParseTargetArchitectureResult(""));
-                }
-            }
-            else
+            if (cmd != null)
             {
-                SetTargetArch(_launchOptions.TargetArchitecture);
+                commands.Add(new LaunchCommand(cmd, ignoreFailures: false, successHandler: successHandler));
```
gregg-miskelly: Should we ignore failures if the user specified a value? (in case there are GDBs out there that don't like to say)

PR #720, opened 2018-05-09
Adding miDebuggerArgs

WardenGnaw:
Allowing miDebugger args to be appended to miDebuggerPath process
arguments.

E.g. "--interpreter=mi" + " " + miDebuggerArgs


src/AndroidDebugLauncher/Launcher.cs
```diff
@@ -437,7 +437,7 @@ private LaunchOptions SetupForDebuggingWorker(CancellationToken token)
                 LaunchOptions launchOptions = null;
                 if (_targetEngine == TargetEngine.Native)
                 {
-                    launchOptions = new LocalLaunchOptions(_installPaths.GDBPath, string.Format(CultureInfo.InvariantCulture, ":{0}", gdbPortNumber), null);
+                    launchOptions = new LocalLaunchOptions(_installPaths.GDBPath, string.Format(CultureInfo.InvariantCulture, ":{0}", gdbPortNumber), String.Empty, null);
```
gregg-miskelly: >String.Empty [](start = 150, length = 12)

You might want to pass this along from `_launchOptions` instead

gregg-miskelly: Nevermind

---
In reply to: [187206988](https://github.com/Microsoft/MIEngine/pull/720#discussion_r187206988) [](ancestors = 187206988)

gregg-miskelly:
CC @paulmaybee

src/MICore/LaunchOptions.cs
```diff
@@ -377,13 +377,14 @@ public sealed class LocalLaunchOptions : LaunchOptions
 
         private const int DefaultLaunchTimeout = 10 * 1000; // 10 seconds
 
-        public LocalLaunchOptions(string MIDebuggerPath, string MIDebuggerServerAddress, IList<EnvironmentEntry> environmentEntries)
+        public LocalLaunchOptions(string MIDebuggerPath, string MIDebuggerServerAddress, string MIDebuggerArgs, IList<EnvironmentEntry> environmentEntries)
```
gregg-miskelly: >public LocalLaunchOptions(string MIDebuggerPath, string MIDebuggerServerAddress, string MIDebuggerArgs, IList<EnvironmentEntry> environmentEntries) [](start = 8, length = 147)

To be on the safe side, I might suggest making this a new constructor, and leave the old one in place (chaining to the new one). There are some parts of the LaunchOptions classes that are used outside of the MIEngine repo, and this may well be one of those parts. This would also let you back out the change to the Android launcher...

src/MICore/UnixUtilities.cs
```diff
@@ -74,9 +74,16 @@ public static class UnixUtilities
 
         internal static string GetDebuggerCommand(LocalLaunchOptions localOptions)
         {
+            string miDebuggerCommand = localOptions.MIDebuggerPath;
+
+            if (!String.IsNullOrEmpty(localOptions.MIDebuggerArgs))
+            {
+                miDebuggerCommand = String.Concat(miDebuggerCommand, " " + localOptions.MIDebuggerArgs);
+            }
+
```
gregg-miskelly: I don't know if this matters much, but by doing this here instead of in `LaunchLocalDebuggerCommand`, you will wind up putting these args in front of `--interpreter=mi` instead of after.

My suggestion: Find all of the spots in MIEngine where we have hard coded the string `--interpreter=mi` and replace it with a call to a helper that gets the args.

Here are the spots that I can see:
  C:\dd\MIEngine\src\MICore\LaunchOptions.cs(135):            string dbgCmdArguments = String.Format(CultureInfo.InvariantCulture, "{0} {1}", dbgPath, "--interpreter=mi");
  C:\dd\MIEngine\src\MICore\LaunchOptions.cs(791):                        startRemoteDebuggerCommand = "gdb --interpreter=mi";
  C:\dd\MIEngine\src\MICore\LaunchOptions.cs(795):                        startRemoteDebuggerCommand = "lldb-mi --interpreter=mi";
  C:\dd\MIEngine\src\MICore\Transports\LocalTransport.cs(28):            proc.StartInfo.Arguments = "--interpreter=mi";
  C:\dd\MIEngine\src\MICore\UnixUtilities.cs(58):                "{4} --interpreter=mi --tty=$DbgTerm < {1} > {2} & " +

src/MICore/LaunchOptions.cs
```diff
@@ -609,6 +633,11 @@ private static string EnsureDebuggerPath(string miDebuggerPath, string debuggerB
         /// </summary>
         public string MIDebuggerPath { get; private set; }
 
+        /// <summary>
+        /// [Required] Path to the MI Debugger Executable.
```
paulmaybee: fix copied comment

src/MICore/LaunchOptions.xsd
```diff
@@ -228,6 +228,11 @@
               <xs:documentation>Path to the MI Debugger executable. If only the executable name is provided, it will attempt to search $PATH for the location of the executable.</xs:documentation>
             </xs:annotation>
           </xs:attribute>
+          <xs:attribute name="MIDebuggerArgs" type="xs:string" use="optional">
```
paulmaybee: Add to the "ServerOptions" too.

src/MICore/Transports/LocalTransport.cs
```diff
@@ -22,10 +22,11 @@ public override void InitStreams(LaunchOptions options, out StreamReader reader,
         {
             LocalLaunchOptions localOptions = (LocalLaunchOptions)options;
             string miDebuggerDir = System.IO.Path.GetDirectoryName(localOptions.MIDebuggerPath);
+            string miDebuggerArgs = localOptions.GetMiDebuggerArgs();
 
             Process proc = new Process();
             proc.StartInfo.FileName = localOptions.MIDebuggerPath;
-            proc.StartInfo.Arguments = "--interpreter=mi";
+            proc.StartInfo.Arguments = miDebuggerArgs;
```
paulmaybee: We format this string in the PipeLaunchOptions path and the UnixShellPortOptions too. I think it would be useful in all these cases.

paulmaybee: Ah. I see we are inconsistent between the json and Xml handling of the pipe launch options. In the Json case, where we do manipulate the gdb command line explicitly, we should probably add in the new args too.

WardenGnaw: PipeLaunchOptions can modify the gdb/lldb command via pipeArguments.

UnixShellPort does not have a way to get the miDebuggerArgs since it is not a LocalLaunchOption.

src/MICore/LaunchOptions.xsd
WardenGnaw: Am I supposed to add ServerOptions to the xml? 

DebugServer, DebugServerArgs, ServerStarted, FilterStdout, FilterStderr, and ServerLaunchTimeout are already in LocalLaunchOptions

src/MICore/Transports/LocalTransport.cs
paulmaybee: right, so wouldn't the args field would need to be added to the "AttachOptions" and then merged into the construction of the LocalLaunchOptions or UnixShellPortLaunchOptions in CreateForAttachRequest?

I guess we have ways of working around the absence of a specific arg in the pipe case.

src/MICore/LaunchOptions.xsd
paulmaybee: ServerOptions is in the xml. It is used in attach scenarios. It has an "MIDebuggerPath" field. It just needs an instance of your new MIDebuggerArgs field too. Then the field needs to be merge into the final attach launch options (as appropriate) in CreateForAttachRequest  (see below).

WardenGnaw: Ah... I totally missed that field.

src/MICore/Transports/LocalTransport.cs
WardenGnaw: Wouldn't it only be for the LocalLaunchOptions since the miDebuggerArgs only exist if there if isServerMode is true?

PR #724, opened 2018-05-30
Fix bug where BPs were showing up multiple times in VS

pieandcakes:
This bug is because gdb may bind the bp in a different location in the
symbol file and the address of where that is loaded in memory changes
before and after the program is launched, causing us to send multiple
breakpoint binds without an unbind for the same breakpoint id.

Fixed so we will:
A) Clear the SourceLineCache right before we start debuggee launch in a
launch scenario
B) Requery gdb for the source line to address mappings
C) use that updated SourceLineCache to determine the correct line number
for the breakpoint
D) Send a BreakpointUnbound() message with the status of Rebind for the
old breakpoint to VS to update the UI.

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
@@ -263,7 +263,7 @@ namespace Microsoft.MIDebugEngine
                         bindTask = _engine.DebuggedProcess.AddInternalBreakAction(this.BindAsync);
                     });
 
-                    bindTask.Wait(_engine.GetBPLongBindTimeout()); 
+                    bindTask.Wait(_engine.GetBPLongBindTimeout());
                     if (!bindTask.IsCompleted)
                     {
                         //send a low severity warning bp. This will allow the UI to respond quickly, and if the mi debugger doesn't end up binding, this warning will get 
@@ -362,7 +362,7 @@ namespace Microsoft.MIDebugEngine
                 //check can bind one last time. If the pending breakpoint was deleted before now, we need to clean up gdb side
                 if (CanBind())
                 {
-                    foreach( var boundBp in _boundBreakpoints.Where((b) => b.PendingBreakpoint.BreakpointId == boundBreakpoint.PendingBreakpoint.BreakpointId).ToList())
+                    foreach (var boundBp in _boundBreakpoints.Where((b) => b.Number.Equals(boundBreakpoint.Number, StringComparison.Ordinal)).ToList())
                     {
                         _engine.Callback.OnBreakpointUnbound(boundBp, enum_BP_UNBOUND_REASON.BPUR_BREAKPOINT_REBIND);
                         _boundBreakpoints.Remove(boundBp);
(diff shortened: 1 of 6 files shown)


pieandcakes:
@richardstanton who did the original PR

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -362,6 +362,12 @@ internal AD7BoundBreakpoint AddBoundBreakpoint(BoundBreakpoint bp)
                 //check can bind one last time. If the pending breakpoint was deleted before now, we need to clean up gdb side
                 if (CanBind())
                 {
+                    foreach( var boundBp in _boundBreakpoints.Where((b) => b.PendingBreakpoint.BreakpointId == boundBreakpoint.PendingBreakpoint.BreakpointId).ToList())
```
gregg-miskelly: Is there any situation where the `Where` clause will evaluate to false?
Did multiple bound breakpoints for a single pending (ex: templates) work before? This seems like it will break it now if it worked before. Seems like we really need to add `Number` to a bound breakpoint if we want it to work.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -457,7 +457,15 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
...
 
-            BreakChangeEvent += _breakpointManager.BreakpointModified;
+            BreakChangeEvent += async delegate (object o, EventArgs args)
+            {
+                try
+                {
+                    await _breakpointManager.BreakpointModified(o, args);
+                }
+                catch (Exception)
+                { }
```
gregg-miskelly: >{ [](start = 16, length = 1)

Call `when (ExceptionHelper.BeforeCatch(e, Logger, reportOnlyCorrupting: true))`?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1613,6 +1632,9 @@ public void Close()
 
                 if (!attach)
```
gregg-miskelly: >attach [](start = 21, length = 6)

Should we be doing this someplace in attach as well?

src/MIDebugEngine/Engine.Impl/SourceLine.cs
```diff
@@ -52,6 +52,12 @@ public SourceLineCache(DebuggedProcess process)
             _process = process;
             _mapFileToLinenums = new Dictionary<string, SourceLineMap>();
         }
+
+        public void Clear()
+        {
+            this._mapFileToLinenums.Clear();
```
gregg-miskelly: >_mapFileToLinenums [](start = 17, length = 18)

Lock

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
pieandcakes: I can add that to this section and the other try/catch above too.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
pieandcakes: I don't think so because it looks like any requests happen after -target-attach is run, which means that the code is loaded in the final memory location already.

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
pieandcakes: @gregg-miskelly can you see if my attempt at the fix will work?

PR #727, opened 2018-06-02
Dev/waan/fix interop sign

WardenGnaw:


WardenGnaw:
I was testing that commit to see if it worked. I fixed it in a later commit.

PR #729, opened 2018-06-11
Fix for -break-insert main returning multiple bind points

pieandcakes:
This fix is to support -break-insert main returning multiple bind
points, which was causing a dictionary error on multiple keys.

Fix for: https://github.com/Microsoft/vscode-cpptools/issues/2056

@paulmaybee Can you verify that this change won't break your use cases?

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -767,8 +767,10 @@ namespace Microsoft.MIDebugEngine
                             {
                                 bkpt = b as TupleValue;
                             }
-                            else if (b is ValueListValue)
+                            else if (b is ValueListValue) // Used when main breakpoint binds in more than one location
                             {
+                                // Grab the first one as this is usually the <MULTIPLE> one that we can unbind them all with.
+                                // This is usually "1" when the children manifest as "1.1", "1.2", etc
                                 bkpt = (b as ValueListValue).Content[0] as TupleValue;
                             }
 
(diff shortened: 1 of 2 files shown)


pieandcakes:
@paulmaybee I tried converting them but sometimes the results come back as a string and not as a Result, so I thought best to leave it the way it is at this point.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -752,29 +757,31 @@ private List<LaunchCommand> GetInitializeCommands()
...
+                            ResultValue b = bkptResult.Find("bkpt");
+                            TupleValue bkpt = null;
+                            if (b is TupleValue)
+                            {
+                                bkpt = b as TupleValue;
+                            }
+                            else if (b is ValueListValue)
                             {
-                                this._entryPointBreakpoint = dict["number"];
+                                bkpt = (b as ValueListValue).Content[0] as TupleValue;
```
gregg-miskelly: Can you add a comment here describing when this will get used?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -752,29 +757,31 @@ private List<LaunchCommand> GetInitializeCommands()
...
+                            }
+
+                            if (bkpt != null)
+                            {
+                                this._entryPointBreakpoint = bkpt.FindString("number");
```
gregg-miskelly: Does GDB do the `.1` thing with `number` in the multiple bind case? If so, do we need to do anything else to pre-process the number before we send it to breakpoint delete?

pieandcakes:
@gregg-miskelly From other instances where we use this method, the Content[0] index will return the parent breakpoint of `"1"` where the children of it will be `"1.1" "1.2" etc`. so if we delete 1, it looks like it deletes its children too.

PR #733, opened 2018-06-14
Use -- instead of -x for gnome-terminal

dlech:
The `-x` option is deprecated, so use `--` instead.

In Ubuntu 18.04, using the `-x` option results in the program being launched in a new tab in an existing terminal. This terminal is not brought to the foreground, nor is the tab activated, which makes a bunch of work for the user to find where there program is. Using the `--` option, on the other hand, results in the program opening in a new terminal and the new terminal is brought to the foreground as one might expect.


pieandcakes:
We should verify that this doesn't affect backwards compatibility but looks good to me.

dlech:
The `--` option has been around for over 10 years if that is what you mean by backwards compatiblity.

PR #738, opened 2018-06-25
Implement dispose pattern on AD7Engine to cleanup unix port.

richardstanton:
There is a Watson failure coming from the SSHPort finalizer. There was a fix put into liblinux, but this also attempts to make cleanup more reliable.

@pieandcakes @gregg-miskelly @WardenGnaw

### DIFF
--- a/src/SSHDebugPS/AD7Port.cs
+++ b/src/SSHDebugPS/AD7Port.cs
@@ -224,8 +224,12 @@ namespace Microsoft.SSHDebugPS
 
         public void Clean()
         {
-            if (_connection != null)
-                _connection.Clean();
+            try
+            {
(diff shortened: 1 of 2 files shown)


src/SSHDebugPS/AD7Port.cs
```diff
@@ -225,7 +225,14 @@ public bool IsLinux()
         public void Clean()
         {
             if (_connection != null)
-                _connection.Clean();
+            {
+                try
+                {
+                    _connection.Clean();
```
pieandcakes: can we make this the _connection?.Clean() notation?

richardstanton: Ok

PR #740, opened 2018-06-26
Quote exe arguments that also contain '(' and ')'

pieandcakes:
gdb does not like '(' and ')' without the argument being quoted. Added
check to ensure that if the argument is already quoted we don't quote it
again.

### DIFF
--- a/src/OpenDebugAD7/MILaunchOptions.cs
+++ b/src/OpenDebugAD7/MILaunchOptions.cs
@@ -306,10 +306,10 @@ namespace OpenDebugAD7
             {
                 if (argument.IndexOfAny(s_CHARS_TO_QUOTE) >= 0)
                 {
-                    // if argument starts with a quote, escape all quotes before quoting
+                    // if argument starts with a quote, escape start and ending quotes
                     if (argument[0] == '\"')
                     {
-                        argument = argument.Replace("\"", "\\\"");
+                        argument = "\\\"" + argument.Trim('\"') + "\\\"";
                     }
 
                     return '"' + argument + '"';
--- a/src/OpenDebugAD7/MILaunchOptions.cs
+++ b/src/OpenDebugAD7/MILaunchOptions.cs
@@ -302,11 +302,16 @@ namespace OpenDebugAD7
         private static char[] s_CHARS_TO_QUOTE = { ' ', '\t', '(', ')' };
         private static string QuoteArgument(string argument)
         {
-            // Make sure its not empty or already quoted (starts with a quote)
-            if (!string.IsNullOrWhiteSpace(argument) && argument[0] != '"')
+            if (!string.IsNullOrWhiteSpace(argument))
             {
                 if (argument.IndexOfAny(s_CHARS_TO_QUOTE) >= 0)
                 {
+                    // if argument starts with a quote, escape all quotes before quoting
+                    if (argument[0] == '\"')
+                    {
+                        argument = argument.Replace("\"", "\\\"");
+                    }
+
                     return '"' + argument + '"';
                 }
             }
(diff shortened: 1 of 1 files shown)


WardenGnaw:
Any reason why we have this and Terminal.Quote?

src/OpenDebugAD7/MILaunchOptions.cs
```diff
@@ -292,12 +292,27 @@ private static string CreateArgumentList(IEnumerable<string> args)
...
             }
             return stringBuilder.ToString();
         }
 
+        // gdb does not like parenthesis without being quoted
+        private static char[] s_CHARS_TO_QUOTE = { ' ', '\t', '(', ')' };
+        private static string QuoteArgument(string argument)
+        {
+            // Make sure its not empty or already quoted (starts with a quote)
+            if (!string.IsNullOrWhiteSpace(argument) && argument[0] != '"')
```
gregg-miskelly: If an argument is something like: `"args": [ "\"Hello World\"" ]`

Shouldn't we really be giving GDB some some sort of arg that has escaped quotes, rather than raw quotes so that in the C program args[1] is `"\"Hello World\""` instead of `"Hello World"`?

src/OpenDebugAD7/MILaunchOptions.cs
```diff
@@ -292,12 +292,32 @@ private static string CreateArgumentList(IEnumerable<string> args)
...
+        // gdb does not like parenthesis without being quoted
+        private static char[] s_CHARS_TO_QUOTE = { ' ', '\t', '(', ')' };
+        private static string QuoteArgument(string argument)
+        {
+            if (!string.IsNullOrWhiteSpace(argument))
+            {
+                if (argument.IndexOfAny(s_CHARS_TO_QUOTE) >= 0)
+                {
+                    // if argument starts with a quote, escape start and ending quotes
+                    if (argument[0] == '\"')
```
gregg-miskelly: Isn't this still a problem is the args is `[ "This is a string\"With quotes\" in it." ]`? Or does GDB just not have a good way of escaping quotes that might be in it?

pieandcakes: The user would need to escape it as "\\\"" to get it to work. we can replace all instances of "\"" with "\\\"" first and then do a trim and replace.

pieandcakes: I'm worried about users who have escaped it properly today being broken if we start replacing the ones in the middle of the arguments list.

gregg-miskelly: But in the json the user already did escape it. You mean that they would need to use:

`args: [ "This is a string \\\"with quotes\\\" in it." ]`

Personally I think compat is a bad reason to leave it broken (assuming we can easily fix it).

pieandcakes: Ok. I'll make the change and see if anyone complains.

src/OpenDebugAD7/MILaunchOptions.cs
```diff
@@ -292,12 +292,28 @@ private static string CreateArgumentList(IEnumerable<string> args)
...
 
-                    stringBuilder.Append(Terminal.Quote(arg));
+                    stringBuilder.Append(QuoteArgument(arg));
                 }
             }
             return stringBuilder.ToString();
         }
 
+        // gdb does not like parenthesis without being quoted
+        private static char[] s_CHARS_TO_QUOTE = { ' ', '\t', '(', ')' };
```
gregg-miskelly: @pieandcakes shouldn't quote be in this list too?

pieandcakes: Technically because we are replacing the quote with the escaped quote, it isn't necessary. I tested it and it doesn't seem to be an issue.

PR #745, opened 2018-07-16
Change from CultureInfo.CurrentUICulture to CultureInfo.CurrentCulture

pieandcakes:
* msbuild is giving error CA1305 when using CultureInfo.CurrentUICulture
in String.Format.


pieandcakes:
@gregg-miskelly its only the nightlies that are complaining. In reading [CA1305](https://msdn.microsoft.com/en-us/library/ms182190.aspx) this seems like it only accepts CurrentCulture and InvariantCulture. I don't know if Nightlies rules got more strict recently but it started failing last week.

PR #750, opened 2018-07-30
Use VSCodeDebugAdapterHost Protocol Library

pieandcakes:
This is Andrew's changes + some fixes.


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1,392 +1,379 @@
...
+            return true;
+        }
 
-                return _configurationDoneTCS.Task;
-            });
-            RegisterSyncEventHandler(typeof(IDebugProgramDestroyEvent2), (engine, process, program, thread, eventObject) =>
+        private void SetCommonDebugSettings(Dictionary<string, JToken> args, out int sourceFileMappings)
+        {
+            // Save the Just My Code setting. We will set it once the engine is created.
+            m_sessionConfig.JustMyCode = args.GetValueAsBool("justMyCode") ?? m_sessionConfig.JustMyCode;
```
pieandcakes: The other option is to use `GetValueOrDefault()` if we don't want to do `??` notation

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1688,481 +1809,456 @@ public override Task<DebugResult> Evaluate(string context, int frameId, string e
...
+                                    Verified = false,
+                                    Id = (int)ad7BPRequest.Id,
+                                    Line = m_pathConverter.ConvertDebuggerLineToClient(ad7BPRequest.DocumentPosition.Line),
+                                    Message = string.Format(CultureInfo.CurrentCulture, AD7Resources.Error_ConditionBreakpoint, ad7BPRequest.Condition, errorMsg)
+                                };
+                            }
+                        }
+                        else
+                        {
+                            bp = new Breakpoint
```
WardenGnaw: `bp = new Breakpoint()`

WardenGnaw: new InitializeResponse
new OpenDebugStoppedEvent

src/OpenDebugAD7/VariableManager.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    internal class VariableEvaluationData
+    {
+        internal IDebugProperty2 DebugProperty;
+        internal enum_DEBUGPROP_INFO_FLAGS propertyInfoFlags;
+    }
+
+    internal class VariableManager
+    {
+        private readonly HandleCollection<Object> m_variableHandles;
```
gregg-miskelly: >private [](start = 8, length = 7)

Suggestion: Add a comment:
// NOTE: The value being stored can be a IDebugStackFrame2 or a VariableEvaluationData

src/OpenDebugAD7/packages.config
```diff
@@ -4,4 +4,7 @@
   <package id="Microsoft.VisualStudio.Debugger.Interop.Portable" version="1.0.1" targetFramework="net45" />
   <package id="MicroBuild.Core" version="0.2.0" targetFramework="net45" />
   <package id="Mono.Unofficial.pdb2mdb" version="4.2.3.4" />
+
+  <!-- Update the version number VsCodeDebugProtocol HintPath in OpenDebugAD7\OpenDebugAD7.csproj -->
```
gregg-miskelly: > [](start = 2, length = 99)

I would suggest makes a .targets file instead, since .targets files are easier to edit.

gregg-miskelly:
>                     args: JsonConvert.SerializeObject(responder.Arguments.ConfigurationProperties),

It feels inefficient to turn this into a string so that we can turn it back into an object. Any reason we are doing it that way?

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:649 in f576920. [](commit_id = f57692075bf1407da284f45e013b37063e92009a, deletion_comment = False)

gregg-miskelly:
>                     args: JsonConvert.SerializeObject(responder.Arguments.ConfigurationProperties),

I see.

---
In reply to: [409360958](https://github.com/Microsoft/MIEngine/pull/750#issuecomment-409360958) [](ancestors = 409360958)

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:649 in f576920. [](commit_id = f57692075bf1407da284f45e013b37063e92009a, deletion_comment = False)

src/OpenDebugAD7/packages.config
pieandcakes: Theres only one for now and we aren't planning on updating it that often. Leaving it as is for now. 

---
In reply to: [206675075](https://github.com/Microsoft/MIEngine/pull/750#discussion_r206675075) [](ancestors = 206675075)

gregg-miskelly:
>         protected override void HandleInitializeRequestAsync(IRequestResponder<InitializeArguments, InitializeResponse> responder)

Is there a reason you are overriding the async version of these instead of the sync version? The sync versions are a little nicer unless you need the flexibility that the async one provides.

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:528 in f576920. [](commit_id = f57692075bf1407da284f45e013b37063e92009a, deletion_comment = False)

src/OpenDebugAD7/DebugEventLogger.cs
```diff
@@ -114,40 +118,55 @@ public void Write(LoggingCategory category, string message, Dictionary<string, o
...
-        /// <summary>
-        /// Sends the message to the raw output callback if the given category has logging enabled.
-        /// </summary>
-        /// <param name="category">The category of debug event logging.</param>
-        /// <param name="message">The message to log.</param>
-        /// <param name="data">Logging data to send alongside the message, if any.</param>
-        public void WriteRaw(LoggingCategory category, string message, Dictionary<string, object> data = null)
+        public void TraceLogger_EventHandler(object sender, LogEventArgs args)
+        {
+            if (args.Message.Contains("<--  "))
```
gregg-miskelly: >Contains [](start = 29, length = 8)

Should this be 'StartsWith'? Or do you really mean for it to be anywhere in the message string?

andrewcrawley:
I can't seem to reply to Gregg's comment, but be aware that overriding `protected virtual ResponseBody HandleProtocolRequest(string requestType, object requestArgs)` to wrap requests in an exception handler will not work for requests where you override the `HandleAsync` method.

PR #751, opened 2018-08-03
Don't send thread event before program event

paulmaybee:
I have been seeing an issue where the threads window and callstack are never populated in the VS UI. This seems to be because the thread create event comes before the program event and so is ignored by the UI. Now delay sending the thread create event until after sending the program event.


paulmaybee:
ping. @gregg-miskelly @pieandcakes

PR #752, opened 2018-08-13
Modify the logic to allow terminal emulators to be installed outside `/usr/bin`

amtlib-dot-dll:
For example, one may install a newer version of GNOME Terminal or XTerm in `~/.local/bin` or somewhere else

### DIFF
--- a/src/MICore/TerminalLauncher.cs
+++ b/src/MICore/TerminalLauncher.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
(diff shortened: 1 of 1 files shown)


amtlib-dot-dll:
This is a Unix-only patch, and I don't know why the "Windows Debug" check failed

pieandcakes:
@dotnet-bot test this please

amtlib-dot-dll:
@pieandcakes Thanks

src/MICore/TerminalLauncher.cs
```diff
@@ -168,20 +168,36 @@ protected override void SetNewProcessEnvironment(ProcessStartInfo processStartIn
...
-        private const string GnomeTerminalPath = "/usr/bin/gnome-terminal";
-        private const string XTermPath = "/usr/bin/xterm";
+        private const string GnomeTerminalPath = "gnome-terminal";
+        private const string XTermPath = "xterm";
         private string _terminalPath;
         private string _bashCommandPrefix;
 
+        private bool ExistInPath(string fileName)
+        {
+            if (File.Exists(fileName))
```
pieandcakes: This check seems invalid now, unless you are checking /usr/bin/<filename>.

I think the proper thing should be to short circuit it with a /usr/bin/ and if it isn't there, then allow it to search path.

pieandcakes:
Is the goal to support specifically if the user has gnometerminal installed twice, that it uses path to determine where it is? Or are you trying to support checking path for these apps instead of only assuming the one installed is in /usr/bin?

src/MICore/TerminalLauncher.cs
amtlib-dot-dll: You are right, only but it's customary to maintain a backward compatibility 😂

amtlib-dot-dll:
Yeah, I have XTerm installed in `/app/bin/xterm` but nowhere else, so MIEngine complains.

amtlib-dot-dll:
This check may be modified some time in the future because there are complaints about the extra terminal window while Visual Studio Code have its own internal one https://github.com/Microsoft/vscode-cpptools/issues/35

pieandcakes:
I think the right fix would be to return the string that gnometerminal is found in and start the search with `/usr/bin` and fallback to `path` if it isn't found and if it is still not found then try xterm and then error.

amtlib-dot-dll:
What if someone just want to override the one installed under `/usr/bin`? `PATH=/usr/bin:/usr/local/bin` and `PATH=/usr/local/bin:/usr/bin` is totally different. The returned string is only used to spawn the terminal process and a bare executable name rather than a full path is OK for the spawning.

However, your idea will make the code more neat and descriptive.

pieandcakes:
@amtlib-dot-dll We removed the requirement for terminal clients by switching to the use of VSCode Debug Protocol's RunInTerminalRequest. This will allow you to specify the terminal client through the VS Code settings instead and will not need us to set it.

PR #754, opened 2018-08-21
Fix VS 2015 build errors

pieandcakes:
Fixing build warnings/errors for building in VS 2015 Update 8


gregg-miskelly:
Nit: I think your commit description was supposed to be 'VS 2015 Update 3'. That was the last VS 2015 update.

pieandcakes:
@gregg-miskelly I actually meant 2017. I updated the comment.

PR #756, opened 2018-09-10
User control of symbol loading

paulmaybee:
Allow users to specify loading all or no symbols and a list of solibs to load/don't load.

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -1093,6 +1093,44 @@ namespace MICore
             }
         }
 
+        private bool _siLoadAll;
+        /// <summary>
+        /// if true then load all symbols, else load no symbols
+        /// </summary>
+        public bool SymbolInfoLoadAll
+        {
+            get { return _siLoadAll;  }
+            set
+            {
+                VerifyCanModifyProperty("SymbolInfoLoadAll");
+                _siLoadAll = value;
+            }
+        }
+
+        private ReadOnlyCollection<string> _siExceptionList;
+        /// <summary>
+        /// List file names. Wildcards ('*') are allowed. Modifies behaviour of SymbolInfoLoadAll.
+        /// If SymbolInfoLoadAll is true then all symbols except for members of this list are loaded. Otherwise only members of this list are loaded.
+        /// </summary>
+        public ReadOnlyCollection<string> SymbolInfoExceptionList
+        {
+            get { return _siExceptionList; }
+            set
+            {
+                VerifyCanModifyProperty("SymbolInfoExceptionList");
+                _siExceptionList = value;
+            }
+        }
+
+        /// <summary>
+        /// Check is it is Ok to let the debugger load symbols on solib events without intervention
+        /// </summary>
+        /// <returns></returns>
+        public bool CanAutoLoadSymbols()
+        {
+            return SymbolInfoLoadAll && SymbolInfoExceptionList.Count == 0;
+        }
+
         /// <summary>
         /// If true, instead of showing Natvis-DisplayString value as a child of a dummy element, it is shown immediately.
         /// Should only be enabled if debugger is fast enough providing the value.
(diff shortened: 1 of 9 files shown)


src/MICore/LaunchOptions.xsd
```diff
@@ -630,4 +632,25 @@
...
+    <xs:simpleContent>
+      <xs:extension base="xs:string">
+        <xs:attribute name="LoadAll" type="xs:boolean" use="optional">
+          <xs:annotation>
+            <xs:documentation>If true, symbols for all libs will be loaded, otherwise no solib symbols will be loaded. Modified by ExceptionList. Default value is true.</xs:documentation>
+          </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="ExceptionList" type="xs:string" use="optional">
+          <xs:annotation>
+            <xs:documentation>List of regular expressions. Modifies behavior of LoadAll. If LoadAll is true then don't load symbols for libs that match any regexp in the list. Otherwise only load symbols for libs that match.</xs:documentation>
```
gregg-miskelly: >List of regular expressions [](start = 30, length = 27)

You might want to make these wildcards instead of regular expressions. Regular expressions are a little inconvenient to use with file paths or file names since the special RegEx characters are characters that often appear in file names (ex: '.'). Also the VS symbol options uses wildcards instead.

You can implement wildcards by using Regular expressions pretty easily. I will email you a link to the vsdbg code for dealing with include/exclude lists if you like.

paulmaybee: Now using the vsdbg include/exclude lists. Thanks.

pieandcakes: Do you need to change the comment now that its using wildcards?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -13,6 +13,7 @@
 using System.Linq;
 using System.Reflection;
 using System.Text;
+using System.Text.RegularExpressions;
```
gregg-miskelly: nit: I think you can remove this now.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -590,6 +618,9 @@ private List<LaunchCommand> GetInitializeCommands()
                 commands.Add(new LaunchCommand("-interpreter-exec console \"set pagination off\""));
             }
 
+            // When user specifies loading directives then the debugger cannot auto load symbols, the MIEngine must intervene at each solib-load event and make a determination
+            commands.Add(new LaunchCommand("-gdb-set auto-solib-add " + (_launchOptions.CanAutoLoadSymbols() ? "on" : "off")));
```
gregg-miskelly: >commands [](start = 12, length = 8)

Do you know if LLDB supports this?

gregg-miskelly:
> using System;

Need a copy right header

---
Refers to: src/MICore/IncludeExcludeList.cs:1 in 8c4af2f. [](commit_id = 8c4af2fcea6ad8cb7997544388523538031d7ef2, deletion_comment = False)

src/MICore/LaunchOptions.cs
```diff
@@ -1812,6 +1850,18 @@ protected void InitializeCommonOptions(Xml.LaunchOptions.BaseLaunchOptions sourc
...
             if (!String.IsNullOrEmpty(source.CoreDumpPath) && source.ProcessIdSpecified)
                 throw new InvalidLaunchOptionsException(String.Format(CultureInfo.InvariantCulture, MICoreResources.Error_CannotSpecifyBoth, nameof(source.CoreDumpPath), nameof(source.ProcessId)));
+
+            if (source.SymbolLoadInfo != null)
+            {
+                SymbolInfoLoadAll = source.SymbolLoadInfo.LoadAllSpecified ? source.SymbolLoadInfo.LoadAll : true;
+                SymbolInfoExceptionList.SetTo(source.SymbolLoadInfo.ExceptionList == null ? new string[0] : source.SymbolLoadInfo.ExceptionList.Split(';'));
+
+                // Ensure that symbol loading options are consistent
+                if (!WaitDynamicLibLoad && !SymbolInfoExceptionList.IsEmpty)
```
gregg-miskelly: >if [](start = 16, length = 2)

Since the LLDB command factory returns false for `SupportsStopOnDynamicLibLoad` should this be an error to include this with LLDB?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
gregg-miskelly: Since LLDB returns false from `SupportsStopOnDynamicLibLoad` seems like we shouldn't set this for LLDB.

---
In reply to: [216798249](https://github.com/Microsoft/MIEngine/pull/756#discussion_r216798249) [](ancestors = 216798249)

paulmaybee: Disabled the list for LLDB, so will either set this on or off for LLDB but not handle an exception list.

src/MICore/LaunchOptions.cs
paulmaybee: yes. done

PR #757, opened 2018-09-12
Users/piel/vscodeterminal

pieandcakes:
Support RunInTerminal in VSCode
Fixes: https://github.com/microsoft/vscode-cpptools/issues/35


pieandcakes:
Sending it out for code review early.

Outstanding:
* test on Mac
* possibly add a legacy flag/redo flag to trigger the runinterminal scenario
* pipeTransport (this will be another PR if we do it)

src/OpenDebugAD7/OpenDebugAD7.csproj
```diff
@@ -28,7 +28,7 @@
     <OutputType>Library</OutputType>
   </PropertyGroup>
   <PropertyGroup Condition="'$(IsCoreClr)' == 'false'">
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
```
gregg-miskelly: >v4.6.1 [](start = 28, length = 6)

Are you missing an `App.config` change to make sure users get a nice error message if they have an older CLR installed?

gregg-miskelly: And make sure the change makes it to your OpenDebugAD7.exe.config - it should, but good to verify.

---
In reply to: [217172144](https://github.com/Microsoft/MIEngine/pull/757#discussion_r217172144) [](ancestors = 217172144)

src/MICore/Transports/StreamTransport.cs
```diff
@@ -156,10 +156,15 @@ protected void Echo(string cmd)
         }
 
         private string GetLine()
+        {
+            return GetLineFromStream(_reader);
+        }
+
+        protected string GetLineFromStream(StreamReader reader)
```
gregg-miskelly: >string [](start = 18, length = 6)

`static`?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -187,20 +187,20 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
...
                 // For local Linux and OS X launch, use the local Unix transport which creates a new terminal and
                 // uses fifos for debugger (e.g., gdb) communication.
                 if (this.MICommandFactory.UseExternalConsoleForLocalLaunch(localLaunchOptions) &&
                     (PlatformUtilities.IsLinux() || (PlatformUtilities.IsOSX() && localLaunchOptions.DebuggerMIMode != MIMode.Lldb)))
                 {
-                    localTransport = new LocalUnixTerminalTransport();
+                    //localTransport = new LocalUnixTerminalTransport();
 
                     // Only need to clear terminal for Linux and OS X local launch
                     _needTerminalReset = (!localLaunchOptions.ProcessId.HasValue && _launchOptions.DebuggerMIMode == MIMode.Gdb);
```
gregg-miskelly: >_needTerminalReset [](start = 20, length = 18)

Do we still need this

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -187,20 +187,20 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
...
                 }
                 else
                 {
-                    localTransport = new LocalTransport();
```
gregg-miskelly: >localTransport [](start = 20, length = 14)

I think you want to put this line back, right?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -187,20 +187,20 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
                         OutputMessage.Severity.Warning));
                 }
 
-                ITransport localTransport = null;
+                ITransport localTransport = new VsCodeTerminalTransport();
                 // For local Linux and OS X launch, use the local Unix transport which creates a new terminal and
                 // uses fifos for debugger (e.g., gdb) communication.
                 if (this.MICommandFactory.UseExternalConsoleForLocalLaunch(localLaunchOptions) &&
                     (PlatformUtilities.IsLinux() || (PlatformUtilities.IsOSX() && localLaunchOptions.DebuggerMIMode != MIMode.Lldb)))
```
gregg-miskelly: Remove these checks now?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -187,20 +187,20 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
                         OutputMessage.Severity.Warning));
                 }
 
-                ITransport localTransport = null;
+                ITransport localTransport = new VsCodeTerminalTransport();
                 // For local Linux and OS X launch, use the local Unix transport which creates a new terminal and
                 // uses fifos for debugger (e.g., gdb) communication.
```
gregg-miskelly: This comment doesn't seem to make much sense anymore.

src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
```diff
@@ -430,6 +430,24 @@ public static void WriteLaunchError(string outputMessage)
...
+        /// <param name="useExternalConsole">True: External Console, False: Integrated Terminal</param>
+        /// <param name="commandArgs">the list of commands, in order, to run.</param>
+        /// <param name="environmentVariables">Environment Variables</param>
+        /// <param name="success">Success callback. In some cases, id contains the pid of the process that was started.</param>
+        /// <param name="failure">Failure callback. Contains the protocol exception that was thrown.</param>
+        /// <returns>true if the message is sent, false if not.</returns>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "cwd")]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures")]
+        public static bool TryRunInTerminal(string title, string workingDirectory, bool useExternalConsole, IReadOnlyList<string> commandArgs, IReadOnlyDictionary<string, string> environmentVariables, Action<int?> success, Action<string> failure)
```
gregg-miskelly: >bool [](start = 22, length = 4)

I don't think you want a `bool` return value here, because you need to know if this is available up front. I would suggest changing this to `void RunInTerminal` (or better yet `Task<int?> RunInTerminal` and you can get rid of the completion style API :)) and another API for deciding if the host supports a terminal request.

src/MICore/Transports/VsCodeTerminalTransport.cs
```diff
@@ -0,0 +1,305 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.DebugEngineHost;
+
+namespace MICore
+{
+    public class VsCodeTerminalTransport : StreamTransport
```
gregg-miskelly: >VsCode [](start = 17, length = 6)

Up to you. But we might want to not put 'VsCode' in these names since hopefully this feature will make it to VS as well someday...

src/WindowsDebugLauncher/DebugLauncher.cs
```diff
@@ -0,0 +1,233 @@
...
+                    || string.IsNullOrEmpty(PidPipeName)
+                    || string.IsNullOrEmpty(DbgExe));
+            }
+
+            public string DbgExeArgsAsString()
+            {
+                StringBuilder argString = new StringBuilder();
+                foreach (var arg in DbgExeArgs.ToList())
+                {
+                    argString.Append(arg);
```
gregg-miskelly: >arg [](start = 37, length = 3)

Do you need to do any quoting?

src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
```diff
@@ -430,6 +430,24 @@ public static void WriteLaunchError(string outputMessage)
         {
             throw new NotImplementedException();
         }
+
+        /// <summary>
```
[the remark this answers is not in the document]
pieandcakes: I created a HostRunInTerminal

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
pieandcakes: Sort of. If we want to keep the original behavior when RunInTerminal isn't supported, then yes?

src/WindowsDebugLauncher/DebugLauncher.cs
pieandcakes: Wasn't planning on it? but i can if it contains a space.

pieandcakes:
@gregg-miskelly @WardenGnaw I'm done tinkering with it.

I left the old Launcher code in because if RunInTerminal isnt available,  we fallback to local or Unix launchers based on the old logic

gregg-miskelly:
You need to add a 'HostRunInTerminal' class to the src/DebugEngineHost project

src/MICore/TerminalLauncher.cs
```diff
@@ -103,8 +103,7 @@ protected override string GetProcessExecutable()
         protected override string GetProcessArgs()
         {
             string thisModulePath = typeof(MacTerminalLauncher).GetTypeInfo().Assembly.ManifestModule.FullyQualifiedName;
-            string launchScript = Path.Combine(Path.GetDirectoryName(thisModulePath), LaunchTerminalScript);
-            if (!File.Exists(launchScript))
+            string launchScript = Path.Combine(Path.GetDirectoryName(thisModulePath), LaunchTerminalScript); if (!File.Exists(launchScript))
```
gregg-miskelly: You accidentally removed a new line before the `if`

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -187,10 +187,18 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
...
+                if (HostRunInTerminal.IsRunInTerminalAvailable()
+                    && String.IsNullOrWhiteSpace(localLaunchOptions.MIDebuggerServerAddress)
+                    && IsCoreDump == false)
+                {
+                    localTransport = new RunInTerminalTransport();
+                }
                 // For local Linux and OS X launch, use the local Unix transport which creates a new terminal and
                 // uses fifos for debugger (e.g., gdb) communication.
-                if (this.MICommandFactory.UseExternalConsoleForLocalLaunch(localLaunchOptions) &&
+                else if (this.MICommandFactory.UseExternalConsoleForLocalLaunch(localLaunchOptions) &&
```
gregg-miskelly: I feel like we should remove or refactor `UseExternalConsoleForLocalLaunch` since its logic is too redundant with deciding if we should use RunInTerminal transport. So we would be better off either change that method to instead decide which of the three local transports to use, or just delete it all together and move the checks that make sense here.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -760,10 +776,10 @@ private List<LaunchCommand> GetInitializeCommands()
                         commands.Add(new LaunchCommand("-environment-cd " + escapedDir));
                     }
 
-                    // TODO: The last clause for LLDB may need to be changed when we support LLDB on Linux
+                    // TODO: The last clause for LLDB may need to be changed when we support LLDB on Linux as LLDB's tty redirection doesn't work.
```
gregg-miskelly: Should this TODO statement be changed? I am assuming we actually can use LLDB on Linux now

pieandcakes: When we first invested time for 3.8.1 of LLDB I built and tested it on Linux and it "worked". I added the comment because the tty redirection doesn't seem to do anything because LLDB has a notion of lldb-launcher on Darwin only that launches a new terminal. It may be that LLDB on Linux with tty will work but I haven't rebuilt or tested it .

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -187,10 +187,18 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
                         OutputMessage.Severity.Warning));
                 }
 
-                ITransport localTransport = null;
+                ITransport localTransport;
+
+                //Attempt to support RunInTerminal first
+                if (HostRunInTerminal.IsRunInTerminalAvailable()
+                    && String.IsNullOrWhiteSpace(localLaunchOptions.MIDebuggerServerAddress)
+                    && IsCoreDump == false)
```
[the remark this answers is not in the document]
pieandcakes: RE: 1. It didn't work but we can use the same RunInTerminal mechanism to do what we were doing before so i went with that option now. 

For number 2, I'll try it again but i didn't think i had a problem with it.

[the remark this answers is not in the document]
pieandcakes: After thinking about your original question to remove the launcher code some more, it didn't make sense to keep some of it in, so i removed it all and fellback to basic localtransport instead if RunInTerminal didn't exist (which I assume will be rare.

pieandcakes:
Testing:
* Verified that we are not affecting pipeTransport on Linux/Windows
* Verified that using miDebuggerServerAddress avoids RunInTerminal
* Integrated Terminal works on Linux and Windows as does externalConsole.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -187,16 +187,22 @@ public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngin
...
-                if (this.MICommandFactory.UseExternalConsoleForLocalLaunch(localLaunchOptions) &&
-                    (PlatformUtilities.IsLinux() || (PlatformUtilities.IsOSX() && localLaunchOptions.DebuggerMIMode != MIMode.Lldb)))
+                ITransport localTransport;
+
+                // Attempt to support RunInTerminal first when it is a local launch and it is not debugging a coredump.
+                // Also since we use gdb-set new-console on in windows for external console, we don't need to RunInTerminal
+                if (HostRunInTerminal.IsRunInTerminalAvailable()
+                    && String.IsNullOrWhiteSpace(localLaunchOptions.MIDebuggerServerAddress)
+                    && IsCoreDump == false
+                    && (PlatformUtilities.IsWindows() && !localLaunchOptions.UseExternalConsole))
```
gregg-miskelly: Shouldn't this last part be:
`&& (PlatformUtilities.IsWindows() ? !localLaunchOptions.UseExternalConsole : true)

gregg-miskelly:
Do we need to test LLDB?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -760,32 +774,55 @@ private List<LaunchCommand> GetInitializeCommands()
                         commands.Add(new LaunchCommand("-environment-cd " + escapedDir));
                     }
 
-                    // TODO: The last clause for LLDB may need to be changed when we support LLDB on Linux
+                    // TODO: The last clause for LLDB may need to be changed when we support LLDB on Linux as LLDB's tty redirection doesn't work.
                     if (localLaunchOptions != null &&
-                        this.MICommandFactory.UseExternalConsoleForLocalLaunch(localLaunchOptions) &&
-                        (PlatformUtilities.IsWindows() || (PlatformUtilities.IsOSX() && this.MICommandFactory.Mode == MIMode.Lldb)))
+                        ((PlatformUtilities.IsWindows() && localLaunchOptions.UseExternalConsole)
+                        || (PlatformUtilities.IsOSX() && this.MICommandFactory.Mode == MIMode.Lldb)))
```
gregg-miskelly: This test is different than the test on line 197. Is that a problem? If so, I would suggest extract this into a helper (UseDebuggerNewConsoleOption) that can be called from both places (so the last line of 197 would be `&& !UseDebuggerNewConsoleOption(options)`).

gregg-miskelly:
>                     if (!string.IsNullOrWhiteSpace(_launchOptions.ExeArguments))

You lost `if (!string.IsNullOrWhiteSpace(_launchOptions.ExeArguments))` and gained a bunch of commented out code.

---
Refers to: src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs:797 in 0f186c2. [](commit_id = 0f186c2bf26ff84f01807dcb7adafc03bd1176bb, deletion_comment = True)

src/OpenDebugAD7/MILaunchOptions.cs
```diff
@@ -279,6 +293,47 @@ private static string FormatCommand(JsonCommand command)
...
+                }
+                else
+                {
+                    index = arg.TrimStart().IndexOf('<');
+                    if (index == 0)
+                    {
+                        isRedirected = true;
+                        break;
+                    }
+                }
```
gregg-miskelly: 1. I might suggest instead just having the set of allowed redirection suffixes prefixes: "<", ">", "1>", "2>"

gregg-miskelly:
Otherwise LGTM

src/OpenDebugAD7/MILaunchOptions.cs
pieandcakes: RE: 1. since we only allow it at the beginning, do we want to do indexof each of these?

RE: 2. if the user chose to redirect, do we still want to redirect everything else to console? I was thinking about all or none approach. I think where this might be necessary to allow a different option would be `<` or `0>` for stdin in that if it is specified, then still redirect stdout/err

PR #760, opened 2018-09-27
Select symbols tp load only on filename instead of full path

paulmaybee:
Use the filename of the solib rather then the full path in order to match which symbols to load. This is consistent with how modules are displayed in the UI and avoid issues with path differences in gdb when in gdbserver mode.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -101,7 +101,7 @@ namespace Microsoft.MIDebugEngine
             LibraryLoadEvent += delegate (object o, EventArgs args)
             {
                 ResultEventArgs results = args as MICore.Debugger.ResultEventArgs;
-                string file = results.Results.TryFindString("host-name");
+                string file = results.Results.TryFindString("id");
                 if (!string.IsNullOrEmpty(file) && MICommandFactory.SupportsStopOnDynamicLibLoad())
                 {
                     _libraryLoaded.Add(file);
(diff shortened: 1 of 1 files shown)


pieandcakes:
Could you have modules that have the same name from different paths? Will this cause both to be loaded or just one?

paulmaybee:
Yes, both will load. But I think that's probably not a problem.

paulmaybee:
The bigger issue is that you cannot do something like specify all libs in a particular directory should get loaded/not get loaded. Is that a scenario we support on Windows?

pieandcakes:
If we specify the `additionalSOLibSearchPath` property, doesn't gdb load all symbols in that path?

paulmaybee:
That adds the path as a symbol search location, but does not override the specification of specific symbols to load. That is, if you specify LoadAll as 'false' then no symbols will be loaded, even if they are found on that path.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -486,18 +486,19 @@ private async Task EnsureModulesLoaded()
                     {
                         foreach (string file in _libraryLoaded)
                         {
-                            if (_launchOptions.SymbolInfoExceptionList.Contains(file))
+                            string filename = Path.GetFileName(file);
```
gregg-miskelly: Path.GetFileName will throw if the specified file name contains any characters which aren't valid file name characters on Windows. You probably want to either catch ArgumentException or make your own helper function that does the same thing but in a less bad way.

paulmaybee: New PR submitted to address this comment.

PR #764, opened 2018-10-11
Users/piel/fixes

pieandcakes:
Two fixes.

1: Drop the new interop assembly Paul introduced
2: fix a refactoring error by me for Unix command. I accidently removed the `& clear` to force it to run in the background and clear

### DIFF
--- a/src/MIDebugPackage/MIDebugPackage.csproj
+++ b/src/MIDebugPackage/MIDebugPackage.csproj
@@ -268,6 +268,7 @@
     <DropSignedFile Include="$(OutDir)\Microsoft.MICore.XmlSerializers.dll" />
     <DropSignedFile Include="$(OutDir)\Microsoft.MIDebugEngine.dll" />
     <DropUnsignedFile Include="$(OutDir)\Microsoft.MIDebugEngine.pdb" />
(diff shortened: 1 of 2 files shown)


src/MIDebugPackage/MIDebugPackage.csproj
```diff
@@ -268,6 +268,7 @@
     <DropSignedFile Include="$(OutDir)\Microsoft.MICore.XmlSerializers.dll" />
     <DropSignedFile Include="$(OutDir)\Microsoft.MIDebugEngine.dll" />
     <DropUnsignedFile Include="$(OutDir)\Microsoft.MIDebugEngine.pdb" />
+    <DropUnsignedFile Include="$(OutDir)\Microsoft.VisualStudio.Debugger.Interop.15.0.dll" />
```
gregg-miskelly: ><DropUnsignedFile Include="$(OutDir)\Microsoft.VisualStudio.Debugger.Interop.15.0.dll" /> [](start = 4, length = 89)

Will this wind up in the CoreXT package we insert into VS? If so, do we want to add a condition here to exclude it from the MDD build?

PR #787, opened 2018-10-12
Remove MacOS from using IntegratedTerminal

pieandcakes:
This is because the capability in lldb-mi doesn't exist yet. Falling
back to LocalTransport

This is to fix: https://github.com/Microsoft/vscode-cpptools/issues/2646


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -800,7 +801,7 @@ private List<LaunchCommand> GetInitializeCommands()
                     // TODO: The last clause for LLDB may need to be changed when we support LLDB on Linux as LLDB's tty redirection doesn't work.
                     if (localLaunchOptions != null &&
                         ((PlatformUtilities.IsWindows() && localLaunchOptions.UseExternalConsole)
```
gregg-miskelly: Should we move the `&& localLaunchOptions.UseExternalConsole` to the first line since it is true on all OSs?

PR #788, opened 2018-10-12
Fix debug argument parsing

pieandcakes:
This is to fix multiple debugger arguments when using WindowsDebugLauncher.

### DIFF
--- a/src/WindowsDebugLauncher/Program.cs
+++ b/src/WindowsDebugLauncher/Program.cs
@@ -120,6 +120,9 @@ namespace WindowsDebugLauncher
             HelpMessage();
         }
 
+        /// <summary>
+        /// Parse dbgargs for spaces and quoted strings
+        /// </summary>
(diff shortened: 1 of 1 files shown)


src/WindowsDebugLauncher/Program.cs
```diff
@@ -120,6 +120,84 @@ private static void GenerateError(string flag)
...
+                            break;
+
+                        case '"':
+                            builder.Append("\"");
+                            break;
+
+                        case ' ':
+                            builder.Append(" ");
+                            break;
+                    }
```
paulmaybee: default? Do you want to just eat the escape and the following character?

paulmaybee:
What was the issue you were seeing? Is there a bug to reference?

pieandcakes:
@paulmaybee 
>What was the issue you were seeing? Is there a bug to reference?

This is when I did testing with a value in `miDebuggerArgs` and it was causing gdb to hang as it parsed it as an entire value.

PR #789, opened 2018-10-17
Fix Android visualization in VS

pieandcakes:
* Fix std::string natvis structure to properly show Size and Capacity
* Add parameter for parent's Full Name in the VisualizedView refresh message so it will refresh with the parent expression.

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7Property.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Property.cs
@@ -36,7 +36,7 @@ namespace Microsoft.MIDebugEngine
             {
                 if ((variable = _engine.DebuggedProcess.Natvis.Cache.VisualizeOnRefresh(_variableInformation)) == null)
                 {
-                    return AD7ErrorProperty.ConstructErrorPropertyInfo(dwFields, _variableInformation.Name, ResourceStrings.NoSideEffectsVisualizerMessage, this);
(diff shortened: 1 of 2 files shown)


pieandcakes:
Fixes bug: 677524

pieandcakes:
Tested with Android 5.1 and 8

src/MIDebugEngine/AD7.Impl/AD7Property.cs
```diff
@@ -488,10 +488,15 @@ public int GetParent(out IDebugProperty2 ppParent)
...
+        public static DEBUG_PROPERTY_INFO ConstructErrorPropertyInfo(enum_DEBUGPROP_INFO_FLAGS dwFields, string name, string error, IDebugProperty2 prop, string varFullName)
         {
-            DEBUG_PROPERTY_INFO property = new DEBUG_PROPERTY_INFO(); ;
-
+            DEBUG_PROPERTY_INFO property = new DEBUG_PROPERTY_INFO();
+            // Add the parent fullname to fullname so when it is refreshed, it will evaluate.
+            if ((dwFields & enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_FULLNAME) != 0)
+            {
+                property.bstrFullName = varFullName;
+                property.dwFields |= enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_FULLNAME;
```
paulmaybee: Should flag be set when varFullName==null?

PR #759, opened 2018-09-18
Adding list-features and check for async 

amocz:
Added ListTargetFeatures() to check for async status.
If async enabled, send EnableTargetAsyncOption().
If async disabled, don't send.

PR #794, opened 2018-10-19
Revert "Adding list-features and check for async  (#759)"

pieandcakes:
This reverts commit b04d5618164b7efe079b3c2378f843374b13e0f5.

This is being reverted as it hangs in the VS & Android case. gdb for Android supports async but doesnt' list it as capabilities. It is noticed when trying to stop debugging while running.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -579,11 +579,7 @@ namespace Microsoft.MIDebugEngine
 
             try
             {
-                Results res = await this.MICommandFactory.ListTargetFeatures();
-                if (res.Contains("async"))
-                {
-                    await this.MICommandFactory.EnableTargetAsyncOption();
-                }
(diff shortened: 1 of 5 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -579,11 +579,7 @@ public async Task Initialize(HostWaitLoop waitLoop, CancellationToken token)
 
             try
             {
-                Results res = await this.MICommandFactory.ListTargetFeatures();
-                if (res.Contains("async"))
-                {
-                    await this.MICommandFactory.EnableTargetAsyncOption();
-                }
+                await this.MICommandFactory.EnableTargetAsyncOption();
```
WardenGnaw: Add a note why we can't do the check

Trass3r: This check failed because it was done too early.
Needs to come after the initialize commands.

PR #796, opened 2018-10-22
Update locals test as BP bind is different

pieandcakes:
For some reason, gdb binds breakpoint 3 (main.cpp:49) at 49, but when it
is hit, it shows as main.cpp:59.


pieandcakes:
I rolled back to prior to the protocol change and it has the same failures.

pieandcakes:
@chuckries I'll add that into the issue also

PR #801, opened 2018-10-29
Translate octal codes to unicode chars assuming utf8

paulmaybee:
When octal literals are found embedded in a string literal decode them assuming a utf8 encoding.

This replaces the buggy octal literal translation from last PR.

### DIFF
--- a/src/MICore/MIResults.cs
+++ b/src/MICore/MIResults.cs
@@ -899,6 +899,7 @@ namespace MICore
             }
             int i = input.Start + 1;
             bool endFound = false;
+
             for (; i < input.Extent; i++)
             {
                 char c = _resultString[i];
@@ -929,7 +930,12 @@ namespace MICore
                         default:
                             if (c >= '0' && c <= '3')
                             {
-                                c = DecodeOctalLiteral(_resultString, i-1, ref i);
+                                i = i - 1;
+                                if (SpanOctalChars(_resultString, ref i, output))
+                                {
+                                    continue;   // handled the output of the octal-encoded chars
+                                }
+                                c = _resultString[i]; // just emit the '\\'
                             }
                             break;
                     }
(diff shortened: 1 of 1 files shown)


src/MICore/MIResults.cs
```diff
@@ -945,29 +951,51 @@ private ConstValue ParseCString(Span input, out Span rest)
...
+        /// <param name="output"></param>
+        /// <returns></returns>
+        bool SpanOctalChars(string str, ref int i, StringBuilder output)
         {
-            if (i+2 >= str.Length)
+            int s = i;
+            bool error = false;
+            int cChars = 0;
+            byte[] bytes = new byte[str.Length];
+            while (!error && i + 3 < str.Length && str[i] == '\\' && str[i+1] >= '0' && str[i+1] <= '3')
```
pieandcakes: Can you add a comment on what the significance the '3' is?

src/MICore/MIResults.cs
```diff
@@ -929,7 +930,12 @@ private ConstValue ParseCString(Span input, out Span rest)
                         default:
                             if (c >= '0' && c <= '3')
```
gregg-miskelly: >3 [](start = 50, length = 1)

Should this be '7' instead of '3'?

gregg-miskelly: Oh, I see. Because 255 in octal is "377"

src/MICore/MIResults.cs
```diff
@@ -945,29 +951,51 @@ private ConstValue ParseCString(Span input, out Span rest)
...
+                    {
+                        v = (v << 3) + (c - '0');
+                    }
+                    else
+                    {
+                        error = true;
+                        continue;
+                    }
                 }
+                bytes[cChars++] = (byte)v;
```
gregg-miskelly: Might be worth an assert that v < 255.

src/MICore/MIResults.cs
pieandcakes: nm. I get it now based on Gregg's comment.

src/MICore/MIResults.cs
```diff
@@ -945,29 +951,51 @@ private ConstValue ParseCString(Span input, out Span rest)
...
+                i += 4;
+            }
+            if (error)
```
gregg-miskelly: I think this check should be `if (error || cChars == 0)`. As if the input is something like `\01` the success case will wind up decrementing from `i`, which we don't want.

PR #802, opened 2018-10-29
Fix stability issues when an Exception is thrown in WindowsDebugLauncher

pieandcakes:
* Don't add console redirection on attach on Windows
* Add try/catch blocks and a close() call in WindowsDebugLauncher.

### DIFF
--- a/src/WindowsDebugLauncher/DebugLauncher.cs
+++ b/src/WindowsDebugLauncher/DebugLauncher.cs
@@ -71,7 +71,7 @@ namespace WindowsDebugLauncher
 
         private LaunchParameters _parameters;
 
-        private bool isRunning = true;
+        private bool _isRunning = true;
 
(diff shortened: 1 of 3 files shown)


src/WindowsDebugLauncher/DebugLauncher.cs
```diff
@@ -73,14 +73,16 @@ public string ParametersAsString()
...
-        StreamReader _npCommandStream;
-        StreamWriter _npErrorStream;
-        StreamWriter _npOutputStream;
-        StreamWriter _npPidStream;
+        private StreamReader _npCommandStream;
+        private StreamWriter _npErrorStream;
+        private StreamWriter _npOutputStream;
+        private StreamWriter _npPidStream;
+
+        private Process dbgProcess;
```
gregg-miskelly: nit: Member variables should start with a `_`.

PR #803, opened 2018-11-02
Run MakePIAPortable to remove calls

pieandcakes:
This is to workaround a mono issue where it does not like 'internalcall'
in PIAs.

### DIFF
--- a/src/MIDebugPackage/MIDebugPackage.csproj
+++ b/src/MIDebugPackage/MIDebugPackage.csproj
@@ -299,10 +299,10 @@
     <DropUnsignedFile Include="$(OutDir)\Microsoft.MICore.dll.mdb" />
     <DropUnsignedFile Include="$(OutDir)\Microsoft.MIDebugEngine.dll.mdb" />
   </ItemGroup>
-  <!--<ItemGroup Condition="'$(IsCoreCLR)' == 'false' OR '$(Lab)' == 'false'">
-    --><!-- Include file for local VS development builds and all non-VS builds --><!--
+  <ItemGroup Condition="'$(IsCoreCLR)' == 'true' AND '$(Lab)' == 'false'">
+    <!-- Include file for local VS development builds. For Non-VS Builds, it will be dropped by MakePIAPortable project -->
     <DropUnsignedFile Include="$(OutDir)\Microsoft.VisualStudio.Debugger.Interop.15.0.dll" />
-  </ItemGroup>-->
+  </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\MICore\MICore.csproj">
       <Project>{54c33afa-438d-4932-a2f0-d0f2bb2fadc9}</Project>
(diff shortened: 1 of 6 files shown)


src/MIDebugPackage/MIDebugPackage.csproj
```diff
@@ -299,10 +299,10 @@
     <DropUnsignedFile Include="$(OutDir)\Microsoft.MICore.dll.mdb" />
     <DropUnsignedFile Include="$(OutDir)\Microsoft.MIDebugEngine.dll.mdb" />
   </ItemGroup>
-  <ItemGroup Condition="'$(IsCoreCLR)' == 'false' OR '$(Lab)' == 'false'">
-    <!-- Include file for local VS development builds and all non-VS builds -->
+  <!--<ItemGroup Condition="'$(IsCoreCLR)' == 'false' OR '$(Lab)' == 'false'">
```
gregg-miskelly: Remove rather than just commenting out?

src/OpenDebugAD7/OpenDebugAD7.csproj
```diff
@@ -126,10 +126,10 @@
     <DropSignedFile Condition="'$(IsCoreClr)' == 'false'" Include="$(OutputPath)\OpenDebugAD7.exe" />
     <DropUnsignedFile Condition="'$(IsCoreClr)' == 'false'" Include="$(OutputPath)\OpenDebugAD7.exe.config" />
     <DropUnsignedFile Condition="'$(IsCoreClr)' == 'false'" Include="$(OutputPath)\OpenDebugAD7.exe.mdb" />
-    <DropUnsignedFile Condition="'$(IsCoreClr)' == 'false'" Include="$(OutputPath)\Microsoft.VisualStudio.Debugger.Interop.10.0.dll" />
-    <DropUnsignedFile Condition="'$(IsCoreClr)' == 'false'" Include="$(OutputPath)\Microsoft.VisualStudio.Debugger.Interop.11.0.dll" />
-    <DropUnsignedFile Condition="'$(IsCoreClr)' == 'false'" Include="$(OutputPath)\Microsoft.VisualStudio.Debugger.Interop.12.0.dll" />
-    <DropUnsignedFile Condition="'$(IsCoreClr)' == 'false'" Include="$(OutputPath)\Microsoft.VisualStudio.Debugger.InteropA.dll" />
+    <!-- <DropUnsignedFile Condition="'$(IsCoreClr)' == 'false'" Include="$(OutputPath)\Microsoft.VisualStudio.Debugger.Interop.10.0.dll" /> -->
```
gregg-miskelly: Remove instead of just commenting out?

src/MIDebugPackage/MIDebugPackage.csproj
pieandcakes: @gregg-miskelly I commented out because my assumption is that we'll put them back when `mono` resolves their issue. I can remove it if you think it would be better?

gregg-miskelly: If you would rather leave them, that is fine. But in that case I would suggest adding a comment to indicate why you left it. Or you could just take it out since you can revert this change entirely at that point.

pieandcakes: I fixed this one so it still copies it out in non-lab regular builds. Otherwise I removed it.

PR #806, opened 2018-11-09
Load natvis from OpenFolder projects

paulmaybee:


src/DebugEngineHost/HostNatvisProject.cs
```diff
@@ -9,6 +9,14 @@
 using System.Threading;
 using System.Threading.Tasks;
 using System.Collections.Generic;
+using Microsoft.VisualStudio.Workspace;
+using Microsoft.VisualStudio.Workspace.Indexing;
+using Microsoft.VisualStudio.Workspace.VSIntegration.Contracts;
+using Microsoft.VisualStudio.ComponentModelHost;
```
pieandcakes: nit: please sort using statements.

src/DebugEngineHost/HostNatvisProject.cs
```diff
@@ -73,6 +81,84 @@ internal enum VSENUMPROJFLAGS
...
+                        workspaceService = new Lazy<IVsFolderWorkspaceService>(() =>
+                        {
+                            return workspaceServices.First().Value;
+                        });
+                    }
+                    return workspaceService;
+                }
+            }
+
+            private static int GetOpenFolderSourceLocations(string bstrFileName, out bool isIndexCompete, out Array pSourcesArray)
```
pieandcakes: should this be `Complete`?

src/DebugEngineHost/HostNatvisProject.cs
```diff
@@ -73,6 +81,84 @@ internal enum VSENUMPROJFLAGS
...
+                    Lazy<IVsFolderWorkspaceService> workspaceService = null;
+
+                    IComponentModel componentModel = ServiceProvider.GlobalProvider.GetService(typeof(SComponentModel).GUID) as IComponentModel;
+                    IEnumerable<Lazy<IVsFolderWorkspaceService>> workspaceServices = componentModel.DefaultExportProvider.GetExports<IVsFolderWorkspaceService>();
+
+                    if (workspaceServices != null && workspaceServices.Count() == 1)
+                    {
+                        workspaceService = new Lazy<IVsFolderWorkspaceService>(() =>
+                        {
+                            return workspaceServices.First().Value;
```
pieandcakes: could it export an an empty IEnumerable? If so then the First() call will throw.

src/DebugEngineHost/HostNatvisProject.cs
```diff
@@ -73,6 +81,84 @@ internal enum VSENUMPROJFLAGS
...
 
+            /// <summary>
+            /// Gets the WorkspaceService from Microsoft.VisualStudio.Workspace
+            /// </summary>
+            /// <remarks>This package won't be automatically loaded by MEF, so we need to manually acquire exported MEF Parts.</remarks>
+            private static Lazy<IVsFolderWorkspaceService> WorkspaceService
+            {
+                get
+                {
+                    Lazy<IVsFolderWorkspaceService> workspaceService = null;
```
pieandcakes: Is this going to be lazy if the backing field is within the get property? Or am i not understanding the pattern? It seems like with this getter, every time WorkspaceService is accessed, it will call the serviceProvider and get the workplaceService again.

src/DebugEngineHost/HostNatvisProject.cs
pieandcakes: Also, why not make the Array an IEnumerable<string> instead?

src/DebugEngineHost/HostNatvisProject.cs
```diff
@@ -81,6 +167,22 @@ public static void FindNatvisInSolutionImpl(List<string> paths)
...
+                bool isOpenFolderActive = (bool)openFolderMode;
+
+                if (isOpenFolderActive)
+                {
+                    bool isIndexComplete;
+                    Array filenames;
+                    GetOpenFolderSourceLocations(".natvis", out isIndexComplete, out filenames);
+                    foreach (var f in filenames)
+                    {
+                        paths.Add((string)f);
```
pieandcakes: If you use IEnumerable<string> for filenames, then you can also use paths.AddRange(filenames) here.

src/DebugEngineHost/packages.config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.VisualStudio.ComponentModelHost" version="15.0.0" targetFramework="net45" />
```
pieandcakes: I don't think you need this file. I don't think we build this assembly, or we should not be, on Desktop.* builds.

src/DebugEngineHost/HostNatvisProject.cs
paulmaybee: which is why the count is checked.

pieandcakes: Ah, missed that part. sorry!

src/DebugEngineHost/HostNatvisProject.cs
```diff
@@ -73,6 +81,74 @@ internal enum VSENUMPROJFLAGS
...
+                    if (indexWorkspaceService != null)
+                    {
+                        isIndexComplete = indexWorkspaceService.State == IndexWorkspaceState.Completed;
+
+                        var findFilesService = indexWorkspaceService as IFindFilesService;
+                        if (findFilesService != null)
+                        {
+                            FindFileServiceProgress progress = new FindFileServiceProgress();
+                            string toSearch = Path.GetFileName(bstrFileName);
+                            using (var cancellationTokenSource = new CancellationTokenSource())
```
gregg-miskelly: There is no reason to create a cancellation token source if we are never going to cancel it.

src/DebugEngineHost/HostNatvisProject.cs
```diff
@@ -73,6 +81,74 @@ internal enum VSENUMPROJFLAGS
...
+                            {
+                                ThreadHelper.JoinableTaskFactory.Run(async () =>
```
gregg-miskelly: It feels like we should be making this function async instead so that this code can run off the UI thread.

src/DebugEngineHost/HostNatvisProject.cs
```diff
@@ -73,6 +81,74 @@ internal enum VSENUMPROJFLAGS
...
+                isIndexComplete = false;
+                var workspaceService = GetWorkspaceService();
+
+                if (workspaceService != null)
+                {
+                    IWorkspace currentWorkspace = workspaceService.CurrentWorkspace;
+                    IIndexWorkspaceService indexWorkspaceService = currentWorkspace?.GetService<IIndexWorkspaceService>(throwIfNotFound: false);
+                    if (indexWorkspaceService != null)
+                    {
+                        isIndexComplete = indexWorkspaceService.State == IndexWorkspaceState.Completed;
```
gregg-miskelly: If the index is not up to date, do we want to log a warning to the output window?

PR #808, opened 2018-11-26
Fix typo

changeworld:
ture -> true


pieandcakes:
@changeworld thank you for your contribution!

Issue #807, opened 2018-11-22
terminal stays open after process exit


payonel:
Host OS: Ubuntu 16.04 LTS

I believe vscode's cpptools extension uses this project for starting a terminal window. But since a recent update the terminal window doesn't close when the process exists. From our perspective, we definitely do not want "Press any key to continue.." holding these zombie terminals open (we get a new one with each run/debug run)

Another user created a ticket against cpptools, but apparently found that this is due to this project
https://github.com/Microsoft/vscode-cpptools/issues/2838

Please work with the cpptools team and provide them a way to not keep these terminal windows open

payonel:
on vscode stable version, it is not showing "Press any key", but the terminal still stays open. Perhaps their stable version is using an older version of this engine

payonel:
are you saying that MIEngine is _not_ keeping the terminal open?

`"Press any key to continue..."`

https://github.com/Microsoft/MIEngine/blob/29c6569861b2042c43f04d8d78b2566db2e2ef1f/src/OpenDebugAD7/OpenDebug/Terminal.cs#L17

`read -p {6} -n1`

https://github.com/Microsoft/MIEngine/blob/29c6569861b2042c43f04d8d78b2566db2e2ef1f/src/OpenDebugAD7/OpenDebug/Terminal.cs#L135

What am I missing? I don't see ANY options for VSCODE to NOT use your `read` stopper

pieandcakes:
@payonel that is legacy code that was ported over that's actually never called. I will remove that file as it isn't used.

PR #809, opened 2018-11-26
Remove old code

pieandcakes:


pieandcakes:
Removing old code. https://github.com/Microsoft/MIEngine/issues/807#issuecomment-441833065

PR #824, opened 2019-01-24
Updating liblinux.il for VS 2019

pieandcakes:


pieandcakes:
@paulmaybee For the Dev16 issues, I've filed a bug with MSBuild for the compiler path. I talked to someone on the Open Folder team about the VS.Workspace dll dependency and they are aware of their lack of proper authoring on NuGet .org but they don't have a timeline for fixes.

For the liblinux IL issue, I spoke with Elizabeth and Ben and they will be taking the Roslyn Public API checker and if they change anything that will break this contract, they will generate and submit a new IL file to this repo.

pieandcakes:
@gregg-miskelly @paulmaybee I'm moving the Dev16 fixes out of this PR as it seems like the liblinux change might not be going in yet.

PR #828, opened 2019-01-28
Fix localization of SSH provider resource dlls.

richardstanton:
Fix the path to the lcl/lci files to enable proper creation of satellite assemblies.

@pieandcakes @gregg-miskelly @WardenGnaw


pieandcakes:
I asked richard to target release_mdd directly and we will do a reverse merge.

PR #831, opened 2019-01-31
Introducing Docker PS for dotnet core

pieandcakes:
@jander-msft


src/MIDebugPackage/MIDebugPackage.csproj
```diff
@@ -8,12 +8,17 @@
     <!--Visual Studio will prompt to upgrade the project unless $(MinimumVisualStudioVersion)==$(VisualStudioVersion), to make
     VS happy, we will conditionally set MinimumVisualStudioVersion based on what version is opening the project -->
     <MinimumVisualStudioVersion Condition="'$(VisualStudioVersion)' == '15.0'">15.0</MinimumVisualStudioVersion>
-    <MinimumVisualStudioVersion Condition="'$(VisualStudioVersion)' != '15.0'">14.0</MinimumVisualStudioVersion>
+    <MinimumVisualStudioVersion Condition="'$(VisualStudioVersion)' != '15.0'">16.0</MinimumVisualStudioVersion>
```
pieandcakes: change back to 14.0

src/MIDebugEngine/Microsoft.MIDebugEngine.pkgdef
```diff
@@ -46,5 +45,11 @@
 "CLSID"="{0fc2f352-2fc1-4f80-8736-51cd1ab28f16}"
 "GlobalVisualizersDirectory"="$PackageFolder$"
 
+[$RootKey$\AD7Metrics\Engine\{91744D97-430F-42C1-9779-A5813EBD6AB2}\PortSupplier]
+; SSH Port Supplier
+"0"="{3FDDF14E-E758-4695-BE0C-7509920432C9}"
+; Docker Port Supplier
```
pieandcakes: Comment out gdb for Docker as we currently are not supporting it for v1

src/MICore/Debugger.cs
```diff
@@ -712,7 +712,7 @@ public Task CmdBreakInternal()
                 {
                     if (_debuggeePids.Count > 0)
                     {
-                        debuggeePid = _debuggeePids.First().Value;
+                        debuggeePid = _transport.DebuggerPid; // _debuggeePids.First().Value;
```
WardenGnaw: Do we still need this comment?

src/SSHDebugPS/AD7/AD7UnixAsyncCommand.cs
```diff
@@ -0,0 +1,122 @@
...
+using System;
+using System.Threading;
+using Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier;
+
+namespace Microsoft.SSHDebugPS
+{
+    internal class AD7UnixAsyncCommand : IDebugUnixShellAsyncCommand
+    {
+        private readonly object _lock = new object();
+        private int _bClosed = 0;
```
WardenGnaw: Why is this an `int`?

src/SSHDebugPS/RawShell.cs
```diff
@@ -0,0 +1,295 @@
...
+        private void ReadLoop(StreamReader reader, CancellationToken token, Action<string> action)
+        {
+            try
+            {
+                while (true)
+                {
+                    char[] buffer = new char[BUFMAX];
+                    Task<int> task = reader.ReadAsync(buffer, 0, buffer.Length);
+                    task.Wait(token);
+                    //if (task.Result == 0)
```
WardenGnaw: Accidental left over comments?

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -0,0 +1,251 @@
...
+        /// </summary>
+        /// <param name="pipeTransportSettings">Settings</param>
+        /// <param name="outerConnection">[Optional] the SSH connection (or maybe something else in future) used to connect to the target.</param>
+        /// <param name="name">The full name of this connection</param>
+        /// <param name="container">The name of the container. For local, this is the same as 'name'.</param>
+        public PipeConnection(IPipeTransportSettings pipeTransportSettings, Connection outerConnection, string name, int? timeout)
+        {
+            Debug.Assert(pipeTransportSettings != null);
+            Debug.Assert(!string.IsNullOrEmpty(name));
+            //Debug.Assert(!string.IsNullOrEmpty(container));
```
WardenGnaw: Left over comment

src/SSHDebugPS/AD7/AD7UnixAsyncCommand.cs
pieandcakes: This was an int before. I didn't change the structure of this command

src/MIDebugPackage/MIDebugPackage.csproj
gregg-miskelly: You can use this to make sure VS doesn't put up the bogus upgrade prompts --

```
    <!--Visual Studio will prompt to upgrade the project unless $(MinimumVisualStudioVersion)==$(VisualStudioVersion), to make
    VS happy, we will conditionally set MinimumVisualStudioVersion based on what version is opening the project -->
    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">14.0</VisualStudioVersion>
    <MinimumVisualStudioVersion>$(VisualStudioVersion)</MinimumVisualStudioVersion>
```



---
In reply to: [252509323](https://github.com/Microsoft/MIEngine/pull/831#discussion_r252509323) [](ancestors = 252509323)

src/MIDebugPackage/MIDebugPackage.csproj
```diff
@@ -8,12 +8,17 @@
...
     <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">14.0</VisualStudioVersion>
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <TargetFrameworkProfile />
     <OutputPath>$(MIDefaultOutputPath)</OutputPath>
+    <FileUpgradeFlags>
+    </FileUpgradeFlags>
+    <UpgradeBackupLocation>
+    </UpgradeBackupLocation>
+    <OldToolsVersion>14.0</OldToolsVersion>
```
gregg-miskelly: Remove

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -1,31 +1,68 @@
...
+                // If SSH connection dialog was cancelled, we should cancel this connection.
+                if (remoteConnection == null)
+                    return null;
+                
+                settings = new DockerExecShellSettings(containerName, isUnix: true); // assume all remote is Unix for now.
+                displayName = remoteConnection.Name + '/' + containerName;
+            }
+            else
+            {
+                throw new ArgumentException("Argument format is incorrect");
```
gregg-miskelly: >ArgumentException("Argument format is incorrect"); [](start = 26, length = 50)

Does this get shown to the user?

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -1,31 +1,68 @@
...
+            // Assume format is <server>/<container> where if <server> is specified, it is for SSH
+            string[] connectionStrings = name.Split('/');
+
+            string displayName;
+            string containerName;
+            if (connectionStrings.Length == 1)
+            {
+                // local connection
+                containerName = connectionStrings[0];
+                settings = new DockerExecShellSettings(containerName, isUnix: false);
```
gregg-miskelly: >isUnix [](start = 70, length = 6)

I am assuming this is supposed to indicate if the host is unix or Windows? If so, I might suggest naming this `isHostUnix`. If not, this is wrong - Windows can host Linux containers. I would assume it is the most common option.

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -0,0 +1,251 @@
...
+        protected IPipeTransportSettings TransportSettings => _settings;
+        protected bool IsClosed => _isClosed;
+
+        /// <summary>
+        /// Create a new pipe connection object
+        /// </summary>
+        /// <param name="pipeTransportSettings">Settings</param>
+        /// <param name="outerConnection">[Optional] the SSH connection (or maybe something else in future) used to connect to the target.</param>
+        /// <param name="name">The full name of this connection</param>
+        /// <param name="container">The name of the container. For local, this is the same as 'name'.</param>
```
gregg-miskelly: >container [](start = 25, length = 9)

This method doesn't take a 'container' argument. But it does time a timeout.

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -0,0 +1,251 @@
...
+            Debug.Assert(pipeTransportSettings != null);
+            Debug.Assert(!string.IsNullOrEmpty(name));
+            //Debug.Assert(!string.IsNullOrEmpty(container));
+
+            _name = name;
+            _settings = pipeTransportSettings;
+            if (outerConnection != null)
+            {
+                _outerConnection = outerConnection;
+                DefaultTimeout = 15000;
```
gregg-miskelly: >15000 [](start = 33, length = 5)

Do we need a progress dialog anywhere? 15 seconds is long enough for people to think things are hung.

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -0,0 +1,251 @@
...
+            {
+                DefaultTimeout = timeout.HasValue ? timeout.Value : 5000;
+            }
+
+            _shellExecutionManager = new ShellExecutionManager(CreateShellFromSettings(_settings, _outerConnection));
+        }
+
+        public override void BeginExecuteAsyncCommand(string commandText, bool runInShell, IDebugUnixShellCommandCallback callback, out IDebugUnixShellAsyncCommand asyncCommand)
+        {
+            throw new NotImplementedException();
```
gregg-miskelly: >throw new NotImplementedException(); [](start = 12, length = 36)

Is there a reason we override this just to throw NotImplementedException? Should we add a comment?

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -0,0 +1,110 @@
...
+            string tmpFile = null;
+
+            if(!Directory.Exists(source) && !File.Exists(source))
+            {
+                throw new ArgumentException(FormattableString.Invariant($"Local path: '{source}' does not exist"), nameof(source));
+            }
+
+            if (OuterConnection != null)
+            {
+                tmpFile = "/tmp" + "/" + Path.GetRandomFileName();
```
gregg-miskelly: >/tmp [](start = 27, length = 4)

I feel like '/tmp' is not the right place for this since it is a shared namespace, and we aren't doing anything to protect it. Maybe use `$XDG_RUNTIME_DIR` and fallback to ~/.cache if that doesn't exist.

gregg-miskelly: >Path.GetRandomFileName() [](start = 41, length = 24)

I don't think using Path.GetRandomFileName is random enough. Just use a GUID? Also we probably want some prefix to identify the app.

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        /// <summary>
+        /// Docker uses its own command system to copy files and directories into and out of the container.
+        /// </summary>
+        /// <param name="sourcePath"></param>
+        /// <param name="destinationPath"></param>
+        public override void CopyDirectory(string sourcePath, string destinationPath)
+        {
+            Copy(sourcePath, destinationPath);
```
gregg-miskelly: >Copy [](start = 12, length = 4)

Do we need to slap on '-r' if we are copying a directory?

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            if(!Directory.Exists(source) && !File.Exists(source))
+            {
+                throw new ArgumentException(FormattableString.Invariant($"Local path: '{source}' does not exist"), nameof(source));
+            }
+
+            if (OuterConnection != null)
+            {
+                tmpFile = "/tmp" + "/" + Path.GetRandomFileName();
+                OuterConnection.CopyFile(source, tmpFile);
```
gregg-miskelly: >CopyFile [](start = 32, length = 8)

Don't you need to Copy either a file or a directory here?

src/SSHDebugPS/PSOutputParser.cs
```diff
@@ -74,36 +56,36 @@ public string Extract(string line)
             }
         }
 
-        private const string UserNamePrefix = "CurrentUserName: ";
-        private const string PSCommandLine = "ps -axww -o pid=A,ruser=B,args=C";
-        public const string CommandText = "echo " + UserNamePrefix + "$USER; " + PSCommandLine;
+        // Use padding to expand column width. 10 for pid and 32 for userid
+        private const string PSCommandLineFormat = "ps{0}-o pid=ppppppppppp -o ruser=rrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr -o args";
```
andrewcrawley: You might want to add some notes here as to the problems you ran into with various "ps" implementations and the reason you settled on doing it this way so whoever touches this next doesn't have to rediscover it.

src/SSHDebugPS/PSOutputParser.cs
```diff
@@ -74,36 +56,36 @@ public string Extract(string line)
...
         private string _currentUserName;
         private ColumnDef _pidCol;
         private ColumnDef _ruserCol;
         private ColumnDef _argsCol;
 
-        public static List<Process> Parse(string output)
+        private static string PSCommandLine = String.Format(CultureInfo.InvariantCulture, PSCommandLineFormat, " -axww ");
+        private static string AltPSCommandLine = String.Format(CultureInfo.InvariantCulture, PSCommandLineFormat, " ");
+        public static string CommandText = PSCommandLine;
+        public static string AltCommandText = AltPSCommandLine;
```
andrewcrawley: Can you get rid of these and make the other set public (or vice versa)

src/SSHDebugPS/PSOutputParser.cs
```diff
@@ -133,48 +118,44 @@ private List<Process> ParseInternal(string output)
...
-            // pid column is right justified so the pid column stops at the 'A' index
-            if (headerLine[index] != 'A')
-                return false;
 
-            _pidCol = new ColumnDef(0, index);
+            // pid column is right justified so the pid column stops at the last letter index
+            while (index < strLen && !char.IsWhiteSpace(headerLine[index]))
+            {
+                index++;
+            }
```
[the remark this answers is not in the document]
andrewcrawley: I think this is doing the opposite - moving the "index" forward until it hits a non-whitespace char.  Could extract this out as "ConsumeNonWhitespace" or similar, or replace this entire function with a regex.

---
In reply to: [253710450](https://github.com/Microsoft/MIEngine/pull/831#discussion_r253710450) [](ancestors = 253710450)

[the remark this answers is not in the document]
gregg-miskelly: Ah. +1

---
In reply to: [253735238](https://github.com/Microsoft/MIEngine/pull/831#discussion_r253735238) [](ancestors = 253735238,253710450)

src/MICore/Debugger.cs
pieandcakes: Sorry, this was a customer investigation that i was doing. Customer says we should be sending signal to GDB and not the debuggee and that would pause the debuggee. I was making sure it worked. I need to move this into its own PR.

src/SSHDebugPS/AD7/AD7Port.cs
```diff
@@ -1,26 +1,24 @@
...
-    internal class AD7Port : IDebugPort2, IDebugUnixShellPort, IDebugPortCleanup, IDebugGdbServerAttach, IConnectionPointContainer, IConnectionPoint
+    internal abstract class AD7Port : IDebugPort2, IDebugUnixShellPort, IDebugPortCleanup, IConnectionPointContainer, IConnectionPoint
     {
         private readonly object _lock = new object();
         private readonly AD7PortSupplier _portSupplier;
-        private string _name;
+        protected string _name;
         private readonly Lazy<Guid> _id = new Lazy<Guid>(() => Guid.NewGuid(), LazyThreadSafetyMode.ExecutionAndPublication);
-        private Connection _connection;
+        protected Connection _connection;
```
andrewcrawley: >protected Connection _connection; [](start = 8, length = 33)

Having this as a protected field is kind of weird - how are derived classes supposed to know to set it in their implementation of GetConnection? Unless there's a good reason not to, I'd have the base class handle the caching.

src/SSHDebugPS/Docker/DockerConnection.cs
andrewcrawley: `$XDG_CACHE_HOME` resolves to ~/.cache by default, maybe use that and fall back to /tmp for non-interactive users that don't have a home directory?

---
In reply to: [253714264](https://github.com/Microsoft/MIEngine/pull/831#discussion_r253714264) [](ancestors = 253714264)

src/SSHDebugPS/ConnectionManager.cs
andrewcrawley: Might consider including the correct format in the error message so users know what to do, e.g. "Expected 'container-id' or 'remote-host/container-id'"

---
In reply to: [253703674](https://github.com/Microsoft/MIEngine/pull/831#discussion_r253703674) [](ancestors = 253703674)

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -0,0 +1,251 @@
...
+            _name = name;
+            _settings = pipeTransportSettings;
+            if (outerConnection != null)
+            {
+                _outerConnection = outerConnection;
+                DefaultTimeout = 15000;
+            }
+            else
+            {
+                DefaultTimeout = timeout.HasValue ? timeout.Value : 5000;
```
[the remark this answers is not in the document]
andrewcrawley: Or `timeout ?? 5000`

---
In reply to: [253705530](https://github.com/Microsoft/MIEngine/pull/831#discussion_r253705530) [](ancestors = 253705530)

src/SSHDebugPS/RawShell.cs
```diff
@@ -0,0 +1,295 @@
...
+        public event EventHandler<string> OutputReceived;
+        public event EventHandler<int> Closed;
+        public event EventHandler ErrorOccured;
+
+        private static int BUFMAX = 4096;
+        private void ReadLoop(StreamReader reader, CancellationToken token, Action<string> action)
+        {
+            try
+            {
+                while (true)
```
andrewcrawley: >while (true) [](start = 16, length = 12)

What causes this loop to terminate?  The only "return" is commented out.  Does the `ReadAsync` call throw after the underlying command exits?  If so, won't every command execution end with an error event being raised?

src/SSHDebugPS/RawShell.cs
```diff
@@ -0,0 +1,295 @@
...
+            {
+                _localProcess.Close();
+            }
+
+            if (_stdoutWriter != null)
+            {
+                _stdoutWriter.Close();
+                _stdoutWriter = null;
+            }
+        }
```
andrewcrawley: I've found it very helpful to have a `DisposableObject` base class that implements `IDisposable` correctly and provides utilities like `EnsureNotDisposed` so you don't have to write that test everywhere.  Richard wrote a good implementation that we're using in PZ if you want it.

src/SSHDebugPS/RawShell.cs
```diff
@@ -0,0 +1,295 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using liblinux;
+using liblinux.Shell;
+using Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier;
+
+namespace Microsoft.SSHDebugPS
+{
+    public interface IRawShell : IDisposable
```
andrewcrawley: >IRawShell [](start = 21, length = 9)

I find this name very confusing, since many of the actual implementations of this interface don't involve shells.  Maybe `IRawProcess` or `IProcess`?

src/SSHDebugPS/RawShell.cs
```diff
@@ -0,0 +1,295 @@
...
+            if (_isRunning)
+            {
+                _asyncCommand.Abort();
+                _isRunning = false;
+            }
+        }
+
+        public void Write(string text)
+        {
+            if (_isRunning)
```
andrewcrawley: >if (_isRunning) [](start = 12, length = 15)

Could be factored out into an `EnsureRunning` method so you don't have duplicated logic in each method body

src/SSHDebugPS/RawShell.cs
```diff
@@ -0,0 +1,295 @@
...
+
+namespace Microsoft.SSHDebugPS
+{
+    public interface IRawShell : IDisposable
+    {
+        event EventHandler<string> OutputReceived;
+        event EventHandler<int> Closed;
+        event EventHandler ErrorOccured;
+
+        void WriteCommandStart(string startCommand);
```
andrewcrawley: >void WriteCommandStart(string startCommand); [](start = 8, length = 44)

It looks like all implementations of this just call `WriteLine` - remove from interface?

src/SSHDebugPS/RawShell.cs
```diff
@@ -0,0 +1,295 @@
...
+        private void OnProcessExited(object sender, EventArgs e)
+        {
+            Closed?.Invoke(this, _localProcess.ExitCode);
+        }
+    }
+
+    /// <summary>
+    /// Launches a raw command that doesn't have shell-like input/output
+    /// </summary>
+    internal class CommandShell : IRawShell
```
andrewcrawley: >CommandShell [](start = 19, length = 12)

Is the difference between this and RawLocalShell that RawLocalShell only provides full lines as output, and this will also provide data that doesn't end with a newline?

src/SSHDebugPS/RawShell.cs
```diff
@@ -0,0 +1,295 @@
...
+            _localProcess = System.Diagnostics.Process.Start(processStartInfo);
+            _localProcess.Exited += OnProcessExited;
+            _localProcess.EnableRaisingEvents = true;
+
+            _stdoutWriter = new StreamWriter(_localProcess.StandardInput.BaseStream, new UTF8Encoding(encoderShouldEmitUTF8Identifier: false), 4096, leaveOpen: true);
+            _processReader = _localProcess.StandardOutput;
+            _processError = _localProcess.StandardError;
+
+            Thread outputThread = new Thread(() => ReadLoop(_processReader, _cancellationSource.Token, (msg) => { OutputReceived?.Invoke(this, msg); }));
+            Thread errorThread = new Thread(() => ReadLoop(_processError, _cancellationSource.Token, (msg) => { ErrorOccured?.Invoke(this, null); }));
```
andrewcrawley: >null [](start = 139, length = 4)

Why are you just throwing away stderr output?

src/SSHDebugPS/ShellExecutionManager.cs
```diff
@@ -0,0 +1,94 @@
...
+            AD7UnixAsyncShellCommand command = new AD7UnixAsyncShellCommand(_shell, commandCallback, false);
+
+            try
+            {
+                _currentCommand = command;
+                _currentCommand.Start(commandText);
+                if (!_commandCompleteEvent.WaitOne(timeout))
+                {
+                    commandOutput = "Command Timeout";
+                    return 55;
```
andrewcrawley: >55 [](start = 27, length = 2)

Is this value meaningful?  If it really needs to be '55', I'd make it a constant somewhere.  Otherwise, maybe use a constant that already exists, like ERROR_TIMEOUT?

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -0,0 +1,110 @@
...
+            if (runInShell)
+            {
+                AD7UnixAsyncShellCommand command = new AD7UnixAsyncShellCommand(CreateShellFromSettings(TransportSettings, OuterConnection), callback, closeShellOnComplete: true);
+                command.Start(commandText);
+
+                asyncCommand = command;
+            }
+            else
+            {
+                /// TODO: figure out what RunInShell means?
```
andrewcrawley: >RunInShell [](start = 42, length = 10)

I added the `runInShell` parameter to allow consumers to run a command directly rather than having to start a shell and have the shell run the command to avoid the output mangling that sometimes entails.

src/SSHDebugPS/Docker/DockerTransportSettings.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        private static string WindowsExe => "docker.exe";
+        private static string UnixExe => "docker";
+
+        #region IPipeTransportSettings
+        public abstract string ExeCommandArgs { get; }
+
+        public string ExeCommand => IsUnix ? UnixExe : WindowsExe;
+
+        public string ExeNotFoundErrorMessage => string.Format(CultureInfo.InvariantCulture, "{0} not found.", IsUnix ? UnixExe : WindowsExe);
```
andrewcrawley: >IsUnix ? UnixExe : WindowsExe [](start = 111, length = 29)

Replace with `ExeCommand`?

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        public override string MakeDirectory(string path)
+        {
+            Debug.Assert(!string.IsNullOrWhiteSpace(path), nameof(path));
+
+            if (!DoesCommandExist("mkdir"))
+            {
+                throw new CommandFailedException("mkdir");
+            }
```
andrewcrawley: I think this `DoesCommandExist` pattern is a bad idea in general.  Just try to run the command and handle "doesn't exist" like any other failure: by presenting the error text to the user.

src/SSHDebugPS/IConnection.cs
```diff
@@ -0,0 +1,82 @@
...
+        #endregion
+
+        public bool DoesCommandExist(string command)
+        {
+            int exitCode = ExecuteCommand("which " + command, DefaultTimeout, out string commandOutput);
+
+            if (exitCode != 0 || string.IsNullOrWhiteSpace(commandOutput))
+                return false;
+
+            return true;
```
andrewcrawley: I don't think this works universally.  POSIX doesn't specify the `which` command, and I *think* the MacOS implementation doesn't return a non-zero result if the command doesn't exist, it just prints a message.

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -0,0 +1,251 @@
...
+namespace Microsoft.SSHDebugPS
+{
+    internal abstract class PipeConnection : Connection
+    {
+        private readonly object _lock = new object();
+        private readonly IPipeTransportSettings _settings;
+        private readonly Connection _outerConnection;
+
+        private readonly ShellExecutionManager _shellExecutionManager;
+        private readonly LinkedList<IRawShell> _shellList = new LinkedList<IRawShell>();
```
andrewcrawley: >LinkedList [](start = 64, length = 10)

Why is this a `LinkedList<T>` instead of just `List<T>`?

src/SSHDebugPS/SSH/SSHConnection.cs
```diff
@@ -86,12 +105,17 @@ internal void CopyFile(string sourcePath, string destinationPath)
             _remoteSystem.FileSystem.UploadFile(sourcePath, destinationPath);
         }
 
+        public override void CopyDirectory(string sourcePath, string destinationPath)
+        {
+            throw new NotImplementedException();
```
andrewcrawley: >throw new NotImplementedException(); [](start = 12, length = 36)

Why is this not implemented?  I believe `IRemoteFileSystem.UploadDirectory` will do it.

src/SSHDebugPS/ConnectionManager.cs
pieandcakes: No it doesn't. I'm returning null (no connection) for now and added a comment that we will use the dialog in this instance.

src/SSHDebugPS/PipeConnection.cs
pieandcakes: I think it was left from when i had it non-abstract. i'll remove it.

src/SSHDebugPS/Docker/DockerConnection.cs
pieandcakes: Removing the CopyDirectory as we never actually use it and we can implement it in the future if needed.

src/SSHDebugPS/RawShell.cs
pieandcakes: In the dispose method i call Stream.Close(). I'll change the loop to be `while(!token.IsCancellationRequested)

---
In reply to: [254079310](https://github.com/Microsoft/MIEngine/pull/831#discussion_r254079310) [](ancestors = 254079310)

src/SSHDebugPS/PipeConnection.cs
pieandcakes: Andrew pointed out that we should allow the user to cancel and set the timeout to something high. I'm looking at making that happen.

---
In reply to: [253705761](https://github.com/Microsoft/MIEngine/pull/831#discussion_r253705761) [](ancestors = 253705761)

src/SSHDebugPS/ConnectionManager.cs
pieandcakes: I missed one. I had the other places where it was renamed. I'll fix it.

---
In reply to: [253704386](https://github.com/Microsoft/MIEngine/pull/831#discussion_r253704386) [](ancestors = 253704386)

src/SSHDebugPS/Docker/DockerConnection.cs
pieandcakes: I'm planning on using `Microsoft.VisualStudio.DockerPS.FileCopy.<GUID>` and leaving it in the `/tmp` directory. 

---
In reply to: [254075353](https://github.com/Microsoft/MIEngine/pull/831#discussion_r254075353) [](ancestors = 254075353,253714264)

src/SSHDebugPS/IConnection.cs
pieandcakes: removed

---
In reply to: [254106158](https://github.com/Microsoft/MIEngine/pull/831#discussion_r254106158) [](ancestors = 254106158)

src/SSHDebugPS/RawShell.cs
pieandcakes: This one doesn't do readline and uses a readloop witha buffer instead. basically avoiding the readline implementation for the above.

---
In reply to: [254097210](https://github.com/Microsoft/MIEngine/pull/831#discussion_r254097210) [](ancestors = 254097210)

pieandcakes: I need a better name for these two classes but i don't know what to call it. The other option is to allow a boolean and implement both in the same class?

---
In reply to: [256226539](https://github.com/Microsoft/MIEngine/pull/831#discussion_r256226539) [](ancestors = 256226539,254097210)

pieandcakes:
@andrewcrawley @gregg-miskelly I think I addressed everything.

src/MICore/Debugger.cs
gregg-miskelly: I think you forgot to fix this

src/SSHDebugPS/CommandRunner.cs
```diff
@@ -0,0 +1,281 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using liblinux;
+using liblinux.Shell;
+using Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier;
+
+namespace Microsoft.SSHDebugPS
+{
+    public interface ICommandRunner : IDisposable
```
gregg-miskelly: >public [](start = 4, length = 6)

Should these be public from this assembly?

src/SSHDebugPS/CommandRunner.cs
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        private void OnErrorOutput(object sender, DataReceivedEventArgs e)
+        {
+            ErrorOccured?.Invoke(sender, new ErrorOccuredEventArgs(new Exception(e.Data)));
+        }
+
+        private void OnProcessOutput(object sender, DataReceivedEventArgs e)
+        {
+            OutputReceived?.Invoke(this, e?.Data + '\n');
```
gregg-miskelly: >e? [](start = 41, length = 2)

Could `e` actually be null?

src/SSHDebugPS/Microsoft.SSHDebugPS.pkgdef
```diff
@@ -2,14 +2,24 @@
 "CLSID"="{1326FB0D-EA51-4F90-BEDF-5948588B0FE1}"
 "Name"="SSH"
 
+[$RootKey$\AD7Metrics\PortSupplier\{A2BBC114-47E4-473F-A49C-69EE89711243}]
+"CLSID"="{18F976B7-D66D-4C36-9CAE-A9217E0E3DF4}"
+"Name"="Docker"
```
gregg-miskelly: >Docker [](start = 8, length = 6)

I wonder if we should call this `Docker (Linux)` to make it clear that it doesn't support Windows scenarios?

pieandcakes: I was thinking about how to name it to specify that we support only Docker Linux containers for now.

Docker (Linux) might lead users to believe Docker On Linux?

andrewcrawley: Could go with "Docker (Linux Container)" or something, to indicate that you can attach to any linux-based container, regardless of the host OS.

pieandcakes:
@dotnet-bot test this please

pieandcakes:
@gregg-miskelly Are you ok with this going in?

PR #832, opened 2019-01-31
Set encoding with UTF8 with UTF8Identifier false

pieandcakes:
Users asking for this fix: https://github.com/Microsoft/vscode-cpptools/issues/2816


pieandcakes:
I don't see any issue with taking this change, but I'll see if @gregg-miskelly thinks differently. I think this is the fix that the commenters in the above issue are requesting.

pieandcakes:
@gregg-miskelly seems like it works for both.

PR #834, opened 2019-02-04
Send LoadCompleteEvent from MIEngine. Add event logging.

paulmaybee:
Multi-process debugging was failing and VS was hanging because the debugger failed to Stop all processes before attempting to request data (stack info).  The stop behavior is triggered in the SDM by the LoadCompleteEvent from the debug engine. This fix sends the LoadCompleteEvent on process startup.

Also adding event logging messages to the engine log.

See https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems/edit/776626

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/EngineCallback.cs
+++ b/src/MIDebugEngine/Engine.Impl/EngineCallback.cs
@@ -29,7 +29,7 @@ namespace Microsoft.MIDebugEngine
         {
             uint attributes;
             Guid riidEvent = new Guid(iidEvent);
-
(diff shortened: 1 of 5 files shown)


src/MIDebugEngine/Engine.Impl/EngineCallback.cs
```diff
@@ -241,7 +241,7 @@ public void OnAsyncBreakComplete(DebuggedThread thread)
 
         public void OnLoadComplete(DebuggedThread thread)
```
gregg-miskelly: I would suggest removing the `thread` argument since it will always be null.

PR #835, opened 2019-02-05
Don't log log messages

paulmaybee:
Check that event is not an outputWindw mesage before logging it. Last checkin broke OutputWindow logging.


paulmaybee:
Unfortunately if someone tries to log to the output window VS will die with a stack overflow. A lot of people know to use logging now to diagnose issue, so it would be best if we got it fixed before we released bit to the world. That said, it is a diagnostics feature rather than core functionality that is affected. Guess I would do a bar check with Charlie.

PR #841, opened 2019-03-20
Support environment for pipelaunchoptions

paulmaybee:
environment variables were only supported for local launch options. now supported for all launch options.


src/MICore/LaunchOptions.cs
```diff
@@ -1877,6 +1871,8 @@ protected void InitializeCommonOptions(Xml.LaunchOptions.BaseLaunchOptions sourc
                     throw new InvalidLaunchOptionsException(MICoreResources.Error_InvalidSymbolInfo);
                 }
             }
+
+            this.Environment = new ReadOnlyCollection<EnvironmentEntry>(GetEnvironmentEntries(source.Environment) ?? new List<EnvironmentEntry>());
```
WardenGnaw: ?? will never return false since GetEnvironmentEntries seems to always return a non-null.

src/MICore/LaunchOptions.cs
```diff
@@ -1898,6 +1894,8 @@ public void InitializeLaunchOptions(Json.LaunchOptions.LaunchOptions launch)
             {
                 this.LaunchCompleteCommand = launch.LaunchCompleteCommand.Value;
             }
+
+            this.Environment = new ReadOnlyCollection<EnvironmentEntry>(GetEnvironmentEntries(launch.Environment) ?? new List<EnvironmentEntry>());
```
WardenGnaw: ?? will never return false since GetEnvironmentEntries seems to always return a non-null.

PR #843, opened 2019-04-02
Add Azure-Pipeline support and update version

WardenGnaw:


build/version.settings.targets
```diff
@@ -2,9 +2,9 @@
   <PropertyGroup>    
     <!--SxS: These three properties should be changed at the start of a new version, VersionZeroYear should be the year
     before the start of the project. When updating the version, also update MIEngine\metadata.json.-->
-    <MajorVersion>14</MajorVersion>
-    <MinorVersion>0</MinorVersion>
-    <VersionZeroYear>2013</VersionZeroYear>
+    <MajorVersion>16</MajorVersion>
```
gregg-miskelly: NOTE: This will change the MIEngine assemblies' AssemblyVersion -- `<AssemblyVersion Condition="'$(AssemblyVersion)'==''">$(MajorVersion).$(MinorVersion).0</AssemblyVersion>`.

But there are a few MIEngine assemblies that are used outside of this repo -- I *think* just MICore @paulmaybee is there anything else I am missing?

I believe you want to add something like this to MICore:
C:\dd\MIEngine\src\DebugEngineHost.Stub\DebugEngineHost.Stub.csproj(6):    <AssemblyVersion>1.0.0</AssemblyVersion>

But setting the version to 14.0.0

paulmaybee: MICore is used in the Azure Sphere VS Extension. No others AFAIK.

gregg-miskelly: Does anyone remember if there is some sort of dependency from the Linux project system?

PR #847, opened 2019-04-10
Temporarily ignore libasan source file information

paulmaybee:
This change causes the MIEngine to ignore source line information coming from libasan (AddressSanitizer). The purpose is to stop VS from popping up a source dialog when an ASAN error occurs. It is seen as a feedback blocker for the first demonstration of ASAN integrated into VS. We plan to revisit the implementation post Update1 and will remove this temporary kludge at that point.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
@@ -313,7 +313,19 @@ namespace Microsoft.MIDebugEngine
         private ThreadContext CreateContext(TupleValue frame)
         {
             ulong? pc = frame.TryFindAddr("addr");
-            MITextPosition textPosition = MITextPosition.TryParse(this._debugger, frame);
+
+            // don't report source line info for modules marked as IgnoreSource
+            DebuggedModule module = null;
+            if (pc != null)
+            {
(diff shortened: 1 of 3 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -313,7 +313,19 @@ private async Task<List<ThreadContext>> WalkStack(DebuggedThread thread)
         private ThreadContext CreateContext(TupleValue frame)
         {
             ulong? pc = frame.TryFindAddr("addr");
-            MITextPosition textPosition = MITextPosition.TryParse(this._debugger, frame);
+
+            // don't report source line info for modules marked as IgnoreSource
+            DebuggedModule module = null;
```
gregg-miskelly: nit: I had to read this code a few times to understand what it was doing. How about --

```
            // don't report source line info for modules marked as IgnoreSource
            bool ignoreSource = false;
            if (pc != null)
            {
                module = _debugger.FindModule(pc.Value);
                if (module != null && module.IgnoreSource)
                {
                    ignoreSource = true;
                }
            }
            MITextPosition textPosition = !ignoreSource ? MITextPosition.TryParse(this._debugger, frame) : null;
```

PR #851, opened 2019-04-25
VSOperationWaiter MPF to JTF

WardenGnaw:
There is a hanging issue with VC tools in SSH attach due to VSOperationWaiter not using JTF. 

This PR upgrades the VSOperationWaiter to use JTF by upgrading shell to 15.0 which requires Framework v4.7.2


gregg-miskelly:
Is there a reason you need to update to v4.7.2? Updating to 15 is probably reasonable at this point, but v4.7.2 means that you are exceeding the version of .NET pre-req'ed by VS 15, which would affectively make this a Dev16-only product.

WardenGnaw:
@gregg-miskelly Updating the Shell to use 15.0 and the new framework lib, pulled in a dependency that was compiled with 4.7.2 and the current fix was to upgrade the framework.

gregg-miskelly:
Do you know what dependency that was? From what I recall, VS 2017 didn't pre-req 4.7.2 -- that was a Dev16 thing. So I would guess (maybe incorrectly?) that as along as we can switch to the VS 2017 version of whatever that is, we should be able to stay on something older than 4.7.2.

WardenGnaw:
> The primary reference "MIEngine\src\packages\Microsoft.VisualStudio.Shell.Framework.15.6.27413\lib\net45\Microsoft.VisualStudio.Shell.Framework.dll" could not be resolved because it has an indirect dependency on the assembly "Microsoft.Build.Framework, Version=15.1.0.0, Culture=neutral, PublicKeyToken=...." which was built against the ".NETFramework,Version=v4.7.2" framework. This is a higher version than the currently targeted framework ".NETFramework,Version=v4.6.2"

gregg-miskelly:
https://www.nuget.org/packages/Microsoft.Build.Framework/15.6.85 targets 'net46'. Are you building in Dev16? If so, I suspect it is picking up a Dev16 version of msbuild from that and causing your problem. Also, do you need the 15.6 version of things? You might want to use 15.0 if you just need 15.x.

WardenGnaw:
Is there a way when building on Dev16 to use the net46 target?

pieandcakes:
you can update nuget to pull in shell 15.0.26606 . I suspect its finding the one installed in VS 2019 and is using that version which is bringing up the 4.7.2 requirement. I ran into this problem before i had left.

src/SSHDebugPS/VSOperationWaiter.cs
```diff
@@ -1,21 +1,14 @@
...
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.IO;
-using System.Linq;
-using System.Runtime.InteropServices;
-using System.Text;
-using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.VisualStudio.Threading;
+
```
pieandcakes: remove extra line

src/SSHDebugPS/VSOperationWaiter.cs
```diff
@@ -30,59 +23,35 @@ internal static class VSOperationWaiter
...
                         if (throwOnCancel)
                         {
                             throw;
                         }
-
-                        return false;
                     }
 
-                    if (cancellationTokenSource.IsCancellationRequested)
+                    if (canellationToken.IsCancellationRequested)
```
pieandcakes: spelling

src/SSHDebugPS/SSHDebugPS.csproj
```diff
@@ -68,7 +68,7 @@
     <Reference Include="Microsoft.VisualStudio.OLE.Interop, Version=7.1.40304.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
       <Private>False</Private>
     </Reference>
```
pieandcakes: I think you removed the only reference to this in this project. you can probably remove this now

src/SSHDebugPS/VSOperationWaiter.cs
```diff
@@ -1,21 +1,13 @@
...
-using System.Diagnostics;
-using System.IO;
-using System.Linq;
-using System.Runtime.InteropServices;
-using System.Text;
-using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.VisualStudio.Threading;
+using Microsoft.VisualStudio.Shell;
 using Task = System.Threading.Tasks.Task;
```
pieandcakes: you have two empty lines below

pieandcakes:
@paulmaybee is added to the review. #Closed

src/DebugEngineHost/HostNatvisProject.cs
```diff
@@ -52,7 +52,7 @@ public static string FindSolutionRoot()
             string path = null;
             try
             {
-                ThreadHelper.Generic.Invoke(() => { path = Internal.FindSolutionRootImpl(); });
+                ThreadHelper.JoinableTaskFactory.Run(async () => { path = Internal.FindSolutionRootImpl(); });
```
AArnott: This is *not* equivalent. 
`ThreadHelper.Generic.Invoke` invokes the delegate on the UI thread.
`JoinableTaskFactory.Run` invokes the delegate on the caller's thread.
If you need `FindSolutionRootImpl` to run on the main thread, you should change it to this:

```cs
ThreadHelper.JoinableTaskFactory.Run(async () => { 
   await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
   path = Internal.FindSolutionRootImpl();
});
```

src/SSHDebugPS/VSOperationWaiter.cs
```diff
@@ -30,39 +22,21 @@ internal static class VSOperationWaiter
...
-            {
-                // Visual Studio is not installed on this box
-            }
-
-            if (waiterImpl != null)
-            {
-                using (ManualResetEvent completeEvent = new ManualResetEvent(false))
+            ThreadHelper.JoinableTaskFactory.Run(
+                actionName,
+                async (progress, cancellationTokenSource) =>
```
AArnott: rename `cancellationTokenSource` -> `cancellationToken`

src/SSHDebugPS/VSOperationWaiter.cs
```diff
@@ -71,18 +45,12 @@ public static bool Wait(string actionName, bool throwOnCancel, Action action)
...
-                    {
-                        throw t.Exception.InnerException;
                     }
+                    success = true;
                 }
-            }
+            );
 
-            return true;
+            return success;
```
AArnott: As you have it, I believe `success` is guaranteed to be `true`, so I'm not sure why you bother having the local variable at all (or return `bool` from this method). It used to be that you'd return `false`  when OCE was thrown and `throwOnCancel==false`, but now you set `success=true` before returning it.

AArnott:
> This PR upgrades the VSOperationWaiter to use JTF by upgrading shell to 15.0 which requires Framework v4.7.2

I don't understand why upgrading shell to 15.0 was necessary. `ThreadHelper.JoinableTaskFactory` [existed in 14.0](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.threadhelper.joinabletaskfactory?view=visualstudiosdk-2015) as well.

WardenGnaw:
@AArnott The TWD extension methods did not seem to exist in 14.0.

src/SSHDebugPS/VSOperationWaiter.cs
WardenGnaw: I was hoping that if throwOnCancel was true and an exception was thrown or cancel was set, the async method would throw and never set success to true.

AArnott: It would -- but you have code here to catch that exception and swallow it if `!throwOnCancel`.

AArnott: Oh, also: your `action` doesn't actually even accept a `CancellationToken`, so in fact the way you're coding this up, you're offering the user a Cancel button which is a no-op. It would only cancel if the user happened to click it before the `action` delegate were even invoked, which is *highly* unlikely. You should either call the `JTF.Run` overload that does not provide the delegate with a token, or modify your `Wait` method to take an `Action<CancellationToken>` so that the delegates can respond to cancellation requests.

AArnott:
> The TWD extension methods did not seem to exist in 14.0.

Ah yes, good point @WardenGnaw. And anyway I fully support folks targeting Dev15 over Dev14 nowadays. :)

src/SSHDebugPS/VSOperationWaiter.cs
```diff
@@ -30,39 +22,21 @@ internal static class VSOperationWaiter
...
-            catch (FileNotFoundException)
-            {
-                // Visual Studio is not installed on this box
-            }
-
-            if (waiterImpl != null)
-            {
-                using (ManualResetEvent completeEvent = new ManualResetEvent(false))
+            ThreadHelper.JoinableTaskFactory.Run(
+                actionName,
```
gregg-miskelly: >actionName [](start = 16, length = 10)

Does this overload of JTF.Run put up a threaded wait dialog if things take too long?

src/SSHDebugPS/VSOperationWaiter.cs
```diff
@@ -30,39 +22,21 @@ internal static class VSOperationWaiter
...
+                async (progress, cancellationTokenSource) =>
                 {
-                    CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
-
-                    t.ContinueWith((System.Threading.Tasks.Task unused) => completeEvent.Set(), TaskContinuationOptions.ExecuteSynchronously);
-
-                    try
-                    {
-                        waiterImpl.Wait(completeEvent, cancellationTokenSource);
+                    try {
```
gregg-miskelly: >{ [](start = 24, length = 1)

You somehow forgot to add a newline before this :)

src/SSHDebugPS/VSOperationWaiter.cs
```diff
@@ -30,39 +22,21 @@ internal static class VSOperationWaiter
...
+                        await Task.Run(action, cancellationTokenSource);
                     }
                     catch (OperationCanceledException) // VS Wait dialog implementation always throws on cancel
```
gregg-miskelly: > // VS Wait dialog implementation always throws on cancel [](start = 54, length = 57)

Is this still the correct exception? The comment is wrong now.

src/SSHDebugPS/VSOperationWaiter.cs
```diff
@@ -30,39 +22,21 @@ internal static class VSOperationWaiter
...
                     {
                         if (throwOnCancel)
                         {
                             throw;
                         }
```
gregg-miskelly: nit: either add a 'when' handler and only catch the exception when `!throwOnCancel`, or remove this and let the below code throw.

src/SSHDebugPS/VSOperationWaiter.cs
AArnott: Yes. But it also gives a Cancel button on the dialog. So we should either switch to an overload that does not offer the delegate a `CancellationToken` (and doesn't display the Cancel button to the user) or we should make `action` be cancelable.

src/SSHDebugPS/VSOperationWaiter.cs
```diff
@@ -22,15 +22,21 @@ internal static class VSOperationWaiter
...
-                    await Task.Run(action);
-                }
-            );
+            try
+            {
+                //TODO: Once 'action' is cancelable, provide token to the action instead.
+                ThreadHelper.JoinableTaskFactory.Run(
+                    actionName,
+                    async (progress, cancellationToken) =>
+                    {
```
gregg-miskelly: I don't think you want to use `Task.Run(...).Wait()` as this will block two different threads.

I think you want something like this:
```
        // Kicks off the specified action, and returns a Task which is completed when either the action finishes or the
        // cancellation token is set.
        static Task RunUntilCancelled(Action action, CancellationToken cancellationToken)
        {
            TaskCompletionSource<object> tcs = new TaskCompletionSource<object>();

            CancellationTokenRegistration cancellationTokenRegistration = cancellationToken.Register(() =>
            {
                // Mark the task returned from this method as cancelled, but we cannot cancel the underlying operation,
                // so we will just let it do nothing.
                tcs.TrySetCanceled();
            });

            Task runResult = Task.Run(action);
            runResult.ContinueWith((Task t) =>
            {
                tcs.TrySetResult(null);
                cancellationTokenRegistration.Dispose();
            });

            return tcs.Task;
        }
```

src/SSHDebugPS/VSOperationWaiter.cs
```diff
@@ -22,15 +22,21 @@ internal static class VSOperationWaiter
...
-                {
-                    await Task.Run(action);
-                }
-            );
+            try
+            {
+                //TODO: Once 'action' is cancelable, provide token to the action instead.
+                ThreadHelper.JoinableTaskFactory.Run(
+                    actionName,
+                    async (progress, cancellationToken) =>
```
gregg-miskelly: Note: you don't want to use `async` on your delegate. You aren't awaiting anything, so it isn't an async delegate.

src/SSHDebugPS/VSOperationWaiter.cs
AArnott: Wow. That's really complicated. I agree you don't want to use `Task.Wait()` here though, since that's just what JTF.Run is supposed to take care of for you. So just replace your Wait() line with this:
```cs
await Task.Run(action).WithCancellation(cancellationToken);
```

That extension method will take care of the rest for you to get what you wanted. But as your code comment suggests, this still leaves `action` running on a background thread, with the risk perhaps that the user will repeat the operation and you'll end up stacking up multiple threads all running `action`. So you should change `action` to take a `CancellationToken` so that your cancellation is _real_ as soon as you can.

src/SSHDebugPS/VSOperationWaiter.cs
AArnott: You'll `await` something once you implement my suggestion below. :)

src/SSHDebugPS/VSOperationWaiter.cs
gregg-miskelly: Glad you are on the review so we didn't need to duplicate `WithCancellation` :)

Issue #844, opened 2019-04-03
MIMode "gdb" wrongly assumes the type of Inferior; debugger pause does not work


haneefdm:
Okay, here is the scenario. Sorry, long

I am using GDB on my Mac (via MIEngine/cpptools) to debug a program/image running on a board connected to my USB/serial port. Pretty much everything works except for one huge thing. I cannot pause a running target. I can pause it using any other tool (Eclipse, gdb command line, even another VSCode extension, etc.). This is the 99.9% use case for embedded/IoT applications where they can't actually afford to have an entire OS/gdb running on the board. When I try to pause, nothing happens. I pretty much have to kill the VSCode Debug session, find and kill the gdb-server and/or gdb itself.

I think I know why this is happening. In [MICore/src/Debugger.cs, around line 707](https://github.com/Microsoft/MIEngine/blob/5ab89ebe0c8f3a45b1a49c2304923c19a549f02f/src/MICore/Debugger.cs#L707), you check to see if it is local gdb and assumes program is local too.

Then there is https://sourceware.org/bugzilla/show_bug.cgi?id=20035.

Well, that is only applicable to gdb executing native prorgrams. In our case, gdb is debugging a quasi-remote program in async mode. It is called "remote serial" or "extended remote serial". The local gdb talks to a gdb-server either using a serial port, local/remote IP address See [-target-select](https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI-Target-Manipulation.html#GDB_002fMI-Target-Manipulation) doc.

Problem is that gdb-servers can create fake pids and thread-groups if they have to and gdb hasn't got a clue they are fake. Since MIEngine monitors and collects these pids, it tries to send a SIGINT to the debuggee which will never work. **Might even kill an innocent process on local machine? :-)**.

Using the gdb command ```info target``` When you debug something with gdb, you can figure out what kind of target it is even though gdb and executable are local...

With gdb running on my Mac and board connected to a USB port, it shows.

```
Symbols from "/Users/hdm/tmp/cystuff/demo/main_final.elf".
Extended remote serial target in gdb-specific protocol:
Debugging a target over a serial line.
	While running this, GDB does not access memory from...
Local exec file:
	`/Users/hdm/tmp/cystuff/demo/main_final.elf', file type elf32-littlearm.
	Entry point: 0x100802ce
	0x10080000 - 0x100811e4 is .text
        ...
```

With gdb running a local process on Windows/Cygwin, it will say 'Native process'

```
[...]

PR #855, opened 2019-05-07
Fix for Issue #844 where pause does not work on Linux and Mac when using gdbservers

haneefdm:
There are two ways of having local gdb but running with a gdbserver. You can either use the `miDebuggerServerAddress` or `debugServerPath` in launch configuration. Most of us the `debugServerPath` to start and end the server with the debug session because of additional configuration that is needed to get the server running properly.

When using a gdbserver, the logic was wrong for the case of `debugServerPath` on linux/Mac and we did kill a process that was not running locally -- ending up potentially killing a random process.

The fix is to treat `miDebuggerServerAddress` and `debugServerPath` similarly. I refactored some code and put into a single function `IsLocalLaunchUsingServer()` as this kind of logic was used in multiple places. I also found that the checks for the use of these properties were inconsistent. Sometimes it was `string. IsNullOrWhiteSpace()` and sometimes `string. IsNullOrEmpty()`. Now, they are all `IsNullOrWhiteSpace()` for these two options.

I renamed two methods IsLocalGdb and IsRemoteGdb to be more meaningful. Add a 'Target' suffix to  them

Here is my [testing matrix](https://docs.google.com/spreadsheets/d/1l7t73YQePvi91gI1SbknqpVVickWuwgGOnpOcvGAISk/edit?usp=sharing).

Basically, I had to test on all three platforms in multiple configurations. Many fails turned into passes with these changes.

My first PR so, let me know if I missed anything. I could not run the CodeFormatter.exe -- it was crashing due to a missing dll, but I formatted my changes with the VStudio editor.

In the process of this, I found a couple more issues but I want to treat them as separate things. I documented them in my 'testing matrix' link above.


src/MICore/Debugger.cs
```diff
@@ -578,20 +578,22 @@ public Task CmdBreak(BreakRequest request)
...
 
-        internal bool IsLocalGdb()
+        protected bool IsLocalLaunchUsingServer()
         {
-            return (this.MICommandFactory.Mode == MIMode.Gdb &&
-               this._launchOptions is LocalLaunchOptions &&
-               String.IsNullOrEmpty(((LocalLaunchOptions)this._launchOptions).MIDebuggerServerAddress));
+            return (_launchOptions is LocalLaunchOptions &&
+                (!String.IsNullOrWhiteSpace(((LocalLaunchOptions)_launchOptions).MIDebuggerServerAddress) ||
+                 !String.IsNullOrWhiteSpace(((LocalLaunchOptions)_launchOptions).DebugServer)));
```
pieandcakes: Do we want to do a quick/dirty string compare for `"setupCommands"` also to see if any start with `-target-select remote` or `-target-select extended-remote` and add that as an option here?

Also, do we want to do this one validation one time instead of any time that this method is called?

src/MICore/Debugger.cs
```diff
@@ -578,20 +578,22 @@ public Task CmdBreak(BreakRequest request)
...
 
-        private bool IsRemoteGdb()
+        private bool IsRemoteGdbTarget()
         {
-            return this.MICommandFactory.Mode == MIMode.Gdb &&
-               (this._launchOptions is PipeLaunchOptions || this._launchOptions is UnixShellPortLaunchOptions ||
-               (this._launchOptions is LocalLaunchOptions
-                    && !String.IsNullOrEmpty(((LocalLaunchOptions)this._launchOptions).MIDebuggerServerAddress)));
+            return MICommandFactory.Mode == MIMode.Gdb &&
+               (_launchOptions is PipeLaunchOptions || _launchOptions is UnixShellPortLaunchOptions || IsLocalLaunchUsingServer());
```
pieandcakes: nit: if you can add newlines after the `||`, it makes it a little easier to parse

pieandcakes:
@haneefdm Thank you for submitting this!

pieandcakes:
resolves #844

src/MICore/Debugger.cs
haneefdm: Actually, there is a way to know if someone did a `remote` or `extended-remote`. You will get a response from gdb `^connected`. Scanning in `setupCommands` is not a reliable method, I thought. They could have done that using 'mi' commands or even non-mi commands. Looking for `^connected` was my first implementation and it worked but we would know about that after the session starts. My recent implementation using launch properties, we will know even before the session starts.

Yes, I suppose, this could be validated once. Did not want to change too much from what the existing code did. I just refactored it and added the additional check. Do you want me to do that? This method is used so infrequently so....

src/MICore/Debugger.cs
haneefdm: Bothered me too I wanted to do it originally but the old line I replaced was about that long.

Do you want me to do it now?

src/MICore/Debugger.cs
pieandcakes: >Yes, I suppose, this could be validated once. Did not want to change too much from what the existing code did. I just refactored it and added the additional check. Do you want me to do that? This method is used so infrequently so....

The value won't change in subsequent checks so my gut is to do it one time and cache it but we can see if this way is performant enough and make a change in the future if necessary.

src/MICore/Debugger.cs
pieandcakes: If you don't mind

src/MICore/Debugger.cs
haneefdm: Yup, the same thing can be said about the old `IsLocalGdb()` and `IsRemoteGdb()`. This code is only executed during startup and when the user hits pause/terminate. Perhaps keeping less state is also a good thing?

src/MICore/Debugger.cs
haneefdm: Will do. Perhaps I should wait until others comment? Do I close this PR and start a whole new one, or do a separate PR?

pieandcakes: you can add commits to your branch and push it and it will show up in the PR.

src/MICore/Debugger.cs
pieandcakes: True. Like you mentioned in the testing document, theres a lot of iffy state management that we try and do such as the stopAtEntry work that isn't 100% foolproof.

src/MICore/Debugger.cs
haneefdm: Done.

src/MICore/Debugger.cs
```diff
@@ -578,20 +578,23 @@ public Task CmdBreak(BreakRequest request)
...
-               this._launchOptions is LocalLaunchOptions &&
-               String.IsNullOrEmpty(((LocalLaunchOptions)this._launchOptions).MIDebuggerServerAddress));
+            return (_launchOptions is LocalLaunchOptions &&
+                (!String.IsNullOrWhiteSpace(((LocalLaunchOptions)_launchOptions).MIDebuggerServerAddress) ||
+                 !String.IsNullOrWhiteSpace(((LocalLaunchOptions)_launchOptions).DebugServer)));
+        }
 
+        internal bool IsLocalGdbTarget()
+        {
+            return (MICommandFactory.Mode == MIMode.Gdb && !IsLocalLaunchUsingServer());
```
gregg-miskelly: Unless I am missing something, this will give a different result than the old `IsLocalGdb` function would give if `this._launchOptions is LocalLaunchOptions` is false.

src/MICore/Debugger.cs
```diff
@@ -578,20 +578,23 @@ public Task CmdBreak(BreakRequest request)
             return CmdBreakInternal();
         }
 
-        internal bool IsLocalGdb()
+        protected bool IsLocalLaunchUsingServer()
         {
-            return (this.MICommandFactory.Mode == MIMode.Gdb &&
-               this._launchOptions is LocalLaunchOptions &&
-               String.IsNullOrEmpty(((LocalLaunchOptions)this._launchOptions).MIDebuggerServerAddress));
+            return (_launchOptions is LocalLaunchOptions &&
```
gregg-miskelly: nit: C# now has a better way of writing this code. Since we are touching it, we might as well use it:

```
            return (_launchOptions is LocalLaunchOptions localLaunchOptions &&
                (!String.IsNullOrWhiteSpace(localLaunchOptions.MIDebuggerServerAddress) ||
                 !String.IsNullOrWhiteSpace(localLaunchOptions.DebugServer));
```

src/MICore/Debugger.cs
gregg-miskelly: In other words, I think you want to change this to `MICommandFactory.Mode == MIMode.Gdb && this._launchOptions is LocalLaunchOptions && !IsLocalLaunchUsingServer()`

haneefdm: Ummm, yes and no, and hence the change in the method name. The old `IsLocalGdn()` also wanted to make sure that we are not using `MIDebuggerServerAddress` meaning a server. So does this one but using it also makes sure that `DebugServer` is not being used. _And that was actually the bug_. It was never looking for a 'is this a purely locally running gdb?`. I looked at every place `IsLocalGdb()` and `IsRemoteGdb()` is/was used and what we really needed to know was if there is a server in play or gdb controlling the target directly.

gregg-miskelly: Yup, I got that, but now this function will incorrectly return true in a different case - if we aren't dealing with local launch options.

src/MICore/Debugger.cs
haneefdm: Ohh nice!! Did not know that. Done, will wait for more comments before another round of testing.

src/MICore/Debugger.cs
haneefdm: Yes, I see that now.

haneefdm: Done. It now has the true meaning of what `IsLocalGdbTarget()` should mean. Did another round of testing and pushed changes.

src/MICore/Debugger.cs
haneefdm: Done, that is such a nice feature, bit obscure syntax -- declaring a variable in the middle of an expression -- but the scope of that var was reasonable. Thanks for the education.

src/MICore/Debugger.cs
haneefdm: Just to acknowledge, what @gregg-miskelly pointed out was a serious defect

gregg-miskelly:
@pieandcakes Do you have any additional comments? @paulmaybee Do you want to review this since it touches some features that you added?

gregg-miskelly:
@haneefdm Thanks again for submitting this!

haneefdm:
Thank you all.

PR #857, opened 2019-05-09
Add FormatProvider for FXCop error CA1305. Fix failing tests and a bug

pieandcakes:
Once I fixed the format provider, the SSH PS tests were failing, which showed a bug. Updated tests with new PS format and fixed the bug.

### DIFF
--- a/src/SSHDebugPS/PSOutputParser.cs
+++ b/src/SSHDebugPS/PSOutputParser.cs
@@ -123,7 +123,7 @@ namespace Microsoft.SSHDebugPS
             int index = 0;
             int strLen = headerLine.Length;
 
-            // pid column is right justified so the pid column stops at the last letter index
+            // pid column is a fixed length column
             if (!SkipNonWhitespace(headerLine, ref index))
                 return false;
             _pidCol = new ColumnDef(0, index);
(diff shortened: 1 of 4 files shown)


pieandcakes:
@WardenGnaw @gregg-miskelly Apparently some test are failing too so I need to fix those also.

pieandcakes:
@gregg-miskelly @WardenGnaw Can you take another look. This should pass the build now.

src/SSHDebugPS/PSOutputParser.cs
```diff
@@ -147,7 +147,7 @@ private bool ProcessHeaderLine(/*OPTIONAL*/ string headerLine)
             SkipNonWhitespace(headerLine, ref index);
 
             // make sure the line is now empty, aside from whitespace
-            if (SkipWhitespace(headerLine, ref index))
+            if (SkipWhitespace(headerLine, ref index) && index < strLen)
```
gregg-miskelly: Why did you add the condition? `SkipWhitespace` will return false if there is no more whitespace before the end.

src/SSHDebugPS/SSH/SSHConnection.cs
```diff
@@ -46,7 +46,7 @@ public override List<Process> ListProcesses()
             var usernameCommand = _remoteSystem.Shell.ExecuteCommand("id -u -n");
             if (usernameCommand.ExitCode == 0)
             {
-                username = usernameCommand.Output;
+                username = usernameCommand.Output.TrimEnd('\n'); // trim line endings because 'id' command ends with a newline
```
gregg-miskelly: Should we trim `\r` also?

pieandcakes: I can add it. Will we have '\r' on Linux?

PR #864, opened 2019-05-15
Update signing to the new SHA256 cert

pieandcakes:


andrewcrawley:
LGTM, just check the release build when it's done and make sure the right cert is being used.

gregg-miskelly:
Do you need to fix this code also:

`\MIEngine\build\miengine.csharp.localization.targets(18):        <Authenticode>Microsoft</Authenticode>`

gregg-miskelly:
Otherwise LGTM

PR #866, opened 2019-05-22
Replace returning NotImplementedException with a protocol failure

pieandcakes:
See: Microsoft/vscode-cpptools#3662


andrewcrawley:
LGTM.

PR #868, opened 2019-05-24
Attach to Docker containers UI 

pieandcakes:


src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -46,7 +55,12 @@ public static Connection GetDockerConnection(string name)
                 // If SSH connection dialog was cancelled, we should cancel this connection.
                 if (remoteConnection == null)
                     return null;
-                
+
+                // Verify container exists on remote machine.
```
WardenGnaw: Did you want to check in these comments?

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -76,25 +89,31 @@ public static Connection GetSSHConnection(string name)
...
-                }
+                    string userName;
+                    string hostName;
 
-                PasswordConnectionInfo newConnectionInfo = new PasswordConnectionInfo(hostName, userName, new System.Security.SecureString());
+                    int atSignIndex = name.IndexOf('@');
+                    if (atSignIndex > 0)
+                    {
+                        userName = name.Substring(0, atSignIndex);
+                        hostName = name.Substring(atSignIndex + 1);
```
WardenGnaw: Is it possible for the user to supply ```user@``` and hostname would Index out of bounds?

src/SSHDebugPS/ConnectionManager.cs
pieandcakes: yes. I'll add TODO in front of it

src/SSHDebugPS/ConnectionManager.cs
pieandcakes: Added a check and will default to string.empty for the hostname in that case as that is what the user wanted.

src/SSHDebugPS/SSHDebugPS.csproj
```diff
@@ -83,6 +88,14 @@
     <Reference Include="$(NugetPackagesDirectory)\Microsoft.VisualStudio.Shell.Framework.$(Microsoft_VisualStudio_Shell_Framework_Version)\lib\net45\Microsoft.VisualStudio.Shell.Framework.dll">
       <Private>False</Private>
     </Reference>
+    <Reference Include="Microsoft.VisualStudio.Utilities, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.VisualStudio.Utilities.15.0.26606\lib\net45\Microsoft.VisualStudio.Utilities.dll</HintPath>
+    </Reference>
+    <!-- NOTE: MIEngine's usage of Newtonsoft.Json's flows through public interfaces exposed by MICore (IPlatformAppLauncher). So this version needs to be within the binding redirect range used by devenv.exe (currently 4.5.0.0-9.0.0.0) -->
+    <Reference Include="Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <HintPath>$(NuGetPackagesDirectory)/Newtonsoft.Json.6.0.8/lib/portable-net45+wp80+win8+wpa81+aspnetcore50/Newtonsoft.Json.dll</HintPath>
+      <Private>True</Private>
```
gregg-miskelly: >True [](start = 15, length = 4)

Don't you want False here?

src/SSHDebugPS/SSHDebugPS.csproj
```diff
@@ -55,13 +55,18 @@
...
   <ItemGroup>
-    <None Include="project.json" />
     <Reference Include="liblinux">
       <HintPath>$(GeneratedAssembliesDir)liblinux.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.VisualStudio.Debugger.InteropA, Version=9.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
       <Private>False</Private>
     </Reference>
+    <Reference Include="Microsoft.VisualStudio.ImageCatalog, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
+      <HintPath>$(NuGetPackagesDirectory)\Microsoft.VisualStudio.ImageCatalog.$(Microsoft_VisualStudio_Shell_15_0_Version)\lib\net45\Microsoft.VisualStudio.ImageCatalog.dll</HintPath>
```
gregg-miskelly: >HintPath [](start = 7, length = 8)

Should all of these have `<Private>False</Private>` to avoid msbuild's copy-local behavior?

src/SSHDebugPS/Microsoft.SSHDebugPS.pkgdef
```diff
@@ -23,4 +23,17 @@
...
+
+[$RootKey$\CLSID\{91BDF293-E6A0-49C4-B033-6F36CFC4FF98}]
+"Assembly"="Microsoft.SSHDebugPS"
+"Class"="Microsoft.SSHDebugPS.Docker.DockerPortPicker"
+"InprocServer32"="$WinDir$\SYSTEM32\MSCOREE.DLL"
+"CodeBase"="$PackageFolder$\Microsoft.SSHDebugPS.dll"
+
+[$RootKey$\RuntimeConfiguration\dependentAssembly\codeBase\{7E3052B2-FB42-4E38-B22C-1FD281BD4413}]
+"name"="Microsoft.SSHDebugPS"
+"publicKeyToken"=""
```
gregg-miskelly: Do we need this? It feels wrong to set this to empty string when in fact this dll has a public key token.

src/SSHDebugPS/ConnectionManager.cs
gregg-miskelly: I think that is just invalid input.

---
In reply to: [287481980](https://github.com/microsoft/MIEngine/pull/868#discussion_r287481980) [](ancestors = 287481980)

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -149,5 +174,140 @@ public static Connection GetSSHConnection(string name)
...
+            ConnectionInfoStore store = new ConnectionInfoStore();
+
+            return store.Connections.ToList().Select(item => (ConnectionInfo)item);
+        }
+
+        private const string dockerCommand = "docker";
+        // --no-trunc avoids parameter truncation
+        private const string dockerPSArgs = "ps --no-trunc --format \"{{json .}}\"";
+        public static IEnumerable<IContainerInstance> GetLocalDockerContainers()
+        {
```
gregg-miskelly: nit: move to a class in the Docker folder

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -149,5 +174,140 @@ public static Connection GetSSHConnection(string name)
...
+                        }
+
+                        var containerInstance = DockerContainerInstance.Create(args);
+                        if (containerInstance != null)
+                            containers.Add(containerInstance);
+                    }
+                });
+
+                commandRunner.Run();
+                resetEvent.WaitOne();
```
gregg-miskelly: >resetEvent.WaitOne(); [](start = 16, length = 21)

Add TODO comment to use a cancellable wait?

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -149,5 +174,140 @@ public static Connection GetSSHConnection(string name)
...
+
+            commandRunner.OutputReceived += ((sender, args) =>
+            {
+                if (!string.IsNullOrWhiteSpace(args))
+                {
+                    // If it isn't json, assume its an error message
+                    if (args.Trim()[0] != '{')
+                    {
+                        errorSB.Append(args);
+                    }
```
gregg-miskelly: Can this event be fired more than once?

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -149,5 +174,140 @@ public static Connection GetSSHConnection(string name)
...
+            }
+
+            List<DockerContainerInstance> containers = new List<DockerContainerInstance>();
+            RemoteCommandRunner commandRunner = new RemoteCommandRunner(dockerCommand, dockerPSArgs, sshConnection);
+
+            ManualResetEvent resetEvent = new ManualResetEvent(false);
+            int exitCode = 0;
+            commandRunner.ErrorOccured += ((sender, args) =>
+            {
+                resetEvent.Set();
```
gregg-miskelly: >resetEvent [](start = 16, length = 10)

Should this be appending to errorSB?

gregg-miskelly: >resetEvent.Set(); [](start = 16, length = 17)

Should we be waiting to see if there is additional error output?

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -149,5 +174,140 @@ public static Connection GetSSHConnection(string name)
...
+        {
+            ConnectionInfoStore store = new ConnectionInfoStore();
+
+            return store.Connections.ToList().Select(item => (ConnectionInfo)item);
+        }
+
+        private const string dockerCommand = "docker";
+        // --no-trunc avoids parameter truncation
+        private const string dockerPSArgs = "ps --no-trunc --format \"{{json .}}\"";
+        public static IEnumerable<IContainerInstance> GetLocalDockerContainers()
```
gregg-miskelly: >GetLocalDockerContainers [](start = 54, length = 24)

Are you planning to someday better unify the local and remote cases? They are sure very similar.

src/SSHDebugPS/IConnection.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Microsoft.SSHDebugPS
 {
-    internal interface IConnection
+    public interface IConnection
```
gregg-miskelly: >public [](start = 4, length = 6)

Do you know why you changed this?

pieandcakes: We had an issue with the Xaml not liking a property i was using it on because it wasn't public and i couldn't make it public without making the interface public. Is it bad to have this as public?

src/SSHDebugPS/ConnectionManager.cs
pieandcakes: Yes. I ran into a timing issue on the local case which forced me to create a .Run() method on the local and then i had to handle error output slightly different. 

My plan is when I go back and clean up the CommandRunner and the ShellExecutionManager to work that out too.

src/SSHDebugPS/AD7/AD7PortSupplier.cs
```diff
@@ -44,18 +36,33 @@ public int GetPortSupplierName(out string name)
...
+        }
+
+        int IDebugPortSupplierDescription2.GetDescription(enum_PORT_SUPPLIER_DESCRIPTION_FLAGS[] flags, out string text)
+        {
+            text = Description;
             return HR.S_OK;
         }
 
-        public int RemovePort(IDebugPort2 pPort)
+        public virtual int EnumPersistedPorts(BSTR_ARRAY portNames, out IEnumDebugPorts2 portEnum)
```
gregg-miskelly: >EnumPersistedPorts [](start = 27, length = 18)

Is there a reason we aren't implementing this for Docker? If we aren't going to, we should probably stop implementing IDebugPortSupplier3, and move all of that just to the SSH implementation.

src/SSHDebugPS/SSH/SSHPortSupplier.cs
```diff
@@ -95,6 +95,12 @@ private bool IsLibLinuxAvailable()
             return (ServiceProvider.GlobalProvider.GetService(typeof(IVsConnectionManager)) as IVsConnectionManager) != null;
         }
 
+        public override int CanPersistPorts()
+        {
+            return HR.S_OK;
+        }
+
```
gregg-miskelly: The SSH port supplier doesn't persist ports. This should be returning S_FALSE. So that the SDM will handle persistence.

pieandcakes: Pretty sure they (liblinux) persist the SSH connections so we by default do because we query them.

gregg-miskelly: Got it. In that case we could stop implementing IDebugPortSupplier3, and EnumPersistedPorts. EnumPersistedPorts will only ever be called if CanPersistPorts returns S_FALSE.

---
In reply to: [287537299](https://github.com/microsoft/MIEngine/pull/868#discussion_r287537299) [](ancestors = 287537299)

src/SSHDebugPS/IConnection.cs
gregg-miskelly: Sounds reasonable. But I would add a comment then as to why this is public.

---
In reply to: [287534458](https://github.com/microsoft/MIEngine/pull/868#discussion_r287534458) [](ancestors = 287534458)

src/SSHDebugPS/AD7/AD7PortSupplier.cs
gregg-miskelly: Actually, the SSH doesn't implement it either. So we should remove IDebugPortSupplier3 and its two methods from the base class, and either implement EnumPersistedPorts in the Docker side, or remove EnumPersistedPorts/CanPersistedPorts all together.

---
In reply to: [287536095](https://github.com/microsoft/MIEngine/pull/868#discussion_r287536095) [](ancestors = 287536095)

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
```diff
@@ -0,0 +1,309 @@
...
+        </ScrollViewer>
+
+        <!-- OK / Cancel buttons -->
+        <StackPanel Grid.Row="5"
+                    Grid.ColumnSpan="2"
+                    Margin="4"
+                    Orientation="Horizontal"
+                    HorizontalAlignment="Right">
+
+            <Button x:Uid="uidOkayButton" x:Name="OkayButton"
```
gregg-miskelly: >Button [](start = 13, length = 6)

Do we need to disable the okay button if no containers are selected?

src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
```diff
@@ -0,0 +1,238 @@
...
+            OnPropertyChanged(nameof(SupportedConnections));
+        }
+
+        internal void RefreshContainersList()
+        {
+            IsRefreshEnabled = false;
+
+            try
+            {
+                StatusText = UIResources.SearchingStatusText; // Change text to reflect finding?
```
gregg-miskelly: >Change text to reflect finding? [](start = 65, length = 31)

I don't understand what this comment means

src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
```diff
@@ -0,0 +1,238 @@
...
+                Debug.Fail("AddSSHConnection cannot be called.");
+            }
+        }
+
+        private Lazy<bool> _sshAvailable;
+        public bool CanAddConnection
+        {
+            get
+            {
+                return _sshAvailable?.Value ?? false;
```
gregg-miskelly: >?. [](start = 36, length = 2)

Shouldn't `_sshAvailable` never be null?

src/SSHDebugPS/Microsoft.SSHDebugPS.pkgdef
pieandcakes: @gregg-miskelly Since we have a different public key token for dev sign, do you know a way to do that? I'll test without that field set too.

---
In reply to: [287511218](https://github.com/microsoft/MIEngine/pull/868#discussion_r287511218) [](ancestors = 287511218)

src/SSHDebugPS/SSHDebugPS.csproj
pieandcakes: I added it for the two I added and for the Linux ConnectionManager one. 

---
In reply to: [287510686](https://github.com/microsoft/MIEngine/pull/868#discussion_r287510686) [](ancestors = 287510686)

src/SSHDebugPS/Microsoft.SSHDebugPS.pkgdef
gregg-miskelly: My hope was that we could just remove it. If not, I would probably suggest a comment.

src/SSHDebugPS/ConnectionManager.cs
pieandcakes: Technically yes, in my test scenario no.

---
In reply to: [287513281](https://github.com/microsoft/MIEngine/pull/868#discussion_r287513281) [](ancestors = 287513281)

pieandcakes: Which means i need an output string builder and then to process the string afterwards.

---
In reply to: [287545722](https://github.com/microsoft/MIEngine/pull/868#discussion_r287545722) [](ancestors = 287545722,287513281)

pieandcakes: well now that i think about it, i think this is fine. liblinux returns lines so this *should* be ok

---
In reply to: [287545780](https://github.com/microsoft/MIEngine/pull/868#discussion_r287545780) [](ancestors = 287545780,287545722,287513281)

src/SSHDebugPS/ConnectionManager.cs
pieandcakes: we can do that.

---
In reply to: [287519036](https://github.com/microsoft/MIEngine/pull/868#discussion_r287519036) [](ancestors = 287519036)

src/SSHDebugPS/AD7/AD7PortSupplier.cs
pieandcakes: With SSH returning true for CanPersistPorts, can i do this? I feel like i need to keep it in place. 


---
In reply to: [287537861](https://github.com/microsoft/MIEngine/pull/868#discussion_r287537861) [](ancestors = 287537861,287536095)

src/SSHDebugPS/SSH/SSHPortSupplier.cs
pieandcakes: I need IDebugPortSupplier3 for CanPersistPorts though. I thought EnumPersistedPorts is only called if CanPersistPorts is true?

---
In reply to: [287537461](https://github.com/microsoft/MIEngine/pull/868#discussion_r287537461) [](ancestors = 287537461,287537299)

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
pieandcakes: currently we treat it like a cancel where nothing is selected so we return no data. I can talk to Leslie and see if that is acceptable or if we should just grey out "ok". Doing it would probably be trivial.

---
In reply to: [287538546](https://github.com/microsoft/MIEngine/pull/868#discussion_r287538546) [](ancestors = 287538546)

src/SSHDebugPS/AD7/AD7PortSupplier.cs
gregg-miskelly: The only way to have the SDM persist ports is to both implement IDebugPortSupplier3 and also return S_FALSE from CanPersistPorts. Otherwise the SDM will just call EnumPorts.

gregg-miskelly: In other words - there isn't really a point in implementing IDebugPortSupplier3 unless your implementation of CanPersistPorts return S_FALSE, at least in some cases.

src/SSHDebugPS/SSH/SSHPortSupplier.cs
gregg-miskelly: No, EnumPersistedPorts is only called if CanPersistPorts returns S_FALSE.

gregg-miskelly: I updated the comments in msdbg.idl to make this less confusing:

```C++
// --------------------------------------------------------
// IDebugPortSupplier3
//
// Optional interface which can be implemented by port suppliers to opt-into
// have the SDM persist a MRU of port names.
[
	object,
	uuid(5b5eec44-51aa-4210-b84f-1938b8576d8d),
	pointer_default(unique)
]
interface IDebugPortSupplier3: IDebugPortSupplier2
{
	// Can the port supplier remember new ports? S_OK == yes; S_FALSE == no
        //
        // Returning S_FALSE from this method indicates that the SDM should persist
        // port names for the port supplier, and that the SDM can call EnumPersistedPorts
        // with the names that it has persisted.
	HRESULT CanPersistPorts();
	
	// Convert the port names that the SDM has persisted into an enumerator of IDebugPort2
        // objects. This method will only be called if `CanPersistPorts` returns S_FALSE.
	HRESULT EnumPersistedPorts(
		[in] BSTR_ARRAY PortNames,
		[out] IEnumDebugPorts2** ppEnum);
};
```

pieandcakes: That makes more sense. I think in the Docker scenario we should return FALSE which will force a call to EnumPersistedPorts. If that isn't the behavior today, I'll fix it for GA

src/SSHDebugPS/CommandRunner.cs
```diff
@@ -34,25 +35,173 @@ public ErrorOccuredEventArgs(Exception e)
...
+
+        public void Run()
+        {
+            _localProcess = new System.Diagnostics.Process();
+            _localProcess.StartInfo = processStartInfo;
+            _localProcess.Exited += OnProcessExited;
+            _localProcess.EnableRaisingEvents = true;
+            _localProcess.Start();
+
+            _stdoutWriter = new StreamWriter(_localProcess.StandardInput.BaseStream, new UTF8Encoding(encoderShouldEmitUTF8Identifier: false), 4096, leaveOpen: false);
```
WardenGnaw: NIT: BUFMAX

PR #870, opened 2019-05-30
Fixes bug 823041

pieandcakes:
When no documentContext is available when calling
IDebugCodeContext.GetDocumentContext() return S_FALSE instead.


pieandcakes:
The fix was @andrewcrawley 's suggestion

PR #871, opened 2019-05-30
Fix for hang when VS loses connection with linux

pieandcakes:
Calling _remoteSystem.Dispose() will hang VS until a timeout (30sec-1min) if VS has lost connection to the remote Linux machine.

### DIFF
--- a/src/SSHDebugPS/SSH/SSHConnection.cs
+++ b/src/SSHDebugPS/SSH/SSHConnection.cs
@@ -172,7 +172,8 @@ namespace Microsoft.SSHDebugPS.SSH
 
             if (_remoteSystem != null)
             {
-                _remoteSystem.Dispose();
+                if (_remoteSystem.IsConnected)
+                    _remoteSystem.Disconnect();
                 _remoteSystem = null;
             }
         }


pieandcakes:
This fixes Feedback item 736065

src/SSHDebugPS/SSH/SSHConnection.cs
```diff
@@ -172,7 +172,8 @@ public override void Close()
 
             if (_remoteSystem != null)
             {
-                _remoteSystem.Dispose();
+                if (_remoteSystem.IsConnected)
+                    _remoteSystem.Disconnect();
```
andrewcrawley: I checked liblinux, and the implementation of `IRemoteSystem.Dispose` just calls `Disconnect`, so if you're seeing a hang when calling `Dispose`, I'd expect to see it when calling `Disconnect` as well.  If checking `IsConnected` first avoids the problem, I'd leave this as a call to `Dispose`, just to maintain the pattern that `IDisposable` objects should be disposed.

pieandcakes: I sent them email to find out what i should be doing

PR #872, opened 2019-05-31
Fix fxcop formattable string error

pieandcakes:


src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -62,7 +63,7 @@ public static IEnumerable<IContainerInstance> GetLocalDockerContainers()
                 // might need to throw an exception here too??
                 if (exitCode != 0)
                 {
-                    Debug.Fail($"Exit Code: {exitCode}");
+                    Debug.Fail(FormattableString.Invariant($"Exit Code: {exitCode}"));
```
andrewcrawley: You can put `using static System.FormattableString` in your usings, then just write `Invariant($"...")`.  Not a big deal, since there's only the one change, but handy if you want to save typing.

PR #884, opened 2019-06-25
Users/piel/noshell

pieandcakes:
This change is to remove the launching of a "shell" and running all commands inside of it. Also added better error message handling and checks.


pieandcakes:
Tested in VS:
* SSH Attach (gdb)
* Docker attach local (dotnet)
* Docker attach over SSH (dotnet)
* Docker attach invalid container (local/remote)

src/SSHDebugPS/SSHDebugPS.csproj
```diff
@@ -214,6 +218,7 @@
   <ItemGroup>
     <None Include="packages.config" />
   </ItemGroup>
+  <ItemGroup />
```
WardenGnaw: Extra ItemGroup?

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -129,7 +78,7 @@ private string GetFullPath(string path)
             string pwd;
             ExecuteCommand("pwd", Timeout.Infinite, throwOnFailure: true, commandOutput: out pwd);
             ExecuteCommand($"cd \"{path}\"; pwd", Timeout.Infinite, throwOnFailure: true, commandOutput: out fullpath);
-            ExecuteCommand($"cd \"{pwd}\"", Timeout.Infinite, throwOnFailure: false, commandOutput: out output); //This might fail in some instances, so ignore a failure and ignore output
+            ExecuteCommand($"cd '{pwd}'", Timeout.Infinite, throwOnFailure: false, commandOutput: out output); //This might fail in some instances, so ignore a failure and ignore output
```
WardenGnaw: Why did this change and not the one above it too?

pieandcakes: It was during some testing. I will change it to match 

---
In reply to: [296972331](https://github.com/microsoft/MIEngine/pull/884#discussion_r296972331) [](ancestors = 296972331)

src/SSHDebugPS/Utilities/VSMessageBoxHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+    internal class VSMessageBoxHelper
+    {
+        /// <summary>
+        /// Switches to main thread and shows a VS message box.
+        /// </summary>
+        public static void ShowMessage(string title, string message, OLEMSGICON icon, OLEMSGBUTTON button, OLEMSGDEFBUTTON defaultButton)
+        {
+            ThreadingTasks.Task.Run(async () =>
+                {
+                    await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
gregg-miskelly: I can't tell if this is lambda is an `async Task` or `async void` method. I would probably suggest defining a local function instead to make it clear (`async void` is bad).

src/SSHDebugPS/Utilities/VSMessageBoxHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+                    VsShellUtilities.ShowMessageBox(ServiceProvider.GlobalProvider,
+                    message,
```
gregg-miskelly: nit: indent args

src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
```diff
@@ -226,7 +228,7 @@ public IConnectionViewModel SelectedConnection
             }
             set
             {
-                if (!(_selectedConnection != null && _selectedConnection.DisplayName.Equals(value.DisplayName, StringComparison.Ordinal)))
+                if ((_selectedConnection != value))
```
gregg-miskelly: nit: remove extra `(` `)`

src/SSHDebugPS/CommandRunner.cs
```diff
@@ -9,304 +9,304 @@
...
-            _localProcess.Start();
-
-            _stdoutWriter = new StreamWriter(_localProcess.StandardInput.BaseStream, new UTF8Encoding(encoderShouldEmitUTF8Identifier: false), 4096, leaveOpen: false);
-            _processReader = new StreamReader(_localProcess.StandardOutput.BaseStream, new UTF8Encoding(encoderShouldEmitUTF8Identifier: false), false, BUFMAX, false);
-            _outputReadTask = System.Threading.Tasks.Task.Run(() => ReadLoop(_processReader, _cancellationSource.Token, (line) => { OutputReceived?.Invoke(this, line); }));
-
-            _localProcess.ErrorDataReceived += OnErrorOutput;
-            _localProcess.Exited += OnProcessExited;
-            _localProcess.BeginErrorReadLine();
+            CreateProcessStartInfo(command, commandArgs);
```
andrewcrawley: >CreateProcessStartInfo [](start = 12, length = 22)

Should not call a virtual method from the constructor of a non-sealed class.  In this case, I don't think the method needs to be virtual anyway, so you can either make it non-virtual or just inline it into the constructor.

src/SSHDebugPS/CommandRunner.cs
```diff
@@ -9,304 +9,304 @@
...
-                ErrorOccured?.Invoke(this, new ErrorOccuredEventArgs(e));
-                Dispose();
+                throw new InvalidOperationException("Unable to create process. Process start info does not exist");
             }
-        }
 
-        private void OnErrorOutput(object sender, DataReceivedEventArgs e)
-        {
-            if (!string.IsNullOrWhiteSpace(e.Data))
+            lock (_lock)
```
andrewcrawley: >lock (_lock) [](start = 12, length = 12)

I think this lock should be around the entire method - otherwise, if you could end up with two threads both sitting at this point, where the first one gets the lock, initializes _process, and releases the lock, at which point the second thread will overwrite _process.

src/SSHDebugPS/CommandRunner.cs
```diff
@@ -9,304 +9,304 @@
...
-        public event EventHandler<string> OutputReceived;
-        public event EventHandler<int> Closed;
-        public event EventHandler<ErrorOccuredEventArgs> ErrorOccured;
+        protected virtual void OnOutputReceived(string line)
+        {
+            OutputReceived?.Invoke(this, line);
+        }
 
-        public void Dispose()
+        protected virtual void CreateProcessStartInfo(string command, string commandArgs)
```
andrewcrawley: I don't think this method needs to be virtual, but if you want it to be, it needs to return the ProcessStartInfo instead of setting the private field directly, or else an implementation that overrides this method will blow up when you call Start, since the private field isn't set.

src/SSHDebugPS/CommandRunner.cs
```diff
@@ -9,304 +9,304 @@
...
                 }
-
-                if (_stdoutWriter != null)
+                else
                 {
-                    _stdoutWriter.Close();
-                    _stdoutWriter = null;
+                    CleanUpProcess();
                 }
             }
```
andrewcrawley: If you make this method non-virtual, this block should no longer be necessary, as it's only ever called from the constructor, so there's no way the process could exist yet.

src/SSHDebugPS/CommandRunner.cs
```diff
@@ -9,304 +9,304 @@
...
+
+            if (_processStartInfo != null)
+            {
+                Debug.Fail("ProcessStartInfo is already set.");
+                _processStartInfo = null;
+            }
```
andrewcrawley: If you make this method non-virtual, this block should no longer be necessary, as it's only ever called from the constructor, so there's no way the process or start info could exist yet.

src/SSHDebugPS/Docker/DockerExecutionManager.cs
```diff
@@ -0,0 +1,121 @@
...
+            errorMessage = string.Empty;
+            if (_currentCommand != null)
+            {
+                throw new InvalidOperationException("already a command processing");
+            }
+            _commandCompleteEvent.Reset();
+
+            using (ICommandRunner commandRunner = GetExecCommandRunner(commandText))
+            {
+                //LocalCommandRunner commandRunner = runner as LocalCommandRunner;
```
gregg-miskelly: Remove commented out code?

src/SSHDebugPS/Utilities/VSMessageBoxHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+using ThreadingTasks = System.Threading.Tasks;
+
+namespace Microsoft.SSHDebugPS.Utilities
+{
+    internal class VSMessageBoxHelper
+    {
+        /// <summary>
+        /// Switches to main thread and shows a VS message box.
+        /// </summary>
+        public static void ShowMessage(string title, string message, OLEMSGICON icon, OLEMSGBUTTON button, OLEMSGDEFBUTTON defaultButton)
```
gregg-miskelly: nit: I might suggest renaming this to something that makes it clear that the message box will be put up async. Maybe `PostUIMessage`.

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -21,47 +26,66 @@ public static DockerConnection GetDockerConnection(string name)
...
-                displayName = remoteConnection.Name + '/' + containerName;
+            }
+            else if (connectionStrings.Length == 1)
+            {
+                // local connection
+                dockerString = connectionStrings[0];
             }
             else
             {
                 // TODO: This will be replaced by the docker container selection dialog
```
gregg-miskelly: Do we need to do something to display an error message for 16.2?

src/SSHDebugPS/CommandRunner.cs
pieandcakes: I'll remove the virtual

---
In reply to: [297443578](https://github.com/microsoft/MIEngine/pull/884#discussion_r297443578) [](ancestors = 297443578)

andrewcrawley:
>                 }

Does this override need the same logic as in the base class to handle when the the process has exited, but there's still data waiting to be read from the buffer? Wouldn't it just continue to spin here reading zero bytes over and over until someone signals the cancellation token?

---
Refers to: src/SSHDebugPS/CommandRunner.cs:321 in f43337d. [](commit_id = f43337d55481aa56ffcae83b7eb4de7eaf6c0973, deletion_comment = False)

src/SSHDebugPS/CommandRunner.cs
```diff
@@ -9,304 +9,304 @@
...
         }
 
-        private void OnProcessOutput(object sender, DataReceivedEventArgs e)
+        public void Write(string text)
         {
             lock (_lock)
             {
-                OutputReceived?.Invoke(this, e.Data);
-                if (_hasExited && _localProcess.StandardOutput.EndOfStream)
+                if (IsRunning())
```
andrewcrawley: >if (IsRunning()) [](start = 16, length = 16)

Is there a reason this silently drops input if the command isn't running, but the SSH version throws an exception?  Seems like the behavior should be consistent.

andrewcrawley:
>             EnsureRunning();

Is there a reason this implementation does no locking, but the local process version does?

---
Refers to: src/SSHDebugPS/CommandRunner.cs:372 in f43337d. [](commit_id = f43337d55481aa56ffcae83b7eb4de7eaf6c0973, deletion_comment = False)

src/SSHDebugPS/CommandRunner.cs
```diff
@@ -9,304 +9,304 @@
...
-                            }
-                        }
-                        return; // end of stream
-                    }
-
-                    action(task.Result);
+                    throw new InvalidOperationException("Process already running");
                 }
+                CleanUpProcess();
             }
```
andrewcrawley: Personally, I'd replace this with a check that throws an ObjectDisposedException if the object has been disposed - it's unusual to have an object that implements IDisposable but is still usable after disposing.

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -180,12 +129,27 @@ public override List<Process> ListProcesses()
             TryGetUsername(out username);
 
             string commandOutput;
+            int exitCode;
 
             if (!ExecuteCommand(PSOutputParser.PSCommandLine, Timeout.Infinite, false, out commandOutput))
             {
-                if (!ExecuteCommand(PSOutputParser.AltPSCommandLine, Timeout.Infinite, false, out commandOutput))
+                commandOutput = string.Empty;
+                if (!ExecuteCommand(PSOutputParser.AltPSCommandLine, Timeout.Infinite, false, out commandOutput, out errorMessage, out exitCode))
```
pieandcakes: Extra parameter. Removed

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -21,47 +26,66 @@ public static DockerConnection GetDockerConnection(string name)
...
-            if (connectionStrings.Length == 1)
-            {
-                // local connection
-                containerName = connectionStrings[0];
-                settings = new DockerExecShellSettings(containerName, hostIsUnix: false);
-                displayName = name;
-                // TODO: Verify container exists on local machine
-            }
-            else if (connectionStrings.Length == 2)
+            // Assume format is <server>/<hostname>::<container> where if <server> is specified, it is for SSH
```
andrewcrawley: ><server>/<hostname>::<container> [](start = 32, length = 32)

typo?  Should be a single colon, right?

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -196,17 +160,23 @@ public override List<Process> ListProcesses()
...
                 throw new CommandFailedException(error);
             }
             else
                 return exitCode == 0;
         }
+
+        protected bool ExecuteCommand(string command, int timeout, bool throwOnFailure, out string commandOutput)
+        {
+            int exitCode;
+            return ExecuteCommand(command, timeout, throwOnFailure, out commandOutput, out exitCode);
```
andrewcrawley: >exitCode [](start = 91, length = 8)

Not sure if you're on C# 7, but if you are, you can use "_" as a built-in "discard" variable so you don't have to declare one just for the out parameter.

andrewcrawley:
>                 //TODO: Once 'action' is cancelable, provide token to the action instead.

TODO is now accomplished!

---
Refers to: src/SSHDebugPS/VSOperationWaiter.cs:27 in f43337d. [](commit_id = f43337d55481aa56ffcae83b7eb4de7eaf6c0973, deletion_comment = False)

src/SSHDebugPS/PipeConnection.cs
pieandcakes: Good to know. I don't know our entire build chain (Azure pipelines, devops, etc) but I do know we aren't supporting C#7.2 for named parameters.

---
In reply to: [297459596](https://github.com/microsoft/MIEngine/pull/884#discussion_r297459596) [](ancestors = 297459596)

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -46,33 +54,43 @@ public static IEnumerable<IContainerInstance> GetLocalDockerContainers()
                         if (args.Trim()[0] != '{')
                         {
                             // output isn't json, command Error
-                            string errorMessage = string.Format(CultureInfo.CurrentCulture, UIResources.CommandExecutionErrorFormat, dockerCommand, args);
-                            throw new CommandFailedException(errorMessage);
+                            //string errorMessage = UIResources.CommandExecutionErrorFormat.FormatCurrentCultureWithArgs("{0} {1}".FormatInvariantWithArgs(settings.Command, settings.CommandArgs), args);
```
WardenGnaw: Commented out code?

src/SSHDebugPS/CommandRunner.cs
```diff
@@ -9,304 +9,286 @@
...
+        protected void VerifyProcessExitedAndNotify()
         {
             lock (_lock)
             {
-                if (!_hasExited && _outputReadTask.IsCompleted)
+                if (!_hasExited && _process.HasExited)
                 {
                     _hasExited = true;
-                    Closed?.Invoke(this, _localProcess.ExitCode);
+                    Closed?.Invoke(this, _process.ExitCode);
```
gregg-miskelly: >Closed [](start = 20, length = 6)

Do you want to fire the Closed event under the lock?

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -46,33 +55,55 @@ public static IEnumerable<IContainerInstance> GetLocalDockerContainers()
...
-                // might need to throw an exception here too??
-                if (exitCode != 0)
+                bool cancellationRequested = false;
+                VS.VSOperationWaiter.Wait(UIResources.QueryingForContainersMessage, false, (cancellationToken) =>
                 {
-                    Debug.Fail(FormattableString.Invariant($"Exit Code: {exitCode}"));
-                    return null;
+                    while (!resetEvent.WaitOne(2000) && !cancellationToken.IsCancellationRequested)
+                    { }
+                    cancellationRequested = cancellationToken.IsCancellationRequested;
```
gregg-miskelly: >cancellationRequested = cancellationToken.IsCancellationRequested; [](start = 20, length = 66)

I don't understand, why are you checking this inside the lambda?

gregg-miskelly: Pierson explained it.

---
In reply to: [297897088](https://github.com/microsoft/MIEngine/pull/884#discussion_r297897088) [](ancestors = 297897088)

src/SSHDebugPS/AD7/AD7Port.cs
```diff
@@ -65,24 +66,26 @@ public bool IsConnected
...
         {
+            int hr = HR.S_OK;
             IEnumDebugProcesses2 result = null;
             var connection = GetConnection();
 
             if (connection == null)
             {
-                processEnum = null;
-                return HR.E_REMOTE_CONNECT_USER_CANCELED;
+                // Don't return a failure to prevent vsdebug.dll from showing an error message
```
pieandcakes: @gregg-miskelly I verified that this only gives me one error message. `null` doesn't work as it generates a different error.

PR #885, opened 2019-06-27
Add support for querying /proc fs on Linux for Docker scenario

pieandcakes:
Add a fallback if ps isn't available to query /proc file system on Linux to obtain process list.

### DIFF
--- a/src/SSHDebugPS/ProcFSOutputParser.cs
+++ b/src/SSHDebugPS/ProcFSOutputParser.cs
@@ -1,0 +1,110 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using System.Windows.Documents;
+
+namespace Microsoft.SSHDebugPS
+{
+    public class ProcFSOutputParser
+    {
+        public static string EscapedCommandText => CommandText.Replace("$", "\\$");
+        public static string CommandText => @"echo shell-process:$$; for filename in /proc/[0-9]*; do echo $filename,cmdline:$(tr '\0' ' ' < $filename/cmdline 2>/dev/null),ls:$(ls -lh $filename/exe 2>/dev/null); done";
+
+        private const string ShellProcessPrefix = "shell-process:";
+        private readonly Regex _linePattern = new Regex(@"^/proc/([0-9]+),cmdline:(.*),ls:(.*)$", RegexOptions.None);
+
+        private List<Process> _processList = new List<Process>();
+        private uint _shellProcess = uint.MaxValue;
+
+        public static List<Process> Parse(string output, string username)
+        {
+            return new ProcFSOutputParser().ParseInternal(output, username);
+        }
+
+        private ProcFSOutputParser()
+        { }
+
+        private List<Process> ParseInternal(string output, string username)
+        {
+            using (var reader = new StringReader(output))
+            {
+                while (true)
+                {
+                    var line = reader.ReadLine();
+                    if (line == null)
+                        break;
+
+                    ProcessLine(line.Trim(), username);
+                }
+
+                if (_processList.Count == 0)
+                {
+                    throw new CommandFailedException(StringResources.Error_PSFailed);
+                }
+
+                _processList.Sort(
+                    (x, y) => x.Id < y.Id ? -1
+                            : x.Id == y.Id ? 0
+                            : 1);
+            }
+
+            return _processList;
+        }
+
+        private void ProcessLine(string line, string username)
+        {
+            if (string.IsNullOrWhiteSpace(line))
+                return;
+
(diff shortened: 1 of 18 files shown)


src/SSHDebugPS/ProcFSOutputParser.cs
```diff
@@ -0,0 +1,110 @@
+﻿using System;
```
pieandcakes: Missing Header

gregg-miskelly:
>         public override List<Process> ListProcesses()

Is there a reason that we are using different code for enumerating processes in SSH vs Pipe?

---
Refers to: src/SSHDebugPS/PipeConnection.cs:131 in a8d5324. [](commit_id = a8d5324a85dcb83e208d29ea6340a90114d08bcf, deletion_comment = False)

src/SSHDebugPS/StringResources.resx
```diff
@@ -172,16 +172,32 @@
...
   </data>
+  <data name="Error_OperationTimedOut" xml:space="preserve">
+    <value>Operation timed out.</value>
+  </data>
+  <data name="Error_ProcessListFailedFormat" xml:space="preserve">
+    <value>{0}
+
+{1}</value>
+    <comment>{0} = 'ps' error message, {1} = procfs error message</comment>
+  </data>
```
gregg-miskelly: Do we need some context text to indicate what the problem was? If not - no reason to localize this.

src/SSHDebugPS/StringResources.resx
```diff
@@ -172,16 +172,32 @@
...
+  <data name="Error_ProcessListFailedTitle" xml:space="preserve">
+    <value>Unable to obtain process list</value>
```
gregg-miskelly: >list [](start = 36, length = 4)

Add a period at the end?

pieandcakes: This is the title. Do you want a period in the title?

src/SSHDebugPS/StringResources.resx
pieandcakes: I put it here to get the new lines to show up in the message dialog.

src/SSHDebugPS/ProcFSOutputParser.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using System.Windows.Documents;
+
+namespace Microsoft.SSHDebugPS
+{
+    public class ProcFSOutputParser
+    {
+        public static string EscapedCommandText => CommandText.Replace("$", "\\$");
+        public static string CommandText => @"echo shell-process:$$; for filename in /proc/[0-9]*; do echo $filename,cmdline:$(tr '\0' ' ' < $filename/cmdline 2>/dev/null),ls:$(ls -lh $filename/exe 2>/dev/null); done";
```
andrewcrawley: >(ls -lh $filename/exe 2>/dev/null) [](start = 176, length = 34)

Not saying this needs to change, but if you run into issues with parsing the output of "ls" on different platforms, you might try using "stat" to get the username instead, e.g.:

andrew@andrewcr-lnx-1:~$ stat -c %U /proc/70532/exe
andrew

pieandcakes:
> > ```
> >     public override List<Process> ListProcesses()
> > ```
> 
> Is there a reason that we are using different code for enumerating processes in SSH vs Pipe?
> 
> Refers to: src/SSHDebugPS/PipeConnection.cs:131 in [a8d5324](https://github.com/microsoft/MIEngine/commit/a8d5324a85dcb83e208d29ea6340a90114d08bcf). [](commit_id = [a8d5324](https://github.com/microsoft/MIEngine/commit/a8d5324a85dcb83e208d29ea6340a90114d08bcf), deletion_comment = False)

@gregg-miskelly the proc filesystem doesn't exist on macOS but SSH works to a macOS device

src/SSHDebugPS/ProcFSOutputParser.cs
gregg-miskelly: >$filename [](start = 107, length = 9)

I am not sure if it is worth it, but shellcheck.net suggests quoting all these various commands/variables to avoid problems with filenames that have spaces...

gregg-miskelly:
Otherwise LGTM

src/SSHDebugPS/StringResources.resx
pieandcakes: I rewrote it

src/SSHDebugPS/ProcFSOutputParser.cs
pieandcakes: added a comment so it doesn't get lost

pieandcakes: i think your regex covers that and there shouldn't be a space in `filename`

src/SSHDebugPS/StringResources.resx
pieandcakes: @andrewcrawley convinced me why i should period it.

PR #891, opened 2019-07-08
Add Advanced Settings field and Hostname support for Docker

pieandcakes:


src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -142,5 +150,44 @@ public static SSHConnection GetSSHConnection(string name)
...
         }
+
+        /// <summary>
+        /// Open the ContainerPickerDialog
+        /// </summary>
+        /// <param name="hwnd">Parent hwnd or IntPtr.Zero</param>
+        /// <param name="connectionString">[out] connection string obtained by the dialog</param>
+        /// <returns></returns>
+        public static bool ShowContainerPickerWindow(IntPtr hwnd, out string connectionString)
+        {
```
gregg-miskelly: >{ [](start = 8, length = 1)

Throw if we aren't on the UI thread?

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -142,5 +150,44 @@ public static SSHConnection GetSSHConnection(string name)
...
+            if (hwnd == IntPtr.Zero) // get the VS main window hwnd
+            {
+                try
+                {
+                    // parent to the global VS window
+                    DTE dte = (DTE)Package.GetGlobalService(typeof(SDTE));
+                    hwnd = new IntPtr(dte?.MainWindow?.HWnd ?? 0);
+                }
+                catch // No DTE?
+                { }
```
gregg-miskelly: >{ [](start = 16, length = 1)

Assert?

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -35,6 +34,16 @@ public static DockerConnection GetDockerConnection(string name)
...
             string[] connectionStrings = name.Split('/');
 
+            // Format is wrong, ask the user to select from the dialog
+            if (connectionStrings.Length > 2 || connectionStrings.Length < 1)
+            {
+                string connectionString;
+                if (ShowContainerPickerWindow(IntPtr.Zero, out connectionString))
+                {
+                    return GetDockerConnection(connectionString);
+                }
```
gregg-miskelly: >} [](start = 16, length = 1)

if this does not return true, do you really want to fall through to the rest of this function?

gregg-miskelly: Oh, I see, maybe you do and are trying to fall through to the error block? Might be worth a comment if so.

---
In reply to: [301297533](https://github.com/microsoft/MIEngine/pull/891#discussion_r301297533) [](ancestors = 301297533)

src/SSHDebugPS/packages.config
```diff
@@ -4,4 +4,6 @@
   <package id="Microsoft.VisualStudio.Imaging" version="15.0.26606" targetFramework="net45" userInstalled="true" />
   <package id="Microsoft.VisualStudio.Utilities" version="15.0.26606" targetFramework="net462" />
   <package id="Newtonsoft.Json" version="6.0.8" targetFramework="net45" />
+  <package id="VSSDK.DTE" version="7.0.4" targetFramework="net461" />
```
gregg-miskelly: >VSSDK.DTE [](start = 15, length = 9)

I think you want this guy instead: [https://www.nuget.org/packages/EnvDTE/](https://www.nuget.org/packages/EnvDTE/). `VSSDK.DTE` is probably fine, but it isn't the official copy.

pieandcakes: @andrewcrawley gave me this one. I'll update it.

src/SSHDebugPS/Docker/DockerContainerInstance.cs
```diff
@@ -0,0 +1,124 @@
...
+                    other.Id.StartsWith(Id, StringComparison.OrdinalIgnoreCase) ? true : false;
+            }
+
+            return false;
+        }
+
+        protected override int GetHashCodeInternal()
+        {
+            // Id should be unique between Docker containers. 
+            return Id.GetHashCode();
```
gregg-miskelly: >return Id.GetHashCode(); [](start = 12, length = 24)

This doesn't agree with `EqualsInternal` -- strings that are prefix equal will get different hash codes, and strings that differ just by case will get different hash codes. Is that going to be a problem for your usage? It might not be possible to correctly implement GetHashCode (aside from maybe just returning the `ToLowerInvariant` of the first character of the name).

src/SSHDebugPS/UI/ContainerInstance.cs
```diff
@@ -9,123 +9,80 @@
...
-            }
-            set
+            if (obj is IContainerInstance instance)
             {
-                if (_isSelected != value)
-                {
-                    _isSelected = value;
-                    OnPropertyChanged(nameof(IsSelected));
-                }
+                this.Equals(instance);
```
gregg-miskelly: >this [](start = 16, length = 4)

return

pieandcakes:
@dustincoleman please review xaml.

gregg-miskelly:
Otherwise LGTM

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml.cs
```diff
@@ -89,24 +90,38 @@ private void DialogWindow_KeyDown(object sender, System.Windows.Input.KeyEventAr
             }
         }
 
+        private void HostnameTextbox_KeyDown(object sender, KeyEventArgs e)
```
dustincoleman: Is this used?

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
```diff
@@ -222,28 +326,64 @@
...
-                    Grid.Column="0"
-                    Margin="0,10,0,10"
-                    Padding="10,0,10,0"
-                    Text="{ Binding StatusText, Mode=OneWay}"
-                    TextWrapping="Wrap"/>
+                <TextBox x:Uid="uidHostnameTextBlock"
+                         Grid.Row="0"
+                         Grid.Column="1"
+                         x:Name="hostnameTextBlock"
+                         Text="{Binding Path=Hostname, Mode=TwoWay}"
```
dustincoleman: Should this binding have and UpdateSourceTrigger of TextChanged instead of the default of LostFocus?

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
```diff
@@ -222,28 +326,64 @@
...
 
-        <TextBlock x:Uid="uidStatusLabel"
-                    Style="{StaticResource StatusLabelStyle}"
-                    Grid.Row="2"
-                    Grid.Column="0"
-                    Margin="0,10,0,10"
-                    Padding="10,0,10,0"
-                    Text="{ Binding StatusText, Mode=OneWay}"
-                    TextWrapping="Wrap"/>
+                <TextBox x:Uid="uidHostnameTextBlock"
```
dustincoleman: Should pressing enter trigger a refresh? I noticed that there is a handler in the codebehind, but it doesn't appear to be hooked up.

src/SSHDebugPS/Docker/DockerContainerInstance.cs
pieandcakes: It looks like it gets called a bunch when the dialog starts because the existing list that it is put into is compared against itself. I'm taking the first character implementation.

src/SSHDebugPS/Docker/DockerContainerInstance.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        protected override bool EqualsInternal(ContainerInstance instance)
+        {
+            if (instance is DockerContainerInstance other)
+            {
+                // the id can be a partial on a container
+                return String.Equals(Id, other.Id, StringComparison.OrdinalIgnoreCase) ? true :
+                    Id.StartsWith(other.Id, StringComparison.OrdinalIgnoreCase) ? true :
+                    other.Id.StartsWith(Id, StringComparison.OrdinalIgnoreCase) ? true : false;
```
andrewcrawley: Writing this as a nested ternary seems kind of confusing - I had to read it a couple of times to make sure I understood what it was checking.  Could this be written as your 3 checks just ||'ed together?

src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
```diff
@@ -211,7 +244,9 @@ public IContainerInstance SelectedContainerInstance
             }
             set
             {
-                if (!(_selectedContainerInstance != null && _selectedContainerInstance.Name.Equals(value.Name, StringComparison.Ordinal)))
+                // checking cases that they are not equal
+                if (((_selectedContainerInstance != null) && (value == null || !string.Equals(_selectedContainerInstance.Name, value.Name, StringComparison.Ordinal)))
+                    || (_selectedContainerInstance == null && value != null))
```
andrewcrawley: Now that `IContainerInstance` implements `IEquatable<T>`, can you use that here to simplify this check?

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
```diff
@@ -176,24 +278,25 @@
...
         <Grid.RowDefinitions>
-            <RowDefinition Height="Auto" />
-            <RowDefinition Height="Auto" />
-            <RowDefinition Height="Auto" />
-            <RowDefinition Height="Auto" />
-            <RowDefinition Height="*" />
-            <RowDefinition Height="Auto" />
+            <RowDefinition Height="Auto" /> <!--0:Connectionlabel-->
+            <RowDefinition Height="Auto" /> <!--1:Connection combo box/add new connection-->
+            <RowDefinition Height="Auto" /> <!--4:Advanced Settings-->
```
andrewcrawley: > [](start = 44, length = 26)

These comments look like helpful reminders for what content is in which row, but the row numbers and content descriptions don't seem to match up (for example. shouldn't the ContainerList be row 4?).  Did I misinterpret?

dustincoleman:
> @dustincoleman please review xaml.

Signed Off

PR #901, opened 2019-07-26
UX Board recommended fixes for the Attach to Process Docker dialog

pieandcakes:
@CarpeDiEmma

### DIFF
--- a/src/SSHDebugPS/UI/ContainerInstance.cs
+++ b/src/SSHDebugPS/UI/ContainerInstance.cs
@@ -15,11 +15,10 @@ using System.Threading.Tasks;
 
 namespace Microsoft.SSHDebugPS.Docker
 {
-    public interface IContainerInstance : IEquatable<IContainerInstance>, INotifyPropertyChanged
+    public interface IContainerInstance : IEquatable<IContainerInstance>
     {
         string Id { get; }
         string Name { get; }
-        bool GetResult(out string selectedQualifier);
     }
 
     public abstract class ContainerInstance : IContainerInstance
@@ -27,19 +26,11 @@ namespace Microsoft.SSHDebugPS.Docker
         public abstract string Id { get; set; }
         public abstract string Name { get; set; }
 
-        public abstract bool GetResult(out string selectedQualifier);
-
-        #region INotifyPropertyChanged
-
-        public abstract event PropertyChangedEventHandler PropertyChanged;
-
-        #endregion
-
         #region IEquatable
 
         public static bool operator ==(ContainerInstance left, ContainerInstance right)
         {
-            if (ReferenceEquals(null, left) || ReferenceEquals(null, right))
+            if (left is null || right is null)
             {
                 return ReferenceEquals(left, right);
             }
--- a/src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
+++ b/src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
@@ -239,6 +239,7 @@
                         Grid.Column="0"
                         AutomationProperties.Name="ContainerDetailsExpander"
                         Command="{Binding ExpandCommand}"
+                        ToolTip="{Binding ExpandCommand.ToolTip}"
                         Background="Transparent"
                         BorderThickness="0">
                     <DockPanel>
@@ -380,7 +381,6 @@
     <!-- Main layout grid -->
     <Grid Margin="12">
         <Grid.RowDefinitions>
-            <RowDefinition Height="Auto" />
             <!--0:Connection label-->
             <RowDefinition Height="Auto" />
             <!--1:Connection combo box/add new connection-->
@@ -390,12 +390,13 @@
             <!--3:Hostname textbox -->
             <RowDefinition Height="Auto" />
             <!--4:Container status/refresh button-->
-            <RowDefinition Height="*" />
-            <!--5:ContainerList-->
             <RowDefinition Height="Auto" />
+            <!--5:ContainerList-->
+            <RowDefinition Height="*" />
             <!--6:Warning Icon and Status Text -->
             <RowDefinition Height="Auto" />
             <!--7: OK/Cancel -->
+            <RowDefinition Height="Auto" />
         </Grid.RowDefinitions>
         <Grid.ColumnDefinitions>
             <ColumnDefinition Width="*" />
@@ -461,17 +462,6 @@
                  PreviewKeyDown="HostnameTextbox_PreviewKeyDown"
                  Text="{Binding Path=Hostname, Mode=TwoWay,UpdateSourceTrigger=PropertyChanged}"
                  ToolTip="{x:Static local:UIResources.HostnameTip}">
-            <TextBox.Style>
-                <Style TargetType="TextBox">
-                    <Style.Triggers>
-                        <DataTrigger Binding="{Binding ElementName=hostnameTextBox, Path=Text.IsEmpty}"
-                                     Value="True">
-                            <Setter Property="Text"
-                                    Value="{x:Static local:UIResources.OptionalHintText}" />
-                        </DataTrigger>
-                    </Style.Triggers>
-                </Style>
-            </TextBox.Style>
         </TextBox>
 
         <!-- Container Count Text and Refresh Button -->
@@ -523,10 +513,10 @@
                                   Visibility="{Binding StatusIsError, Converter={StaticResource BoolToVisibilityConverter}, Mode=OneWay}"
                                   Moniker="{x:Static catalog:KnownMonikers.StatusWarning}" />
             <TextBlock DockPanel.Dock="Right"
-                       x:Name="StatusText"
+                       x:Name="StatusMessage"
                        TextWrapping="WrapWithOverflow"
                        Margin="6,12,9,6"
-                       Text="{Binding StatusText}" />
+                       Text="{Binding StatusMessage}" />
         </DockPanel>
 
         <!-- OK / Cancel buttons -->
--- a/src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
+++ b/src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
@@ -234,8 +234,10 @@
                     <RowDefinition Height="*" />
                     <RowDefinition Height="*" />
                 </Grid.RowDefinitions>
-                <Button Grid.Row="0"
+                <Button x:Name="ContainerDetailsExpander" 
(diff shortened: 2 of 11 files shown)


pieandcakes:
@dustincoleman please review also

pieandcakes:
image

pieandcakes:
image

src/SSHDebugPS/UI/ContainerInstance.cs
```diff
@@ -15,31 +15,22 @@
 
 namespace Microsoft.SSHDebugPS.Docker
 {
-    public interface IContainerInstance : IEquatable<IContainerInstance>, INotifyPropertyChanged
+    public interface IContainerInstance : IEquatable<IContainerInstance>
```
gregg-miskelly: >public [](start = 4, length = 6)

Should these classes still be public?

src/SSHDebugPS/UI/ViewModels/ContainerViewModel.cs
```diff
@@ -0,0 +1,180 @@
...
+    {
+        string Name { get; }
+        string Id { get; }
+        WindowsInput.ICommand ExpandCommand { get; }
+        bool GetResult(out string selectedQualifier);
+        bool IsExpanded { get; set; }
+        bool IsSelected { get; set; }
+    }
+
+    public abstract class ContainerViewModel<T> : IContainerViewModel
```
gregg-miskelly: ><T> [](start = 44, length = 3)

Out of curiosity - are you getting anything out of the fact that this is generic? It seems like you could just replace `T` with `IContainerInstance`, but maybe I am missing something.

src/SSHDebugPS/UI/ViewModels/ContainerViewModel.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        public override WindowsInput.ICommand ExpandCommand => new ContainerUICommand(Expand, "ExpandDetails", "ExpandToolTip");
+
+        private void Expand(object parameter)
+        {
+            IsExpanded = !IsExpanded;
+        }
+
+        // Gets the first 12 characters and appends an ellipsis
+        public string ShortId { get => Id.Substring(0, 12); }
```
gregg-miskelly: >Id.Substring(0, 12); [](start = 39, length = 20)

Any chance `Id` is less than or equal to 12 characters?

gregg-miskelly:
>                     OnPropertyChanged(nameof(StatusIsError));

Assert that either the new value is false, or _statusText isn't empty? If one assigns to this before StatusText the wrong notifications will be sent.

---
Refers to: src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs:225 in d89c949. [](commit_id = d89c949c56cf951c4cd1e9c6c8c29becaa135c2d, deletion_comment = False)

gregg-miskelly:
>                     OnPropertyChanged(nameof(StatusIsError));

Or remove the setter and add a method to update these instead, so that it isn't possible to get this wrong...

---
In reply to: [516089862](https://github.com/microsoft/MIEngine/pull/901#issuecomment-516089862) [](ancestors = 516089862)

---
Refers to: src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs:225 in d89c949. [](commit_id = d89c949c56cf951c4cd1e9c6c8c29becaa135c2d, deletion_comment = False)

src/SSHDebugPS/UI/UIResources.resx
```diff
@@ -209,4 +209,22 @@
   <data name="OKLabel" xml:space="preserve">
     <value>_OK</value>
   </data>
+  <data name="RefreshHyperlink" xml:space="preserve">
+    <value>Refresh</value>
+  </data>
+  <data name="RefreshToolTip" xml:space="preserve">
+    <value>Refresh container list.</value>
```
dustincoleman: >. [](start = 33, length = 1)

Nit: Not a sentence

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
```diff
@@ -146,144 +127,275 @@
...
     <Grid Margin="12">
         <Grid.RowDefinitions>
-            <RowDefinition Height="Auto" /> <!--0:Connectionlabel-->
-            <RowDefinition Height="Auto" /> <!--1:Connection combo box/add new connection-->
-            <RowDefinition Height="Auto" /> <!--2:Advanced Settings-->
-            <RowDefinition Height="Auto" /> <!--3:status/refresh button-->
-            <RowDefinition Height="*" />    <!--4:ContainerList-->
-            <RowDefinition Height="Auto" /> <!--5:OK/Cancel-->
+            <RowDefinition Height="Auto" />
+            <!--0:Connection label-->
```
dustincoleman: > [](start = 12, length = 25)

These should probably be to the right or above (not below) what they are labeling

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
```diff
@@ -292,154 +404,157 @@
...
+                 Text="{Binding Path=Hostname, Mode=TwoWay,UpdateSourceTrigger=PropertyChanged}"
+                 ToolTip="{x:Static local:UIResources.HostnameTip}">
+            <TextBox.Style>
+                <Style TargetType="TextBox">
+                    <Style.Triggers>
+                        <DataTrigger Binding="{Binding ElementName=hostnameTextBox, Path=Text.IsEmpty}"
+                                     Value="True">
+                            <Setter Property="Text"
+                                    Value="{x:Static local:UIResources.OptionalHintText}" />
+                        </DataTrigger>
```
dustincoleman: This looks like a watermark? In that case should the text color get lighter?

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml.cs
```diff
@@ -20,6 +24,34 @@ public ContainerPickerDialogWindow()
...
+                    {
+                        ContainerListBox.SelectedIndex = 0;
+                    }
+                    ((ListBoxItem)ContainerListBox.ItemContainerGenerator.ContainerFromItem(ContainerListBox.SelectedItem)).Focus();
+                    listItemFocused = true;
+                }
+            }
+            catch (Exception)
+            {
+                // Ignore if focus was failed to be set
```
dustincoleman: > // Ignore if focus was failed to be set [](start = 15, length = 40)

Assert?

src/SSHDebugPS/UI/ViewModels/ContainerViewModel.cs
pieandcakes: There shouldn't be but i'll add a guard just in case

src/SSHDebugPS/UI/ViewModels/ContainerViewModel.cs
pieandcakes: It was so I can access concrete type info in derived classes without having to cast it.

src/SSHDebugPS/UI/ContainerInstance.cs
pieandcakes: Yes, because WPF is awesome and needs it to be. If its internal, they won't load.

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
pieandcakes: Oops. That ended up not working so i removed it and added (Optional) to the label

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
```diff
@@ -146,144 +127,276 @@
...
+                    Value="Center" />
+            <Setter Property="VerticalAlignment"
+                    Value="Center" />
+            <Setter Property="SnapsToDevicePixels"
+                    Value="True" />
+
+            <Setter Property="Template">
+                <Setter.Value>
+                    <ControlTemplate TargetType="{x:Type Button}">
+                        <!--Style="{StaticResource TextBlockStyle}"-->
```
CarpeDiEmma: Should this be deleted?

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
```diff
@@ -292,154 +405,146 @@
 
         <!--Connection Type Picker-->
         <Label Grid.Row="0"
+               Grid.ColumnSpan="2"
                x:Uid="uidConnectionTypeLabel"
                x:Name="ConnectionTypeLabel"
+               Margin="0"
```
CarpeDiEmma: Do you need to have margin and padding out if they are both just 0?

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
```diff
@@ -292,154 +405,146 @@
...
-                         Margin="0,5,0,5"
-                         Padding="2" />
-                </Grid>
-        </Expander>
+        <Label Grid.Row="2"
+               Grid.ColumnSpan="2"
+               x:Uid="uidHostnameLabel"
+               x:Name="HostnameLabel"
+               Margin="0,9,0,0"
+               Padding="0"
```
CarpeDiEmma: I think you can remove this too

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
```diff
@@ -292,154 +405,146 @@
...
+               VerticalAlignment="Center"
+               Target="{Binding ElementName=hostnameTextBox}"
+               Content="{x:Static local:UIResources.HostnameLabel}" />
+        <TextBox Grid.Row="3"
+                 Grid.ColumnSpan="2"
+                 Margin="0,4,0,0"
+                 Padding="2"
+                 x:Uid="uidHostnameTextBox"
+                 x:Name="hostnameTextBox"
+                 AutomationProperties.Name="DockerHostnameOptional"
```
CarpeDiEmma: You can have spaces in automation properties. Also, you probably want to have it be "Optional Docker Hostname"

CarpeDiEmma: Also, don't you need to have this in a resource file for localization?

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
pieandcakes: This isn't the default which is why i set it to 0

PR #908, opened 2019-08-06
Fix Debug assertion where value can be 255

### LINKED ISSUE #895: comparison in Debug.Assert of MIResults.cs
https://github.com/microsoft/MIEngine/blob/7637a9b6f6a081ed5b03e85b2ba252ddbb16c642/src/MICore/MIResults.cs#L983
`v <= 255`
i have catch this in str = `...value=\"-1 '\\377'\"...`

pieandcakes:
Fixes #895

### DIFF
--- a/src/MICore/MIResults.cs
+++ b/src/MICore/MIResults.cs
@@ -980,7 +980,7 @@ namespace MICore
                         continue;
                     }
                 }
-                Debug.Assert(v < 255, "Value too large");
+                Debug.Assert(v <= 255, "Value too large");
                 bytes[cChars++] = (byte)v;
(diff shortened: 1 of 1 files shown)


src/MICore/MIResults.cs
```diff
@@ -980,7 +980,7 @@ bool SpanOctalChars(string str, ref int i, StringBuilder output)
                         continue;
                     }
                 }
-                Debug.Assert(v < 255, "Value too large");
+                Debug.Assert(v <= 255, "Value too large");
```
WardenGnaw: nit: use Byte.MaxValue

PR #909, opened 2019-08-07
Fix way we check containers are running

pieandcakes:
Attempt to check container name against ps list of containers

### DIFF
--- a/src/SSHDebugPS/Docker/DockerContainerInstance.cs
+++ b/src/SSHDebugPS/Docker/DockerContainerInstance.cs
@@ -63,8 +63,7 @@ namespace Microsoft.SSHDebugPS.Docker
             {
                 // the id can be a partial on a container
                 return String.Equals(Id, other.Id, StringComparison.Ordinal) ||
-                    Id.StartsWith(other.Id, StringComparison.Ordinal
-                    ) ||
+                    Id.StartsWith(other.Id, StringComparison.Ordinal) ||
                     other.Id.StartsWith(Id, StringComparison.Ordinal);
             }
 
--- a/src/SSHDebugPS/Docker/DockerContainerInstance.cs
+++ b/src/SSHDebugPS/Docker/DockerContainerInstance.cs
@@ -56,14 +56,16 @@ namespace Microsoft.SSHDebugPS.Docker
 
         #endregion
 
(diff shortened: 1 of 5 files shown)


pieandcakes:
Also fixes the container id/name check as it is case sensitive.

src/SSHDebugPS/Docker/DockerContainerInstance.cs
```diff
@@ -56,14 +56,16 @@ public static DockerContainerInstance Create(string json)
...
         {
             if (instance is DockerContainerInstance other)
             {
                 // the id can be a partial on a container
-                return String.Equals(Id, other.Id, StringComparison.OrdinalIgnoreCase) ||
-                    Id.StartsWith(other.Id, StringComparison.OrdinalIgnoreCase) ||
-                    other.Id.StartsWith(Id, StringComparison.OrdinalIgnoreCase);
+                return String.Equals(Id, other.Id, StringComparison.Ordinal) ||
+                    Id.StartsWith(other.Id, StringComparison.Ordinal
+                    ) ||
```
andrewcrawley: Formatting error?

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -102,6 +103,38 @@ public static IEnumerable<DockerContainerInstance> GetLocalDockerContainers(stri
...
+            }
+            else
+            {
+                containers = GetLocalDockerContainers(hostName);
+            }
+
+            if (containers != null)
+            {
+                // Check if the user entered the containerName or possibly part of the Id. 
+                if (containers.Any(container => string.Equals(container.Name, containerName, StringComparison.Ordinal) ? true
```
andrewcrawley: I think this would be logically equivalent if you just had the `.StartsWith` condition, since any string "starts with" itself.  Also, what happens if the prefix you enter matches more than one container?

pieandcakes: The first comparison is against the name, which has to be fully specified. the second (startswith) is against the id which is partial. 

Good point on the duplicate but then i want `docker` to do what it defaults to when there are more than 1?

andrewcrawley: Ah, I missed that they were checking separate properties.  Sorry!

PR #915, opened 2019-08-16
Bugfixes For Docker Host and remote connections

pieandcakes:
* Fix parsing of connectionstrings
* change connectionstring format to
`<containerName>;ssh=<sshConnectionString;host=<dockerhost>`
* fix labelling and tool tips
* fix instance of single quoting to only use it when host is linux. the
ingle quoting does not work on Windows.

### DIFF
--- a/src/SSHDebugPS/ConnectionManager.cs
+++ b/src/SSHDebugPS/ConnectionManager.cs
@@ -20,73 +20,44 @@ namespace Microsoft.SSHDebugPS
 {
     internal class ConnectionManager
     {
+
         public static DockerConnection GetDockerConnection(string name)
         {
-            if (string.IsNullOrWhiteSpace(name))
+           if (string.IsNullOrWhiteSpace(name))
                 return null;
 
-            Connection remoteConnection = null;
-            DockerContainerTransportSettings settings = null;
+            DockerContainerTransportSettings settings;
+            Connection remoteConnection;
 
-            string containerName;
-            string dockerString;
-            string hostName = string.Empty;
-
-            // Assume format is <server>/<hostname>::<container> where if <server> is specified, it is for SSH
-            string[] connectionStrings = name.Split('/');
-
-            // Format is wrong, ask the user to select from the dialog
-            if (connectionStrings.Length > 2 || connectionStrings.Length < 1)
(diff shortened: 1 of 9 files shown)


src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -20,73 +20,44 @@ namespace Microsoft.SSHDebugPS
 {
     internal class ConnectionManager
     {
+
```
pieandcakes: Remove empty space

pieandcakes:
Tested: 
* local docker connections
* remote docker connections via SSH
* verified strings are being built correctly when docker_host is specified.

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -12,15 +12,85 @@ namespace Microsoft.SSHDebugPS.Docker
 {
     internal class DockerConnection : PipeConnection
     {
+        #region Statics
+
+        internal const string SshPrefix = "ssh=";
```
gregg-miskelly: These strings are user visible? If so, I would suggest being a bit more forgiving about parsing them - we should probably allow whitespace after all the tokens which gets removed.

pieandcakes: so trim before and after the '='? I can do that.

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -12,15 +13,92 @@ namespace Microsoft.SSHDebugPS.Docker
...
+                        Match match = dockerHostRegex.Match(segment);
+                        hostName = segment.Substring(match.Length);
+                    }
+                    else if (segment.Contains("="))
+                    {
+                        invalidString = true;
+                    }
+                    else
+                    {
+                        Debug.Assert(string.IsNullOrWhiteSpace(containerName), "containerName should be empty");
```
gregg-miskelly: >Debug.Assert(string.IsNullOrWhiteSpace(containerName), "containerName should be empty"); [](start = 24, length = 88)

Set `invalidString = true;` instead of just asserting?

PR #918, opened 2019-08-20
Add MIMode telemetry

pieandcakes:
We would like to see the number of users using gdb versus lldb MIMode.

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -963,7 +963,7 @@ namespace OpenDebugAD7
                 }
 
                 var properties = new Dictionary<string, object>(StringComparer.Ordinal);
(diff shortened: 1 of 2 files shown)


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -960,6 +963,7 @@ protected override void HandleAttachRequestAsync(IRequestResponder<AttachArgumen
                 }
 
                 var properties = new Dictionary<string, object>(StringComparer.Ordinal);
+                properties.Add(DebuggerTelemetry.TelemetryMIMode, string.Empty);
```
gregg-miskelly: >string.Empty [](start = 66, length = 12)

Don't you want `miMode` here rather than `string.Empty`?

pieandcakes: missed a placeholder! thanks for catching it.

PR #920, opened 2019-08-20
Add quoting to paths for launch command

pieandcakes:
This fixes vscode-cpptools#4030 where user had `(` in their working directory.


pieandcakes:
I also quoted all the `tmp` files we create just in case. 

Tested on Linux that the files are cleaned up and that the folder path with `(` in it works.

PR #934, opened 2019-09-05
Add OpenDebugAD7 support for disable auto-solib

WardenGnaw:
Add entry for SymbolLoadInfo for OpenDebugAD7 BaseLaunchOptions.

Then read the launchJson and convert it into XML.

### DIFF
--- a/src/OpenDebugAD7/MILaunchOptions.cs
+++ b/src/OpenDebugAD7/MILaunchOptions.cs
@@ -167,7 +167,7 @@ namespace OpenDebugAD7
         private class SymbolLoadInfo
         {
             [JsonProperty]
-            public bool LoadAll { get; set; }
+            public bool? LoadAll { get; set; }
 
             [JsonProperty]
             public string ExceptionList { get; set; }
@@ -509,7 +509,16 @@ namespace OpenDebugAD7
                     string miDebuggerArgs = XmlSingleQuotedAttributeEncode(jsonLaunchOptions.MIDebuggerArgs);
                     xmlLaunchOptions.Append(String.Concat("  MIDebuggerArgs='", miDebuggerArgs, "'\n"));
                 }
-                xmlLaunchOptions.Append(String.Concat("  WaitDynamicLibLoad='false'\n"));
+
+                // If we get SymbolLoadInfo and an ExceptionList. We will need to have WaitDynamicLibLoad.
+                if (jsonLaunchOptions.SymbolLoadInfo != null && !String.IsNullOrWhiteSpace(jsonLaunchOptions.SymbolLoadInfo.ExceptionList))
+                {
+                    xmlLaunchOptions.Append(String.Concat("  WaitDynamicLibLoad='true'\n"));
+                }
+                else
+                {
+                    xmlLaunchOptions.Append(String.Concat("  WaitDynamicLibLoad='false'\n"));
+                }
 
                 if (jsonLaunchOptions.MIDebuggerServerAddress != null)
                 {
@@ -596,7 +605,18 @@ namespace OpenDebugAD7
 
                 if (jsonLaunchOptions.SymbolLoadInfo != null)
                 {
-                    xmlLaunchOptions.Append("    <SymbolLoadInfo LoadAll='" + jsonLaunchOptions.SymbolLoadInfo.LoadAll.ToString().ToLower() + "' ExceptionList='" + XmlSingleQuotedAttributeEncode(jsonLaunchOptions.SymbolLoadInfo.ExceptionList) + "' />\n");
+                    xmlLaunchOptions.Append("    <SymbolLoadInfo ");
+                    if (jsonLaunchOptions.SymbolLoadInfo.LoadAll.HasValue)
+                    {
+                        xmlLaunchOptions.Append("LoadAll = '" + (jsonLaunchOptions.SymbolLoadInfo.LoadAll.Value ? "true" : "false") + "' ");
+                    }
+
+                    if (!String.IsNullOrWhiteSpace(jsonLaunchOptions.SymbolLoadInfo.ExceptionList))
+                    {
+                        xmlLaunchOptions.Append("ExceptionList='" + XmlSingleQuotedAttributeEncode(jsonLaunchOptions.SymbolLoadInfo.ExceptionList) + "' ");
+                    }
+
+                    xmlLaunchOptions.Append("/>\n");
                 }
 
                 xmlLaunchOptions.Append("</LocalLaunchOptions>");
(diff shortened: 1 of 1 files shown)


src/OpenDebugAD7/MILaunchOptions.cs
```diff
@@ -496,7 +509,16 @@ private static string GetLLDBMIPath()
                     string miDebuggerArgs = XmlSingleQuotedAttributeEncode(jsonLaunchOptions.MIDebuggerArgs);
                     xmlLaunchOptions.Append(String.Concat("  MIDebuggerArgs='", miDebuggerArgs, "'\n"));
                 }
-                xmlLaunchOptions.Append(String.Concat("  WaitDynamicLibLoad='false'\n"));
+
+                // If we get SymbolLoadInfo and an ExceptionList. We will need to have WaitDynamicLibLoad.
+                if (jsonLaunchOptions.SymbolLoadInfo != null && !String.IsNullOrWhiteSpace(jsonLaunchOptions.SymbolLoadInfo.ExceptionList))
```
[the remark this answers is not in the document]
pieandcakes: I think this behavior is wrong too. The comment in vscode-cpptools says that this list with either be an inclusion (if LoadAll is false) or an exclusion(LoadAll is true). How does setting WaitDynamicLibLoad need to change with each instance?

src/OpenDebugAD7/MILaunchOptions.cs
```diff
@@ -581,6 +603,22 @@ private static string GetLLDBMIPath()
                     xmlLaunchOptions.Append("    </Environment>\n");
                 }
 
+                if (jsonLaunchOptions.SymbolLoadInfo != null)
+                {
+                    xmlLaunchOptions.Append("    <SymbolLoadInfo ");
+                    if (jsonLaunchOptions.SymbolLoadInfo.LoadAll.HasValue)
+                    {
+                        xmlLaunchOptions.Append("LoadAll = '" + (jsonLaunchOptions.SymbolLoadInfo.LoadAll.Value ? "true" : "false") + "' ");
```
pieandcakes: SymbolLoadInfo.LoadAll.Value.ToString().ToLowerInvariant()

src/OpenDebugAD7/MILaunchOptions.cs
```diff
@@ -581,6 +603,22 @@ private static string GetLLDBMIPath()
...
+                    }
+
+                    if (!String.IsNullOrWhiteSpace(jsonLaunchOptions.SymbolLoadInfo.ExceptionList))
+                    {
+                        xmlLaunchOptions.Append("ExceptionList='" + XmlSingleQuotedAttributeEncode(jsonLaunchOptions.SymbolLoadInfo.ExceptionList) + "' ");
```
pieandcakes: What happens if you don't specify LoadAll but there is an exception list? Is the behavior that LoadAll is true or false?

WardenGnaw: > List of filenames (wildcards allowed). Modifies behavior of LoadAll. If LoadAll is true then don't load symbols for libs that match any name in the list. Otherwise only load symbols for libs that match.

src/OpenDebugAD7/MILaunchOptions.cs
```diff
@@ -160,6 +163,16 @@ private class JsonLocalLaunchOptions : JsonBaseLaunchOptions
             public bool AvoidWindowsConsoleRedirection { get; set; }
         }
 
+        [JsonObject]
+        private class SymbolLoadInfo
+        {
+            [JsonProperty]
+            public bool? LoadAll { get; set; }
```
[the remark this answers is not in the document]
WardenGnaw: Yes. Tested with a SymbolLoadInfo object with no fields. Then SymbolLoadInfo with LoadAll set, and then just SymbolLoadInfo with just ExceptionList set.

src/OpenDebugAD7/MILaunchOptions.cs
```diff
@@ -496,7 +509,17 @@ private static string GetLLDBMIPath()
                     string miDebuggerArgs = XmlSingleQuotedAttributeEncode(jsonLaunchOptions.MIDebuggerArgs);
                     xmlLaunchOptions.Append(String.Concat("  MIDebuggerArgs='", miDebuggerArgs, "'\n"));
                 }
-                xmlLaunchOptions.Append(String.Concat("  WaitDynamicLibLoad='false'\n"));
+
+                // If we get SymbolLoadInfo and an ExceptionList or LoadAll is set. We will need to have WaitDynamicLibLoad.
+                if (jsonLaunchOptions.SymbolLoadInfo != null && 
+                    (!String.IsNullOrWhiteSpace(jsonLaunchOptions.SymbolLoadInfo.ExceptionList) || jsonLaunchOptions.SymbolLoadInfo.LoadAll.HasValue))
+                {
+                    xmlLaunchOptions.Append(String.Concat("  WaitDynamicLibLoad='true'\n"));
```
pieandcakes: Did you test this to see if it behaves correctly?

src/OpenDebugAD7/MILaunchOptions.cs
```diff
@@ -581,6 +604,22 @@ private static string GetLLDBMIPath()
...
+                {
+                    xmlLaunchOptions.Append("    <SymbolLoadInfo ");
+                    if (jsonLaunchOptions.SymbolLoadInfo.LoadAll.HasValue)
+                    {
+                        xmlLaunchOptions.Append("LoadAll = '" + jsonLaunchOptions.SymbolLoadInfo.LoadAll.Value.ToString().ToLowerInvariant() + "' ");
+                    }
+
+                    if (!String.IsNullOrWhiteSpace(jsonLaunchOptions.SymbolLoadInfo.ExceptionList))
+                    {
+                        xmlLaunchOptions.Append("ExceptionList='" + XmlSingleQuotedAttributeEncode(jsonLaunchOptions.SymbolLoadInfo.ExceptionList) + "' ");
```
pieandcakes: can you validate that if individual strings are quoted in there that it works properly?

say my exclusionList is `"/foo bar/fooobar.so";"/baz baz/barfoo.so"` or some combination that we pass the right things to `gdb` ?

src/OpenDebugAD7/MILaunchOptions.cs
WardenGnaw: Tested with a program and a static library, with LoadAll set to false, it does not step into the library code. With LoadAll false and ExpectionList with the shared library name it does step in. With LoadAll true and ExceptionList set to the shared library name, it does not step in. 
    LoadAll true and exception list empty, it does what it does normally

src/OpenDebugAD7/MILaunchOptions.cs
WardenGnaw: If you add quotes, you are looking for the library name foobar.so", not foobar.so. I think it is filenames, and not paths so users should not be quoting between semicolons.

PR #945, opened 2019-11-22
Fixes for parsing the connection string for SSH

pieandcakes:
We were not handling a custom SSH port. Added support to display port in
the connection string if the port is not the default (Port 22)

Fixes: 
[Feedback] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1025875
[Bug] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1026403

### DIFF
--- a/src/SSHDebugPS/ConnectionManager.cs
+++ b/src/SSHDebugPS/ConnectionManager.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 using System.Windows.Interop;
 using EnvDTE;
 using liblinux;
@@ -22,7 +23,7 @@ namespace Microsoft.SSHDebugPS
     {
         public static DockerConnection GetDockerConnection(string name, bool supportSSHConnections)
         {
-           if (string.IsNullOrWhiteSpace(name))
+            if (string.IsNullOrWhiteSpace(name))
                 return null;
 
             DockerContainerTransportSettings settings;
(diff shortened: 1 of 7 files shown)


src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -153,5 +140,67 @@ public static bool ShowContainerPickerWindow(IntPtr hwnd, out string connectionS
             connectionString = string.Empty;
             return false;
         }
+
+        /// <summary>
+        /// Parses the SSH connection string. Expected format is some permutation of username@hostname:portnumber.
+        /// If not defined, will provide default values.
+        /// </summary>
+        internal static void ParseSSHConnectionString(string connectionString, out string userName, out string hostName, out int port)
```
WardenGnaw: nit: Unit Test this?

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -153,5 +140,67 @@ public static bool ShowContainerPickerWindow(IntPtr hwnd, out string connectionS
...
+                    return;
+                }
+            }
+            else
+            {
+                hostName = connectionString;
+            }
+
+            // Find if a port is specified. Handle possible IPV6 by grabbing the last colon
+            int lastColonIndex = hostName.LastIndexOf(':');
```
andrewcrawley: I think this is wrong - if I specify a connection string of "user@1:2:3:4:5:6:7:8", it returns host = "1:2:3:4:5:6:7" and port = "8".  I believe the only valid way to specify a port in an IPv6 address is to disambiguate by putting the entire address in square brackets, and the port afterwards, e.g. "[1:2:3:4:5:6:7:8]:22"

pieandcakes: Colon count: 
* 0, no port
* 1, has a port
* >2 ipv6, check for open and close `[]`

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -153,5 +140,67 @@ public static bool ShowContainerPickerWindow(IntPtr hwnd, out string connectionS
...
+            if (lastColonIndex >= 0)
+            {
+                int portStartPos = lastColonIndex + 1;
+                if (lastColonIndex > 0 && portStartPos < hostName.Length)
+                {
+                    string portString = hostName.Substring(portStartPos);
+                    int tempPort;
+                    if (Int32.TryParse(portString, out tempPort) && tempPort != port)
```
gregg-miskelly: >Int32.TryParse(portString, out tempPort) [](start = 24, length = 40)

I feel like if `int.TryParse` fails, we shouldn't just drop the rest off the hostname -- either reset hostname to the placeholder, or leave it as all the text after the `@`.

gregg-miskelly: >TryParse [](start = 30, length = 8)

Use the overload of TryParse that takes a format provider and use the invariant locale.

gregg-miskelly: You might want to just use a RegEx -- `Regex pattern = new Regex(@"@([^:]+)(:[0-9]+)?$", RegexOptions.None);`

Issue #946, opened 2019-12-30
break-insert for lldb-mi no longer requires 'on'


puremourning:
In this line, we have a comment that `break-insert` requires `on` for lldb-mi: https://github.com/microsoft/MIEngine/blob/2bb9706607edf0ce839bab4eb131adc8a555b6ec/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs#L872

This is no longer true (presumably, it has been fixed upstream in lldb-mi): https://github.com/lldb-tools/lldb-mi (maybe this change: https://github.com/lldb-tools/lldb-mi/commit/e448e35a9e4da5b5a4f0f80530ba30d5a619cb3e#diff-16ecaed41b16ef0b1fd18b3f74b1a2f2).

I think fixing this will allow us to fix debugging on macOS using vscode-cpptools: https://github.com/microsoft/vscode-cpptools/issues/3829

Taking the latest version of lldb-mi from GitHub, and building it:

* Including `-break-insert -f on main` does not work:

```
lldbmitest) BenMBP:attach ben$ ../../build/src/lldb-mi
(gdb)
-file-exec-and-symbols ../testdata/simple/simple
^done
(gdb)
=library-loaded,id="/Users/ben/Development/llvm-project/lldb-mi/test/testdata/simple/simple",target-name="/Users/ben/Development/llvm-project/lldb-mi/test/testdata/simple/simple",host-name="/Users/ben/Development/llvm-project/lldb-mi/test/testdata/simple/simple",symbols-loaded="1",symbols-path="/Users/ben/Development/llvm-project/lldb-mi/test/testdata/simple/simple.dSYM/Contents/Resources/DWARF/simple",loaded_addr="-",size="4096"
-break-insert -f on main
^done,bkpt={number="1",type="breakpoint",disp="keep",enabled="y",addr="0xffffffffffffffff",func="??",file="??",fullname="??/??",line="0",pending=["on"],times="0",original-location="on"}
(gdb)
=breakpoint-modified,bkpt={number="1",type="breakpoint",disp="keep",enabled="y",addr="0xffffffffffffffff",func="??",file="??",fullname="??/??",line="0",pending=["on"],times="0",original-location="on"}
(gdb)

```

As you can see, the breakpoint is pending and has no file/line info.

* Including `-break-insert -f main` works:

```
(lldbmitest) BenMBP:attach ben$ ../../build/src/lldb-mi
(gdb)
[...]

PR #950, opened 2020-01-04
Fix hang when checking for lldb-mi break-insert command syntax

puremourning:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/miengine/950)

PR #952, opened 2020-01-06
Check to find '-break-insert' versions.

puremourning:
If '-break-insert -f main' returns unknown command, we revert to use
'-break-insert -f on main'

We need to do all this break-insert command fetching async so that we
run the command and get the result. Otherwise we just hang and
never actually process the result.


puremourning:
Replaces #950 fixes #946

WardenGnaw:
Thank you for your contribution @puremourning !

PR #954, opened 2020-01-09
Add option to disable SSH Connections

pieandcakes:
* Add ability to launch DockerPortPicker UI with SSH Disabled
* Fixed bug where user can enter hostname and hit ok which would return
a local container wiht a hostname, creating an invalid connection
string.
* Fixed it so you can't hit ok until a container has been selected

### DIFF
--- a/src/SSHDebugPS/ConnectionManager.cs
+++ b/src/SSHDebugPS/ConnectionManager.cs
@@ -20,7 +20,7 @@ namespace Microsoft.SSHDebugPS
 {
     internal class ConnectionManager
     {
-        public static DockerConnection GetDockerConnection(string name)
+        public static DockerConnection GetDockerConnection(string name, bool supportSSHConnections)
         {
            if (string.IsNullOrWhiteSpace(name))
                 return null;
@@ -32,7 +32,7 @@ namespace Microsoft.SSHDebugPS
             {
                 string connectionString;
 
-                bool success = ShowContainerPickerWindow(IntPtr.Zero, out connectionString);
+                bool success = ShowContainerPickerWindow(IntPtr.Zero, supportSSHConnections, out connectionString);
                 if (success)
                 {
                     success = DockerConnection.TryConvertConnectionStringToSettings(connectionString, out settings, out remoteConnection);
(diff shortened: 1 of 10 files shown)


pieandcakes:
@dustincoleman

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -146,7 +146,7 @@ public static bool ShowContainerPickerWindow(IntPtr hwnd, out string connectionS
             bool? dialogResult = dialog.ShowModal();
             if (dialogResult.GetValueOrDefault(false))
             {
-                connectionString = dialog.SelectedContainerConnectionString;
+                connectionString = dialog.Model.SelectedContainerConnectionString;
```
dustincoleman: >Model.SelectedContainerConnectionString [](start = 42, length = 39)

nit: This feels pretty weird. Does the view model need to be public or could you just keep the SelectedContainerConnectionString property and return the view models value?

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -146,7 +146,7 @@ public static bool ShowContainerPickerWindow(IntPtr hwnd, out string connectionS
             bool? dialogResult = dialog.ShowModal();
             if (dialogResult.GetValueOrDefault(false))
```
dustincoleman: >dialogResult [](start = 16, length = 12)

nit: I find if (dialogResult == true) a lot easier to read.

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
```diff
@@ -40,6 +39,25 @@
...
+        <Style x:Key="ConnectionTypeComboBoxStyle"
+               TargetType="{x:Type ComboBox}">
+            <Style.Triggers>
+                <DataTrigger Binding="{Binding SupportSSHConnections}"
+                             Value="true">
+                    <Setter Property="Grid.ColumnSpan"
+                            Value="1" />
+                    <Setter Property="IsEnabled"
+                            Value="True" />
+                </DataTrigger>
```
dustincoleman: nit: Do you need this? I think you could at least get rid of this data trigger and just have these be the default setters for the style

src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
```diff
@@ -160,9 +228,27 @@ private bool IsLibLinuxAvailable()
...
+            get
+            {
+                return _supportSSHConnections;
+            }
+            private set
+            {
+                _supportSSHConnections = value;
+                OnPropertyChanged(nameof(SupportSSHConnections));
+            }
+        }
```
dustincoleman: Does this ever change or can it just be a read only property that is set in the constructor?

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -118,10 +118,10 @@ public static SSHConnection GetSSHConnection(string name)
         /// <param name="hwnd">Parent hwnd or IntPtr.Zero</param>
         /// <param name="connectionString">[out] connection string obtained by the dialog</param>
```
WardenGnaw: nit: Update comments for `supportSSHConnections`

src/SSHDebugPS/ConnectionManager.cs
pieandcakes: won't fix

src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
pieandcakes: this still builds against 7.3. I'm going to leave it as it is for now.

Issue #948, opened 2020-01-02
Cannot connect to containers based on mcr.microsoft.com/dotnet/core/aspnet:3.0 with VS2019 remote debug - error 0x80004005


MichaelSimons:
_Copied from https://github.com/dotnet/dotnet-docker/issues/1530 by @ktos_

## Steps to reproduce the issue
1. Create a new container where base image is `mcr.microsoft.com/dotnet/core/aspnet:3.0` or `mcr.microsoft.com/dotnet/core/aspnet:3.1` (or probably older, haven't tested) on a remote Docker host,
2. Try to connect to process using remote debug in VS2019 16.4.10,
3. Error: `Operation not supported. Unknown error: 0x80004005.`

## Expected behavior
List of processes to be attached to appears.

## Actual behavior

An error similar to this:

```
---------------------------
Microsoft Visual Studio
---------------------------
Unable to connect to 'temperaturektosdev_temperatureservice3_1;ssh=ktos@[redacted]'.  Operation not supported. Unknown error: 0x80004005.
---------------------------
OK   
---------------------------
```

![](https://user-images.githubusercontent.com/1633261/70851877-41beaa00-1e9b-11ea-8666-d3f6be86a342.png)

## Additional information (e.g. issue happens only occasionally)

Described the whole problem in Microsoft/DockerTools#222, especially https://github.com/microsoft/DockerTools/issues/222#issuecomment-565748326 - everything is because the lack of `ps` program in the image. Installing it via `apt install procps` fixes the issue.

Problem appears only in **Buster** images, at least not in **Alpine** (there is `ps` installed, I confirmed).

Still not sure why it works flawlessly for my LOCAL Docker host (Windows) and not for my REMOTE Docker host (Ubuntu). Maybe there is some difference in getting processes info?

However, **I think that `ps` should be a part of a base image for ASP.NET Core web apps.**

## Output of `docker version`

```
Client:
 Version:           18.09.7
 API version:       1.39
 Go version:        go1.10.1
 Git commit:        2d0083d
 Built:             Fri Aug 16 14:20:06 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.09.7
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.1
  Git commit:       2d0083d
  Built:            Wed Aug 14 19:41:23 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

## Output of `docker info`

```
[...]

PR #956, opened 2020-01-10
Fix escaping with the procfs script

pieandcakes:
It turns out that we didn't need to escape the '$' but instead we needed
to escape the inner ''' character because it is inside an already quoted
string.

Fixes: #948

### DIFF
--- a/src/SSHDebugPS/PipeConnection.cs
+++ b/src/SSHDebugPS/PipeConnection.cs
@@ -201,8 +201,11 @@ namespace Microsoft.SSHDebugPS
 
             int exitCode;
             string commandOutput;
-            // If used accross SSH, assume host is running Linux and escape the command (specifically the '$')
-            if (!ExecuteCommand(this.OuterConnection == null ? ProcFSOutputParser.CommandText : ProcFSOutputParser.EscapedCommandText, Timeout.Infinite, true, out commandOutput, out errorMessage, out exitCode))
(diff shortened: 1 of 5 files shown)


src/SSHDebugPS/PipeConnection.cs
```diff
@@ -201,8 +201,7 @@ private bool ProcFSListProcess(string username, out string errorMessage, out Lis
 
             int exitCode;
             string commandOutput;
-            // If used accross SSH, assume host is running Linux and escape the command (specifically the '$')
-            if (!ExecuteCommand(this.OuterConnection == null ? ProcFSOutputParser.CommandText : ProcFSOutputParser.EscapedCommandText, Timeout.Infinite, true, out commandOutput, out errorMessage, out exitCode))
+            if (!ExecuteCommand(ProcFSOutputParser.CommandText, Timeout.Infinite, false, out commandOutput, out errorMessage, out exitCode))
```
pieandcakes: This should have been false all long. Fixed it.

src/SSHDebugPS/StringResources.resx
```diff
@@ -133,11 +133,6 @@
     <value>Microsoft-SSHPortSupplier-FileCopy-</value>
     <comment>Do Not localize</comment>
   </data>
-  <data name="DockerExecCommandFormat" xml:space="preserve">
```
pieandcakes: This is never used so i removed it.

PR #957, opened 2020-01-10
Revert "Check to find '-break-insert' versions. (#952)"

pieandcakes:
per @WardenGnaw this breaks existing functionality with lldb 3.8. Reverting until the correct fix can go in.

This reverts commit 119956f50a4fba624d5fb69e131e8441e8796c54.

### DIFF
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -418,7 +418,7 @@ namespace MICore
 
         #region Breakpoints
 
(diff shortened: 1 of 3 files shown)


WardenGnaw:
@puremourning Unfortunately, I was wrong about the check. This causes `-break-insert` to fail on the `lldb-mi` we ship. I've been mostly investigating with the new `lldb-mi` we need to ship to fix macOS Catalina and we will need this change when we have that finally building. 

I will investigate another way to do this check.

I will reopen https://github.com/microsoft/MIEngine/issues/946

PR #959, opened 2020-01-10
Enable Lab.* builds for PR

pieandcakes:


pieandcakes:
Looks like this won't work externally.

PR #960, opened 2020-01-10
Fix for vscode-cpptools/#2492

pieandcakes:
in lldb-mi there appears to be "message" instead of "msg" when an error is returned in the `-exec-*` command results. This is to handle both cases. 

Fixes https://github.com/microsoft/vscode-cpptools/issues/2492

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+﻿    // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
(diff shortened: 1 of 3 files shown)


pieandcakes:
The other option is to refactor TryFindString to return a boolean and take an out parameter, but thats a more invasive fix, but it would be the right thing to do. If we want to go that route instead, let me know and I can update the code instead for that solution.

@gregg-miskelly @WardenGnaw

src/MICore/Debugger.cs
```diff
@@ -1129,7 +1129,26 @@ internal void OnComplete(Results results, MICommandFactory commandFactory)
...
+                            {
+                                // TODO: Remove after update to mainline lldb-mi
+                                // lldb-mi has certain instances (such as the -exec-* commands) that calls the message
+                                // "message" instead of "msg"
+                                miError = results.FindString("message");
+                            }
+                            catch
+                            {
+                                // Rethrow the original exception if we can't find "message"
+                                throw ex;
```
gregg-miskelly: >throw ex; [](start = 32, length = 9)

I don't think you want to throw here -- seems like instead we should do something like `miError = "<Unknown Error>";` (where that string should be localizable.

PR #961, opened 2020-01-10
Use version number to detect previous lldb-mi

WardenGnaw:
Use the `version` command to detect which version of lldb-mi to use the
 `on` keyword for `-break-insert`.

 Currently shipped lldb-mi 3.5 will return `lldb-350.99.0`.

 Other version may return something else, for example currently building
 out of the master branch of lldb-mi returns
 `~"lldb-1100.0.30.11\nApple Swift version 5.1.3 (swiftlang-1100.0.282.1
 clang-100.033.15)\n"`

Addresses https://github.com/microsoft/MIEngine/issues/946

Thank you @puremourning for working on the original proposed fix.

### DIFF
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -418,7 +418,7 @@ namespace MICore
 
         #region Breakpoints
 
-        protected virtual Task<StringBuilder> BuildBreakInsert(string condition, bool enabled)
+        public virtual Task<StringBuilder> BuildBreakInsert(string condition, bool enabled)
         {
             StringBuilder cmd = new StringBuilder("-break-insert -f ");
             if (condition != null)
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -679,11 +679,8 @@ namespace MICore
             return Task<bool>.FromResult(false);
         }
 
-        public async Task<string> GetVersion()
-        {
-            return await _debugger.ConsoleCmdAsync("version");
-        }
 
+        public abstract Task<string> Version();
         #endregion
     }
 }
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -679,6 +679,11 @@ namespace MICore
             return Task<bool>.FromResult(false);
         }
 
+        public async Task<string> GetVersion()
+        {
+            return await _debugger.ConsoleCmdAsync("version");
+        }
+
         #endregion
     }
 }
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -673,14 +673,6 @@ namespace MICore
         {
             return false;
         }
-
-        public virtual Task<bool> RequiresOnKeywordForBreakInsert()
-        {
-            return Task<bool>.FromResult(false);
-        }
-
-
-        public abstract Task<string> Version();
         #endregion
     }
 }
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -418,7 +418,7 @@ namespace MICore
 
         #region Breakpoints
 
-        protected virtual StringBuilder BuildBreakInsert(string condition, bool enabled)
+        protected virtual Task<StringBuilder> BuildBreakInsert(string condition, bool enabled)
         {
             StringBuilder cmd = new StringBuilder("-break-insert -f ");
             if (condition != null)
@@ -431,7 +431,7 @@ namespace MICore
             {
                 cmd.Append("-d ");
             }
-            return cmd;
+            return Task<StringBuilder>.FromResult(cmd);
         }
 
         internal bool PreparePath(string path, bool useUnixFormat, out string pathMI)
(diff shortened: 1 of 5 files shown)


src/MICore/CommandFactories/lldb.cs
```diff
@@ -46,12 +46,15 @@ public override bool SupportsFrameFormatting
...
         {
-            // LLDB's use of the pending flag requires an optional parameter or else it fails.
+            // LLDB's 3.5 use of the pending flag requires an optional parameter or else it fails.
             // We will use "on" for now. 
-            // TODO: Fix this on LLDB-MI's side
-            string pendingFlag = "-f on ";
+            string pendingFlag = "-f ";
+            if (await RequiresOnKeywordForBreakInsert())
+            {
+                pendingFlag = "-f on ";
```
pieandcakes: Create a `const string pendingFlagParameter = "on";`  and after the `cmd.Append(pendingFlag);` check `RequiresOnKeywordForBreakInsert()` and if it comes back as true, then append `pendingFlagParameter`.

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -674,6 +674,16 @@ public virtual bool SupportsBreakpointChecksums()
...
         }
 
+        public virtual Task<bool> RequiresOnKeywordForBreakInsert()
+        {
+            return Task<bool>.FromResult(false);
+        }
+
+        public async Task<string> GetVersion()
+        {
+            return await _debugger.ConsoleCmdAsync("version");
```
pieandcakes: you should make this abstract and then no-op in non `lldb` cases. this will fail in `gdb` so if you want  to do it, then also implement the correct command in `gdb`.

```
(gdb)
interpreter-exec console version
&"interpreter-exec console version\n"
Undefined command: "version".  Try "help".
&"error in command: \"version\".\n"
^error,msg="error in command: \"version\"."
(gdb)
version
&"version\n"
&"Undefined command: \"version\".  Try \"help\".\n"
^error,msg="Undefined command: \"version\".  Try \"help\"."
(gdb)
```

src/MICore/CommandFactories/lldb.cs
```diff
@@ -192,5 +195,28 @@ public async override Task<string> VarAssign(string variableName, string express
...
+        // Later versions do not require the 'on' keyword.
+        public override async Task<bool> RequiresOnKeywordForBreakInsert()
+        {
+            if (!_requiresOnKeywordForBreakInsert.HasValue)
+            {
+                _requiresOnKeywordForBreakInsert = false;
+
+                // Query for the version.
+                string version = await GetVersion();
+                if (!string.IsNullOrEmpty(version) && version.Trim().Equals(OldLLDBMIVersionString))
```
pieandcakes: `StringComparison.Ordinal`

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -418,7 +418,7 @@ public virtual async Task Terminate()
 
         #region Breakpoints
 
-        protected virtual StringBuilder BuildBreakInsert(string condition, bool enabled)
+        protected virtual Task<StringBuilder> BuildBreakInsert(string condition, bool enabled)
```
pieandcakes: what if you call this from `GetBreakInsertMainCommand` so that there isn't duplicate `-break-insert` and checking for the `on` keyword?

src/MICore/CommandFactories/lldb.cs
```diff
@@ -192,5 +197,33 @@ public async override Task<string> VarAssign(string variableName, string express
...
+        // Later versions do not require the 'on' keyword.
+        public override async Task<bool> RequiresOnKeywordForBreakInsert()
+        {
+            if (!_requiresOnKeywordForBreakInsert.HasValue)
+            {
+                _requiresOnKeywordForBreakInsert = false;
+
+                // Query for the version.
+                string version = await Version();
+                if (!string.IsNullOrEmpty(version) && version.Trim().Equals(OldLLDBMIVersionString, StringComparison.Ordinal))
```
pieandcakes: probably want IsNullOrWhitespace in case spaces are returned, you don't need to do the comparison.

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -674,6 +674,13 @@ public virtual bool SupportsBreakpointChecksums()
             return false;
         }
 
+        public virtual Task<bool> RequiresOnKeywordForBreakInsert()
+        {
+            return Task<bool>.FromResult(false);
+        }
+
+
+        public abstract Task<string> Version();
```
gregg-miskelly: Is there a reason to put either of these in the base class? Seems like we only care about these within the LLDB command factory, so they should just be private methods there.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -834,7 +834,11 @@ private List<LaunchCommand> GetInitializeCommands()
                         return Task.FromResult(0);
                     };
 
-                    commands.Add(new LaunchCommand(GetBreakInsertMainCommand(), ignoreFailures: true, successResultsHandler: breakMainSuccessResultsHandler));
+                    // Builds '-break-insert' for 'main'.
+                    StringBuilder breakInsertCommand = await this.MICommandFactory.BuildBreakInsert(null, true);
```
gregg-miskelly: >null, true [](start = 100, length = 10)

nit: used named arguments

src/MICore/CommandFactories/lldb.cs
```diff
@@ -192,5 +197,33 @@ public async override Task<string> VarAssign(string variableName, string express
...
+        private const string OldLLDBMIVersionString = "lldb-350.99.0";
+
+        // In LLDB 3.5, -break-insert -f requires a string before the actual method name.
+        // We use a placeholder 'on' for this.
+        // Later versions do not require the 'on' keyword.
+        public override async Task<bool> RequiresOnKeywordForBreakInsert()
+        {
+            if (!_requiresOnKeywordForBreakInsert.HasValue)
+            {
+                _requiresOnKeywordForBreakInsert = false;
```
gregg-miskelly: >_requiresOnKeywordForBreakInsert = false; [](start = 16, length = 41)

If we wind up trying to set two breakpoints at the same time this is going to do the wrong thing. I think you should only assign once you know the answer.

PR #962, opened 2020-01-11
Enable Lab.* builds for PR

WardenGnaw:
Created TEST_LAB_BUILD variable that will skip Signing Targets

If building Lab.* without TEST_LAB_BUILD set then it will do the signing tasks.

### DIFF
--- a/build/miengine.settings.targets
+++ b/build/miengine.settings.targets
@@ -14,8 +14,8 @@
             <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)..\Keys\DevDivPublicKey.snk</AssemblyOriginatorKeyFile>
 
             <!-- Use PublicSign when TEST_LAB_BUILD is enabled -->
-            <PublicSign Condition="'$(TEST_LAB_BUILD)'==''">true</PublicSign>
-            <DelaySign Condition="'$(TEST_LAB_BUILD)'!=''">true</DelaySign>
+            <PublicSign Condition="'$(TEST_LAB_BUILD)'!=''">true</PublicSign>
+            <DelaySign Condition="'$(TEST_LAB_BUILD)'==''">true</DelaySign>
         </PropertyGroup>
     </When>
     <Otherwise>
--- a/build/miengine.settings.targets
+++ b/build/miengine.settings.targets
@@ -8,11 +8,14 @@
   <Import Project="version.settings.targets" />
 
   <Choose>
-    <When Condition="'$(Lab)' == 'true' and '$(TEST_LAB_BUILD)'==''">
+    <When Condition="'$(Lab)' == 'true'">
         <PropertyGroup>
             <!--NOTE:DevDivPublicKey.snk comes from src\\tools\\devdiv\\FinalPublicKey.snk -->
             <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)..\Keys\DevDivPublicKey.snk</AssemblyOriginatorKeyFile>
-            <DelaySign>true</DelaySign>
+
+            <!-- Use PublicSign when TEST_LAB_BUILD is enabled -->
+            <PublicSign Condition="'$(TEST_LAB_BUILD)'==''">true</PublicSign>
+            <DelaySign Condition="'$(TEST_LAB_BUILD)'!=''">true</DelaySign>
         </PropertyGroup>
     </When>
     <Otherwise>
--- a/build/miengine.settings.targets
+++ b/build/miengine.settings.targets
@@ -8,7 +8,7 @@
   <Import Project="version.settings.targets" />
 
   <Choose>
-    <When Condition="'$(Lab)' == 'true'">
+    <When Condition="'$(Lab)' == 'true' and '$(TEST_LAB_BUILD)'==''">
         <PropertyGroup>
             <!--NOTE:DevDivPublicKey.snk comes from src\\tools\\devdiv\\FinalPublicKey.snk -->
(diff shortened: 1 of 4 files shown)


build/miengine.settings.targets
```diff
@@ -8,7 +8,7 @@
   <Import Project="version.settings.targets" />
 
   <Choose>
-    <When Condition="'$(Lab)' == 'true'">
+    <When Condition="'$(Lab)' == 'true' and '$(TEST_LAB_BUILD)'==''">
```
gregg-miskelly: > and '$(TEST_LAB_BUILD)'=='' [](start = 39, length = 28)

Do you need this change? Feels like you might be missing out on some validation.

pieandcakes: i think this can be left as is. Its only the microbuild portion that needs to be handled for external lab builds.

WardenGnaw: Having this enabled leads to
```
##[error]EXEC(0,0): Error : Could not load file or assembly 'Microsoft.MICore, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. Strong name validation failed. (Exception from HRESULT: 0x8013141A)
EXEC : error : Could not load file or assembly 'Microsoft.MICore, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. Strong name validation failed. (Exception from HRESULT: 0x8013141A) [d:\a\1\s\src\MICore\MICore.csproj]
    - Strong name validation failed. (Exception from HRESULT: 0x8013141A)
```
I was not sure how to get around this without changing the key.

gregg-miskelly: Got it. Here is ideally what we should do: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/publicsign-compiler-option

build/miengine.settings.targets
```diff
@@ -12,7 +12,10 @@
         <PropertyGroup>
             <!--NOTE:DevDivPublicKey.snk comes from src\\tools\\devdiv\\FinalPublicKey.snk -->
             <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)..\Keys\DevDivPublicKey.snk</AssemblyOriginatorKeyFile>
-            <DelaySign>true</DelaySign>
+
+            <!-- Use PublicSign when TEST_LAB_BUILD is enabled -->
+            <PublicSign Condition="'$(TEST_LAB_BUILD)'==''">true</PublicSign>
```
pieandcakes: I think you have this backwards.

build/miengine.settings.targets
WardenGnaw: sgen does not like the public signed assembly

> "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\roslyn\csc.exe" LaunchOptions.xsd.types.designer.cs obj\Lab.Debug\GeneratedAssemblyInfo.cs /out:obj\Lab.Debug\sgen/Microsoft.MICore.dll /target:library /noconfig /keyfile:d:\a\1\s\build\..\Keys\DevDivPublicKey.snk /r:C:\windows\Microsoft.NET\Framework\v4.0.30319\/System.dll /r:C:\windows\Microsoft.NET\Framework\v4.0.30319\/System.Xml.dll" exited with code 1

I did a quick check if it would be happy with /delaysign, but that also failed. SGEN does not seem to have a /publickey flag. 

This is part of the GenerateXmlSerializersAssembly target.

gregg-miskelly: I am guessing you tried, but changing `<SerializationSigningCompilerOptions>` to have '/publicsign' and NOT have '/delaysign' didn't work?

Assuming that didn't work -- is it worth enabling this? I am wondering if we have made the 'Lab' configuration different enough from what will run in the lab that it might not be worth it. If we think so, I guess just go back with your original fix, but include a comment.

gregg-miskelly: Actually, I guess you would need two different versions of `SerializationSigningCompilerOptions` since it is both used to compile the input, and also used to pass to sgen.exe. I believe sgen.exe would use the version of csc.exe from the Windows directory, which wouldn't understand `/publicsign` so you would need `/delaysign` there. Where as it sounds like the build steps that runs the compiler would need `/publicsign`.

gregg-miskelly: In other words try adding `/publicsign` to the first of these commands, and `/delaysign` to the second. If that doesn't work -- I give up.

```
    <!-- Compile the types file into an assembly that we can use as input to sgen. We don't want to pass our real assembly as we have some things that sgen doesn't like. -->
    <Exec Command="&quot;$(CompilerPath)&quot; $(XsdCodeFile) $(GeneratedAssemblyInfoFile) /out:$(IntermediateOutputPath)sgen/$(AssemblyName).dll /target:library /noconfig $(SerializationSigningCompilerOptions) /r:$(RefPath)/System.dll /r:$(RefPath)/System.Xml.dll" />

    <!-- Now generate the serialization assembly -->
    <Exec Command="&quot;$(SDK40ToolsPath)sgen.exe&quot; $(IntermediateOutputPath)sgen/$(AssemblyName).dll /force /compiler:&quot;$(SerializationSigningCompilerOptions)&quot; /keep" />
```

WardenGnaw: @gregg-miskelly sgen does not have a `/publicsign` flag. I tried that locally and it failed. However it seems to work if we do not pass a signedfile flag to sgen.

PR #977, opened 2020-03-02
Allow consoleCommand to exec while process run

WardenGnaw:
Added flag allowWhileRunning to consoleAsyncCmd since certain commands
such as 'version' do not require the process to be stopped in order to execute.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -794,9 +794,9 @@ namespace MICore
             return outStr.ToString();
         }
 
-        public async Task<string> ConsoleCmdAsync(string cmd, bool ignoreFailures = false)
+        public async Task<string> ConsoleCmdAsync(string cmd, bool allowWhileRunning, bool ignoreFailures = false)
         {
-            if (this.ProcessState != ProcessState.Stopped && this.ProcessState != ProcessState.NotConnected)
+            if (!allowWhileRunning && this.ProcessState != ProcessState.Stopped && this.ProcessState != ProcessState.NotConnected)
             {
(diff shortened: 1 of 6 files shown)


src/MICore/Debugger.cs
```diff
@@ -794,9 +794,9 @@ private string Escape(string str)
             return outStr.ToString();
         }
 
-        public async Task<string> ConsoleCmdAsync(string cmd, bool ignoreFailures = false)
+        public async Task<string> ConsoleCmdAsync(string cmd, bool allowWhileRunning, bool ignoreFailures = false)
         {
-            if (this.ProcessState != ProcessState.Stopped && this.ProcessState != ProcessState.NotConnected)
+            if (!allowWhileRunning && this.ProcessState != ProcessState.Stopped && this.ProcessState != ProcessState.NotConnected)
```
pieandcakes: I think you still want NotConnected to fail.

src/MICore/Debugger.cs
```diff
@@ -811,7 +811,7 @@ public async Task<string> ConsoleCmdAsync(string cmd, bool ignoreFailures = fals
             using (ExclusiveLockToken lockToken = await _commandLock.AquireExclusive())
             {
                 // check again now that we have the lock
-                if (this.ProcessState != MICore.ProcessState.Stopped && this.ProcessState != ProcessState.NotConnected)
+                if (!allowWhileRunning && this.ProcessState != MICore.ProcessState.Stopped && this.ProcessState != ProcessState.NotConnected)
```
pieandcakes: Same here.

PR #982, opened 2020-03-24
Remove project.json files and clean up build

andrewcrawley:
This commit removes the "project.json" files that were still being used in some build configurations.  Where necessary, NuGet package references are moved to the relevant "packages.config" files.

This commit also gets rid of the confusingly-named `IsCoreClr` MSBuild variable and the `CORECLR` compile-time flag.  These are replaced with a new `IsXPlat` MSBuild variable and `XPLAT` compile-time flag.  Note that the meaning of this flag is inverted - code that shouldn't be included in the cross-plat build used for the VS Code C++ extension used to be excluded via `#if CORECLR`, while the new exclusion would be `#if !XPLAT`.  `IsXPlat` and `XPLAT` are both defined when building one of the `Desktop.*` build configurations and are undefined for `Debug`, `Release`, and `Lab.*`.


build/miengine.settings.targets
```diff
@@ -29,16 +29,14 @@
   <Choose>
     <When Condition="$(Configuration.Contains('Desktop'))">
       <PropertyGroup>
-        <IsCoreClr>false</IsCoreClr>
-        <NugetPackagesDirectory>$(NugetPackagesDirectory).Desktop</NugetPackagesDirectory>
-        <!-- Skip resolution from project.json -->
-        <ResolveNuGetPackages>false</ResolveNuGetPackages>
+		<!-- "Desktop.*" configs are used to build the cross-platform bits that are used in the VS Code C++ extension -->
+		<IsXPlat>true</IsXPlat>
```
pieandcakes: nit: extra indention?

src/DebugEngineHost.Stub/DebugEngineHost.Stub.csproj
```diff
@@ -27,9 +27,6 @@
     <DocumentationFile>$(DropRootDir)\ReferenceAssemblies\$(AssemblyName).xml</DocumentationFile>
     <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(IsCoreClr)' == 'true'">
-    <ProjectTypeGuids>$(ProjectTypeGuids);{786C830F-07A1-408B-BD7F-6EE04809D6DB};</ProjectTypeGuids>
```
pieandcakes: Note: this was PortableClassLibrary.

src/MIDebugPackage/MIDebugPackage.csproj
```diff
@@ -203,10 +201,6 @@
   <ItemGroup>
     <Content Include="Resources\Package.ico" />
   </ItemGroup>
-  <ItemGroup>
```
pieandcakes: I don't think you want to remove RequiredNetFX46FacadeAssemblies as these need to be in the VSIX.

pieandcakes: And I don't see you adding them back in. Are those making it in the VSIX thats used for Local F5 debugging (experimental instance)?

andrewcrawley: Those are no longer needed now that nothing is being built against them.  I verified that F5-ing the experimental instance still works (and that the VSCode extension still works on Linux).

PR #996, opened 2020-04-09
Modify ParseResultsList to throw instead of null

WardenGnaw:
Callers to ParseResultList expect the return value to be non-null.
However, if there is remaining characters in the span, we will return
null.

This change modifies this call to throw MIResultFormatException so it
will report an error instead of crashing.

Example error:
image

### DIFF
--- a/src/MICore/MIResults.cs
+++ b/src/MICore/MIResults.cs
@@ -730,12 +730,20 @@ namespace MICore
                     {
                         return i == s.Extent;
                     }, listStr, out rest);
+
+            Results results = new Results(resultClass, list);
+
             if (!rest.IsEmpty)
             {
                 ParseError("trailing chars", rest);
-                return null;
+
+                if (rest.Length > 1000)
+                {
+                    rest = new Span(rest.Start, 1000);    // don't show more than 1000 chars
+                }
+                throw new MIResultFormatException(rest.Extract(_resultString), results);
             }
-            return new Results(resultClass, list);
+            return results;
         }
 
         public string ParseCString(string input)
--- a/src/MICore/MIResults.cs
+++ b/src/MICore/MIResults.cs
@@ -596,6 +596,9 @@ namespace MICore
 
     public class MIResults
     {
+        // The amount of characters to send to the UI upon an error.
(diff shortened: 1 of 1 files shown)


src/MICore/MIResults.cs
```diff
@@ -730,12 +730,20 @@ private Results ParseResultList(Span listStr, ResultClass resultClass = ResultCl
                     {
                         return i == s.Extent;
                     }, listStr, out rest);
+
+            Results results = new Results(resultClass, list);
+
             if (!rest.IsEmpty)
```
pieandcakes, line 736: I would change this if statement so it isn't a negation since you are now doing more work here.

```
if (rest.IsEmpty)
{ 
    return results;
}

// your other stuff now
```

src/MICore/MIResults.cs
```diff
@@ -730,12 +730,20 @@ private Results ParseResultList(Span listStr, ResultClass resultClass = ResultCl
...
             {
                 ParseError("trailing chars", rest);
-                return null;
+
+                if (rest.Length > 1000)
+                {
+                    rest = new Span(rest.Start, 1000);    // don't show more than 1000 chars
```
pieandcakes, line 742: is there a reason for 1000 characters? should it be a const value you use in both places? is that too much regardless?

WardenGnaw: Was doing the same limit in ParseError https://github.com/microsoft/MIEngine/blob/95f12798e92a046b8a0a47b087319b6c569f4f7e/src/MICore/MIResults.cs#L1206

pieandcakes: can we make that a static so you we can change that if need to?

src/MICore/MIResults.cs
```diff
@@ -1203,9 +1209,9 @@ private ResultListValue ParseResultList(Span input, out Span rest)
 
         private void ParseError(string message, Span input)
         {
```
pieandcakes, line 1211: knowing this method exists, can it be reused for the above? It adds more logging in this case. 

I think getting rid of the `#if DEBUG` block might be good too as this error would be extremely useful in a user's bug case.

PR #1001, opened 2020-05-16
Support checking gdb.exe on Windows

WardenGnaw:
This PR adds additional checks to look for gdb on Windows.
If a user does not specify miDebuggerPath, we will check for gdb and
gdb.exe in PATH. If a miDebuggerPath is provided and does not supply any
directory, we will search that binary and binary.exe in PATH.

https://github.com/microsoft/vscode-cpptools/issues/3076

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -541,6 +541,8 @@ namespace MICore
         {
             string pathVar = System.Environment.GetEnvironmentVariable("PATH");
 
+            bool checkForWindowsExe = PlatformUtilities.IsWindows() && String.IsNullOrEmpty(Path.GetExtension(command));
+
             // Check each portion of the PATH environment variable to see if it contains the requested file
             foreach (string pathPart in pathVar.Split(Path.PathSeparator))
             {
@@ -552,7 +554,7 @@ namespace MICore
                     return candidate;
                 }
 
-                if (PlatformUtilities.IsWindows() && String.IsNullOrEmpty(Path.GetExtension(command)))
+                if (checkForWindowsExe)
                 {
                     string exeCandidate = candidate + ".exe";
                     if (File.Exists(exeCandidate))
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -551,6 +551,13 @@ namespace MICore
                 {
                     return candidate;
                 }
+
+                string exeCandidate = candidate + ".exe";
+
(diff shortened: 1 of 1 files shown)


src/MICore/LaunchOptions.cs
```diff
@@ -551,6 +551,13 @@ private static string ResolveFromPath(string command)
                 {
                     return candidate;
                 }
+
+                string exeCandidate = candidate + ".exe";
```
pieandcakes, line 555: I don't think this is what you want. I think you want 
```
if (String.IsNullOrEmpty(Path.GetExtension(command)) && PlatformUtilities.IsWindows()) 
{ 
    command.Append(".exe"); 
}
```

above if its windows and doesn't contain a file extension. This code checks what they put in AND whatever they enter + .exe

gregg-miskelly:
>                 if (File.Exists(candidate))

On Windows, does it make sense to look for `gdb` (no extension)? I didn't think you could have a no-extension executable.

---
Refers to: src/MICore/LaunchOptions.cs:550 in 6760a4c. [](commit_id = 6760a4cc2a25beb47f249d08056e8d452d113323, deletion_comment = False)

src/MICore/LaunchOptions.cs
```diff
@@ -551,6 +551,15 @@ private static string ResolveFromPath(string command)
                 {
                     return candidate;
                 }
+
+                if (PlatformUtilities.IsWindows() && String.IsNullOrEmpty(Path.GetExtension(command)))
```
gregg-miskelly, line 555: >String.IsNullOrEmpty(Path.GetExtension(command))) [](start = 53, length = 49)

nit: hoist this outside of the loop so that a new string doesn't need to be allocated for each `pathPart`.

```
bool addExtension = PlatformUtilities.IsWindows() && String.IsNullOrEmpty(Path.GetExtension(command));
foreach (...
```

src/MICore/LaunchOptions.cs
```diff
@@ -551,6 +553,15 @@ private static string ResolveFromPath(string command)
                 {
```
gregg-miskelly, line 553: I think you may have missed my previous question: do we actually want to check for `gdb` (no extension) on Windows? I would have thought that that wouldn't work.

WardenGnaw: I modified my `MinGW\bin\gdb.exe` to be `MinGW\bin\gdb` and it still works.

PR #1004, opened 2020-06-06
Add placeholder AdapterArgs for OpenDebugAD7 in VS

WardenGnaw:


src/MIDebugEngine/Microsoft.MIDebugEngine.pkgdef
```diff
@@ -61,6 +61,7 @@
 "Adapter"="$PackageFolder$\vscode\OpenDebugAD7.exe"
 "Language"="C++"
 "LanguageId"="{3A12D0B7-C26C-11D0-B442-00A0244A1DD2}"
+; "AdapterArgs"=""
```
pieandcakes, line 64: Can this be right after `Adapter`? Its not an engine metric, so having it at the top of that list doesn't make sense.

PR #1005, opened 2020-06-06
Replace 'miMode' to 'MIMode'

WardenGnaw:
The MIMode
Telemetry broke because we try to retrieve the string as "miMode" instead of "MIMode". Fixing it to be "MIMode" as state in our launch.json schema and MIEngineLaunchOptions also looks for "MIMode".

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -601,7 +601,7 @@ namespace OpenDebugAD7
             int hr;
             DateTime launchStartTime = DateTime.Now;
 
-            string mimode = responder.Arguments.ConfigurationProperties.GetValueAsString("miMode");
+            string mimode = responder.Arguments.ConfigurationProperties.GetValueAsString("MIMode");
             string program = responder.Arguments.ConfigurationProperties.GetValueAsString("program")?.Trim();
             if (string.IsNullOrEmpty(program))
             {
@@ -814,7 +814,7 @@ namespace OpenDebugAD7
             bool isLocal = string.IsNullOrEmpty(miDebuggerServerAddress) && !isPipeTransport;
             bool visualizerFileUsed = false;
             int sourceFileMappings = 0;
-            string mimode = responder.Arguments.ConfigurationProperties.GetValueAsString("miMode");
+            string mimode = responder.Arguments.ConfigurationProperties.GetValueAsString("MIMode");
 
             if (isLocal)
             {


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -601,8 +601,19 @@ protected override void HandleLaunchRequestAsync(IRequestResponder<LaunchArgumen
...
-            string mimode = responder.Arguments.ConfigurationProperties.GetValueAsString("miMode");
-            string program = responder.Arguments.ConfigurationProperties.GetValueAsString("program")?.Trim();
+            Dictionary<string, JToken> caseInsensitiveLaunchConfiguration;
+            try
+            {
+                caseInsensitiveLaunchConfiguration = new Dictionary<string, JToken>(responder.Arguments.ConfigurationProperties, StringComparer.OrdinalIgnoreCase);
+            } 
+            catch (ArgumentException e)
+            {
+                responder.SetError(CreateProtocolExceptionAndLogTelemetry(telemetryEventName, 1013, "Duplicate keys found in launch configuration."));
```
pieandcakes, line 611: The format is `<type>: <message>` so `launch: duplicate keys found in configuration`

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -805,16 +816,29 @@ protected override void HandleAttachRequestAsync(IRequestResponder<AttachArgumen
...
             const string telemetryEventName = DebuggerTelemetry.TelemetryAttachEventName;
 
+            Dictionary<string, JToken> caseInsensitiveAttachConfiguration;
+            try
+            {
+                caseInsensitiveAttachConfiguration = new Dictionary<string, JToken>(responder.Arguments.ConfigurationProperties, StringComparer.OrdinalIgnoreCase);
+            }
+            catch (ArgumentException e)
+            {
+                responder.SetError(CreateProtocolExceptionAndLogTelemetry(telemetryEventName, 1013, "Duplicate keys found in attach configuration."));
```
pieandcakes, line 826: Same message as above. the format is `attach: duplicate keys found in configuration`

WardenGnaw: True, I incorrectly read how MIEngine handled JSON parsing. It looks like it also explicitally looks for `MIMode`.
https://github.com/microsoft/MIEngine/blob/403592c998c736fa75fa7aa6cd1deca259d8837a/src/MICore/JsonLaunchOptions.cs#L54

pieandcakes:
Can you verify that it gets sent?

WardenGnaw:
Yes. The string variable we store MIMode in is not null anymore.

PR #1006, opened 2020-06-08
Support QuoteArgs in OpenDebugAD7

WardenGnaw:
This PR adds in the QuoteArgs toggle that we have in
MIEngineLaunchOptions. This will toggle if we quote the debugger command
or not.

**Examples:**
**Configuration:**
```
"pipeTransport": {
                "debuggerPath": "/usr/bin/gdb",
                "pipeProgram": "${env:windir}\\system32\\bash.exe",
                "pipeArgs": ["-c", "${debuggerCommand}"],
                "pipeCwd": "",
                "quoteArgs": true
},
```
**Output:**
`1: (8103) Starting: "C:\windows\sysnative\bash.exe" -c "/usr/bin/gdb --interpreter=mi"`
**Configuration:**
```
"pipeTransport": {
                "debuggerPath": "/usr/bin/gdb",
                "pipeProgram": "${env:windir}\\system32\\wsl.exe",
                "pipeArgs": ["${debuggerCommand}"],
                "pipeCwd": "",
                "quoteArgs": false
},
```
**Output:**
`
1: (6374) Starting: "C:\windows\sysnative\wsl.exe" /usr/bin/gdb --interpreter=mi
`

In this case, we will not be quoting anything at all in the arguments and it is up to the user to quote their arguments. E.g. pipeArgs: ["\"quoted argument with spaces\""]

### DIFF
--- a/src/OpenDebugAD7/MILaunchOptions.cs
+++ b/src/OpenDebugAD7/MILaunchOptions.cs
@@ -204,6 +204,9 @@ namespace OpenDebugAD7
 
             [JsonProperty]
             public Dictionary<string, string> PipeEnv { get; set; }
+
+            [JsonProperty("quoteArgs", DefaultValueHandling = DefaultValueHandling.Ignore)]
+            public bool? QuoteArgs { get; set; }
         }
 
         [JsonObject]
@@ -307,7 +310,7 @@ namespace OpenDebugAD7
             // ExeArguments
             // Build the exe's argument list as a string
             StringBuilder exeArguments = new StringBuilder();
-            exeArguments.Append(CreateArgumentList(exeArgsArray));
+            exeArguments.Append(CreateArgumentList(exeArgsArray, quoteArgs: true)); // Always quote exe arguments
             XmlSingleQuotedAttributeEncode(exeArguments);
             xmlLaunchOptions.Append(String.Concat("  ExeArguments='", exeArguments, "'\n"));
 
@@ -381,7 +384,7 @@ namespace OpenDebugAD7
             return arguments;
         }
 
-        private static string CreateArgumentList(IEnumerable<string> args)
+        private static string CreateArgumentList(IEnumerable<string> args, bool quoteArgs)
         {
             StringBuilder stringBuilder = new StringBuilder();
             if (args != null)
@@ -391,7 +394,7 @@ namespace OpenDebugAD7
                     if (stringBuilder.Length != 0)
                         stringBuilder.Append(' ');
 
-                    stringBuilder.Append(QuoteArgument(arg));
+                    stringBuilder.Append(quoteArgs ? QuoteArgument(arg) : arg);
                 }
             }
             return stringBuilder.ToString();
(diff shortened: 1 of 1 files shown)


src/OpenDebugAD7/MILaunchOptions.cs
```diff
@@ -391,7 +394,7 @@ private static string CreateArgumentList(IEnumerable<string> args)
                     if (stringBuilder.Length != 0)
                         stringBuilder.Append(' ');
 
-                    stringBuilder.Append(QuoteArgument(arg));
+                    stringBuilder.Append(quoteArgs ? QuoteArgument(arg) : arg);
```
pieandcakes, line 397: I think in the case of special characters you will still want to quote the arguments regardless.

Which of these scenarios are you trying to drive:
1. All Arguments are quoted
2. Quote only args that contain special characters
3. No arguments are quoted

I think your change will allow 3 and 2, but not 1. Quote Arguments will quote the args if they are empty or contain one of the characters its looking for (https://github.com/microsoft/MIEngine/blob/4849c9416da13242db687ceaf68b857d4c83e2fd/src/OpenDebugAD7/MILaunchOptions.cs#L405) . 

It might be useful to clarify the scenario you are enabling in the notes section of the PR.

pieandcakes: Also this is the wrong place. I think this changes the option for the user's executable args and not what you are trying to change.

WardenGnaw: This is the current description in MIEngine:

> If individual arguments contain characters (such as spaces or tabs), should it be quoted? If 'false', the debugger command will no longer be automatically quoted. \nDefault is 'true'.

I was copying what we already do in MIEngine.
https://github.com/microsoft/MIEngine/blob/9743dd79f858a47f0cecc5eccd52bae55624b8de/src/MICore/LaunchOptions.cs#L2115

I wonder if the AnyCode scenario is not doing the parenthesis arguments correctly.

pieandcakes: This is putting it into XML that will get unpacked in MIEngine.

WardenGnaw: It is only in PipeTransportOptions.
https://github.com/microsoft/MIEngine/blob/29c6569861b2042c43f04d8d78b2566db2e2ef1f/src/MIDebugPackage/OpenFolderSchema.json#L219

gregg-miskelly: > I think in the case of special characters you will still want to quote the arguments regardless.

I think the purpose of `"qutoeArgs":false` is to say "please don't mess with my args, I have this". So I think if qutoeArgs is false, we shouldn't do anything.

src/OpenDebugAD7/MILaunchOptions.cs
```diff
@@ -384,7 +384,13 @@ private static string[] TryAddWindowsDebuggeeConsoleRedirection(string[] argumen
             return arguments;
         }
 
-        private static string CreateArgumentList(IEnumerable<string> args, bool quoteArgs)
+        /// <summary>
+        /// Converts a list of strings arguments to a string representation.
+        /// </summary>
+        /// <param name="args">The list of arguments to convert into a string.</param>
+        /// <param name="quoteArgs">Only used for PipeTransports. This disables quote handling if the user requests it.</param>
```
pieandcakes, line 391: `Setting this to false disables quote handling....`

PR #1008, opened 2020-06-10
Support ClipboardContext for Evaluate Request

WardenGnaw:
Users who want to copy the full string from the UI will use "copy value"
from the EE windows. This will send an EvaluateRequest with the context
of Clipboard.

This PR noticies the context in OpenDebugAD7 and sets the
HostEvaluateRequestContext's isClipboardContext for MIEngine to query.
If it is true, MIEngine will save the previous value of 'show print
elements', set it to unlimited, and reset it back to the original value
after it evaluates the variable. Then it will provide the full string
back to the user to copy.

**NOTE**: Requires VS Code 1.46.0 to work. Previous engines set the `expression` to be the _value_ of the original expression instead of the _actual expression_.
See https://github.com/microsoft/vscode/issues/97444#issuecomment-630049720

### DIFF
--- a/src/DebugEngineHost/HostEvaluateRequestContext.cs
+++ b/src/DebugEngineHost/HostEvaluateRequestContext.cs
@@ -1,0 +1,32 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.DebugEngineHost
+{
+    /// <summary>
+    /// Provides acceess for MIEngine to determine if it is in an EvaluateRequest with specific contexes that do not
+    /// exist in enum_PARSEFLAGS.
+    /// </summary>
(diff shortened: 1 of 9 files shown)


src/DebugEngineHost/HostEvaluateRequestContext.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// </summary>
+    public sealed class HostEvaluateRequestContext
+    {
+        /// <summary>
+        /// Returns the boolean if the current EvaluareRequest's context is a clipboard context.
+        /// </summary>
+        /// <returns>true if it is in an evaluate request with a clipboard context</returns>
+        public static bool IsClipboardContext()
+        {
+            throw new NotImplementedException();
```
WardenGnaw, line 20: Don't want VS to throw an exception during an evaluate request, should return `false`.

WardenGnaw:
Closing to rework using `IDebugExpressionDAP`

PR #1009, opened 2020-06-11
Support ClipboardContext for Evaluate Request

WardenGnaw:
Users who want to copy the full string from the UI will use "copy value"
    from the EE windows. This will send an EvaluateRequest with the context
    of Clipboard.

    This PR noticies the context in OpenDebugAD7 and uses IDebugExpressionDAP
    with DAPEvalFlag ClipboardContext set for MIEngine to query.
    If it is set, MIEngine will save the previous value of 'show print
    elements', set it to unlimited, and reset it back to the original value
    after it evaluates the variable. Then it will provide the full string
    back to the user to copy.

NOTE: Requires VS Code 1.46.0 to work. Previous engines set the expression to be the value of the original expression instead of the actual expression.

### DIFF
--- a/src/DebugEngineHost.VSCode/VSCode/EngineConfiguration.cs
+++ b/src/DebugEngineHost.VSCode/VSCode/EngineConfiguration.cs
@@ -32,6 +32,7 @@ namespace Microsoft.DebugEngineHost.VSCode
         private readonly ExceptionSettings _exceptionSettings = new ExceptionSettings();
         private bool _conditionalBP;
         private bool _functionBP;
+        private bool _clipboardContext;
 
         // NOTE: CoreCLR doesn't support providing a code base when loading assemblies. So all debug engines
         // must be placed in the directory of OpenDebugAD7.exe
@@ -66,6 +67,12 @@ namespace Microsoft.DebugEngineHost.VSCode
             get { return _functionBP; }
             set { SetProperty(out _functionBP, value); }
         }
+
+        public bool ClipboardContext { 
+            get { return _clipboardContext; } 
+            set { SetProperty(out _clipboardContext, value); }
+        }
+
 
         /// <summary>
         /// Provides the directory of the debug adapter. This is the directory where
--- a/src/DebugEngineHost.VSCode/VSCode/EngineConfiguration.cs
+++ b/src/DebugEngineHost.VSCode/VSCode/EngineConfiguration.cs
@@ -68,7 +68,8 @@ namespace Microsoft.DebugEngineHost.VSCode
             set { SetProperty(out _functionBP, value); }
         }
 
-        public bool ClipboardContext { 
+        public bool ClipboardContext
+        { 
             get { return _clipboardContext; } 
             set { SetProperty(out _clipboardContext, value); }
         }
--- a/src/MIDebugEngine/AD7.Impl/AD7Expression.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Expression.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Text;
 using Microsoft.VisualStudio.Debugger.Interop;
 using System.Threading.Tasks;
+using Microsoft.VisualStudio.Debugger.Interop.DAP;
 
 namespace Microsoft.MIDebugEngine
 {
@@ -13,7 +14,7 @@ namespace Microsoft.MIDebugEngine
     // It is returned as a result of a successful call to IDebugExpressionContext2.ParseText
     // It allows the debugger to obtain the values of an expression in the debuggee. 
     // For the purposes of this sample, this means obtaining the values of locals and parameters from a stack frame.
-    public class AD7Expression : IDebugExpression2
+    public class AD7Expression : IDebugExpression2, IDebugExpressionDAP
     {
         private AD7Engine _engine;
         private IVariableInformation _var;
@@ -88,5 +89,33 @@ namespace Microsoft.MIDebugEngine
         }
 
         #endregion
+
+        #region IDebugExpressionDAP
+
+        int IDebugExpressionDAP.EvaluateSync(enum_EVALFLAGS dwFlags, DAPEvalFlags dapFlags, uint dwTimeout, IDebugEventCallback2 pExprCallback, out IDebugProperty2 ppResult)
+        {
+            ppResult = null;
+            if ((dwFlags & enum_EVALFLAGS.EVAL_NOSIDEEFFECTS) != 0 && _var.IsVisualized)
+            {
+                IVariableInformation variable = _engine.DebuggedProcess.Natvis.Cache.Lookup(_var);
+                if (variable == null)
+                {
+                    ppResult = new AD7ErrorProperty(_var.Name, ResourceStrings.NoSideEffectsVisualizerMessage);
+                }
+                else
+                {
+                    _var = variable;
+                    ppResult = new AD7Property(_engine, _var);
+                }
+                return Constants.S_OK;
+            }
+
+            _var.SyncEval(dapFlags, dwFlags);
+            ppResult = new AD7Property(_engine, _var);
+            return Constants.S_OK;
+        }
+
+        #endregion
+
     }
 }
--- a/src/MIDebugEngine/AD7.Impl/AD7Expression.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Expression.cs
@@ -67,7 +67,7 @@ namespace Microsoft.MIDebugEngine
         // This method evaluates the expression synchronously.
         int IDebugExpression2.EvaluateSync(enum_EVALFLAGS dwFlags, uint dwTimeout, IDebugEventCallback2 pExprCallback, out IDebugProperty2 ppResult)
         {
-            return EvaluateSync(dwFlags, DAPEvalFlags.NONE, dwTimeout, pExprCallback, out ppResult);
+            return EvaluateSyncInternal(dwFlags, DAPEvalFlags.NONE, dwTimeout, pExprCallback, out ppResult);
         }
 
         #endregion
@@ -76,12 +76,12 @@ namespace Microsoft.MIDebugEngine
 
         int IDebugExpressionDAP.EvaluateSync(enum_EVALFLAGS dwFlags, DAPEvalFlags dapFlags, uint dwTimeout, IDebugEventCallback2 pExprCallback, out IDebugProperty2 ppResult)
         {
-            return EvaluateSync(dwFlags, dapFlags, dwTimeout, pExprCallback, out ppResult);
+            return EvaluateSyncInternal(dwFlags, dapFlags, dwTimeout, pExprCallback, out ppResult);
         }
 
         #endregion
 
-        private int EvaluateSync(enum_EVALFLAGS dwFlags, DAPEvalFlags dapFlags, uint dwTimeout, IDebugEventCallback2 pExprCallback, out IDebugProperty2 ppResult)
+        private int EvaluateSyncInternal(enum_EVALFLAGS dwFlags, DAPEvalFlags dapFlags, uint dwTimeout, IDebugEventCallback2 pExprCallback, out IDebugProperty2 ppResult)
         {
             ppResult = null;
             if ((dwFlags & enum_EVALFLAGS.EVAL_NOSIDEEFFECTS) != 0 && _var.IsVisualized)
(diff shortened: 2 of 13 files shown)


src/MIDebugEngine/MIDebugEngine.csproj
```diff
@@ -169,6 +169,11 @@
       <Project>{54c33afa-438d-4932-a2f0-d0f2bb2fadc9}</Project>
       <Name>MICore</Name>
     </ProjectReference>
+    <ProjectReference Include="..\Microsoft.VisualStudio.Debugger.Interop.DAP\Microsoft.VisualStudio.Debugger.Interop.DAP.csproj">
+      <Project>{23DE943C-1F84-4E23-9490-50FE1DEDE858}</Project>
+      <Name>Microsoft.VisualStudio.Debugger.Interop.DAP</Name>
+	  <EmbedInteropTypes>True</EmbedInteropTypes>
```
gregg-miskelly, line 175: >	   [](start = 0, length = 3)

nit: you have a tab here.

gregg-miskelly:
>             return Constants.S_OK;

Call the new EvaluateSync passing `DAPEvalFlags.NONE` (or an internal helper method they both call)

---
Refers to: src/MIDebugEngine/AD7.Impl/AD7Expression.cs:88 in 070c3a9. [](commit_id = 070c3a9e092aebe714964e721d236d987456d27b, deletion_comment = False)

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -469,6 +476,20 @@ internal async Task Eval(enum_EVALFLAGS dwFlags = 0)
...
                 else
                 {
+                    int numElements = 200;
+                    if (dwDAPFlags.HasFlag(DAPEvalFlags.CLIPBOARD_CONTEXT))
+                    {
+                        string showPrintElementsResult = await MIDebugCommandDispatcher.ExecuteCommand("show print elements", _debuggedProcess, ignoreFailures: true);
+                        // Possible values for 'numElementsStr'
+                        // "Limit on string chars or array elements to print is <number>."
+                        // "Limit on string chars or array elements to print is unlimited."
+                        string numElementsStr = Regex.Match(showPrintElementsResult, @"\d+").Value;
```
gregg-miskelly, line 486: >. [](start = 92, length = 1)

Don't ye need to check if the match was successful before you call '.Value'? Also, should you verify there is just one match?

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -469,6 +476,20 @@ internal async Task Eval(enum_EVALFLAGS dwFlags = 0)
...
+                        if (int.TryParse(numElementsStr, out numElements))
```
gregg-miskelly, line 487: >) [](start = 73, length = 1)

`&& numElements != 0`

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -33,6 +34,7 @@ internal interface IVariableInformation : IDisposable
         void AsyncEval(IDebugEventCallback2 pExprCallback);
         void AsyncError(IDebugEventCallback2 pExprCallback, IDebugProperty2 error);
         void SyncEval(enum_EVALFLAGS dwFlags = 0);
```
gregg-miskelly, line 36: >void SyncEval(enum_EVALFLAGS dwFlags = 0); [](start = 8, length = 42)

Should we get rid of this method? Or is something calling it?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1862,7 +1864,10 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
             hr = frame.GetExpressionContext(out expressionContext);
             eb.CheckHR(hr);
 
+            System.Diagnostics.Debugger.Launch();
```
gregg-miskelly, line 1867: >System.Diagnostics.Debugger.Launch(); [](start = 12, length = 37)

Just so you don't forget: remove this before you checkin

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1887,8 +1892,20 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
...
             }
 
+            if (context == EvaluateArguments.ContextValue.Clipboard)
+            {
+                dapEvalFlags |= DAPEvalFlags.CLIPBOARD_CONTEXT;
+            }
+
             IDebugProperty2 property;
-            hr = expressionObject.EvaluateSync(flags, Constants.EvaluationTimeout, null, out property);
+            if (dapEvalFlags != DAPEvalFlags.NONE)
```
gregg-miskelly, line 1901: > if (dapEvalFlags != DAPEvalFlags.NONE) [](start = 11, length = 39)

This check doesn't seem to make sense -- either you should do: `if (expressionObject is IDebugExpressionDAP expressionDapObject)` or you should just always call it. I would probably argue for the `if` check since someday we may use OpenDebugAD7 for hosting other engines.

src/OpenDebugAD7/OpenDebugAD7.csproj
```diff
@@ -161,6 +161,11 @@
       <Project>{81de2423-fb5e-4069-b3c5-4c13ce76dc0a}</Project>
       <Name>DebugEngineHost.VSCode</Name>
     </ProjectReference>
+    <ProjectReference Include="..\Microsoft.VisualStudio.Debugger.Interop.DAP\Microsoft.VisualStudio.Debugger.Interop.DAP.csproj">
+      <Project>{23DE943C-1F84-4E23-9490-50FE1DEDE858}</Project>
+      <Name>Microsoft.VisualStudio.Debugger.Interop.DAP</Name>
+	  <EmbedInteropTypes>True</EmbedInteropTypes>
```
gregg-miskelly, line 167: >	 [](start = 0, length = 1)

nit: tab

gregg-miskelly:
Otherwise LGTM.

src/DebugEngineHost.VSCode/VSCode/EngineConfiguration.cs
```diff
@@ -67,6 +68,12 @@ public bool FunctionBP
             set { SetProperty(out _functionBP, value); }
         }
 
+        public bool ClipboardContext {
```
pieandcakes, line 71: nit: Open `{` needs to be on a newline

src/MIDebugEngine/AD7.Impl/AD7Expression.cs
```diff
@@ -88,5 +89,33 @@ int IDebugExpression2.EvaluateSync(enum_EVALFLAGS dwFlags, uint dwTimeout, IDebu
...
+                }
+                else
+                {
+                    _var = variable;
+                    ppResult = new AD7Property(_engine, _var);
+                }
+                return Constants.S_OK;
+            }
+
+            _var.SyncEval(dapFlags, dwFlags);
```
pieandcakes, line 113: Could you have the top method call this method with dapFlags.None and then fix the order of your flags to have SyncEval use dwFlags and then dapFlags? i think the switching of order of which the flags are in different methods is going to cause confusion.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -400,10 +402,15 @@ public void AsyncError(IDebugEventCallback2 pExprCallback, IDebugProperty2 error
...
+        {
+            SyncEval(DAPEvalFlags.NONE, dwFlags);
+        }
+
+        public void SyncEval(DAPEvalFlags dwDAPFlags, enum_EVALFLAGS dwFlags = 0)
         {
             Task eval = Task.Run(async () =>
             {
-                await Eval(dwFlags);
+                await Eval(dwFlags, dwDAPFlags);
```
pieandcakes, line 413: per my above comment: this is where it gets confusing. Eval takes dwFlags, DAP flags but SyncEval takes dwDAPFlags, dwflasg. I would change this method to support dapFlags.NONE as a default and keep consistent ordering.

src/MIDebugEngine/Engine.Impl/Variables.cs
pieandcakes: If you can't parse it, then what is the behavior that happens?

src/MIDebugEngine/Engine.Impl/Variables.cs
WardenGnaw: There is one capture context, so it should have only one Capture's value.

src/Microsoft.VisualStudio.Debugger.Interop.DAP/Microsoft.VisualStudio.Debugger.Interop.DAP.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.VisualStudio.Debugger.Interop.DAP
+{
+    /// <summary>
+    /// An extended set of evaluate flags that are in Debug Adapter Protocol but not listed in AD7
+    /// </summary>
+    public enum DAPEvalFlags
```
gregg-miskelly, line 9: `[Flags]`

src/MIDebugEngine/Engine.Impl/Variables.cs
WardenGnaw: Then TryParse will fail and we will just default to the current print element value.

src/MIDebugEngine/AD7.Impl/AD7Expression.cs
```diff
@@ -65,6 +66,22 @@ int IDebugExpression2.EvaluateAsync(enum_EVALFLAGS dwFlags, IDebugEventCallback2
...
+        #region IDebugExpressionDAP
+
+        int IDebugExpressionDAP.EvaluateSync(enum_EVALFLAGS dwFlags, DAPEvalFlags dapFlags, uint dwTimeout, IDebugEventCallback2 pExprCallback, out IDebugProperty2 ppResult)
+        {
+            return EvaluateSync(dwFlags, dapFlags, dwTimeout, pExprCallback, out ppResult);
+        }
+
+        #endregion
+
+        private int EvaluateSync(enum_EVALFLAGS dwFlags, DAPEvalFlags dapFlags, uint dwTimeout, IDebugEventCallback2 pExprCallback, out IDebugProperty2 ppResult)
```
pieandcakes, line 84: The pattern with the internal implementation is to add Internal to the name like `EvaluateSyncInternal`

src/Microsoft.VisualStudio.Debugger.Interop.DAP/Microsoft.VisualStudio.Debugger.Interop.DAP.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// </summary>
+        NONE = 0,
+        /// <summary>
+        /// Evaluation is for a clipboard context
+        /// </summary>
+        CLIPBOARD_CONTEXT = 1
+    }
+
+    /// <summary>
+    /// IDebugExpresssion for Debug Adapter Protocol
```
pieandcakes, line 23: sp: `Expression`

src/Microsoft.VisualStudio.Debugger.Interop.DAP/Microsoft.VisualStudio.Debugger.Interop.DAP.cs
```diff
@@ -0,0 +1,43 @@
...
+    [ComVisible(true)]
+    [Guid("76C710C4-CE66-4422-AC7C-1E41B3FC0BE3")]
+    [InterfaceType(1)]
+    public interface IDebugExpressionDAP
+    {
+        /// <summary>
+        /// Evaluate for IDebugExpressionDAP
+        /// </summary>
+        /// <param name="dwFlags"></param>
+        /// <param name="dapFlags"></param>
```
pieandcakes, line 35: Theres no point in having these `param` comments here if they are empty. Either add some text or remove them. Also explain why we want this in the summary.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -535,6 +552,11 @@ internal async Task Eval(enum_EVALFLAGS dwFlags = 0)
                     {
                         Debug.Fail("Weird msg from -var-create");
```
pieandcakes, line 553: nit: Can you change this to say `"Unexpected format of msg from -var-create"` ?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1887,8 +1890,21 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
...
             }
 
+            if (context == EvaluateArguments.ContextValue.Clipboard)
+            {
+                dapEvalFlags |= DAPEvalFlags.CLIPBOARD_CONTEXT;
+            }
+
             IDebugProperty2 property;
-            hr = expressionObject.EvaluateSync(flags, Constants.EvaluationTimeout, null, out property);
+            if (expressionObject is IDebugExpressionDAP expressionDapObject)
```
pieandcakes, line 1899: Can you just check the context value in this if block?

PR #1010, opened 2020-06-12
Fixing Conditional Breakpoints for VS Code

WardenGnaw:
VS Code has modified the way it handles conditional breakpoints.
It used to send a SetBreakpoints that removed the breakpoint and a new
SetBreakpoints request with the updated condition.

This change now keeps track of breakpoint conditions and recreate them
if the condition has changed.

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -1594,12 +1594,11 @@ namespace OpenDebugAD7
                         {
                             // already created
                             IDebugBreakpointRequest2 breakpointRequest;
-                            if (dict[bp.Line].GetBreakpointRequest(out breakpointRequest) == 0)
+                            if (dict[bp.Line].GetBreakpointRequest(out breakpointRequest) == 0 && 
+                                breakpointRequest is AD7BreakPointRequest ad7BPRequest)
                             {
-                                var ad7BPRequest = (AD7BreakPointRequest)breakpointRequest;
-
                                 // Check to see if this breakpoint has a condition that has changed.
-                                if (ad7BPRequest.Condition != bp.Condition)
+                                if (!StringComparer.InvariantCulture.Equals(ad7BPRequest.Condition, bp.Condition))
                                 {
                                     // Condition has been modified. Delete breakpoint so it will be recreated with the updated condition.
                                     var toRemove = dict[bp.Line];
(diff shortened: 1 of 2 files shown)


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -40,6 +40,7 @@ internal class AD7DebugSession : DebugAdapterBase, IDebugPortNotify2, IDebugEven
 
         private readonly DebugEventLogger m_logger;
         private readonly Dictionary<string, Dictionary<int, IDebugPendingBreakpoint2>> m_breakpoints;
```
gregg-miskelly, line 42: >IDebugPendingBreakpoint2 [](start = 60, length = 24)

Instead of making a second map, why not just use IDebugPendingBreakpoint2.GetBreakpointRequest to check the condition?

gregg-miskelly:
>                     IDebugBreakpointRequest2 breakpointRequest;

Do function breakpoints have the problem also?

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:1808 in 1741bfa. [](commit_id = 1741bfa9a04c94c9435321c6ae9622077025cb43, deletion_comment = False)

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1590,6 +1590,45 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
                     var resBreakpoints = new List<Breakpoint>();
                     foreach (var bp in breakpoints)
                     {
+                        if (dict.ContainsKey(bp.Line))
```
pieandcakes, line 1593: would dict.TryGetValue(key, out value) be a better way to do this in case this becomes async?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1590,6 +1590,45 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                        {
+                            // already created
+                            IDebugBreakpointRequest2 breakpointRequest;
+                            if (dict[bp.Line].GetBreakpointRequest(out breakpointRequest) == 0)
+                            {
+                                var ad7BPRequest = (AD7BreakPointRequest)breakpointRequest;
+
+                                // Check to see if this breakpoint has a condition that has changed.
+                                if (ad7BPRequest.Condition != bp.Condition)
```
pieandcakes, line 1602: Will this need a StringComparer?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1590,6 +1590,45 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                                        });
+                                    }
+                                    continue;
+                                }
+                            }
+                        }
+
+
+                        // Create a new breakpoint
                         if (!dict.ContainsKey(bp.Line))
```
pieandcakes, line 1632: change this to be an else to the above `if` statement. I don't see you modifying the dictionary above.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1590,6 +1590,45 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
                     var resBreakpoints = new List<Breakpoint>();
                     foreach (var bp in breakpoints)
                     {
+                        if (dict.ContainsKey(bp.Line))
+                        {
+                            // already created
+                            IDebugBreakpointRequest2 breakpointRequest;
+                            if (dict[bp.Line].GetBreakpointRequest(out breakpointRequest) == 0)
+                            {
+                                var ad7BPRequest = (AD7BreakPointRequest)breakpointRequest;
```
pieandcakes, line 1599: In case we end up supporting other types of breakpoints, it might be better to add in the if statement `&& breakpointRequest is AD7BreakPointRequest ad7BPRquest)`

pieandcakes: I realize you copied this from where it was below but i think it might be worth fixing now with the new language changes.

gregg-miskelly:
Do you need to do the same for function breakpoints?

WardenGnaw:
> Do you need to do the same for function breakpoints?

Is there a way to set function breakpoints with conditions in VS Code?

gregg-miskelly:
> > Do you need to do the same for function breakpoints?
> 
> Is there a way to set function breakpoints with conditions in VS Code?

Oh, I guess you are right. I didn't notice. Though it is possible to do in VS and we now use this code in VS, so maybe worth fixing anyway?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1590,6 +1590,44 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
                     {
+                        if (dict.ContainsKey(bp.Line))
+                        {
+                            // already created
+                            IDebugBreakpointRequest2 breakpointRequest;
+                            if (dict[bp.Line].GetBreakpointRequest(out breakpointRequest) == 0 && 
+                                breakpointRequest is AD7BreakPointRequest ad7BPRequest)
+                            {
+                                // Check to see if this breakpoint has a condition that has changed.
+                                if (!StringComparer.InvariantCulture.Equals(ad7BPRequest.Condition, bp.Condition))
```
gregg-miskelly, line 1601: nit: Is there a reason you are using `InvariantCulture` instead of an ordinal string compare?

PR #1011, opened 2020-06-12
Log ClientId in Telemetry

WardenGnaw:
ClientId will be useful to know when repoducing crashes on specific clients.

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -32,6 +32,8 @@ namespace OpenDebugAD7
         // This is a general purpose lock. Don't hold it across long operations.
         private readonly object m_lock = new object();
 
+        private string m_clientId;
+
         private IDebugProcess2 m_process;
         private string m_processName;
         private IDebugEngineLaunch2 m_engineLaunch;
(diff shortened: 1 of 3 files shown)


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1890,14 +1894,14 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
...
-            {
-                dapEvalFlags |= DAPEvalFlags.CLIPBOARD_CONTEXT;
-            }
-
             IDebugProperty2 property;
             if (expressionObject is IDebugExpressionDAP expressionDapObject)
             {
+                if (context == EvaluateArguments.ContextValue.Clipboard)
+                {
+                    dapEvalFlags |= DAPEvalFlags.CLIPBOARD_CONTEXT;
```
pieandcakes, line 1902: Scope dapEvalFlags to the context of the `if` statement here. its not necessary to have it scoped outside of this as it isn't used except for the call below.

pieandcakes:
Two things: 
* what are the possible values for the AdapterID?
* don't forget that you will need to classify it before it is available.

WardenGnaw:
> what are the possible values for the ClientID?
Looking to capture vscode and visualstudio.

andrewcrawley:
> 
> 
> > what are the possible values for the ClientID?
> 
> Looking to capture vscode and visualstudio.

Might also be worth capturing "liveshare-server-host".

gregg-miskelly:
We currently already know what IDE MIEngine is running in because each IDE logs telemetry to a different Kusto table. Is there a reason we need this?

WardenGnaw:
I was looking at the telemetry and it seems like we wont log which IDE is being used if it causes a crash in the OpenDebugAD7 side?

Also there may be clientId specific things to do in AD7DebugSession so I wanted to have access to that variable.

gregg-miskelly:
But each IDE is ultimately responsible for logging all telemetry. So we never need to log which IDE we are talking to -- its implied.

PR #1013, opened 2020-06-19
Add support for LogPoints in OpenDebugAD7

WardenGnaw:
This PR includes a TracepointManager for OpenDebugAD7 which will
keep track of LogMessages included in BreakpointRequests.

LogMessages can include expression inside of { } for the
engine to evaluate.

TODO: $PID since all we have is an AD_PROCESS_ID.


src/OpenDebugAD7/TracepointManager.cs
```diff
@@ -0,0 +1,229 @@
...
+        private readonly enum_DEBUGPROP_INFO_FLAGS propertyInfoFlags;
+        private readonly enum_EVALFLAGS flags;
+
+        internal TracepointManager()
+        {
+            m_breakpointLogMessages = new Dictionary<uint, string>();
+
+            // Matches strings that are in { } 
+            // or upper case strings that begin with $.
+            regex = new Regex("\\$[A-Z]+|{.*?}", RegexOptions.Compiled);
```
gregg-miskelly, line 27: >regex = new Regex("\\$[A-Z]+|{.*?}", RegexOptions.Compiled); [](start = 12, length = 60)

I am skeptical that you want to use Regex. Here are some test cases to try:

`{wcscmp(L"{hi}", argv[0])}`
`\{Escape Test\}`

src/OpenDebugAD7/TracepointManager.cs
```diff
@@ -0,0 +1,229 @@
...
+                (enum_EVALFLAGS)enum_EVALFLAGS110.EVAL110_FORCE_REAL_FUNCEVAL |
+                enum_EVALFLAGS.EVAL_NOSIDEEFFECTS;
+
+        }
+
+        internal void Add(uint breakpointId, string logMessage)
+        {
+            lock (_lock)
+            {
+                m_breakpointLogMessages.Add(breakpointId, logMessage);
```
gregg-miskelly, line 49: >logMessage [](start = 58, length = 10)

Do you want to parse and validate the log message up front?

gregg-miskelly:
>                                 breakpointRequest is AD7BreakPointRequest ad7BPRequest)

Is there a reason to maintain your own collection instead of just storing this in the breakpoint request object?

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:1604 in df983e1. [](commit_id = df983e11fc402348dbe89438d0ac993e00c50e42, deletion_comment = False)

src/OpenDebugAD7/TracepointManager.cs
```diff
@@ -0,0 +1,229 @@
...
+            }
+
+            return logMessage;
+        }
+
+        private string GetInterpolatedLogMessage(string logMessage, IDebugThread2 pThread, uint radix)
+        {
+            if (pThread == null)
+            {
+                return logMessage;
```
gregg-miskelly, line 106: >logMessage [](start = 23, length = 10)

Shouldn't these cases be returning an error rather than the log message?

src/OpenDebugAD7/TracepointManager.cs
```diff
@@ -0,0 +1,229 @@
...
+                eventProperties.Add(DebuggerTelemetry.TelemetryTracepointTokens, tokensUsed);
+            }
+            DebuggerTelemetry.ReportEvent(DebuggerTelemetry.TelemetryTracepointEventName, eventProperties);
+
+            return interpolatedLogMessage;
+        }
+
+        private string InterpolateToken(string token)
+        {
+            return string.Format(CultureInfo.CurrentCulture, "<Not Implemented: ${0}>", token);
```
gregg-miskelly, line 184: >Implemented [](start = 67, length = 11)

Should this be localized?

src/OpenDebugAD7/TracepointManager.cs
```diff
@@ -0,0 +1,229 @@
...
+                        {
+                            case "FILEPOS":
+                            case "FUNCTION":
+                            case "ADDRESS":
+                            case "TID":
+                            case "TNAME":
+                            case "PID":
+                            case "PNAME":
+                            case "CALLER":
+                            case "CALLSTACK":
```
gregg-miskelly, line 144: Personally I would implement these - they are the main reason I use tracepoints. Even more than evaluating expressions.

src/OpenDebugAD7/TracepointManager.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            IDebugExpression2 expressionObject;
+            hr = expressionContext.ParseText(variable, enum_PARSEFLAGS.PARSE_EXPRESSION, radix, out expressionObject, out string _, out uint _);
+            eb.CheckHR(hr);
+            eb.CheckOutput(expressionObject);
+
+            IDebugProperty2 property;
+            if (expressionObject is IDebugExpressionDAP expressionDapObject)
+            {
+                DAPEvalFlags dapEvalFlags = DAPEvalFlags.CLIPBOARD_CONTEXT; // Use full string for tracepoints.
```
gregg-miskelly, line 205: >DAPEvalFlags dapEvalFlags = DAPEvalFlags.CLIPBOARD_CONTEXT; // Use full string for tracepoints. [](start = 16, length = 95)

Does GDB have particularly small strings? I am a bit skeptical that you want gigantic strings here.

src/OpenDebugAD7/TracepointManager.cs
```diff
@@ -0,0 +1,229 @@
...
+            }
+            eb.CheckHR(hr);
+            eb.CheckOutput(property);
+
+            DEBUG_PROPERTY_INFO[] propertyInfo = new DEBUG_PROPERTY_INFO[1];
+            hr = property.GetPropertyInfo(propertyInfoFlags, Constants.EvaluationRadix, Constants.EvaluationTimeout, null, 0, propertyInfo);
+            eb.CheckHR(hr);
+
+            // If the expression evaluation produces an error result and we are trying to get the expression for data tips
+            // return a failure result so that VS code won't display the error message in data tips
```
gregg-miskelly, line 220: Remove

src/OpenDebugAD7/TracepointManager.cs
```diff
@@ -0,0 +1,229 @@
...
+            int hr = HRConstants.S_OK;
+            string errorMessage = string.Format(CultureInfo.CurrentCulture, "<Evaluation Error: {{{0}}}>", variable);
+            ErrorBuilder eb = new ErrorBuilder(() => errorMessage);
+
+            IDebugExpressionContext2 expressionContext;
+            hr = topFrame.GetExpressionContext(out expressionContext);
+            eb.CheckHR(hr);
+
+            IDebugExpression2 expressionObject;
+            hr = expressionContext.ParseText(variable, enum_PARSEFLAGS.PARSE_EXPRESSION, radix, out expressionObject, out string _, out uint _);
```
gregg-miskelly, line 198: >_ [](start = 129, length = 1)

This is the error text. We should probably not ignore it.

src/OpenDebugAD7/TracepointManager.cs
```diff
@@ -0,0 +1,229 @@
...
+                    else
+                    {
+                        string expression = match.Value.Substring(1, match.Value.Length - 2);
+                        try
+                        {
+                            return InterpolateVariable(expression, topFrame[0].m_pFrame, radix);
+                        }
+                        catch (AD7Exception e)
+                        {
+                            return e.ToString();
```
gregg-miskelly, line 161: >e.ToString(); [](start = 35, length = 13)

Are there cases where this is useful? Sholud we just have some sort of generic 'Unable to exvaluate ...' instead?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2094,7 +2114,46 @@ public void HandleIDebugEntryPointEvent2(IDebugEngine2 pEngine, IDebugProcess2 p
...
+                        if (ppPendingBreakpoint.GetBreakpointRequest(out IDebugBreakpointRequest2 ppBPRequest) == HRConstants.S_OK)
+                        {
+                            if (ppBPRequest is AD7BreakPointRequest ad7BreakpointRequest)
+                            {
+                                if (m_tracepointManager.Contains(ad7BreakpointRequest.Id))
+                                {
+                                    fireStopEvent = false;
+
+                                    ThreadPool.QueueUserWorkItem((o) =>
+                                    {
```
gregg-miskelly, line 2136: Do we need to kick this to the thread pool?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2094,7 +2114,46 @@ public void HandleIDebugEntryPointEvent2(IDebugEngine2 pEngine, IDebugProcess2 p
...
+                {
+                    if (boundBp[0].GetPendingBreakpoint(out IDebugPendingBreakpoint2 ppPendingBreakpoint) == HRConstants.S_OK)
+                    {
+                        if (ppPendingBreakpoint.GetBreakpointRequest(out IDebugBreakpointRequest2 ppBPRequest) == HRConstants.S_OK)
+                        {
+                            if (ppBPRequest is AD7BreakPointRequest ad7BreakpointRequest)
+                            {
+                                if (m_tracepointManager.Contains(ad7BreakpointRequest.Id))
+                                {
+                                    fireStopEvent = false;
```
gregg-miskelly, line 2133: >fireStopEvent [](start = 36, length = 13)

We need to either break out of the enumeration if we find one, or change this code to find multiple breakpoints up front. As you have it, if we ever found multiple breakpoints you would continue the process multiple times.

src/OpenDebugAD7/TracepointManager.cs
gregg-miskelly: In case it isn't obvious what I was pointing to, I meant `$CALLSTACK`.

src/OpenDebugAD7/TracepointManager.cs
WardenGnaw: The default size GDB prints is 200. I was worried if users were going to evaluate long strings and not get the full string.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: Yes. EnumFrameInfo runs `_engine.DebuggedProcess.WorkerThread.RunOperation(async () => stackFrames = await _engine.DebuggedProcess.ThreadCache.StackFrames(_debuggedThread));` and `RunOperation` which has an assert `!IsPollThread()`. If it does not run in the thread pool, it deadlocks.

src/OpenDebugAD7/TracepointManager.cs
gregg-miskelly: I guess. Though if the string is a megabyte, you probably don't want that either :). Doesn't seem like there is a great answer, so pick whichever I guess.

WardenGnaw: I agree with you. I think we will assume the majority of printed strings will be 200 or less. If users want to see more, they can add `set print elements 0` in `setupCommands`.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: Is there a case where we would get multiple breakpoints where there are logpoints and regular breakpoints?

gregg-miskelly: In DAP, the only way it could happen is if you had both multiple types of breakpoint (ex: function breakpoint and line) on the same instruction.

gregg-miskelly: Actually, I guess that isn't 100% true, because you could have multiple line breakpoints that all wound up pointing at the same instruction...

src/OpenDebugAD7/TracepointManager.cs
WardenGnaw: I do this now in CreateTracepoint(). Do if it fails do we just not bind the tracepoint?

src/OpenDebugAD7/TracepointManager.cs
WardenGnaw: Not sure what error to say, `unable to evaluate logMessage`?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1893,12 +1955,10 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
             hr = frame.GetExpressionContext(out expressionContext);
             eb.CheckHR(hr);
 
-            const uint InputRadix = 10;
-            DAPEvalFlags dapEvalFlags = DAPEvalFlags.NONE;
             IDebugExpression2 expressionObject;
             string error;
             uint errorIndex;
-            hr = expressionContext.ParseText(expression, enum_PARSEFLAGS.PARSE_EXPRESSION, InputRadix, out expressionObject, out error, out errorIndex);
+            hr = expressionContext.ParseText(expression, enum_PARSEFLAGS.PARSE_EXPRESSION, Constants.EvaluationRadix, out expressionObject, out error, out errorIndex);
```
gregg-miskelly, line 1961: >EvaluationRadix [](start = 101, length = 15)

nit: I might suggest calling this `InputRadix` or `ParseRadix` since `EvaluationRadix` doesn't indicate what this does.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2094,7 +2154,32 @@ public void HandleIDebugEntryPointEvent2(IDebugEngine2 pEngine, IDebugProcess2 p
...
+                    foreach (var tp in tracepoints)
+                    {
+                        string logMessage = tp.GetLogMessage(pThread, Constants.EvaluationRadix, m_processName);
+
+                        m_logger.WriteLine(LoggingCategory.DebuggerStatus, logMessage);
+                    }
+                });
+            }
+
+            if (!m_isStepping && tracepoints.Any())
```
gregg-miskelly, line 2171: >m_isStepping [](start = 17, length = 12)

Your idea here is that if we are stepping, stopping at the tracepoint is going to mess up the step, so we should probably stop? Probably worth a comment at least.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2094,7 +2154,32 @@ public void HandleIDebugEntryPointEvent2(IDebugEngine2 pEngine, IDebugProcess2 p
...
+                ThreadPool.QueueUserWorkItem((o) =>
+                {
+                    BeforeContinue();
+                    m_program.Continue(pThread);
+                });
+            }
+            else
+            {
+                FireStoppedEvent(pThread, StoppedEvent.ReasonValue.Breakpoint);
+            }
```
gregg-miskelly, line 2182: If there were tracepoints, you want to wait for the work you scheduled to the thread pool to finish before doing either of these. I would suggest moving this into the first lambda.

src/OpenDebugAD7/TracepointManager.cs
gregg-miskelly: Yup

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,523 @@
...
+                    {
+                        return processName;
+                    }
+                case "$CALLSTACK":
+                    {
+                        IEnumDebugFrameInfo2 frameInfoEnum;
+                        int hr = pThread.EnumFrameInfo(enum_FRAMEINFO_FLAGS.FIF_FRAME | enum_FRAMEINFO_FLAGS.FIF_FLAGS, Constants.EvaluationRadix, out frameInfoEnum);
+                        int count = 0;
+                        StringBuilder sb = new StringBuilder();
+                        while (count < 5)
```
gregg-miskelly, line 209: >5 [](start = 39, length = 1)

5 seems like a pretty small frame limit...

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: Oh, I see. Don't reuse `EvaluationRadix` here. The input to `ParseText` should always be `10`. AD7 was a bit crazy and supported something that we will never use - having a different input radix. So, in other words, some debugger where evaluating `10 + 6` would result in 22 (0x16) rather than 16. Instead make a new constant for `InputRadix` since someday we might want to fix MIEngine to allow hex output (it is part of the DAP, but VS Code doesn't send it). But we will never allow hex input.

---
In reply to: [444418266](https://github.com/microsoft/MIEngine/pull/1013#discussion_r444418266) [](ancestors = 444418266)

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,523 @@
...
+            int hr = HRConstants.S_OK;
+            string errorMessage = string.Format(CultureInfo.CurrentCulture, "<Evaluation Error: {{{0}}}>", variable);
+            ErrorBuilder eb = new ErrorBuilder(() => errorMessage);
+
+            IDebugExpressionContext2 expressionContext;
+            hr = topFrame.GetExpressionContext(out expressionContext);
+            eb.CheckHR(hr);
+
+            IDebugExpression2 expressionObject;
+            hr = expressionContext.ParseText(variable, enum_PARSEFLAGS.PARSE_EXPRESSION, radix, out expressionObject, out string errStr, out uint errIdx);
```
gregg-miskelly, line 241: >radix [](start = 89, length = 5)

Input radix

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: Yep. Ran into a bug where we would get a step event. Fire the step event in the debugger. Debugger sends a BreakpointEvent and instead of stoppping for the step, we would continue.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1641,12 +1683,32 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                                    });
+                                }
+                                else
+                                {
+                                    resBreakpoints.Add(new Breakpoint()
+                                    {
+                                        Id = (int)pBPRequest.Id,
+                                        Verified = verified,
+                                        Line = bp.Line,
+                                        Message = "Unable to parse logMessage."
```
gregg-miskelly, line 1709: >"Unable to parse logMessage." [](start = 50, length = 29)

Localize?

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,523 @@
...
+                        {
+                            nested--;
+                        }
+                        else
+                        {
+                            endIndex = index;
+                            return true;
+                        }
+                    }
+                    else if (c == '"')
```
gregg-miskelly, line 495: >c == '"') [](start = 29, length = 9)

Does VS special case quoted strings also? This feels like it isn't really necessary -- if I evaluated `"{myVarWithValue1}"`, getting my `"1"` seems okay to me. But if VS does it also, I don't have a problem doing the same.

gregg-miskelly: Sorry, wrong spot.

---
In reply to: [444434372](https://github.com/microsoft/MIEngine/pull/1013#discussion_r444434372) [](ancestors = 444434372)

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,523 @@
...
+                        if (FindEndQuote(input.Substring(index), out int length))
+                        {
+                            length = length + 1;
+                        }
+                        else
+                        {
+                            length = input.Length - index;
+                        }
+                        replace.Append(input.Substring(index, length));
+                        index += length - 1;
```
gregg-miskelly, line 369: Does VS special case quoted strings also? This feels like it isn't really necessary -- if I evaluated `"{myVarWithValue1}"`, getting my `"1"` seems okay to me. But if VS does it also, I don't have a problem doing the same.

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: Got it. GDB also has other issues if you mix tracepoints and stepping, so sounds reasonable.

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,509 @@
...
+
+        private Tracepoint(string logMessage)
+        {
+            m_indexToExpressions = new Dictionary<int, string>();
+            LogMessage = Parse(logMessage);
+        }
+
+        internal static Tracepoint CreateTracepoint(string logMessage)
+        {
+            return new Tracepoint(logMessage); ;
```
WardenGnaw, line 30: nit: extra semicolon

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,509 @@
...
+                case "$ADDRESS":
+                case "$TID":
+                case "$TNAME":
+                case "$PID":
+                case "$PNAME":
+                case "$CALLER":
+                case "$CALLSTACK":
+                case "$TICK":
+                    return true;
+                    break;
```
WardenGnaw, line 395: nit: never hit

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -513,6 +545,7 @@ private void StepInternal(int threadId, enum_STEPKIND stepKind, enum_STEPUNIT st
                 ErrorBuilder builder = new ErrorBuilder(() => errorMessage);
                 try
                 {
+                    m_isStepping = true;
```
pieandcakes, line 548: why is this in the `try` block

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,509 @@
...
+using System.Text.RegularExpressions;
+
+namespace OpenDebugAD7
+{
+    internal class Tracepoint
+    {
+        // LogMessage after it has been Parsed()
+        public string LogMessage { get; set; }
+
+        private readonly IDictionary<int, string> m_indexToExpressions;
```
pieandcakes, line 20: are 'Expressions' the tokens that you have to evaluate separately?

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,509 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.Debugger.Interop;
+using Microsoft.VisualStudio.Debugger.Interop.DAP;
+using Microsoft.VisualStudio.Shared.VSCodeDebugProtocol.Messages;
+using System;
```
pieandcakes, line 7: nit: Sort System Usings first

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,509 @@
...
+                                TEXT_POSITION[] textPosBeg = new TEXT_POSITION[1];
+                                TEXT_POSITION[] textPosEnd = new TEXT_POSITION[1];
+                                hr = pDocumentContext.GetStatementRange(textPosBeg, textPosEnd);
+                                if (hr >= 0)
+                                {
+                                    return string.Format(CultureInfo.InvariantCulture, "{0}({1})", fileName, textPosBeg[0].dwLine + 1);
+                                }
+                            }
+                        }
+                        return string.Format(CultureInfo.InvariantCulture, "<No File Posiition>");
```
pieandcakes, line 151: spelling `Position`

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,509 @@
...
+                        {
+                            FRAMEINFO[] frames = new FRAMEINFO[2];
+                            uint fetched = 0;
+                            hr = frameInfoEnum.Next(2, frames, ref fetched);
+                            if (hr < 0 && fetched == 2)
+                            {
+                                return frames[1].m_bstrFuncName;
+                            }
+                        }
+                        return string.Format(CultureInfo.InvariantCulture, "<No caller avaliable>");
```
pieandcakes, line 167: why do your other errors have the first letter of each word capitalized but not here?

src/OpenDebugAD7/Tracepoint.cs
WardenGnaw: yes, it is a map of the index in LogMessage to insert the evaluated expression such as `$CALLSTACK` or `{1 + 5}`

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,509 @@
...
+            }
+
+            sb.Append(LogMessage.Substring(currIndex, LogMessage.Length - currIndex));
+
+            return sb.ToString();
+        }
+
+        private string InterpolateToken(string token, IDebugThread2 pThread, IDebugStackFrame2 topFrame, uint radix, string processName)
+        {
+            switch (token)
```
pieandcakes, line 104: can I use lowercase tokens?

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,509 @@
...
+                        index += length;
+                    }
+                }
+            }
+
+            return false;
+        }
+
+        #endregion
+    }
```
pieandcakes, line 501: nit: newline

src/OpenDebugAD7/Telemetry/DebuggerTelemetry.cs
```diff
@@ -55,6 +56,7 @@ internal sealed class DebuggerTelemetry
         public const string TelemetrySourceFileMappings = "SourceFileMappings";
         public const string TelemetryMIMode = "MIMode";
         public const string TelemetryStackFrameId = TelemetryExecuteInConsole + ".ExecuteInConsole";
+        public const string TelemetryTracepointTokens = ".Tokens";
```
pieandcakes, line 59: where is this being used?

pieandcakes: also your property shouldn't start with a `.`. I think you are missing a concatenation.

src/OpenDebugAD7/Telemetry/DebuggerTelemetry.cs
```diff
@@ -55,6 +56,7 @@ internal sealed class DebuggerTelemetry
         public const string TelemetrySourceFileMappings = "SourceFileMappings";
         public const string TelemetryMIMode = "MIMode";
         public const string TelemetryStackFrameId = TelemetryExecuteInConsole + ".ExecuteInConsole";
```
pieandcakes, line 58: not your change, but why is TelemetryStackFrameId = "Evaluate.ExecuteInConsole.ExecuteInConsole" ?

pieandcakes: Should be maybe ".StackFrameId"

src/OpenDebugAD7/Telemetry/DebuggerTelemetry.cs
```diff
@@ -28,6 +28,7 @@ internal sealed class DebuggerTelemetry
         public const string TelemetryAttachEventName = "Attach";
         public const string TelemetryEvaluateEventName = "Evaluate";
         public const string TelemetryPauseEventName = "Pause";
+        public const string TelemetryTracepointEventName = "Tracepoint";
```
pieandcakes, line 31: You should add telemetry to the session telemetry to show how often tracepoints are used. 

Also add failure telemetry to see how often your exception gets thrown and add some information on why its failing to parse w/o adding the actual text.

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,509 @@
...
+                    {
+                        return processName;
+                    }
+                case "$CALLSTACK":
+                    {
+                        IEnumDebugFrameInfo2 frameInfoEnum;
+                        int hr = pThread.EnumFrameInfo(enum_FRAMEINFO_FLAGS.FIF_FRAME | enum_FRAMEINFO_FLAGS.FIF_FLAGS, Constants.EvaluationRadix, out frameInfoEnum);
+                        int count = 0;
+                        StringBuilder sb = new StringBuilder();
+                        while (count < Constants.DefaultNumCallStacksShownToken)
```
pieandcakes, line 209: Is this going to be user configurable? can you allow format of `$CALLSTACK,<num>` if they want to override it?

pieandcakes: Also, what does this look like? i imagine its not going to look great.

pieandcakes:
@WardenGnaw Please add an example of input/output in your original comments so we can get an idea on what the output looks like in VS Code

src/OpenDebugAD7/Tracepoint.cs
WardenGnaw: No, these tokens are the same as the one in VS which are all uppercase.
image

https://docs.microsoft.com/en-us/visualstudio/debugger/using-tracepoints?view=vs-2019#actions-menu

src/OpenDebugAD7/Tracepoint.cs
WardenGnaw: I was going to do a max of 10 frames for now, if users want more I'll have it as a feature request.

WardenGnaw:
Example usage:
image

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,546 @@
...
+                message = string.Format(CultureInfo.InvariantCulture, "Unable to interpolate logMessage because current thread is missing.");
+                return HRConstants.E_FAIL;
+            }
+
+            // Get topFrame
+            IEnumDebugFrameInfo2 frameInfoEnum;
+            hr = pThread.EnumFrameInfo(enum_FRAMEINFO_FLAGS.FIF_FRAME | enum_FRAMEINFO_FLAGS.FIF_FLAGS, Constants.EvaluationRadix, out frameInfoEnum);
+            if (hr < 0)
+            {
+                message = "Unable to interpolate logMessage because frames could not be retrieved.";
```
WardenGnaw, line 62: string.format

pieandcakes: Should your warnings/errors be in a resx in case we decide to do localization in the future?

pieandcakes:
Thanks for the example. Is there a way to hide that breakpoint hit message in the middle of that message?

WardenGnaw:
That will be part of the `logging.threadEvents` flag work.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -237,6 +238,37 @@ private ProtocolException VerifyProcessId(string processId, string telemetryEven
...
+
+            if (debugEvent != null)
+            {
+                debugEvent.EnumBreakpoints(out IEnumDebugBoundBreakpoints2 pBoundBreakpoints);
+                IDebugBoundBreakpoint2[] boundBp = new IDebugBoundBreakpoint2[1];
+
+                uint numReturned = 0;
+                while (pBoundBreakpoints.Next(1, boundBp, ref numReturned) == HRConstants.S_OK && numReturned == 1)
+                {
+                    if (boundBp[0].GetPendingBreakpoint(out IDebugPendingBreakpoint2 ppPendingBreakpoint) == HRConstants.S_OK)
```
pieandcakes, line 253: Nit: would this be better as a multiline AND statement? the open/close curlies make it very tall.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1637,16 +1679,37 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                                }
+                                else
+                                {
+                                    resBreakpoints.Add(new Breakpoint()
+                                    {
+                                        Id = (int)pBPRequest.Id,
+                                        Verified = verified,
+                                        Line = bp.Line,
+                                        Message = string.Format(CultureInfo.CurrentCulture, AD7Resources.Error_UnableToSetTracepoint)
+                                    }); ;
```
pieandcakes, line 1711: remove extra `;`

src/OpenDebugAD7/AD7Impl/AD7BreakPointRequest.cs
```diff
@@ -112,5 +112,58 @@ public int GetChecksum(ref Guid guidAlgorithm, CHECKSUM_DATA[] checksumData)
...
+            catch (InvalidTracepointException e)
+            {
+                DebuggerTelemetry.ReportError(DebuggerTelemetry.TelemetryTracepointEventName, e.Message);
+                return false;
+            }
+            m_logMessage = logMessage;
+            return true;
+        }
+
+        public string LogMessage
```
pieandcakes, line 143: Would you be able to use the `public string LogMessage => m_logMessage;` notation for this one and the below?

src/OpenDebugAD7/AD7Resources.resx
```diff
@@ -192,6 +192,9 @@
   <data name="Error_UnableToSetBreakpoint" xml:space="preserve">
     <value>Error setting breakpoint. {0}</value>
   </data>
+  <data name="Error_UnableToSetTracepoint" xml:space="preserve">
```
pieandcakes, line 195: use a name that describes the error message you are giving, not the action `Error_UnableToParseLogMessage`

src/OpenDebugAD7/Telemetry/DebuggerTelemetry.cs
```diff
@@ -54,7 +55,7 @@ internal sealed class DebuggerTelemetry
         public const string TelemetryVisualizerFileUsed = "VisualizerFileUsed";
         public const string TelemetrySourceFileMappings = "SourceFileMappings";
         public const string TelemetryMIMode = "MIMode";
-        public const string TelemetryStackFrameId = TelemetryExecuteInConsole + ".ExecuteInConsole";
+        public const string TelemetryStackFrameId = TelemetryExecuteInConsole + ".StackFrameId";
```
pieandcakes, line 58: note: remember to classify this after its been shipped.

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -0,0 +1,546 @@
...
+            return hr;
+        }
+
+        private int GetInterpolatedLogMessage(string logMessage, IDebugThread2 pThread, uint radix, string processName, out string message)
+        {
+            int hr = HRConstants.S_OK;
+
+            if (pThread == null)
+            {
+                message = string.Format(CultureInfo.InvariantCulture, "Unable to interpolate logMessage because current thread is missing.");
```
pieandcakes, line 53: why are you using string.format if theres no formatting happening?

src/OpenDebugAD7/Constants.cs
```diff
@@ -28,7 +28,9 @@ internal static class Constants
     {
         // POST_PREVIEW_TODO: no-func-eval support, radix, timeout
         public const uint EvaluationRadix = 10;
+        public const uint ParseRadix = 10;
         public const uint EvaluationTimeout = 5000;
         public const int DisconnectTimeout = 2000;
+        public const int DefaultNumCallStacksShownToken = 10;
```
pieandcakes, line 34: why `Token`?

src/OpenDebugAD7/Telemetry/DebuggerTelemetry.cs
pieandcakes: Same for the tracepoint one.

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -68,13 +68,14 @@ private int GetInterpolatedLogMessage(string logMessage, IDebugThread2 pThread,
...
             {
-                message = string.Format(CultureInfo.InvariantCulture, "Unable to interpolate logMessage because there is no top frame.");
+                message = AD7Resources.Error_InterpolateMissingTopFrame;
                 return hr;
             }
 
             Dictionary<string, string> seenExpressions = new Dictionary<string, string>();
             StringBuilder sb = new StringBuilder();
             int currIndex = 0;
+            bool failureSeen = false;
```
pieandcakes, line 78: why not just create int hr = Constants.S_OK and set it to Constants.E_FAIL in your bool area and return hr at the bottom?

src/OpenDebugAD7/AD7Impl/AD7BreakPointRequest.cs
```diff
@@ -112,5 +112,40 @@ public int GetChecksum(ref Guid guidAlgorithm, CHECKSUM_DATA[] checksumData)
...
+        public bool SetLogMessage(string logMessage)
+        {
+            try
+            {
+                m_Tracepoint = Tracepoint.CreateTracepoint(logMessage);
+                DebuggerTelemetry.ReportEvent(DebuggerTelemetry.TelemetryTracepointEventName);
+            }
+            catch (InvalidTracepointException e)
+            {
+                DebuggerTelemetry.ReportError(DebuggerTelemetry.TelemetryTracepointEventName, e.Message);
```
pieandcakes, line 136: will this have a meaningful message? I don't remember you setting the message when you were throwing.

WardenGnaw: The only time this exception is thrown is when we cant find a matching '}' for interpolation.

src/OpenDebugAD7/Tracepoint.cs
```diff
@@ -94,10 +93,9 @@ private int GetInterpolatedLogMessage(string logMessage, IDebugThread2 pThread,
                     else
                     {
                         string toInterpolate = keyValuePair.Value;
-                        hr = InterpolateVariable(toInterpolate.Substring(1, toInterpolate.Length - 2), topFrame[0].m_pFrame, radix, out value);
-                        if (hr < 0)
+                        if (InterpolateVariable(toInterpolate.Substring(1, toInterpolate.Length - 2), topFrame[0].m_pFrame, radix, out value) < 0)
```
pieandcakes, line 96: why not make this return value the HR? and then ...?

WardenGnaw: I did not want to stop the log message from interpolating half way through. This allows it to completly interpolate the expressions and indicate an error has occured for the ones that failed (if there are multiple)

PR #1015, opened 2020-06-26
Eliminated ambiguity in the ListProcesses() error message

gc46:
Eliminated ambiguity in the ListProcesses() error message for the Docker 'Attach to Process' scenario. The change was tested by reproducing the scenario with the more effective error message dialog output.

**Before**
image

**After**
image

### DIFF
--- a/src/SSHDebugPS/Docker/DockerConnection.cs
+++ b/src/SSHDebugPS/Docker/DockerConnection.cs
@@ -2,14 +2,14 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 using Microsoft.SSHDebugPS.Utilities;
 using Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier;
-using System.Collections.Generic;
-using System.Text;
 
 namespace Microsoft.SSHDebugPS.Docker
 {
@@ -245,104 +245,12 @@ namespace Microsoft.SSHDebugPS.Docker
                 return new RemoteCommandRunner(settings, OuterConnection);
         }
 
-        private const string dockerVersionCommand = "version";
-        private const string dockerVersionArgs = "--format {{.Server.Os}}";
-
-        private const string dockerInfoCommand = "info";
-        private const string dockerInfoArgs = "--format {{.Driver}}";
-
-        public override List<Process> ListProcesses()
+        protected override string ProcFSErrorMessage
         {
-            string username;
-            TryGetUsername(out username);
-
-            List<Process> processes;
-            string psErrorMessage;
-            // Try using 'ps' first
-            if (!PSListProcess(username, out psErrorMessage, out processes))
+            get
             {
-                string procErrorMessage;
-                // try using the /proc file system
-                if (!ProcFSListProcess(username, out procErrorMessage, out processes))
-                {
-                    StringBuilder sb = new StringBuilder();
-                    sb.Append(psErrorMessage);
-                    sb.AppendLine(string.Empty);
-                    sb.AppendLine(procErrorMessage);
-
-                    string dockerContainer = DockerHelper.GetDockerOutputString(username, dockerVersionCommand, dockerVersionArgs);
-
-                    DockerPortSupplier portSupplier = new DockerPortSupplier();
-                    string connectionType;
-                    portSupplier.GetPortSupplierName(out connectionType);
-
-                    if (dockerContainer.Equals("windows") && connectionType.Equals(StringResources.Docker_PSName))
-                    {
-                            sb.AppendLine(StringResources.Error_DockerTypesNotMatching);
-                    }
-
-                    VSMessageBoxHelper.PostErrorMessage(StringResources.Error_ProcessListFailedTitle, sb.ToString());
-
-                    return new List<Process>(0);
-                }
+                return String.Concat(base.ProcFSErrorMessage, "\n", StringResources.Error_DockerTypesNotMatching);
             }
-
-            return processes;
-        }
-
-        /// <summary>
-        /// Query 'ps' command for a list of processes
-        /// </summary>
-        private bool PSListProcess(string username, out string errorMessage, out List<Process> processes)
-        {
-            errorMessage = string.Empty;
-            string commandOutput;
-            int exitCode;
-
-            if (!ExecuteCommand(PSOutputParser.PSCommandLine, Timeout.Infinite, false, out commandOutput, out errorMessage))
-            {
-                // Clear output and errorMessage
-                commandOutput = string.Empty;
-                errorMessage = string.Empty;
-                if (!ExecuteCommand(PSOutputParser.AltPSCommandLine, Timeout.Infinite, false, out commandOutput, out errorMessage, out exitCode))
-                {
-                    if (exitCode == 127)
-                    {
-                        //command doesn't Exist
-                        errorMessage = StringResources.Error_PSMissing;
-                    }
-                    else
-                    {
-                        errorMessage = StringResources.Error_PSErrorFormat.FormatCurrentCultureWithArgs(exitCode, errorMessage);
-                    }
-
-                    processes = null;
-                    return false;
-                }
-            }
-
-            processes = PSOutputParser.Parse(commandOutput, username);
-            return true;
-        }
-
-        /// <summary>
-        /// Query /proc for a list of processes
-        /// </summary>
-        private bool ProcFSListProcess(string username, out string errorMessage, out List<Process> processes)
-        {
-            errorMessage = string.Empty;
-            processes = null;
-
-            int exitCode;
-            string commandOutput;
-            if (!ExecuteCommand(ProcFSOutputParser.CommandText, Timeout.Infinite, false, out commandOutput, out errorMessage, out exitCode))
-            {
-                errorMessage = StringResources.Error_ProcFSError.FormatCurrentCultureWithArgs(errorMessage);
-                return false;
-            }
-
-            processes = ProcFSOutputParser.Parse(commandOutput, username);
-            return true;
         }
     }
 }
--- a/src/SSHDebugPS/Docker/DockerConnection.cs
+++ b/src/SSHDebugPS/Docker/DockerConnection.cs
@@ -8,6 +8,8 @@ using System.Text.RegularExpressions;
 using System.Threading;
 using Microsoft.SSHDebugPS.Utilities;
 using Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier;
+using System.Collections.Generic;
+using System.Text;
 
 namespace Microsoft.SSHDebugPS.Docker
 {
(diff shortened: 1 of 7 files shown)


src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -242,5 +244,105 @@ private ICommandRunner GetCommandRunner(DockerContainerTransportSettings setting
...
+        private const string dockerVersionCommand = "version";
+        private const string dockerVersionArgs = "--format {{.Server.Os}}";
+
+        private const string dockerInfoCommand = "info";
+        private const string dockerInfoArgs = "--format {{.Driver}}";
+
+        public override List<Process> ListProcesses()
+        {
+            string username;
+            TryGetUsername(out username);
```
WardenGnaw, line 257: What happens if we fail to get a username?

WardenGnaw: Would also do the same fix for https://github.com/microsoft/MIEngine/blob/4d8d626c7b0ef5150b44bafd9c6475c92b13c9e9/src/SSHDebugPS/PipeConnection.cs#L135

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -242,5 +244,105 @@ private ICommandRunner GetCommandRunner(DockerContainerTransportSettings setting
...
+
+            List<Process> processes;
+            string psErrorMessage;
+            // Try using 'ps' first
+            if (!PSListProcess(username, out psErrorMessage, out processes))
```
WardenGnaw, line 262: How about the case where PSListProcess fails but ProcFSListProcess succeeded? Might want to handle showing `psErrorMessage`.

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -242,5 +244,105 @@ private ICommandRunner GetCommandRunner(DockerContainerTransportSettings setting
             else
                 return new RemoteCommandRunner(settings, OuterConnection);
         }
+
+        private const string dockerVersionCommand = "version";
+        private const string dockerVersionArgs = "--format {{.Server.Os}}";
+
+        private const string dockerInfoCommand = "info";
```
WardenGnaw, line 251: Where are we checking for LCOW?

pieandcakes:
@gc46 thank you for your contribution!

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -10,13 +10,111 @@
...
+                    path = GetDockerOutputStringImpl(hostname, dockerCommand, dockerArgs);
+                });
+            }
+            catch (Exception)
+            {
+            }
+            return path;
+        }
+
+        public static string GetDockerOutputStringImpl(string hostname, string dockerCommand, string dockerArgs)
```
pieandcakes, line 41: If you are going to make an Impl method to back the public method, mark it `private`. If your intention is to be able to call either from the public surface, then don't create the impl method.

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -10,13 +10,111 @@
...
+            try
+            {
+                ThreadHelper.JoinableTaskFactory.Run(async () =>
+                {
+                    await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                    path = GetDockerOutputStringImpl(hostname, dockerCommand, dockerArgs);
+                });
+            }
+            catch (Exception)
+            {
```
pieandcakes, line 36: A blanket catch statement shouldn't be done here. You should handle `CommandFailedException` messages separately and then write an error.

src/SSHDebugPS/Docker/DockerConnection.cs
pieandcakes: In the previous case and in this case we shouldn't care. by returning no username, it should pull all processes. This is the correct behavior.

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -8,6 +8,8 @@
 using System.Threading;
 using Microsoft.SSHDebugPS.Utilities;
 using Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier;
```
pieandcakes, line 10: nit: sort usings. System usings should be above namespace usings.

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -242,5 +244,105 @@ private ICommandRunner GetCommandRunner(DockerContainerTransportSettings setting
             else
                 return new RemoteCommandRunner(settings, OuterConnection);
         }
+
+        private const string dockerVersionCommand = "version";
```
pieandcakes, line 248: Move this and the check to DockerHelper.cs

src/SSHDebugPS/Docker/DockerHelper.cs
pieandcakes: Call this GetOutputFromDockerCommand or something and reuse it for the other calls in this file.

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -10,13 +10,111 @@
 using Microsoft.SSHDebugPS.SSH;
 using Microsoft.SSHDebugPS.Utilities;
```
pieandcakes, line 12: remove extra newline

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -242,5 +244,105 @@ private ICommandRunner GetCommandRunner(DockerContainerTransportSettings setting
...
+                string procErrorMessage;
+                // try using the /proc file system
+                if (!ProcFSListProcess(username, out procErrorMessage, out processes))
+                {
+                    StringBuilder sb = new StringBuilder();
+                    sb.Append(psErrorMessage);
+                    sb.AppendLine(string.Empty);
+                    sb.AppendLine(procErrorMessage);
+
+                    string dockerContainer = DockerHelper.GetDockerOutputString(username, dockerVersionCommand, dockerVersionArgs);
```
pieandcakes, line 273: This won't work if your connecting to a remote SSH connection to get the file list. This method looks like it only runs the command locally.

I think you want to use `ExecuteCommand` but will need to play around with it to see what happens. 

Also you are mixing up `hostname` and `username`. `hostname` is for the `-H` flag to pass to docker if the user  wants to connect to a different docker daemon. `username` is the username associated with the docker session to get the processes for that user.

pieandcakes: Also think about a better way to do this as the only thing you are adding is the error handling but the rest of the method is copied over. maybe make an interface change that allows you to modify the error message (if necessary) so that there isn't duplicate code here.

src/SSHDebugPS/StringResources.resx
```diff
@@ -162,6 +162,9 @@
     <value>Source file '{0}' not found.</value>
     <comment>{0} = source file name</comment>
   </data>
+  <data name="Error_DockerTypesNotMatching" xml:space="preserve">
+    <value>Make sure your Docker Desktop container type matches the selected Connection type.</value>
```
pieandcakes, line 166: Change message to say `Ensure the target Docker container target type is Linux.` You know that for this connection it has to be a Linux connect.

src/SSHDebugPS/UI/UIResources.resx
```diff
@@ -190,6 +190,9 @@
   <data name="QueryingForContainersMessage" xml:space="preserve">
     <value>Querying for containers...</value>
   </data>
+  <data name="QueryingDockerCommandMessage" xml:space="preserve">
```
pieandcakes, line 193: lets not make this as generic as this.

pieandcakes:
The testing matrix should be somewhat like this:
* SSH -> Linux -> Linux containers
* local -> docker hostname -> [Windows | Linux] containers
* local [Windows | Linux] containers
* SSH -> linux -> docker hostname -> [Windows | Linux] containers

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -242,5 +244,13 @@ private ICommandRunner GetCommandRunner(DockerContainerTransportSettings setting
             else
                 return new RemoteCommandRunner(settings, OuterConnection);
         }
+
+        protected override string ProcFSErrorMessage
+        {
+            get
+            {
+                return String.Concat(base.ProcFSErrorMessage, "\n", StringResources.Error_DockerTypesNotMatching);
```
pieandcakes, line 252: Should this be `\r\n`  on Windows?

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -17,6 +18,107 @@ public class DockerHelper
...
+                ThreadHelper.JoinableTaskFactory.Run(async () =>
+                {
+                    await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                    path = GetDockerOutputStringImpl(hostname, dockerCommand, dockerArgs);
+                });
+            }
+            catch (CommandFailedException ex)
+            {
+                string errorMessage = UIResources.CommandExecutionErrorFormat.FormatCurrentCultureWithArgs(dockerCommand, ex.Message);
+                throw new CommandFailedException(errorMessage, ex);
```
pieandcakes, line 40: Why catch and throw a new one? can you format the error where the original throw is and then just do `throw` here to rethrow? throwing new exceptions is expensive and this seems to be for control flow.

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -160,7 +155,7 @@ public override List<Process> ListProcesses()
         /// <summary>
         /// Query 'ps' command for a list of processes
         /// </summary>
-        private bool PSListProcess(string username, out string errorMessage, out List<Process> processes)
+        protected bool PSListProcess(string username, out string errorMessage, out List<Process> processes)
```
pieandcakes, line 158: change back to private

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -191,10 +186,12 @@ private bool PSListProcess(string username, out string errorMessage, out List<Pr
             return true;
         }
 
+        protected virtual string ProcFSErrorMessage => StringResources.Error_ProcFSError;
+
         /// <summary>
         /// Query /proc for a list of processes
         /// </summary>
-        private bool ProcFSListProcess(string username, out string errorMessage, out List<Process> processes)
+        protected bool ProcFSListProcess(string username, out string errorMessage, out List<Process> processes)
```
pieandcakes, line 194: change back to private

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -17,6 +17,7 @@ public class DockerHelper
         private const string dockerPSCommand = "ps";
         // --no-trunc avoids parameter truncation
         private const string dockerPSArgs = "-f status=running --no-trunc --format \"{{json .}}\"";
+
```
WardenGnaw, line 20: Undo changes in this file.

WardenGnaw: You want to run `git checkout HEAD~6 src/SSHDebugPS/Docker/DockerHelper.cs`

src/SSHDebugPS/StringResources.resx
```diff
@@ -162,6 +162,9 @@
     <value>Source file '{0}' not found.</value>
     <comment>{0} = source file name</comment>
   </data>
+  <data name="Error_DockerTypesNotMatching" xml:space="preserve">
```
pieandcakes, line 165: change the error name to be more indicative of the actual message. maybe "Error_EnsureDockerContainerIsLinux" or something.  Also use `Ensure` versus `Make sure`.

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -242,5 +242,13 @@ private ICommandRunner GetCommandRunner(DockerContainerTransportSettings setting
             else
                 return new RemoteCommandRunner(settings, OuterConnection);
         }
+
+        protected override string ProcFSErrorMessage
+        {
+            get
+            {
+                return String.Concat(base.ProcFSErrorMessage, "\r\n", StringResources.Error_DockerTypesNotMatching);
```
pieandcakes, line 250: nit: replace '\r\n' with Environment.NewLine

PR #1017, opened 2020-07-07
[Docker] Fix remote containers list, add OutputWindow error support and Telemetry

WardenGnaw:
In the Container Picker Dialog, it uses Name and ShortId. However, if
Id is null, ShortId tries to get the length of the Id and this causes a
NullRefException.

We also are adding null containers for GetRemoteDockerContainers. Added
a null check. We do the same in:
https://github.com/microsoft/MIEngine/blob/185e133c3f051a364252e65dd4159d5e8a6ca408/src/SSHDebugPS/Docker/DockerHelper.cs#L62

Adding a null check on Id so it just returns null.

Added support for logging errors to the output window.

Added telemetry to count for errors in Docker process listings

### DIFF
--- a/src/SSHDebugPS/Docker/DockerContainerInstance.cs
+++ b/src/SSHDebugPS/Docker/DockerContainerInstance.cs
@@ -2,7 +2,11 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
+
+using Microsoft.DebugEngineHost;
+using Microsoft.SSHDebugPS.Utilities;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
@@ -10,13 +14,13 @@ namespace Microsoft.SSHDebugPS.Docker
 {
     public class DockerContainerInstance : ContainerInstance
     {
+        private static readonly string Property_ExceptionName = "ExceptionName";
         /// <summary>
         /// Create a DockerContainerInstance from the results of docker ps in JSON format
         /// </summary>
-        public static bool TryCreate(string json, out DockerContainerInstance instance, out string error)
+        public static bool TryCreate(string json, out DockerContainerInstance instance)
         {
             instance = null;
-            error = string.Empty;
             try
             {
                 JObject obj = JObject.Parse(json);
@@ -24,7 +28,12 @@ namespace Microsoft.SSHDebugPS.Docker
             }
             catch (Exception e)
             {
-                error = e.ToString();
+                List<KeyValuePair<string, object>> eventProperties = new List<KeyValuePair<string, object>>();
+                eventProperties.Add(new KeyValuePair<string, object>(Property_ExceptionName, e.GetType().Name));
+                HostTelemetry.SendEvent(Telemetry.Event_DockerPSParseFailure, eventProperties.ToArray());
+
+                string error = e.ToString();
+                VsOutputWindowWrapper.WriteLine(StringResources.Error_DockerPSParseFailed.FormatCurrentCultureWithArgs(json, error), StringResources.Docker_PSName);
                 Debug.Fail(error);
             }
             return instance != null;
--- a/src/SSHDebugPS/Docker/DockerContainerInstance.cs
+++ b/src/SSHDebugPS/Docker/DockerContainerInstance.cs
@@ -13,20 +13,19 @@ namespace Microsoft.SSHDebugPS.Docker
         /// <summary>
         /// Create a DockerContainerInstance from the results of docker ps in JSON format
         /// </summary>
-        public static DockerContainerInstance Create(string json)
+        public static bool TryCreate(string json, out DockerContainerInstance instance)
         {
+            instance = null;
             try
             {
                 JObject obj = JObject.Parse(json);
-                var instance = obj.ToObject<DockerContainerInstance>();
-                if (instance != null)
-                    return instance;
+                instance = obj.ToObject<DockerContainerInstance>();
             }
             catch (Exception e)
             {
                 Debug.Fail(e.ToString());
             }
-            return null;
+            return instance != null;
         }
 
         private DockerContainerInstance() { }
--- a/src/SSHDebugPS/Docker/DockerContainerInstance.cs
+++ b/src/SSHDebugPS/Docker/DockerContainerInstance.cs
@@ -13,9 +13,10 @@ namespace Microsoft.SSHDebugPS.Docker
         /// <summary>
         /// Create a DockerContainerInstance from the results of docker ps in JSON format
         /// </summary>
-        public static bool TryCreate(string json, out DockerContainerInstance instance)
+        public static bool TryCreate(string json, out DockerContainerInstance instance, out string error)
         {
             instance = null;
+            error = string.Empty;
             try
             {
                 JObject obj = JObject.Parse(json);
@@ -23,7 +24,8 @@ namespace Microsoft.SSHDebugPS.Docker
             }
             catch (Exception e)
             {
-                Debug.Fail(e.ToString());
+                error = e.ToString();
+                Debug.Fail(error);
             }
             return instance != null;
         }
--- a/src/SSHDebugPS/Docker/DockerContainerInstance.cs
+++ b/src/SSHDebugPS/Docker/DockerContainerInstance.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-
 using Microsoft.DebugEngineHost;
 using Microsoft.SSHDebugPS.Utilities;
 using Newtonsoft.Json;
@@ -28,9 +27,9 @@ namespace Microsoft.SSHDebugPS.Docker
             }
             catch (Exception e)
             {
-                List<KeyValuePair<string, object>> eventProperties = new List<KeyValuePair<string, object>>();
-                eventProperties.Add(new KeyValuePair<string, object>(Property_ExceptionName, e.GetType().Name));
-                HostTelemetry.SendEvent(Telemetry.Event_DockerPSParseFailure, eventProperties.ToArray());
+                HostTelemetry.SendEvent(Telemetry.Event_DockerPSParseFailure, new KeyValuePair<string, object>[] {
+                    new KeyValuePair<string, object>(Property_ExceptionName, e.GetType().Name)
+                });
 
                 string error = e.ToString();
                 VsOutputWindowWrapper.WriteLine(StringResources.Error_DockerPSParseFailed.FormatCurrentCultureWithArgs(json, error), StringResources.Docker_PSName);
--- a/src/SSHDebugPS/Docker/DockerHelper.cs
+++ b/src/SSHDebugPS/Docker/DockerHelper.cs
@@ -7,8 +7,7 @@ using System.ComponentModel;
 using System.Linq;
 using System.Text;
 using System.Threading;
-using EnvDTE;
-using Microsoft.DebugEngineHost;
+
 using Microsoft.SSHDebugPS.SSH;
 using Microsoft.SSHDebugPS.Utilities;
 
@@ -60,7 +59,10 @@ namespace Microsoft.SSHDebugPS.Docker
                         }
                         else
                         {
-                            AddOutputLineToContainersList(args, ref containers);
+                            if (DockerContainerInstance.TryCreate(args, out DockerContainerInstance containerInstance))
+                            {
+                                containers.Add(containerInstance);
+                            }
                         }
                     }
                 });
@@ -210,24 +212,14 @@ namespace Microsoft.SSHDebugPS.Docker
 
                 foreach (var item in outputLines)
                 {
-                    AddOutputLineToContainersList(item, ref containers);
+                    if (DockerContainerInstance.TryCreate(item, out DockerContainerInstance containerInstance))
+                    {
+                        containers.Add(containerInstance);
+                    }
                 }
             }
 
             return containers;
-        }
-
-        private static void AddOutputLineToContainersList(string item, ref List<DockerContainerInstance> containers)
-        {
-            if (DockerContainerInstance.TryCreate(item, out DockerContainerInstance containerInstance, out string error))
-            {
-                containers.Add(containerInstance);
-            }
-            else
-            {
-                HostTelemetry.SendEvent(Telemetry.Event_DockerPSParseFailure);
-                OutputWindow.Instance.WriteLine(StringResources.Error_DockerPSParseFailed.FormatCurrentCultureWithArgs(item, error), StringResources.Docker_PSName);
-            }
         }
     }
 }
--- a/src/SSHDebugPS/Docker/DockerHelper.cs
+++ b/src/SSHDebugPS/Docker/DockerHelper.cs
@@ -210,7 +210,11 @@ namespace Microsoft.SSHDebugPS.Docker
 
                 foreach (var item in outputLines)
                 {
-                    containers.Add(DockerContainerInstance.Create(item));
+                    DockerContainerInstance instance = DockerContainerInstance.Create(item);
+                    if (instance != null)
+                    {
+                        containers.Add(instance);
+                    }
                 }
             }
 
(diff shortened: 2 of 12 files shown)


src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -210,7 +210,11 @@ internal static IEnumerable<DockerContainerInstance> GetRemoteDockerContainers(I
 
                 foreach (var item in outputLines)
                 {
-                    containers.Add(DockerContainerInstance.Create(item));
+                    DockerContainerInstance instance = DockerContainerInstance.Create(item);
```
gregg-miskelly, line 213: >Create [](start = 79, length = 6)

nit: Rename to TryCreate?

src/SSHDebugPS/Docker/DockerContainerInstance.cs
```diff
@@ -13,20 +13,19 @@ public class DockerContainerInstance : ContainerInstance
...
-                if (instance != null)
-                    return instance;
+                instance = obj.ToObject<DockerContainerInstance>();
             }
             catch (Exception e)
             {
                 Debug.Fail(e.ToString());
             }
-            return null;
+            return instance != null;
```
pieandcakes, line 28: should add some logging or assertion if it comes back null

pieandcakes: And find some way to let the user know that it can't be parsed. Might be good to add some telemetry too.

src/SSHDebugPS/OutputWindow.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Collections.Generic;
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.SSHDebugPS
+{
+    class OutputWindow
+    {
+        private static OutputWindow _instance;
```
pieandcakes, line 15: you should rename this as VsOutputWindow so that we know thats the output window you are talking about.

Also mark OutputWindow `internal`

src/SSHDebugPS/OutputWindow.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                this.paneId = paneId;
+                this.Shown = false;
+            }
+
+            internal Guid paneId;
+            internal bool Shown { get; set; }
+        }
+
+        private IVsOutputWindow outputWindow;
```
pieandcakes, line 42: rethink the naming to differentiate between `this` object and the actual IVsOutputWindow.

src/SSHDebugPS/OutputWindow.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        private Dictionary<string, PaneInfo> panes = new Dictionary<string, PaneInfo>()
+        {
+            // The 'Debug' pane exists by default
+            { "Debug", new PaneInfo(VSConstants.GUID_OutWindowDebugPane) }
+        };
+
+        private OutputWindow()
+        {
+            outputWindow = Package.GetGlobalService(typeof(SVsOutputWindow)) as IVsOutputWindow;
```
pieandcakes, line 53: if this runs outside of VS (like during tests) this will throw. Think about putting it in a try/catch.

src/SSHDebugPS/OutputWindow.cs
```diff
@@ -0,0 +1,123 @@
...
+            shell = Package.GetGlobalService(typeof(SVsUIShell)) as IVsUIShell;
+            if (outputWindow == null || shell == null)
+            {
+                throw new NullReferenceException("Could not create OutputWindow");
```
pieandcakes, line 57: maybe debug.assert and then no-opt the writing. I don't know if throwing is the best option here since it will crash VS.

src/SSHDebugPS/OutputWindow.cs
```diff
@@ -0,0 +1,123 @@
...
+                            object inputVariant = null;
+                            shell.PostExecCommand(commandSet, (uint)VSConstants.VSStd97CmdID.OutputWindow, 0, ref inputVariant);
+                        }
+                    }
+
+                    // Output the text
+                    outputPane.OutputString(message);
+                }
+                catch (Exception)
+                {
```
pieandcakes, line 110: why blanket catch with no action? are you expecting a throw? Maybe a debug.assert here to call out why its throwing?

This seems different than the experience above where you are throwing NRE.

src/SSHDebugPS/OutputWindow.cs
gregg-miskelly: Make sure to run glass tests. You may actually need to move these calls to a separate class to be able to catch the failure.

---
In reply to: [452387606](https://github.com/microsoft/MIEngine/pull/1017#discussion_r452387606) [](ancestors = 452387606)

gregg-miskelly: Actually, wait, we don't have glass tests for the SSH port supplier - just the engine.

---
In reply to: [452411351](https://github.com/microsoft/MIEngine/pull/1017#discussion_r452411351) [](ancestors = 452411351,452387606)

src/SSHDebugPS/OutputWindow.cs
gregg-miskelly: Also, don't throw NullReferenceException. InvalidOperationException or NotImplementedException seem to make sense.

---
In reply to: [452387943](https://github.com/microsoft/MIEngine/pull/1017#discussion_r452387943) [](ancestors = 452387943)

src/SSHDebugPS/Docker/DockerContainerInstance.cs
```diff
@@ -13,20 +13,21 @@ public class DockerContainerInstance : ContainerInstance
...
-                if (instance != null)
-                    return instance;
+                instance = obj.ToObject<DockerContainerInstance>();
             }
             catch (Exception e)
             {
-                Debug.Fail(e.ToString());
+                error = e.ToString();
+                Debug.Fail(error);
             }
```
gregg-miskelly, line 29: >} [](start = 12, length = 1)

Are you sure you don't want to do the logging in the spot where you have the exception so that you can also include some part of the failure info?

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -210,11 +210,24 @@ internal static IEnumerable<DockerContainerInstance> GetRemoteDockerContainers(I
...
+
+        private static void AddOutputLineToContainersList(string item, ref List<DockerContainerInstance> containers)
+        {
+            if (DockerContainerInstance.TryCreate(item, out DockerContainerInstance containerInstance, out string error))
+            {
+                containers.Add(containerInstance);
+            }
+            else
+            {
+                HostTelemetry.SendEvent(Telemetry.Event_DockerPSParseFailure);
```
gregg-miskelly, line 228: >) [](start = 76, length = 1)

I wonder if you want to include the exception type name in the telemetry?

src/SSHDebugPS/Docker/DockerContainerInstance.cs
```diff
@@ -2,31 +2,41 @@
...
-                if (instance != null)
-                    return instance;
+                instance = obj.ToObject<DockerContainerInstance>();
             }
             catch (Exception e)
             {
-                Debug.Fail(e.ToString());
+                List<KeyValuePair<string, object>> eventProperties = new List<KeyValuePair<string, object>>();
+                eventProperties.Add(new KeyValuePair<string, object>(Property_ExceptionName, e.GetType().Name));
+                HostTelemetry.SendEvent(Telemetry.Event_DockerPSParseFailure, eventProperties.ToArray());
```
pieandcakes, line 33: why create a list to add an item and then call toArray on it ? can you define an array with one keyvalue pair inline?

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -7,6 +7,7 @@
 using System.Linq;
 using System.Text;
 using System.Threading;
+
```
pieandcakes, line 10: nit: we don't need a new line here

src/SSHDebugPS/Docker/DockerContainerInstance.cs
```diff
@@ -2,31 +2,41 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
+
```
pieandcakes, line 7: nit: we don't need a new line here

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -58,9 +59,10 @@ public static IEnumerable<DockerContainerInstance> GetLocalDockerContainers(stri
                         }
                         else
                         {
-                            var containerInstance = DockerContainerInstance.Create(args);
-                            if (containerInstance != null)
```
pieandcakes, line 62: think about updating the text in the status on the window with something like `Found x containers, Failed to parse y`. or even creating a status message that will tell users to go look in the output window.

src/SSHDebugPS/Utilities/Telemetry.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.SSHDebugPS.Utilities
+{
+    internal class Telemetry
```
pieandcakes, line 6: nit: `static`

src/SSHDebugPS/VsOutputWindowWrapper.cs
```diff
@@ -0,0 +1,138 @@
...
+                    await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                    outputWindow = Package.GetGlobalService(typeof(SVsOutputWindow)) as IVsOutputWindow;
+                }
+                catch (Exception)
+                {
+                }
+                return outputWindow;
+            });
+        });
+        private static Lazy<IVsUIShell> shellLazy = new Lazy<IVsUIShell>(() =>
```
pieandcakes, line 32: nit: newline

src/SSHDebugPS/VsOutputWindowWrapper.cs
```diff
@@ -0,0 +1,138 @@
...
+            return ThreadHelper.JoinableTaskFactory.Run(async delegate
+            {
+                IVsOutputWindow outputWindow = null;
+                try
+                {
+                    await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                    outputWindow = Package.GetGlobalService(typeof(SVsOutputWindow)) as IVsOutputWindow;
+                }
+                catch (Exception)
+                {
```
pieandcakes, line 27: Assert here as this shouldn't throw in most cases.

src/SSHDebugPS/VsOutputWindowWrapper.cs
```diff
@@ -0,0 +1,138 @@
...
+            return ThreadHelper.JoinableTaskFactory.Run(async delegate
+            {
+                IVsUIShell shell = null;
+                try
+                {
+                    await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                    shell = Package.GetGlobalService(typeof(SVsUIShell)) as IVsUIShell;
+                }
+                catch (Exception)
+                {
```
pieandcakes, line 43: same

src/SSHDebugPS/VsOutputWindowWrapper.cs
```diff
@@ -0,0 +1,138 @@
...
+                }
+                return shell;
+            });
+        });
+
```
pieandcakes, line 48: nit: extra newline

src/SSHDebugPS/VsOutputWindowWrapper.cs
```diff
@@ -0,0 +1,138 @@
...
+            }
+
+            internal Guid paneId;
+            internal bool Shown { get; set; }
+        }
+
+        private static Dictionary<string, PaneInfo> panes = new Dictionary<string, PaneInfo>()
+        {
+            // The 'Debug' pane exists by default
+            { "Debug", new PaneInfo(VSConstants.GUID_OutWindowDebugPane) }
```
pieandcakes, line 65: i feel like with the same string being used in 3 places (at least) you should make a const  and refer to it.

src/SSHDebugPS/VsOutputWindowWrapper.cs
```diff
@@ -0,0 +1,138 @@
...
+                            object inputVariant = null;
+                            shell.PostExecCommand(commandSet, (uint)VSConstants.VSStd97CmdID.OutputWindow, 0, ref inputVariant);
+                        }
+                    }
+
+                    // Output the text
+                    outputPane.OutputString(message);
+                }
+                catch (Exception)
+                {
```
pieandcakes, line 125: Assert here too as this shouldn't fail and we should find out when it does.

src/SSHDebugPS/VsOutputWindowWrapper.cs
```diff
@@ -0,0 +1,138 @@
...
+                        paneInfo.Shown = true;
+
+                        // Switch to the pane of the Output window
+                        outputPane.Activate();
+
+                        // Show the output window
+                        IVsUIShell shell = shellLazy.Value;
+                        if (shell != null)
+                        {
+                            Guid commandSet = VSConstants.GUID_VSStandardCommandSet97;
```
pieandcakes, line 115: why do you create a variable here for comandset, but then inline the OutputWindow CmdId?

WardenGnaw:
Example output: 
image

WardenGnaw:
Example output:
image

src/SSHDebugPS/VsOutputWindowWrapper.cs
```diff
@@ -0,0 +1,145 @@
...
+            // The 'Debug' pane exists by default
+            { DefaultOutputPane, new PaneInfo(VSConstants.GUID_OutWindowDebugPane) }
+        };
+
+        /// <summary>
+        /// Writes text directly to the VS Output window.
+        /// </summary>
+        public static void Write(string message, string pane = DefaultOutputPane)
+        {
+            ThreadHelper.JoinableTaskFactory.Run(async delegate
```
gregg-miskelly, line 80: >Run [](start = 45, length = 3)

Why block the thread waiting to write? Use RunAsync(...).FileAndForget(...);

src/SSHDebugPS/VsOutputWindowWrapper.cs
```diff
@@ -0,0 +1,145 @@
...
+    internal static class VsOutputWindowWrapper
+    {
+        private static Lazy<IVsOutputWindow> outputWindowLazy = new Lazy<IVsOutputWindow>(() =>
+        {
+            return ThreadHelper.JoinableTaskFactory.Run(async delegate
+            {
+                IVsOutputWindow outputWindow = null;
+                try
+                {
+                    await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
gregg-miskelly, line 24: ThreadHelper.ThrowIfNotOnUIThread();

src/SSHDebugPS/VsOutputWindowWrapper.cs
```diff
@@ -0,0 +1,145 @@
...
+        }, LazyThreadSafetyMode.PublicationOnly);
+
+        private static Lazy<IVsUIShell> shellLazy = new Lazy<IVsUIShell>(() =>
+        {
+            return ThreadHelper.JoinableTaskFactory.Run(async delegate
+            {
+                IVsUIShell shell = null;
+                try
+                {
+                    await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
gregg-miskelly, line 43: Same

src/SSHDebugPS/VsOutputWindowWrapper.cs
```diff
@@ -0,0 +1,138 @@
...
+                    }
+
+                    // Output the text
+                    outputPane.OutputString(message);
+                }
+                catch (Exception)
+                {
+                    Debug.Fail("Failed to write to output pane.");
+                }
+            }).FileAndForget("vs/SSHDebugPS/VsOutputWindowWrapper/Write");
```
gregg-miskelly, line 127: >vs [](start = 30, length = 2)

I think you want to use the same namespace as the rest of our telemetry - "VS/Diagnostics/Debugger/SSHDebugPS/VsOutputWindowWrapper/Write"

PR #1021, opened 2020-07-20
Added Platform field to the container header and expanded view in the "Select Docker Container" window

gc46:
**This enhancement assists the user in selecting the correct Connection target for the chosen Connection type.**

**Before**

**After**

### DIFF
--- a/src/SSHDebugPS/Docker/DockerConnection.cs
+++ b/src/SSHDebugPS/Docker/DockerConnection.cs
@@ -2,10 +2,12 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Text.RegularExpressions;
 using System.Threading;
+using Microsoft.DebugEngineHost;
 using Microsoft.SSHDebugPS.Utilities;
 using Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier;
 
@@ -243,10 +245,14 @@ namespace Microsoft.SSHDebugPS.Docker
                 return new RemoteCommandRunner(settings, OuterConnection);
         }
 
+        private static readonly string IncorrectContainerType = "IncorrectContainerType";
         protected override string ProcFSErrorMessage
         {
             get
             {
+                HostTelemetry.SendEvent(Telemetry.Event_DockerConnectionTypesNotMatching, new KeyValuePair<string, object>[] {
+                    new KeyValuePair<string, object>(IncorrectContainerType, 1)
+                });
                 return String.Concat(base.ProcFSErrorMessage, Environment.NewLine, StringResources.Error_EnsureDockerContainerIsLinux);
             }
         }
--- a/src/SSHDebugPS/Docker/DockerConnection.cs
+++ b/src/SSHDebugPS/Docker/DockerConnection.cs
@@ -245,14 +245,11 @@ namespace Microsoft.SSHDebugPS.Docker
                 return new RemoteCommandRunner(settings, OuterConnection);
         }
 
-        private static readonly string IncorrectContainerType = "IncorrectContainerType";
         protected override string ProcFSErrorMessage
         {
             get
             {
-                HostTelemetry.SendEvent(Telemetry.Event_DockerConnectionTypesNotMatching, new KeyValuePair<string, object>[] {
-                    new KeyValuePair<string, object>(IncorrectContainerType, 1)
-                });
+                HostTelemetry.SendEvent(TelemetryHelper.Event_ProcFSError);
                 return String.Concat(base.ProcFSErrorMessage, Environment.NewLine, StringResources.Error_EnsureDockerContainerIsLinux);
             }
         }
--- a/src/SSHDebugPS/Docker/DockerHelper.cs
+++ b/src/SSHDebugPS/Docker/DockerHelper.cs
@@ -17,18 +17,13 @@ namespace Microsoft.SSHDebugPS.Docker
         private const string dockerPSCommand = "ps";
         // --no-trunc avoids parameter truncation
         private const string dockerPSArgs = "-f status=running --no-trunc --format \"{{json .}}\"";
-
         private const string dockerInfoCommand = "info";
-        private const string dockerInfoArgs = "-f {{.Driver}}"; // lcow check
+        private const string dockerInfoArgs = "-f {{.Driver}}";
         private const string dockerVersionCommand = "version";
-        private const string dockerVersionArgs = "-f {{.Server.Os}}"; // server OS
+        private const string dockerVersionArgs = "-f {{.Server.Os}}";
         private const string dockerInspectCommand = "inspect";
-        private const string dockerInspectArgs = "-f \"{{json .Platform}}\" "; // container platform
-
+        private const string dockerInspectArgs = "-f \"{{json .Platform}}\" ";
         private static char[] charsToTrim = { ' ', '\"' };
-
-        // test -- need to delete
-        // make docker calls here
 
         public static bool LCOW(string hostname, out bool lcow)
         {
@@ -138,7 +133,6 @@ namespace Microsoft.SSHDebugPS.Docker
                 return false;
             }
         } 
-
 
         public static IEnumerable<DockerContainerInstance> GetLocalDockerContainers(string hostname, out int totalContainers)
         {
--- a/src/SSHDebugPS/Docker/DockerHelper.cs
+++ b/src/SSHDebugPS/Docker/DockerHelper.cs
@@ -19,13 +19,23 @@ namespace Microsoft.SSHDebugPS.Docker
         private const string dockerPSArgs = "-f status=running --no-trunc --format \"{{json .}}\"";
         private const string dockerInspectCommand = "inspect";
         private const string dockerInspectArgs = "-f \"{{json .Platform}}\" ";
+        private const string dockerVersionCommand = "version";
+        private const string dockerVersionArgs = "-f {{.Server.Os}}";
 
-        public static string GetContainerPlatform(string hostname, string containerName)
+        public static string GetContainerPlatform(string hostname, string containerName, bool inspect)
         {
             string containerPlatform = string.Empty;
 
             DockerCommandSettings settings = new DockerCommandSettings(hostname, false);
-            settings.SetCommand(dockerInspectCommand, string.Concat(dockerInspectArgs, containerName));
+
+            if (inspect)
+            {
+                settings.SetCommand(dockerInspectCommand, string.Concat(dockerInspectArgs, containerName));
+            }
+            else
+            {
+                settings.SetCommand(dockerVersionCommand, dockerVersionArgs);
+            }
 
             LocalCommandRunner commandRunner = new LocalCommandRunner(settings);
 
@@ -54,7 +64,8 @@ namespace Microsoft.SSHDebugPS.Docker
                 {
                     if (!string.IsNullOrWhiteSpace(args))
                     {
-                        containerPlatform = args.Trim().Replace("\"", "").Equals("windows") ? "Windows" : "Linux";
+                        string output = args.Trim().Replace("\"", "");
+                        containerPlatform = output.Equals("windows") ? "Windows" : output.Equals("linux") ? "Linux" : "Unknown";
                     }
                 });
 
--- a/src/SSHDebugPS/Docker/DockerHelper.cs
+++ b/src/SSHDebugPS/Docker/DockerHelper.cs
@@ -25,11 +25,8 @@ namespace Microsoft.SSHDebugPS.Docker
         private const string dockerInspectArgs = "-f \"{{json .Platform}}\" ";
         private static char[] charsToTrim = { ' ', '\"' };
 
-        // TO-DO: refactor code
         private static bool TryRunDockerCommand(DockerCommandSettings settings, Action <string> callback)
         {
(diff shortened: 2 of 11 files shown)


src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -17,6 +17,87 @@ public class DockerHelper
         private const string dockerPSCommand = "ps";
         // --no-trunc avoids parameter truncation
         private const string dockerPSArgs = "-f status=running --no-trunc --format \"{{json .}}\"";
+        private const string dockerInspectCommand = "inspect";
+        private const string dockerInspectArgs = "-f \"{{json .Platform}}\" ";
+
+        public static string GetContainerPlatform(string hostname, string containerName)
```
WardenGnaw, line 23: I would convert this to `bool TryGetContainerPlatform(string hostname, string containerName, out string platform)`

src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
```diff
@@ -162,6 +162,11 @@ private void RefreshContainersListInternal()
                     containers = DockerHelper.GetRemoteDockerContainers(connection, Hostname, out totalContainers);
                 }
 
+                foreach(DockerContainerInstance container in containers)
+                {
+                    container.Platform = DockerHelper.GetContainerPlatform(Hostname, container.Name);
```
WardenGnaw, line 167: Are we throwing all the containers away if we cant get a platform? What happened to going back to `docker info -f "{{.Server.OS}}"`?

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -17,6 +17,87 @@ public class DockerHelper
...
+                    { }
+                    cancellationRequested = cancellationToken.IsCancellationRequested;
+                });
+
+                if (!cancellationRequested)
+                {
+                    // might need to throw an exception here too??
+                    if (exitCode.GetValueOrDefault(-1) != 0)
+                    {
+                        // if the exit code is not zero, then the output we received possibly is the error message
```
WardenGnaw, line 76: Do we want to handle Telemetry here to see what invalid messages we get?

```
HostTelemetry.SendEvent(Telemetry.Event.Error_DockerPlatform, new KeyValuePair<string, object>[] {
      new KeyValuePair<string, object>(DockerPlatform_Error, errorSB.ToString())
}); 
```

Add `Error_DockerPlatform` in https://github.com/microsoft/MIEngine/blob/a95f162a88070c323c4976146a1ae16250c4270f/src/SSHDebugPS/Utilities/Telemetry.cs#L8

src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
pieandcakes: also, we shouldn't throw them away. we should support an "Unknown" option in that field

pieandcakes:
After looking at the after screenshot, can you move the platform to the top of the list?

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -17,6 +17,87 @@ public class DockerHelper
...
+                {
+                    exitCode = args;
+                    resetEvent.Set();
+                });
+
+                commandRunner.OutputReceived += ((sender, args) =>
+                {
+                    if (!string.IsNullOrWhiteSpace(args))
+                    {
+                        containerPlatform = args.Trim().Replace("\"", "").Equals("windows") ? "Windows" : "Linux";
```
caslan, line 57: Are we using these string later on for equality comparison etc.? If that's not the case and if this line is for converting the resulting string to title case, you can just use ToTitleCase method for that purpose. If the platform string changes slightly later on for whatever reason, the code will continue to work.

pieandcakes, line 57: You should use `Trim([' ', '\"'])` to remove all leading and trailing spaces and quotes.

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -17,6 +17,87 @@ public class DockerHelper
...
+
+            DockerCommandSettings settings = new DockerCommandSettings(hostname, false);
+            settings.SetCommand(dockerInspectCommand, string.Concat(dockerInspectArgs, containerName));
+
+            LocalCommandRunner commandRunner = new LocalCommandRunner(settings);
+
+            StringBuilder errorSB = new StringBuilder();
+            int? exitCode = null;
+
+            try
```
pieandcakes, line 35: I would suggest refactoring this and the below item as a lot of this code is duplicated.

src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
pieandcakes: maybe think about an enum here that has Unknown, Windows and Linux. This way if you have to do comparisons later with it, you aren't doing string comparisons.

src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
```diff
@@ -162,6 +162,26 @@ private void RefreshContainersListInternal()
                     containers = DockerHelper.GetRemoteDockerContainers(connection, Hostname, out totalContainers);
                 }
 
+                foreach(DockerContainerInstance container in containers)
+                {
+                    try
+                    {
+                        container.Platform = DockerHelper.GetContainerPlatform(Hostname, container.Name, false);
+                        try
```
pieandcakes, line 170: please remove the inner tryblock. That is not necessary. Also, why is this a blanket catch statement?

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -17,6 +17,123 @@ public class DockerHelper
...
         private const string dockerPSArgs = "-f status=running --no-trunc --format \"{{json .}}\"";
+        private const string dockerInfoCommand = "info";
+        private const string dockerInfoArgs = "-f {{.Driver}}";
+        private const string dockerVersionCommand = "version";
+        private const string dockerVersionArgs = "-f {{.Server.Os}}";
+        private const string dockerInspectCommand = "inspect";
+        private const string dockerInspectArgs = "-f \"{{json .Platform}}\" ";
+        private static char[] charsToTrim = { ' ', '\"' };
+
+        public static bool LCOW(string hostname, out bool lcow)
```
WardenGnaw, line 28: nit: Rename to TryGetLCOW and add comments above what it is doing.
"Querying the docker daemon to see if it supports Linux Containers on Windows (LCOW)"

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -17,6 +17,123 @@ public class DockerHelper
...
+        {
+            string driverInfo = string.Empty;
+            bool temp = false;
+            DockerCommandSettings settings = new DockerCommandSettings(hostname, false);
+            settings.SetCommand(dockerInfoCommand, dockerInfoArgs);
+            bool status = TryGetDockerCommandOutput(settings, out driverInfo);
+            if (driverInfo.Contains("lcow"))
```
WardenGnaw, line 35: `private const string LCOW = "lcow"`

Remove temp and simplify to `lcow = driverInfo.Contains(LCOW)`

src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
```diff
@@ -162,6 +165,41 @@ private void RefreshContainersListInternal()
                     containers = DockerHelper.GetRemoteDockerContainers(connection, Hostname, out totalContainers);
                 }
 
+                string serverOS;
+                bool getServerOS = DockerHelper.TryGetServerOS(Hostname, out serverOS);
+
+                if (getServerOS)
```
WardenGnaw, line 171: remove `getServerOs` and change to `if (DockerHelper.TryGetServerOS(Hostname, out serverOS))`

src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
```diff
@@ -162,6 +165,41 @@ private void RefreshContainersListInternal()
...
+                    bool getLCOW = DockerHelper.LCOW(Hostname, out lcow);
+                    TextInfo textInfo = new CultureInfo("en-US", false).TextInfo;
+                    serverOS = textInfo.ToTitleCase(serverOS);
+
+                    if (lcow && serverOS.Contains("Windows"))
+                    {
+                        foreach (DockerContainerInstance container in containers)
+                        {
+                            string containerPlatform = string.Empty;
+                            DockerHelper.TryGetContainerPlatform(Hostname, container.Name, out containerPlatform);
```
WardenGnaw, line 183: Make sure to validate that the docker command executed properly.

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -53,63 +55,142 @@ public static IEnumerable<DockerContainerInstance> GetLocalDockerContainers(stri
...
+                        args = args.Trim(charsToTrim);
+                        callback(args);
                     }
                 });
 
                 commandRunner.Start();
 
                 bool cancellationRequested = false;
-                VS.VSOperationWaiter.Wait(UIResources.QueryingForContainersMessage, false, (cancellationToken) =>
+                VS.VSOperationWaiter.Wait(UIResources.RunningDockerCommandMessage, false, (cancellationToken) =>
```
WardenGnaw, line 66: Can you pass in the correct resource to use? A generic `Running docker command...` gives no information to the user what is actually happening. 

Add an argument to this method so we can still say `Querying for Containers`. Then also have something like `Getting Docker Container Information...` and `Getting Docker Engine Version Information` for the other two commands.

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -243,10 +245,14 @@ private ICommandRunner GetCommandRunner(DockerContainerTransportSettings setting
                 return new RemoteCommandRunner(settings, OuterConnection);
         }
 
+        private static readonly string IncorrectContainerType = "IncorrectContainerType";
```
pieandcakes, line 248: if you are adding this for telemetry, please follow @WardenGnaw 's pattern and add it to the telemtry file and not in line with the code. Telemetry events and properties need to be prefaced correctly.

src/SSHDebugPS/Docker/DockerConnection.cs
```diff
@@ -243,10 +245,14 @@ private ICommandRunner GetCommandRunner(DockerContainerTransportSettings setting
...
         protected override string ProcFSErrorMessage
         {
             get
             {
+                HostTelemetry.SendEvent(Telemetry.Event_DockerConnectionTypesNotMatching, new KeyValuePair<string, object>[] {
```
pieandcakes, line 253: So I want to call out that "wrong container type" isn't the only way ProcFS fails, but just one of them so this telemetry point is misleading. It might be better to change the telemetry to just call it procfs fails.

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -53,63 +55,142 @@ public static IEnumerable<DockerContainerInstance> GetLocalDockerContainers(stri
...
             catch (Win32Exception ex)
             {
-                // docker doesn't exist 
                 string errorMessage = UIResources.CommandExecutionErrorFormat.FormatCurrentCultureWithArgs(settings.CommandArgs, ex.Message);
-                throw new CommandFailedException(errorMessage, ex);
+                return false;
+            }
+        }
+
+        internal static bool TryGetLCOW(string hostname, out bool lcow)
```
pieandcakes, line 103: please add function comments and explain what LCOW is.

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -53,63 +55,142 @@ public static IEnumerable<DockerContainerInstance> GetLocalDockerContainers(stri
...
                     cancellationRequested = cancellationToken.IsCancellationRequested;
                 });
 
                 if (!cancellationRequested)
                 {
-                    // might need to throw an exception here too??
                     if (exitCode.GetValueOrDefault(-1) != 0)
                     {
-                        // if the exit code is not zero, then the output we received possibly is the error message
                         string exceptionMessage = UIResources.CommandExecutionErrorWithExitCodeFormat.FormatCurrentCultureWithArgs(
```
pieandcakes, line 76: So I see this message being created, but since you removed the `throw` this isn't being used anywhere. There needs to be a way to bubble up the error message.

src/SSHDebugPS/Docker/DockerHelper.cs
```diff
@@ -53,63 +55,142 @@ public static IEnumerable<DockerContainerInstance> GetLocalDockerContainers(stri
...
+
+        internal static bool TryGetLCOW(string hostname, out bool lcow)
+        {
+            lcow = false;
+            bool delegateLCOW = false;
+
+            DockerCommandSettings settings = new DockerCommandSettings(hostname, false);
+            settings.SetCommand(dockerInfoCommand, dockerInfoArgs);
+
+            bool result = TryRunDockerCommand(settings, delegate (string args) {
```
pieandcakes, line 111: you can simplify this by doing 
```
if (tryRunDockerCommand(...))
{
    lcow = delegateLCOW;
return true;
}

return false;
```

PR #1025, opened 2020-08-04
Win: Fix encoding issue on ANSI code page 

imba-tjd:
The original change was introduced in https://github.com/microsoft/MIEngine/pull/942 and I believe it causes users who use ANSI code page can't input characters upper than 127 when using internal console because as I understand that code would set the encoding to utf8 regardless the original value.

For example:

```c
#include <stdio.h>
#include <stdlib.h>

int main() {
    //system("chcp 437"); // Sorry, won't work for internal console. Change the chcp in $profile instead.
    char s[10];
    gets(s); // input ü
    puts(s);
    system("pause");
}
```

Here is the result:

图片

图片

But I don't know how to build this project and cpptools, so I can't test it.

If it works, could you also test `chcp 936` and `你好`?
If the issue still exists, then there must be a change between cpptools 0.26.0 and 0.26.1 that caused this problem.
And if it works, `microsoft/vscode-cpptools#1527` also needs testing in case regression.

Squash is recommended.

Fix https://github.com/microsoft/vscode-cpptools/issues/5308
Fix https://github.com/microsoft/vscode-cpptools/issues/5325

### DIFF
--- a/src/OpenDebugAD7/OpenDebug/Program.cs
+++ b/src/OpenDebugAD7/OpenDebug/Program.cs
@@ -144,7 +144,7 @@ namespace OpenDebug
             {
                 // stdin/stdout
                 Console.Error.WriteLine("waiting for v8 protocol on stdin/stdout");
-                if (Utilities.IsWindows())
+                if (Utilities.IsWindows() && Console.InputEncoding.CodePage == 65001)
                 {
                     // Avoid sending the BOM on Windows if the Beta Unicode feature is enabled in Windows 10
                     Console.OutputEncoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
--- a/src/WindowsDebugLauncher/Program.cs
+++ b/src/WindowsDebugLauncher/Program.cs
@@ -20,8 +20,11 @@ namespace WindowsDebugLauncher
             parameters.PipeServer = "."; // Currently only supporting local pipe connections
 
             // Avoid sending the BOM on Windows if the Beta Unicode feature is enabled in Windows 10
-            Console.OutputEncoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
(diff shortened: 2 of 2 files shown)


imba-tjd:
Could you team review this? It's been a week.
Maybe @pieandcakes since you made the original change?

pieandcakes:
@WardenGnaw Can you take a look? 

@imba-tjd I don't think that this change will do what is expected.

imba-tjd:
If this doesn't work, I think a issue should be filed in cpptools repo.

In Summary:

* External console doesn't have this issue, internal console has
* 0.26.0 doesn't have this issue, 0.26.1 and later have

imba-tjd:
Fine. I built this with [GitHub Actions](https://github.com/imba-tjd/MIEngine/blob/build/.github/workflows/build.yml), basically by `msbuild -v:m -r -p:Configuration=Desktop.Release`.
And I replaced `C:\Users\xxx\.vscode\extensions\ms-vscode.cpptools-0.29.0\debugAdapters\bin\WindowsDebugLauncher.exe` with `bin/Desktop.Release/vscode/WindowsDebugLauncher.exe`.
And it fixed my issue.
One problem is that if I replace `OpenDebugAD7.exe` with the file built my me, the debug won't start.

I also found that `system("chcp 437")` won't work in internal console. PowerShell's `$profile` is a way to change it.

Fortunately I found another evidence that your code would change the chcp regardless the original value:

图片

returnL (commented):
Can this problem be solved quickly?

imba-tjd:
Do You mean this won't be fixed?

WardenGnaw:
This will need to be re-targeted to the `main` branch.

imba-tjd:
I'm not going to reopen or open another PR. You team can do the change as you wish.

WardenGnaw:
Added an issue to keep track of problem.

thefatwhale:
@imba-tjd .Thank you for your code, it solved my problem.

PR #1028, opened 2020-08-14
Support HandleReadMemoryRequestAsync

WardenGnaw:
This allows OpenDebugAD7 to support the ReadMemoryRequest for DAP.

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -1372,7 +1372,8 @@ namespace OpenDebugAD7
                                     Dictionary<string, Variable> variablesDictionary = new Dictionary<string, Variable>();
                                     for (uint c = 0; c < count; c++)
                                     {
-                                        var variable = m_variableManager.CreateVariable(ref childProperties[c], variableEvaluationData.propertyInfoFlags);
+                                        string memoryReference = AD7Utils.GetMemoryReferenceFromIDebugProperty(childProperties[c].pProperty);
+                                        var variable = m_variableManager.CreateVariable(ref childProperties[c], variableEvaluationData.propertyInfoFlags, memoryReference);
                                         int uniqueCounter = 2;
                                         string variableName = variable.Name;
                                         string variableNameFormat = "{0} #{1}";
@@ -1389,8 +1390,9 @@ namespace OpenDebugAD7
                                 }
                                 else
                                 {
+                                    string memoryReference = AD7Utils.GetMemoryReferenceFromIDebugProperty(childProperties[0].pProperty);
                                     // Shortcut when no duplicate can exist
-                                    response.Variables.Add(m_variableManager.CreateVariable(ref childProperties[0], variableEvaluationData.propertyInfoFlags));
+                                    response.Variables.Add(m_variableManager.CreateVariable(ref childProperties[0], variableEvaluationData.propertyInfoFlags, memoryReference));
                                 }
                             }
                         }
@@ -2017,20 +2019,9 @@ namespace OpenDebugAD7
                 return;
             }
 
-            string memoryReference = null;
-            if (property.GetMemoryContext(out IDebugMemoryContext2 memoryContext) == HRConstants.S_OK)
-            {
-                CONTEXT_INFO[] contextInfo = new CONTEXT_INFO[1];
-                if (memoryContext.GetInfo(enum_CONTEXT_INFO_FIELDS.CIF_ADDRESS, contextInfo) == HRConstants.S_OK)
-                {
-                    if (contextInfo[0].dwFields.HasFlag(enum_CONTEXT_INFO_FIELDS.CIF_ADDRESS))
-                    {
-                        memoryReference = contextInfo[0].bstrAddress;
-                    }
-                }
-            }
+            string memoryReference = AD7Utils.GetMemoryReferenceFromIDebugProperty(property);
 
-            Variable variable = m_variableManager.CreateVariable(ref propertyInfo[0], propertyInfoFlags);
+            Variable variable = m_variableManager.CreateVariable(ref propertyInfo[0], propertyInfoFlags, memoryReference);
 
             if (context != EvaluateArguments.ContextValue.Hover)
             {
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -2050,7 +2050,7 @@ namespace OpenDebugAD7
             {
                 if (string.IsNullOrEmpty(rma.MemoryReference))
                 {
-                    throw new ArgumentNullException("ReadMemoryArguments.MemoryReference is null or empty.");
+                    throw new ArgumentException("ReadMemoryArguments.MemoryReference is null or empty.");
                 }
 
                 ulong address;
@@ -2075,26 +2075,27 @@ namespace OpenDebugAD7
                     }
                 }
 
-                hr = ((IDebugMemoryBytesDAP)m_engine).CreateMemoryContext(address, out IDebugMemoryContext2 ppMemory);
+                hr = ((IDebugMemoryBytesDAP)m_engine).CreateMemoryContext(address, out IDebugMemoryContext2 memoryContext);
                 eb.CheckHR(hr);
 
                 byte[] data = new byte[rma.Count];
-                uint pdwUnreadable = 0;
+                uint unreadableBytes = 0;
+                uint bytesRead = 0;
 
                 if (rma.Count != 0)
                 {
                     hr = m_program.GetMemoryBytes(out IDebugMemoryBytes2 debugMemoryBytes);
                     eb.CheckHR(hr);
 
-                    hr = debugMemoryBytes.ReadAt(ppMemory, (uint)rma.Count, data, out uint pdwRead, ref pdwUnreadable);
+                    hr = debugMemoryBytes.ReadAt(memoryContext, (uint)rma.Count, data, out bytesRead, ref unreadableBytes);
                     eb.CheckHR(hr);
                 }
 
                 responder.SetResponse(new ReadMemoryResponse()
                 {
                     Address = string.Format(CultureInfo.InvariantCulture, "0x{0:X}", address),
-                    Data = Convert.ToBase64String(data),
-                    UnreadableBytes = (int?)pdwUnreadable
+                    Data = Convert.ToBase64String(data, 0, (int)bytesRead),
+                    UnreadableBytes = (int?)unreadableBytes
                 });
             }
             catch (Exception e)
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -619,7 +619,8 @@ namespace OpenDebugAD7
                 SupportsConditionalBreakpoints = m_engineConfiguration.ConditionalBP,
                 ExceptionBreakpointFilters = m_engineConfiguration.ExceptionSettings.ExceptionBreakpointFilters.Select(item => new ExceptionBreakpointsFilter() { Default = item.@default, Filter = item.filter, Label = item.label }).ToList(),
                 SupportsClipboardContext = m_engineConfiguration.ClipboardContext,
-                SupportsLogPoints = true
+                SupportsLogPoints = true,
+                SupportsReadMemoryRequest = true
             };
 
             responder.SetResponse(initializeResponse);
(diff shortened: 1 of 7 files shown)


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -263,6 +263,58 @@ private IList<Tracepoint> GetTracepoints(IDebugBreakpointEvent2 debugEvent)
...
+            IDebugExpressionContext2 expressionContext;
+            hr = frame.GetExpressionContext(out expressionContext);
+            if (hr >= 0)
+            {
+                IDebugExpression2 expressionObject;
+                uint errorIndex;
+                hr = expressionContext.ParseText(expression, enum_PARSEFLAGS.PARSE_EXPRESSION, Constants.ParseRadix, out expressionObject, out error, out errorIndex);
+                if (!string.IsNullOrEmpty(error))
+                {
+                    // TODO: Is this how errors should be returned?
```
gregg-miskelly, line 291: >// TODO: Is this how errors should be returned? [](start = 20, length = 47)

Looks correct to me

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2029,6 +2059,65 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
...
+            ErrorBuilder eb = new ErrorBuilder(() => AD7Resources.Error_Scenario_ReadMemory);
+            try
+            {
+                IDebugStackFrame2 frame;
+                m_frameHandles.TryGetFirst(out frame);
+                eb.CheckOutput(frame);
+
+                IDebugProperty2 property;
+                string error;
+                hr = GetIDebugPropertyFromExpression(rma.MemoryReference, frame, 0, DAPEvalFlags.NONE, out property, out error);
```
gregg-miskelly, line 2075: >MemoryReference [](start = 57, length = 15)

This isn't really how the protocol is supposed to work. You are supposed to return `memoryReference` values from expressions, and this should be a format that you don't need to parse (ex: just the address string).

src/OpenDebugAD7/AD7DebugSession.cs
pieandcakes: I think this is copied from HandleEvaluateRequestAsync. you will want to use a different error code here so you know its coming through this path and not the previous.

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: BTW you might want to first test your changes by running against Visual Studio as it already fully supports this part of the protocol.

WardenGnaw: Do we send the request in CMake Debug if we have the Memory Window Open?

gregg-miskelly: Can you tell me more about the plan for how we are going to add a memory window to VS Code? If I understand correctly, the idea is that it isn't going to really be part of VS Code, and instead we are going to somehow request memory from the adapter, save the bytes to a file, and then open up a hex editor on the bytes. But a few things I don't understand:
* How is the user going to enter the expression they want evaluated?
* How does the user specify how many bytes they want?
* How does the memory get refreshed after a step (or any other form of exiting and reentering break state)?

gregg-miskelly: In case it is useful, how I can imagine a memory window should use this feature -

1. UI sends an `evaluate` request from the user's expression, and waits for the result. If the result doesn't have a `memoryReference` - show user an error.
    A. Note: since in this case since the UI is part of the extension, maybe the UI could cheat if the expression to evaluate was just an address string and skip this step. But make sure to implement the 'real way' first so that things work in a compliant client.
2. Send a `readMemory` request with the `memoryReference` from the previous value.

WardenGnaw: 1. How is the user going to enter the expression they want evaluated?

The current idea was to make it a command. Users will use `ctrl+shift+p` and there will be a quick input where they can either type an address or a variable. Then the next input it will ask the offset then the size.

Example: https://github.com/haneefdm/cortex-debug/blob/e3faed3353e515f9802ea66fa9980043118344fc/src/frontend/extension.ts#L257

2. How does the user specify how many bytes they want?

Only interface with VS Code we can use is the Quick Input as a command.

3. How does the memory get refreshed after a step (or any other form of exiting and reentering break state)?

There will be a cache of the last request they made and on a stopping event it will refresh the page. We trace the requests using the DebugAdapterTracker.

gregg-miskelly: Thanks. In that case the description above should be correct.

WardenGnaw: Note: I'll be working on the VS Code UI before checking this in.

gregg-miskelly:
>     internal class AD7DebugSession : DebugAdapterBase, IDebugPortNotify2, IDebugEventCallback2

Need to also send a `ProcessEvent` with `PointerSize` defined

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:30 in dbd0fe3. [](commit_id = dbd0fe3c5333608c1e2982f50881076f4b8ac874, deletion_comment = False)

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2026,7 +2027,67 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
                 Type = variable.Type,
                 VariablesReference = variable.VariablesReference
             });
```
gregg-miskelly, line 2029: >}); [](start = 12, length = 3)

Set memoryReference

gregg-miskelly:
>     {

VariableManager.CreateVariable needs to populate the `memoryReference`

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:31 in dbd0fe3. [](commit_id = dbd0fe3c5333608c1e2982f50881076f4b8ac874, deletion_comment = False)

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2026,7 +2027,67 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
...
+        protected override void HandleReadMemoryRequestAsync(IRequestResponder<ReadMemoryArguments, ReadMemoryResponse> responder)
+        {
+            int hr;
+            ReadMemoryArguments rma = responder.Arguments;
+            ErrorBuilder eb = new ErrorBuilder(() => AD7Resources.Error_Scenario_ReadMemory);
+            try
+            {
+                if (string.IsNullOrEmpty(rma.MemoryReference))
+                {
+                    throw new ArgumentNullException("ReadMemoryArguments.MemoryReference is null or empty.");
```
gregg-miskelly, line 2042: >"ReadMemoryArguments.MemoryReference is null or empty." [](start = 52, length = 55)

The argument to `ArgumentNullExecption`'s contructor is a parameter name. So you probably want `ArgumentException` instead.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2026,7 +2027,67 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
...
+                {
+                    address = Convert.ToUInt64(rma.MemoryReference, 10);
+                }
+
+                hr = ((IDebugMemoryBytesDAP)m_engine).CreateMemoryContext(address, out IDebugMemoryContext2 ppMemory);
+                eb.CheckHR(hr);
+
+                if (rma.Offset.HasValue && rma.Offset.Value != 0)
+                {
+                    ulong offset = Convert.ToUInt64(Math.Abs(rma.Offset.Value));
```
gregg-miskelly, line 2060: >Convert.ToUInt64(Math.Abs(rma.Offset.Value)); [](start = 35, length = 45)

Any reason to not just cast? (ex: `.Subtract((ulong)(-rma.Offset.Value)`)

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2026,7 +2027,67 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
...
+                uint pdwUnreadable = 0;
+                hr = m_program.GetMemoryBytes(out IDebugMemoryBytes2 debugMemoryBytes);
+                eb.CheckHR(hr);
+
+                hr = debugMemoryBytes.ReadAt(ppMemory, (uint)rma.Count, data, out uint pdwRead, ref pdwUnreadable);
+                eb.CheckHR(hr);
+
+                responder.SetResponse(new ReadMemoryResponse()
+                {
+                    Address = rma.MemoryReference,
```
gregg-miskelly, line 2082: >rma.MemoryReference [](start = 30, length = 19)

This is wrong. This is supposed to take into account the offset. You need to either ask the memory context for its address or change the code to do the pointer arithmetic first rather than calling `.Add`/`.Subtract`.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: I was trying to reuse offset for both the additon and subtract case.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2026,7 +2027,67 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
...
+                    eb.CheckHR(hr);
+                }
+
+                byte[] data = new byte[rma.Count];
+                uint pdwUnreadable = 0;
+                hr = m_program.GetMemoryBytes(out IDebugMemoryBytes2 debugMemoryBytes);
+                eb.CheckHR(hr);
+
+                hr = debugMemoryBytes.ReadAt(ppMemory, (uint)rma.Count, data, out uint pdwRead, ref pdwUnreadable);
+                eb.CheckHR(hr);
```
gregg-miskelly, line 2078: VS will sometimes request zero-byte reads just to convert an address+offset into an address. Might be worth an if here to skip this part of `rma.Count` is zero

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: I would find the casts simpler, but up to you...

WardenGnaw:
Converting to Draft. I need to finish up the messaging on the VS Code side to fully test this.

WardenGnaw:
@gregg-miskelly Do we want to add memoryReference to all variables created by `VariableManager.CreateVariable`?

Used in:
[HandleVariablesRequestAsync](https://github.com/microsoft/MIEngine/blob/4c21dbdfc7fc7e0985d003830c1415ce97b41683/src/OpenDebugAD7/AD7DebugSession.cs#L1331)

which also calls
[VariablesFromFrame](https://github.com/microsoft/MIEngine/blob/4c21dbdfc7fc7e0985d003830c1415ce97b41683/src/OpenDebugAD7/AD7DebugSession.cs#L487)

gregg-miskelly:
> @gregg-miskelly Do we want to add memoryReference to all variables created by `VariableManager.CreateVariable`?

I think you would want to add them to any variable where GetMemoryContext succeeds.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2030,9 +2035,72 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
...
+                    eb.CheckHR(hr);
+
+                    hr = debugMemoryBytes.ReadAt(ppMemory, (uint)rma.Count, data, out uint pdwRead, ref pdwUnreadable);
+                    eb.CheckHR(hr);
+                }
+
+                responder.SetResponse(new ReadMemoryResponse()
+                {
+                    Address = string.Format(CultureInfo.InvariantCulture, "0x{0:X}", address),
+                    Data = Convert.ToBase64String(data),
```
gregg-miskelly, line 2096: >data [](start = 50, length = 4)

You need: `Convert.ToBase64String(data, 0, pdwRead)`

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2030,9 +2035,72 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
...
+                    {
+                        address += (ulong)rma.Offset.Value;
+                    }
+                    else
+                    {
+                        address -= (ulong)-rma.Offset.Value;
+                    }
+                }
+
+                hr = ((IDebugMemoryBytesDAP)m_engine).CreateMemoryContext(address, out IDebugMemoryContext2 ppMemory);
```
gregg-miskelly, line 2078: >pp [](start = 108, length = 2)

nit: the Hungarian prefixes you are using don't make sense. Might want to remove them.

PR #1031, opened 2020-08-20
parse natvis format specifiers properly

### LINKED ISSUE #4: Format specifiers don't work with natvis
Does not support the format specifiers yet like ,su for unicode strings.  See https://msdn.microsoft.com/en-us/library/75w45ekt%28v=vs.140%29.aspx

Trass3r:
fixes #4

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -361,7 +361,7 @@ namespace Microsoft.MIDebugEngine
                 case "Xb":
                 case "hb":
                 case "Hb":
-                    formatSpecifier = "hexadecimal";
+                    formatSpecifier = "zero-hexadecimal";
                     goto case "";
                 case "o":
                     formatSpecifier = "octal";
@@ -375,7 +375,6 @@ namespace Microsoft.MIDebugEngine
                     goto case "";
                 case "e":
                 case "g":
-                    //formatSpecifier = "natural";
                     goto case "";
                 case "s":
                 case "sb":
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -343,23 +343,70 @@ namespace Microsoft.MIDebugEngine
 
         private string StripFormatSpecifier(string exp, out string formatSpecifier)
         {
-            formatSpecifier = null;
+            formatSpecifier = null; // will be used with -var-set-format
             int lastComma = exp.LastIndexOf(',');
-            if (lastComma > 0)
+            if (lastComma <= 0)
+                return exp;
+
+            // https://docs.microsoft.com/en-us/visualstudio/debugger/format-specifiers-in-cpp
+            string expFS = exp.Substring(lastComma + 1);
+            string trimmed = expFS.Trim();
+            switch (trimmed)
             {
-                string expFS = exp.Substring(lastComma + 1);
-                string trimmed = expFS.Trim();
-                if (trimmed == "x" || trimmed == "X" || trimmed == "h" || trimmed == "H")
-                {
+                case "x":
+                case "X":
+                case "h":
+                case "H":
+                case "xb":
+                case "Xb":
+                case "hb":
+                case "Hb":
                     formatSpecifier = "hexadecimal";
-                    return exp.Substring(0, lastComma);
-                }
-                else if (trimmed == "o")
-                {
+                    goto case "";
+                case "o":
                     formatSpecifier = "octal";
+                    goto case "";
+                case "d":
+                    formatSpecifier = "decimal";
+                    goto case "";
+                case "b":
+                case "bb":
+                    formatSpecifier = "binary";
+                    goto case "";
+                case "e":
+                case "g":
+                    //formatSpecifier = "natural";
+                    goto case "";
+                case "s":
+                case "sb":
+                case "s8":
+                case "s8b":
+                    return "(const char*)(" + exp.Substring(0, lastComma) + ")";
+                case "su":
+                case "sub":
+                    return "(const char16_t*)(" + exp.Substring(0, lastComma) + ")";
+                case "c":
+                    return "(char)(" + exp.Substring(0, lastComma) + ")";
+                // just remove and ignore these
+                case "en":
+                case "na":
+                case "nd":
+                case "nr":
+                case "!":
+                case "":
                     return exp.Substring(0, lastComma);
-                }
             }
+
+            // array with static size
+            // TODO: could return '(T(*)[n])(exp)' but requires T
+            var m = Regex.Match(trimmed, @"^\[?(\d+)\]?$");
+            if (m.Success)
+                return exp.Substring(0, lastComma);
+
+            // array with dynamic size
+            if (Regex.Match(trimmed, @"^\[([a-zA-Z_][a-zA-Z_\d]*)\]$").Success)
+                return exp.Substring(0, lastComma);
+
             return exp;
         }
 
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -343,23 +343,70 @@ public enum NodeType
...
+            if (lastComma <= 0)
+                return exp;
+
+            string expFS = exp.Substring(lastComma + 1);
+            string trimmed = expFS.Trim();
+            switch (trimmed)
+            {
+                case "x":
+                case "X":
+                case "h":
```
pieandcakes, line 357: would it make sense to call trimmed.toLower() instead? I assume `o` and `O` should be the same specifier. this would get rid of some duplicates?

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -343,23 +343,70 @@ public enum NodeType
...
-            if (lastComma > 0)
-            {
-                string expFS = exp.Substring(lastComma + 1);
-                string trimmed = expFS.Trim();
-                if (trimmed == "x" || trimmed == "X" || trimmed == "h" || trimmed == "H")
-                {
+            if (lastComma <= 0)
+                return exp;
+
+            string expFS = exp.Substring(lastComma + 1);
```
pieandcakes, line 351: need to check length such that the expression doesn't end in a comma. this will add +1 and go beyond the length of `exp`.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -343,23 +343,70 @@ public enum NodeType
...
+                    return "(const char16_t*)(" + exp.Substring(0, lastComma) + ")";
+                case "c":
+                    return "(char)(" + exp.Substring(0, lastComma) + ")";
+                // just remove and ignore these
+                case "en":
+                case "na":
+                case "nd":
+                case "nr":
+                case "!":
+                case "":
```
pieandcakes, line 396: use `string.Empty` to represent `""` here and above.

pieandcakes:
Thanks for the PR. Can you tell us which debugger version and distros you tried this on so that we can validate the testing matrix?

@WardenGnaw TODO: add tests for these format specifiers

src/MIDebugEngine/Engine.Impl/Variables.cs
Trass3r: It will return "" in that case: https://docs.microsoft.com/en-us/dotnet/api/system.string.substring?view=netcore-3.1#System_String_Substring_System_Int32_

Trass3r:
> I don't see some of the ones you have on here. Do you have a reference document that specifies some of these?

https://sourceware.org/gdb/current/onlinedocs/gdb/GDB_002fMI-Variable-Objects.html#The-_002dvar_002dset_002dformat-Command

> Thanks for the PR. Can you tell us which debugger version and distros you tried this on so that we can validate the testing matrix?

gdb 9/10 on WSL2 Ubuntu 18.04/20.04.

src/MIDebugEngine/Engine.Impl/Variables.cs
Trass3r: I actually had `string trimmed = expFS.ToLowerInvariant().Trim();` before but removed it because it might cause problems with later usages of `trimmed` and the spec only defines uppercase chars for hex.

Trass3r: `switch (trimmed.ToLowerInvariant())` is an option but I don't know if VS accepts uppercase for other specifiers.
On the other hand they seem to allow weird constructs like

https://github.com/inviwo/inviwo/blob/2c7284bdda7e001578c9b39fb508ca19113a7de3/tools/natvis/qt5.natvis#L234

src/MIDebugEngine/Engine.Impl/Variables.cs
Trass3r: `error CS0150: A constant value is expected`

pieandcakes: ugh forgot about that.

pieandcakes:
> > I don't see some of the ones you have on here. Do you have a reference document that specifies some of these?
> 
> https://docs.microsoft.com/en-us/visualstudio/debugger/format-specifiers-in-cpp
> https://sourceware.org/gdb/current/onlinedocs/gdb/GDB_002fMI-Variable-Objects.html#The-_002dvar_002dset_002dformat-Command
> 
> > Thanks for the PR. Can you tell us which debugger version and distros you tried this on so that we can validate the testing matrix?
> 
> gdb 9/10 on WSL2 Ubuntu 18.04/20.04.

The problem with https://docs.microsoft.com/en-us/visualstudio/debugger/format-specifiers-in-cpp?view=vs-2019 is that it is specifically for the Visual Studio Debugger (`cppvsdbg`) and not for this. I don't mind it being added in this case but that document is targeting something else.

pieandcakes:
@WardenGnaw Assigning to you. I'm approving it but I'll leave it up to you to determine when to commit this change.

Trass3r:
> The problem with https://docs.microsoft.com/en-us/visualstudio/debugger/format-specifiers-in-cpp?view=vs-2019 is that it is specifically for the Visual Studio Debugger (cppvsdbg) and not for this. I don't mind it being added in this case but that document is targeting something else.

The reason for adding these is compatibility with natvis type definitions originally written for VS.
Like Qt visualizers for example.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -343,23 +343,69 @@ public enum NodeType
...
-                {
+            if (lastComma <= 0)
+                return exp;
+
+            string expFS = exp.Substring(lastComma + 1);
+            string trimmed = expFS.Trim();
+            switch (trimmed)
+            {
+                case "x":
+                case "X":
```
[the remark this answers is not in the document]
Trass3r: Yeah but is it worth it? In the end it's just displayed to the user.

Trass3r:
Done.

src/MIDebugEngine/Engine.Impl/Variables.cs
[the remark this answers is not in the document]
gregg-miskelly: I am not sure the casing stuff really matters. I don't know why the `x` vs. `X` stuff was added. Maybe just because programmers are picky :).

I have seen folks really care about the no-quotes stuff through...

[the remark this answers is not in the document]
Trass3r: Yeah to me it's only important that valid natvis doesn't get rejected.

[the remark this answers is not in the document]
Trass3r: Yeah to me it's only important that valid natvis doesn't get rejected.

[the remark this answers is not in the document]
gregg-miskelly: I guess if you don't want to implement it, it would be acceptable to just add a comment indicating that this isn't supported yet, and what a real implementation should do.

Trass3r:
Done.

gregg-miskelly (approved):

WardenGnaw:
@Trass3r Thank you for your contirubtion to MIEngine! 

We will target this for the next VSCode C++ Extension's insiders release.  

It will appear in the [CHANGELOG](https://github.com/microsoft/vscode-cpptools/blob/master/Extension/CHANGELOG.md) when it is avaliabe to use in the extension.

Trass3r:
> It will appear in the [CHANGELOG](https://github.com/microsoft/vscode-cpptools/blob/master/Extension/CHANGELOG.md) when it is avaliabe to use in the extension.

@WardenGnaw I still don't see it in the changelog. So I guess it hasn't been integrated yet?

WardenGnaw:
I must have missed this one when updating the change log.
Do you not see the change in the latest cpptools extension?

PR #1032, opened 2020-08-20
Support $PID in Tracepoint LogMessage

WardenGnaw:
Add support for $PID in LogMessages.

We retrieve the actual PID in IDebugProcessInfoUpdatedEvent158. If users try to use $PID before we get the event, it will return `<Unknown PID>`. 

Also updates m_processName from IDebugProcessInfoUpdatedEvent158 event

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -34,6 +34,7 @@ namespace OpenDebugAD7
 
         private IDebugProcess2 m_process;
         private string m_processName;
+        private int m_processId;
         private IDebugEngineLaunch2 m_engineLaunch;
         private IDebugEngine2 m_engine;
         private EngineConfiguration m_engineConfiguration;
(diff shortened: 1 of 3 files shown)


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -34,6 +34,7 @@ internal class AD7DebugSession : DebugAdapterBase, IDebugPortNotify2, IDebugEven
 
         private IDebugProcess2 m_process;
         private string m_processName;
+        private int m_processId;
```
gregg-miskelly, line 37: >m_processId [](start = 20, length = 11)

Initialize to invalid?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -960,6 +962,10 @@ protected override void HandleAttachRequestAsync(IRequestResponder<AttachArgumen
                     m_isAttach = true;
                 }
 
+                if (int.TryParse(processId, out m_processId))
```
gregg-miskelly, line 965: >, [](start = 42, length = 1)

`NumberStyles.None, CultureInfo.InvariantCulture`?

PR #1036, opened 2020-08-26
display function return values

### LINKED ISSUE #1027: Display function return value
https://github.com/microsoft/vscode-cpptools/issues/3214

PoC for gdb: https://github.com/Trass3r/MIEngine/commit/4fda0e9427bff88c0b0ae0cd11e8f2b5830c2b10

But this really only works when stepping out of a function (also no way to get the name of the callee).
The question is how to intercept other step requests and transform them based on the current line of code, e.g. step over function call => step in and out, step over a return statement => step out.
https://github.com/microsoft/MIEngine/blob/54b7d6b6dcc11d99297bcad8df765870f4e80da0/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs#L1621-L1643

Trass3r:
Tested on gdb 8/10.
Fixes  #1027.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -25,7 +25,7 @@ namespace Microsoft.MIDebugEngine
         public AD7Engine Engine { get; private set; }
         public List<string> VariablesToDelete { get; private set; }
         public List<IVariableInformation> ActiveVariables { get; private set; }
-
+        public VariableInformation ReturnValue { get; private set; }
         public SourceLineCache SourceLineCache { get; private set; }
         public ThreadCache ThreadCache { get; private set; }
         public Disassembly Disassembly { get; private set; }
@@ -1053,6 +1053,7 @@ namespace Microsoft.MIDebugEngine
                     varInfo.Dispose();
                 }
                 this.ActiveVariables.Clear();
+                ReturnValue = null; // already disposed above
             }
 
             ThreadCache.MarkDirty();
@@ -1235,8 +1236,19 @@ namespace Microsoft.MIDebugEngine
                     }
                 }
             }
-            else if (reason == "end-stepping-range" || reason == "function-finished")
+            // step over/into
+            // NB: unfortunately this event does not provide a return value: https://sourceware.org/bugzilla/show_bug.cgi?id=26354
+            else if (reason == "end-stepping-range")
+                _callback.OnStepComplete(thread);
+            // step out
+            else if (reason == "function-finished")
             {
+                string resultVar = results.Results.TryFindString("gdb-result-var"); // a gdb value history var like "$1"
+                if (!string.IsNullOrEmpty(resultVar))
+                {
+                    ReturnValue = new VariableInformation("$ReturnValue", resultVar, cxt, Engine, (AD7Thread)thread.Client, isParameter: false);
+                    await ReturnValue.Eval();
+                }
                 _callback.OnStepComplete(thread);
             }
             else if (reason == "signal-received")
@@ -1873,6 +1885,9 @@ namespace Microsoft.MIDebugEngine
                 VariableInformation vi = await simpleInfo.CreateMIDebuggerVariable(ctx, Engine, thread);
                 variables.Add(vi);
             }
+
+            if (ReturnValue != null && ctx.Level == 0 && ReturnValue.Client.Id == thread.Id)
+                variables.Add(ReturnValue);
 
             return variables;
         }
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -60,7 +60,7 @@ namespace Microsoft.MIDebugEngine
 
         internal async Task<VariableInformation> CreateMIDebuggerVariable(ThreadContext ctx, AD7Engine engine, AD7Thread thread)
         {
-            VariableInformation vi = new VariableInformation(Name, ctx, engine, thread, IsParameter);
+            VariableInformation vi = new VariableInformation(Name, Name, ctx, engine, thread, IsParameter);
             await vi.Eval();
             return vi;
         }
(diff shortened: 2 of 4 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1235,8 +1236,21 @@ private async Task HandleBreakModeEvent(ResultEventArgs results, BreakRequest br
...
                 }
             }
-            else if (reason == "end-stepping-range" || reason == "function-finished")
+            // step over/into
+            else if (reason == "end-stepping-range")
+                _callback.OnStepComplete(thread);
+            // step out
+            else if (reason == "function-finished")
             {
+                string resultVar = results.Results.TryFindString("gdb-result-var"); // a gdb value history var like "$1"
```
Trass3r, line 1245: Unfortunately we only get the name of the caller instead of the callee here.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -75,7 +75,7 @@ public ArgumentList(int level, List<SimpleVariableInformation> args)
 
     internal class VariableInformation : IVariableInformation
     {
-        public string Name { get; private set; }
+        public string Name { get; internal set; }
```
Trass3r, line 78: Was needed to set display name different from the expression.
Should probably change the constructors instead.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1646,6 +1646,12 @@ public async Task Step(int threadId, enum_STEPKIND kind, enum_STEPUNIT unit)
                         await MICommandFactory.ExecStep(threadId);
                         break;
                     case enum_STEPKIND.STEP_OVER:
+                        string line = await ConsoleCmdAsync("f", false, true);
+                        if (line.Contains("return "))
```
Trass3r, line 1650: Simple heuristic that works well in practice.
Unfortunately there's no easy way to catch stepping over function calls.

gregg-miskelly: I don't think I like this idea. It seems like it might be better to just tell users that if they step out they will get this feature.

If others disagree, I think at the least --
1. You want to change this to instead just checking if the line has "return" anywhere in it you should instead check: `^[\s\{]*return\s`. Otherwise stepping over `if (false) return 27;` would result in a step out.
2. This needs to be behind an off-by-default option

Trass3r: Yeah it was more of a test.
In the end it doubles the commands run for every step. This really needs gdb support.
https://sourceware.org/bugzilla/show_bug.cgi?id=26354

Trass3r:
@WardenGnaw cleaned it up.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1235,8 +1236,18 @@ private async Task HandleBreakModeEvent(ResultEventArgs results, BreakRequest br
...
+            // step over/into
+            else if (reason == "end-stepping-range")
+                _callback.OnStepComplete(thread);
+            // step out
+            else if (reason == "function-finished")
             {
+                string resultVar = results.Results.TryFindString("gdb-result-var"); // a gdb value history var like "$1"
+                if (!string.IsNullOrEmpty(resultVar))
+                {
+                    ReturnValue = new VariableInformation("Function return value", resultVar, cxt, Engine, (AD7Thread)thread.Client, isParameter: false);
```
WardenGnaw, line 1248: To keep in sync with what we have in VS, it should be `$ReturnValue`.

```suggestion
ReturnValue = new VariableInformation("$ReturnValue", resultVar, cxt, Engine, (AD7Thread)thread.Client, isParameter: false);
```

WardenGnaw (changes requested):
Will this only support return value on `step-out` until GDB supports getting a return value?

Trass3r:
Yes only works on step out.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1235,8 +1236,18 @@ private async Task HandleBreakModeEvent(ResultEventArgs results, BreakRequest br
                     }
                 }
             }
-            else if (reason == "end-stepping-range" || reason == "function-finished")
+            // step over/into
+            else if (reason == "end-stepping-range")
+                _callback.OnStepComplete(thread);
```
WardenGnaw, line 1241: Also would leave a comment here saying waiting for support for return value in GDB for step over and into. See https://sourceware.org/bugzilla/show_bug.cgi?id=26354

Trass3r:
Done. Yes only works on step out.

gregg-miskelly (approved):

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1874,6 +1886,9 @@ internal async Task<List<VariableInformation>> GetLocalsAndParameters(AD7Thread
                 variables.Add(vi);
             }
 
+            if (ReturnValue != null)
```
gregg-miskelly, line 1889: Wait, we are missing a check here: we only want to do this if we are on the top frame of the thread that stopped.

gregg-miskelly (changes requested):

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1874,6 +1886,9 @@ internal async Task<List<VariableInformation>> GetLocalsAndParameters(AD7Thread
                 variables.Add(vi);
             }
 
+            if (ReturnValue != null && ctx.Level == 0)
```
gregg-miskelly, line 1889: >) [](start = 53, length = 1)

Need to also check the thread

gregg-miskelly (approved):

WardenGnaw (approved):

WardenGnaw:
@Trass3r Thank you again for your contribution!

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
Edord99: Nu11

PR #1037, opened 2020-08-26
implement reverse execution for gdb

Trass3r:
fixes microsoft/vscode-cpptools#2130


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1155,14 +1156,26 @@ protected override void HandleContinueRequestAsync(IRequestResponder<ContinueArg
             }
         }
 
-        protected override void HandleStepInRequestAsync(IRequestResponder<StepInArguments> responder)
+        protected override void HandleReverseContinueRequestAsync(IRequestResponder<ReverseContinueArguments> responder)
         {
-            StepInResponse response = new StepInResponse();
+            try
+            {
+                // FIXME: there is no IDebugProgram2.ReverseContinue
```
WardenGnaw, line 1163: We have [`IDebugReversibleEngineProgram160.SetExecuteDirection`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.debugger.interop.idebugreversibleengineprogram160?view=visualstudiosdk-2019)

For every ReverseRequest we set `SetExecuteDirection` to `backward` and for every regular continue/step request we set it to `foward`.

We will also need to add a reference to `Microsoft.VisualStudio.Debugger.Interop.16.0`

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1624,18 +1624,20 @@ public async Task Step(int threadId, enum_STEPKIND kind, enum_STEPUNIT unit)
...
             await ExceptionManager.EnsureSettingsUpdated();
 
+            bool isForwardStep = kind != enum_STEPKIND.STEP_BACKWARDS;
             if ((unit == enum_STEPUNIT.STEP_LINE) || (unit == enum_STEPUNIT.STEP_STATEMENT))
             {
                 switch (kind)
                 {
                     case enum_STEPKIND.STEP_INTO:
-                        await MICommandFactory.ExecStep(threadId);
+                        await MICommandFactory.ExecStep(threadId, isForwardStep);
```
WardenGnaw, line 1633: instead of using `isForwardStep` check the `exeuctionDirection`

Trass3r:
> We will want to check if `gdb` can do reverse commands. You will need to execute [`-list-target-features`](https://sourceware.org/gdb/current/onlinedocs/gdb/GDB_002fMI-Support-Commands.html#The-_002dlist_002dtarget_002dfeatures-Command) and check for `reverse.`

The problem is we need to advertise the capability before the target is even launched.

WardenGnaw:
There is a [`CapabilitiesEvent`](https://github.com/microsoft/debug-adapter-protocol/blob/4b2eba24dd77f3bcbb3400596358d5cfe737446a/debugAdapterProtocol.json#L525) that can be sent if it has changed. We are not sure if VS Code supports changing the UI for reverse if the capabilities have changed. 

image

src/OpenDebugAD7/AD7DebugSession.cs
Trass3r: But how does it interact with `STEPKIND.STEP_BACKWARDS`?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
Trass3r: `STEP_BACKWARDS`is a valid input though.

WardenGnaw: It is an older enum from `IDebugProcess3` and that is only for stepping. We now use the newer interface `IDebugReversibleSessionProcess160` to handle reverse program control which includes `ReverseStep` and `ReverseExecute`.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -1083,6 +1082,19 @@ public int WatchForThreadStep(IDebugProgram2 pOriginatingProgram, uint dwTid, in
 
         #endregion
 
+        #region IDebugEngineProgram2 Members
+        int IDebugReversibleEngineProgram160.CanReverse()
+        {
+            return DebuggedProcess.TargetFeatures.Contains("reverse") ? 1 : 0;
```
WardenGnaw, line 1088: The `int` here is an `HRESULT`. 

You will want to return `Constants.S_OK` if `DebuggedProcess.TargetFeatures.Contains("reverse")` is true, and `Constants.S_FALSE` otherwise.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -765,6 +776,8 @@ protected override void HandleLaunchRequestAsync(IRequestResponder<LaunchArgumen
                     eb.ThrowHR(hr);
                 }
 
+                Protocol.SendEvent(new CapabilitiesEvent() { Capabilities = new InitializeResponse() { SupportsStepBack = ((IDebugReversibleEngineProgram160)m_engine).CanReverse() != 0 } });
```
WardenGnaw, line 779:
```suggestion
                Protocol.SendEvent(new CapabilitiesEvent() { Capabilities = new InitializeResponse() { SupportsStepBack = ((IDebugReversibleEngineProgram160)m_engine).CanReverse() == HRConstants.S_OK } });
```

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -992,6 +1005,8 @@ protected override void HandleAttachRequestAsync(IRequestResponder<AttachArgumen
...
```
WardenGnaw, line 1008:
```suggestion
                Protocol.SendEvent(new CapabilitiesEvent() { Capabilities = new InitializeResponse() { SupportsStepBack = ((IDebugReversibleEngineProgram160)m_engine).CanReverse() == HRConstants.S_OK } });
```

Trass3r:
Drafted something.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: @Trass3r Have you checked to see if this changes the VS Code UI to show the reverse debug buttons?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -27,6 +26,18 @@
 
 namespace OpenDebugAD7
 {
+    // not in the DAP dll yet
```
WardenGnaw, line 29: Looks like we need to enable this to be generated, it was not created because no one was using this event.

src/OpenDebugAD7/AD7DebugSession.cs
Trass3r: Oh yeah it works fine both ways 🎉

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -618,6 +618,9 @@ public async Task Initialize(HostWaitLoop waitLoop, CancellationToken token)
                         }
                     }
                 }
+                // now the exe is loaded and we can check target features
+                TargetFeatures = await MICommandFactory.GetTargetFeatures();
+                // TODO: only now we know if (TargetFeatures.Contains("async")), see above
```
Trass3r, line 623: What to do about this `async` feature?

WardenGnaw: We had an implementation in https://github.com/microsoft/MIEngine/pull/244 but had to revert with https://github.com/microsoft/MIEngine/pull/794 because of:

> This is being reverted as it hangs in the VS & Android case. gdb for Android supports async but doesnt' list it as capabilities. It is noticed when trying to stop debugging while running.

src/OpenDebugAD7/OpenDebugAD7.csproj
```diff
@@ -104,9 +104,13 @@
...
       <ForceIncludeInVSIX>true</ForceIncludeInVSIX>
     </Reference>
+    <Reference Include="Microsoft.VisualStudio.Debugger.Interop.16.0">
+      <HintPath>$(NuGetPackagesDirectory)/Microsoft.VisualStudio.Debugger.Interop.16.0.16.7.30328.139/lib/net20/Microsoft.VisualStudio.Debugger.Interop.16.0.dll</HintPath>
+      <ForceIncludeInVSIX>true</ForceIncludeInVSIX>
+    </Reference>
     <!-- Same version as in packages.config -->
     <Reference Include="Microsoft.VisualStudio.Shared.VSCodeDebugProtocol">
-      <HintPath>$(NuGetPackagesDirectory)/Microsoft.VisualStudio.Shared.VsCodeDebugProtocol.16.7.40526.2/lib/net45/Microsoft.VisualStudio.Shared.VSCodeDebugProtocol.dll</HintPath>
+      <HintPath>$(NuGetPackagesDirectory)/Microsoft.VisualStudio.Shared.VsCodeDebugProtocol.16.6.40406.1/lib/net45/Microsoft.VisualStudio.Shared.VSCodeDebugProtocol.dll</HintPath>
```
Trass3r, line 113: Not sure why Nuget changed this.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: You can leave this in for now. When theres an update to the protocol, I will remove and update the package.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
Trass3r: See my comment there. It was simply done too early. I ran into the same problem.
So I guess `EnableTargetAsyncOption` could be moved to this line.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: When `STEPKIND.STEP_BACKWARDS` was created, it did not take into the idea of  `step in`, `step out`, and `step over`. We created the IDebugReversibleEngine and Session to address this issue. 

I would remove `STEPKIND.STEP_BACKWARDS`, since even in the Visual Studio Debugger, we dismiss that enum.

Trass3r:
Done, it's a lot cleaner now.
I even implemented stepping granularity.

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -661,6 +670,12 @@ public virtual bool CanDetach()
             return true;
         }
 
+        public virtual async Task<string[]> GetTargetFeatures()
+        {
+            Results results = await _debugger.CmdAsync("-list-target-features", ResultClass.done);
+            return results.Find<ValueListValue>("features").AsStrings;
```
Trass3r, line 676: The [docs](https://sourceware.org/gdb/current/onlinedocs/gdb/GDB_002fMI-Support-Commands.html#The-_002dlist_002dtarget_002dfeatures-Command) say it's named `result` but in my case gdb returned `features`.

src/OpenDebugAD7/packages.config
```diff
@@ -2,7 +2,8 @@
 <packages>
   <package id="Newtonsoft.Json" version="9.0.1" targetFramework="net45" />
   <package id="Microsoft.VisualStudio.Debugger.Interop.Portable" version="1.0.1" targetFramework="net45" />
-  <package id="MicroBuild.Core" version="0.2.0" targetFramework="net45" />
+  <package id="Microsoft.VisualStudio.Debugger.Interop.16.0" version="16.7.30328.139" targetFramework="net462" />
```
Trass3r, line 5: I guess this still needs to be added to some packaging code.

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -661,6 +670,12 @@ public virtual bool CanDetach()
             return true;
         }
 
+        public virtual async Task<string[]> GetTargetFeatures()
```
WardenGnaw, line 673: You will want to move this to `gdb.cs`. It looks like lldb-mi does not support -list-target-features, but only [`-list-features`](https://github.com/lldb-tools/lldb-mi/blob/d69dadb35a859bc0aa4a43d0adb3fc4fe5856947/src/MICmdCmdSupportList.cpp).

You can leave it throw unimplmented for clrdbg.cs since that is not used anymore.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -618,6 +617,10 @@ public async Task Initialize(HostWaitLoop waitLoop, CancellationToken token)
                         }
                     }
                 }
+                // now the exe is loaded and we can check target features
+                TargetFeatures = await MICommandFactory.GetTargetFeatures();
+                if (TargetFeatures.Contains("async"))
+                    await MICommandFactory.EnableTargetAsyncOption();
```
WardenGnaw, line 623: To support the lldb-mi case, if GetTargetFeatures fails, we need to always run `EnableTargetAsync`

src/OpenDebugAD7/packages.config
WardenGnaw: I'll take a look to see what needs to be fixed.

Trass3r:
Done.

WardenGnaw:
@Trass3r For the launch scenario, do you have `-interpreter-exec console \"target record-full\"` in the setup command?

Trass3r:
No. Even in gdb itself there are different methods to record and then there's the next level rr tool which is what I use.
You can't hardcode the method.

WardenGnaw:
Do we need to worry that someone would enable reverse execution in the middle of debugging? We only check and enable Reverse functionality in Launch/Attach. Is there a set of commands we can capture to resend the capabilities?

Trass3r:
That's a good question, is it possible with the gdb way?
Using rr it's a two stage procedure. First you record the entire run, then you run rr disguising as gdb with full forward and backward capability.

WardenGnaw:
I've never used RR, but its good to know that it works in that scenario. 
This is how I was testing with GDB.

```
-list-target-features
^done,features=[]
(gdb)
-exec-step
^running
*running,thread-id="all"
(gdb)
*stopped,reason="end-stepping-range",frame={addr="0x0000000008001224",func="main",args=[],file="main.cpp",fullname="/home/waan/cpp/main.cpp",line="7",arch="i386:x86-64"},thread-id="1",stopped-threads="all",core="0"
(gdb)
-exec-step --reverse
^error,msg="Target native does not support this command."
(gdb)
target record-full
&"target record-full\n"
=record-started,thread-group="i1",method="full"
^done
(gdb)
-list-target-features
^done,features=["reverse"]
(gdb)
-exec-step
^running
*running,thread-id="all"
(gdb)
*stopped,reason="end-stepping-range",frame={addr="0x000000000800139a",func="std::chrono::duration<long, std::ratio<1l, 1000l> >::duration<int, void>",args=[{name="this",value="0x80018bd <__libc_csu_init+77>"},{name="__rep",value="<error reading variable>"}],file="/usr/include/c++/9/chrono",fullname="/usr/include/c++/9/chrono",line="331",arch="i386:x86-64"},thread-id="1",stopped-threads="all",core="0"
(gdb)
-exec-step --reverse
^running
*running,thread-id="1"
(gdb)
*stopped,reason="end-stepping-range",frame={addr="0x000000000800122b",func="main",args=[],file="main.cpp",fullname="/home/waan/cpp/main.cpp",line="7",arch="i386:x86-64"},thread-id="1",stopped-threads="all",core="0"
(gdb)
```

We may want a check in `HandleEvaluateRequestAsync` and if its a `-exec` with `record`, check for `target-features` and update the capabilities.

Trass3r:
I see, makes sense.

pirpyn:
> Do we need to worry that someone would enable reverse execution in the middle of debugging? We only check and enable Reverse functionality in Launch/Attach. Is there a set of commands we can capture to resend the capabilities?

Hello, I'm using also UndoDB which record from the beginning, but with gdb I only record specific portion of code. 
So yeah, as a user, that something i'm likely to do.

WardenGnaw:
@Trass3r Ignore my suggestion above about capturing commands for users in `HandleEvaluateAsync`. We could detect recording when GDB sends the output
`=record-started,thread-group="i1",method="full"` then update the capabilities

gregg-miskelly:
FYI from https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI-Async-Records.html:

```
=record-started,thread-group="id",method="method"[,format="format"]
=record-stopped,thread-group="id"
```
Execution log recording was either started or stopped on an inferior. The id is the GDB identifier of the thread group corresponding to the affected inferior.

The method field indicates the method used to record execution. If the method in use supports multiple recording formats, format will be present and contain the currently used format. See Process Record and Replay, for existing method and format values.

Trass3r:
Something like this. The question is how to proceed from there.

WardenGnaw:
Sorry for the delay, again. I am looking on how this should be implemented with the given API. 

What needs to be done is:
`IDebugReversibleEngineProgram160.CanReverse` is going to be controlled by a variable `m_canReverse`. This can be initially `false`.

When we first call `MICommandFactory.GetTargetFeatures()` we can do this check and assign `m_canReverse`
`m_canReverse = DebuggedProcess.TargetFeatures.Contains("reverse");`

Anytime we get the `record-started`/`record-stopped` events, update `m_canReverse`.

--------------------------------------------------------
For updating the Capabilities to the Client, there will need to be a check every time we stop to see if we should send an updated `CapabilitiesEvent`. This can be before we send off the `OpenDebugStoppedEvent`.

Trass3r:
Ok this sort of works but first of all we get an `Unknown stopping event` exception when reaching the end of the record history.
Also this only updates the UI one time step later. You enable recording but the caps are only after the next step (which would still be kind of ok cause you couldn't step back at this point anyway), you stop recording and the stepping buttons are still there.

Sidenote: no clue where
```
 natvis FormatDisplayString: Value cannot be null.
Parameter name: key
```
comes from. There is no such variable.

<details>
  <summary>Log:</summary>

```js
`record full
<--   C (evaluate-14): {"command":"evaluate","arguments":{"expression":"`record full","frameId":1000,"context":"repl"},"type":"request","seq":14}
    <-1039-interpreter-exec console "record full"
    ->=record-started,thread-group="i1",method="full"
    ->1039^done
    ->(gdb)
    1039: elapsed time 2
    <-1040-list-target-features
    ->1040^done,features=["async","reverse"]
    ->(gdb)
    1040: elapsed time 0
    natvis FormatDisplayString: Value cannot be null.
Parameter name: key
--> R (evaluate-14): {"type":"response","request_seq":14,"success":true,"command":"evaluate","body":{"result":"","variablesReference":0},"seq":574}

<--   C (scopes-15): {"command":"scopes","arguments":{"frameId":1000},"type":"request","seq":15}
--> R (scopes-15): {"type":"response","request_seq":15,"success":true,"command":"scopes","body":{"scopes":[{"name":"Locals","variablesReference":1003,"expensive":false}]},"seq":577}
<--   C (variables-16): {"command":"variables","arguments":{"variablesReference":1003},"type":"request","seq":16}
    <-1041-var-create - * "((s)._M_dataplus)._M_p"
    ->1041^done,name="var7",numchild="1",value="0xa <error: Cannot access memory at address 0xa>",type="std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::pointer",thread-id="1",has_more="0"
    ->(gdb)
    1041: elapsed time 1
    <-1042-var-list-children --simple-values "var7" 0 1000
    ->1042^done,numchild="1",children=[child={name="var7.*((s)._M_dataplus)._M_p",exp="*((s)._M_dataplus)._M_p",numchild="0",value="",type="char16_t",thread-id="1"}],has_more="0"
    ->(gdb)
    1042: elapsed time 0
--> R (variables-16): {"type":"response","request_seq":16,"success":true,"command":"variables","body":{"variables":[{"name":"s","value":"0xa <error: Cannot access memory at address 0xa>","type":"std::u16string","evaluateName":"s","variablesReference":1004},{"name":"argc","value":"1","type":"int","evaluateName":"argc","variablesReference":0,"memoryReference":"0x0000000000000001"},{"name":"argv","value":"0x7fffffffdcc8","type":"const char **","evaluateName":"argv","variablesReference":1005,"memoryReference":"0x00007fffffffdcc8"}]},"seq":596}
<--   C (next-17): {"command":"next","arguments":{"threadId":1342},"type":"request","seq":17}
    <-1043-exec-next
    ->1043^running
    ->*running,thread-id="all"
    ->(gdb)
    1043: elapsed time 0
--> R (next-17): {"type":"response","request_seq":17,"success":true,"command":"next","body":{},"seq":609}
    ->*stopped,reason="end-stepping-range",frame={addr="0x00005555555552aa",func="main",args=[{name="argc",value="1"},{name="argv",value="0x7fffffffdcc8"}],file="test.cpp",fullname="/tmp/test.cpp",line="17",arch="i386:x86-64"},thread-id="1",stopped-threads="all",core="2"
    <-1044-stack-list-frames 0 1000
    ->1044^done,stack=[frame={level="0",addr="0x00005555555552aa",func="main",file="test.cpp",fullname="/tmp/test.cpp",line="17",arch="i386:x86-64"}]
    ->(gdb)
    1044: elapsed time 0
    <-1045-var-delete var3
    ->1045^done,ndeleted="1"
    ->(gdb)
    1045: elapsed time 0
    <-1046-var-delete var4
    ->1046^done,ndeleted="2"
    ->(gdb)
    1046: elapsed time 0
    <-1047-var-delete var5
    ->1047^done,ndeleted="7"
    ->(gdb)
    1047: elapsed time 0
    <-1048-var-delete var6
    ->1048^done,ndeleted="2"
    ->(gdb)
    1048: elapsed time 0
    <-1049-var-delete var7
    ->1049^done,ndeleted="2"
    ->(gdb)
    1049: elapsed time 0
    Send Event AD7StepCompleteEvent
--> E (capabilities): {"type":"event","event":"capabilities","body":{"capabilities":{"exceptionBreakpointFilters":[],"supportsStepBack":true,"completionTriggerCharacters":[],"additionalModuleColumns":[],"supportedChecksumAlgorithms":[]}},"seq":663}
--> E (stopped): {"type":"event","event":"stopped","body":{"reason":"step","threadId":1342,"allThreadsStopped":true,"source":{"name":"test.cpp","path":"/tmp/test.cpp","sources":[],"checksums":[]},"line":17,"column":1},"seq":665}
<--   C (threads-18): {"command":"threads","type":"request","seq":18}
--> R (threads-18): {"type":"response","request_seq":18,"success":true,"command":"threads","body":{"threads":[{"id":1342,"name":"test"}]},"seq":668}
<--   C (stackTrace-19): {"command":"stackTrace","arguments":{"threadId":1342,"startFrame":0,"levels":20},"type":"request","seq":19}
    <-1050-stack-list-arguments 0 0 0
    ->1050^done,stack-args=[frame={level="0",args=[name="argc",name="argv"]}]
    ->(gdb)
    1050: elapsed time 0
    <-1051-var-create - * "argc"
    ->1051^done,name="var8",numchild="0",value="1",type="int",thread-id="1",has_more="0"
    ->(gdb)
    1051: elapsed time 0
    <-1052-var-delete var8
    ->1052^done,ndeleted="1"
    ->(gdb)
    1052: elapsed time 0
    <-1053-var-create - * "argv"
    ->1053^done,name="var9",numchild="1",value="0x7fffffffdcc8",type="const char **",thread-id="1",has_more="0"
    ->(gdb)
    1053: elapsed time 0
    <-1054-var-delete var9
    ->1054^done,ndeleted="1"
    ->(gdb)
    1054: elapsed time 0
--> R (stackTrace-19): {"type":"response","request_seq":19,"success":true,"command":"stackTrace","body":{"stackFrames":[{"id":1000,"name":"main(int argc, const char ** argv)","source":{"name":"test.cpp","path":"/tmp/test.cpp","sources":[],"checksums":[]},"line":17,"column":1}],"totalFrames":1},"seq":711}
<--   C (stackTrace-20): {"command":"stackTrace","arguments":{"threadId":1342,"startFrame":0,"levels":20},"type":"request","seq":20}
--> R (stackTrace-20): {"type":"response","request_seq":20,"success":true,"command":"stackTrace","body":{"stackFrames":[{"id":1001,"name":"main(int argc, const char ** argv)","source":{"name":"test.cpp","path":"/tmp/test.cpp","sources":[],"checksums":[]},"line":17,"column":1}],"totalFrames":1},"seq":714}
<--   C (scopes-21): {"command":"scopes","arguments":{"frameId":1000},"type":"request","seq":21}
    <-1055-stack-list-variables 0
    ->1055^done,variables=[{name="argc",arg="1"},{name="argv",arg="1"},{name="s"}]
    ->(gdb)
    1055: elapsed time 0
    <-1056-var-create - * "argc"
    ->1056^done,name="var10",numchild="0",value="1",type="int",thread-id="1",has_more="0"
    ->(gdb)
    1056: elapsed time 0
    <-1057-var-create - * "argv"
    ->1057^done,name="var11",numchild="1",value="0x7fffffffdcc8",type="const char **",thread-id="1",has_more="0"
    ->(gdb)
    1057: elapsed time 1
    <-1058-var-create - * "s"
    ->1058^done,name="var12",numchild="2",value="{...}",type="std::u16string",thread-id="1",has_more="0"
    ->(gdb)
    1058: elapsed time 0
--> R (scopes-21): {"type":"response","request_seq":21,"success":true,"command":"scopes","body":{"scopes":[{"name":"Locals","variablesReference":1000,"expensive":false}]},"seq":749}
<--   C (variables-22): {"command":"variables","arguments":{"variablesReference":1000},"type":"request","seq":22}
    <-1059-var-list-children --simple-values "var12" 0 1000
    ->1059^done,numchild="2",children=[child={name="var12.public",exp="public",numchild="1",value="",thread-id="1"},child={name="var12.private",exp="private",numchild="3",value="",thread-id="1"}],has_more="0"
    ->(gdb)
    1059: elapsed time 1
    <-1060-var-list-children --simple-values "var12.public" 0 1000
    ->1060^done,numchild="1",children=[child={name="var12.public.npos",exp="npos",numchild="0",value="18446744073709551615",type="const std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::size_type",thread-id="1"}],has_more="0"
    ->(gdb)
    1060: elapsed time 1
    <-1061-var-list-children --simple-values "var12.private" 0 1000
    ->1061^done,numchild="3",children=[child={name="var12.private._M_dataplus",exp="_M_dataplus",numchild="2",type="std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::_Alloc_hider",thread-id="1"},child={name="var12.private._M_string_length",exp="_M_string_length",numchild="0",value="4",type="std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::size_type",thread-id="1"},child={name="var12.private.2_anonymous",exp="<anonymous union>",numchild="1",type="union {...}",thread-id="1"}],has_more="0"
    ->(gdb)
    1061: elapsed time 1
    <-1062-var-create - * "((s)._M_dataplus)._M_p"
    ->1062^done,name="var13",numchild="1",value="0x7fffffffdba0 u\"test\"",type="std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::pointer",thread-id="1",has_more="0"
    ->(gdb)
    1062: elapsed time 1
    <-1063-var-list-children --simple-values "var13" 0 1000
    ->1063^done,numchild="1",children=[child={name="var13.*((s)._M_dataplus)._M_p",exp="*((s)._M_dataplus)._M_p",numchild="0",value="116 u't'",type="char16_t",thread-id="1"}],has_more="0"
    ->(gdb)
    1063: elapsed time 0
    <-1064-var-list-children --simple-values "var11" 0 1000
    ->1064^done,numchild="1",children=[child={name="var11.*argv",exp="*argv",numchild="1",value="0x7fffffffdf65 \"/tmp/test\"",type="const char *",thread-id="1"}],has_more="0"
    ->(gdb)
    1064: elapsed time 0
--> R (variables-22): {"type":"response","request_seq":22,"success":true,"command":"variables","body":{"variables":[{"name":"s","value":"0x7fffffffdba0 u\"test\"","type":"std::u16string","evaluateName":"s","variablesReference":1001},{"name":"argc","value":"1","type":"int","evaluateName":"argc","variablesReference":0,"memoryReference":"0x0000000000000001"},{"name":"argv","value":"0x7fffffffdcc8","type":"const char **","evaluateName":"argv","variablesReference":1002,"memoryReference":"0x00007fffffffdcc8"}]},"seq":800}
<--   C (stepBack-23): {"command":"stepBack","arguments":{"threadId":1342},"type":"request","seq":23}
    <-1065-exec-next --reverse
    ->1065^running
    ->*running,thread-id="1"
    1065: elapsed time 0
    ->(gdb)
--> R (stepBack-23): {"type":"response","request_seq":23,"success":true,"command":"stepBack","body":{},"seq":813}
    ->~"\nNo more reverse-execution history.\n"

No more reverse-execution history.
    ->~"main (argc=1, argv=0x7fffffffdcc8) at test.cpp:16\n"
main (argc=1, argv=0x7fffffffdcc8) at test.cpp:16
    ->~"16\t\tstd::u16string s = u\"test\";\n"
    ->*stopped,frame={addr="0x000055555555527b",func="main",args=[{name="argc",value="1"},{name="argv",value="0x7fffffffdcc8"}],file="test.cpp",fullname="/tmp/test.cpp",line="16",arch="i386:x86-64"},thread-id="1",stopped-threads="all",core="2"
16		std::u16string s = u"test";
    <-1066-stack-list-frames 0 1000
    ->1066^done,stack=[frame={level="0",addr="0x000055555555527b",func="main",file="test.cpp",fullname="/tmp/test.cpp",line="16",arch="i386:x86-64"}]
    ->(gdb)
    1066: elapsed time 0
    <-1067-var-delete var10
    ->1067^done,ndeleted="1"
    ->(gdb)
    1067: elapsed time 0
    <-1068-var-delete var11
    ->1068^done,ndeleted="2"
    ->(gdb)
    1068: elapsed time 0
    <-1069-var-delete var12
    ->1069^done,ndeleted="7"
    ->(gdb)
    1069: elapsed time 0
    <-1070-var-delete var13
    ->1070^done,ndeleted="2"
    ->(gdb)
    1070: elapsed time 0
    Send Event AD7ExceptionEvent
--> E (stopped): {"type":"event","event":"stopped","body":{"reason":"exception","threadId":1342,"text":"Unknown stopping event","allThreadsStopped":true,"source":{"name":"test.cpp","path":"/tmp/test.cpp","sources":[],"checksums":[]},"line":16,"column":1},"seq":871}
<--   C (threads-24): {"command":"threads","type":"request","seq":24}
--> R (threads-24): {"type":"response","request_seq":24,"success":true,"command":"threads","body":{"threads":[{"id":1342,"name":"test"}]},"seq":874}
<--   C (stackTrace-25): {"command":"stackTrace","arguments":{"threadId":1342,"startFrame":0,"levels":20},"type":"request","seq":25}
    <-1071-stack-list-arguments 0 0 0
    ->1071^done,stack-args=[frame={level="0",args=[name="argc",name="argv"]}]
    ->(gdb)
    1071: elapsed time 0
    <-1072-var-create - * "argc"
    ->1072^done,name="var14",numchild="0",value="1",type="int",thread-id="1",has_more="0"
    ->(gdb)
    1072: elapsed time 0
    <-1073-var-delete var14
    ->1073^done,ndeleted="1"
    ->(gdb)
    1073: elapsed time 0
    <-1074-var-create - * "argv"
    ->1074^done,name="var15",numchild="1",value="0x7fffffffdcc8",type="const char **",thread-id="1",has_more="0"
    ->(gdb)
    1074: elapsed time 0
    <-1075-var-delete var15
    ->1075^done,ndeleted="1"
    ->(gdb)
    1075: elapsed time 0
--> R (stackTrace-25): {"type":"response","request_seq":25,"success":true,"command":"stackTrace","body":{"stackFrames":[{"id":1000,"name":"main(int argc, const char ** argv)","source":{"name":"test.cpp","path":"/tmp/test.cpp","sources":[],"checksums":[]},"line":16,"column":1}],"totalFrames":1},"seq":917}
<--   C (stackTrace-26): {"command":"stackTrace","arguments":{"threadId":1342,"startFrame":0,"levels":20},"type":"request","seq":26}
--> R (stackTrace-26): {"type":"response","request_seq":26,"success":true,"command":"stackTrace","body":{"stackFrames":[{"id":1001,"name":"main(int argc, const char ** argv)","source":{"name":"test.cpp","path":"/tmp/test.cpp","sources":[],"checksums":[]},"line":16,"column":1}],"totalFrames":1},"seq":920}
<--   C (scopes-27): {"command":"scopes","arguments":{"frameId":1000},"type":"request","seq":27}
    <-1076-stack-list-variables 0
    ->1076^done,variables=[{name="argc",arg="1"},{name="argv",arg="1"},{name="s"}]
    ->(gdb)
    1076: elapsed time 2
    <-1077-var-create - * "argc"
    ->1077^done,name="var16",numchild="0",value="1",type="int",thread-id="1",has_more="0"
    ->(gdb)
    1077: elapsed time 2
    <-1078-var-create - * "argv"
    ->1078^done,name="var17",numchild="1",value="0x7fffffffdcc8",type="const char **",thread-id="1",has_more="0"
    ->(gdb)
    1078: elapsed time 1
    <-1079-var-create - * "s"
    ->1079^done,name="var18",numchild="2",value="{...}",type="std::u16string",thread-id="1",has_more="0"
    ->(gdb)
    1079: elapsed time 1
--> R (scopes-27): {"type":"response","request_seq":27,"success":true,"command":"scopes","body":{"scopes":[{"name":"Locals","variablesReference":1000,"expensive":false}]},"seq":955}
<--   C (variables-28): {"command":"variables","arguments":{"variablesReference":1000},"type":"request","seq":28}
    <-1080-var-list-children --simple-values "var18" 0 1000
    ->1080^done,numchild="2",children=[child={name="var18.public",exp="public",numchild="1",value="",thread-id="1"},child={name="var18.private",exp="private",numchild="3",value="",thread-id="1"}],has_more="0"
    ->(gdb)
    1080: elapsed time 0
    <-1081-var-list-children --simple-values "var18.public" 0 1000
    ->1081^done,numchild="1",children=[child={name="var18.public.npos",exp="npos",numchild="0",value="18446744073709551615",type="const std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::size_type",thread-id="1"}],has_more="0"
    ->(gdb)
    1081: elapsed time 0
    <-1082-var-list-children --simple-values "var18.private" 0 1000
    ->1082^done,numchild="3",children=[child={name="var18.private._M_dataplus",exp="_M_dataplus",numchild="2",type="std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::_Alloc_hider",thread-id="1"},child={name="var18.private._M_string_length",exp="_M_string_length",numchild="0",value="93824992239101",type="std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::size_type",thread-id="1"},child={name="var18.private.2_anonymous",exp="<anonymous union>",numchild="1",type="union {...}",thread-id="1"}],has_more="0"
    ->(gdb)
    1082: elapsed time 1
    <-1083-var-create - * "((s)._M_dataplus)._M_p"
    ->1083^done,name="var19",numchild="1",value="0xa <error: Cannot access memory at address 0xa>",type="std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::pointer",thread-id="1",has_more="0"
    ->(gdb)
    1083: elapsed time 4
    <-1084-var-list-children --simple-values "var19" 0 1000
    ->1084^done,numchild="1",children=[child={name="var19.*((s)._M_dataplus)._M_p",exp="*((s)._M_dataplus)._M_p",numchild="0",value="",type="char16_t",thread-id="1"}],has_more="0"
    ->(gdb)
    1084: elapsed time 0
    <-1085-var-list-children --simple-values "var17" 0 1000
    ->1085^done,numchild="1",children=[child={name="var17.*argv",exp="*argv",numchild="1",value="0x7fffffffdf65 \"/tmp/test\"",type="const char *",thread-id="1"}],has_more="0"
    ->(gdb)
    1085: elapsed time 0
--> R (variables-28): {"type":"response","request_seq":28,"success":true,"command":"variables","body":{"variables":[{"name":"s","value":"0xa <error: Cannot access memory at address 0xa>","type":"std::u16string","evaluateName":"s","variablesReference":1001},{"name":"argc","value":"1","type":"int","evaluateName":"argc","variablesReference":0,"memoryReference":"0x0000000000000001"},{"name":"argv","value":"0x7fffffffdcc8","type":"const char **","evaluateName":"argv","variablesReference":1002,"memoryReference":"0x00007fffffffdcc8"}]},"seq":1006}




`record stop
<--   C (evaluate-35): {"command":"evaluate","arguments":{"expression":"`record stop","frameId":1000,"context":"repl"},"type":"request","seq":35}
    <-1107-interpreter-exec console "record stop"
    ->~"Process record is stopped and all execution logs are deleted.\n"
    ->=record-stopped,thread-group="i1"
    ->1107^done
    ->(gdb)
    1107: elapsed time 2
    <-1108-list-target-features
Process record is stopped and all execution logs are deleted.
    natvis FormatDisplayString: Value cannot be null.
Parameter name: key
    ->1108^done,features=["async"]
--> R (evaluate-35): {"type":"response","request_seq":35,"success":true,"command":"evaluate","body":{"result":"","variablesReference":0},"seq":1234}

    ->(gdb)
    1108: elapsed time 1
<--   C (scopes-36): {"command":"scopes","arguments":{"frameId":1000},"type":"request","seq":36}
--> R (scopes-36): {"type":"response","request_seq":36,"success":true,"command":"scopes","body":{"scopes":[{"name":"Locals","variablesReference":1003,"expensive":false}]},"seq":1242}
<--   C (variables-37): {"command":"variables","arguments":{"variablesReference":1003},"type":"request","seq":37}
    <-1109-var-create - * "((s)._M_dataplus)._M_p"
    ->1109^done,name="var26",numchild="1",value="0x7fffffffdba0 u\"test\"",type="std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::pointer",thread-id="1",has_more="0"
    ->(gdb)
    1109: elapsed time 1
    <-1110-var-list-children --simple-values "var26" 0 1000
    ->1110^done,numchild="1",children=[child={name="var26.*((s)._M_dataplus)._M_p",exp="*((s)._M_dataplus)._M_p",numchild="0",value="116 u't'",type="char16_t",thread-id="1"}],has_more="0"
    ->(gdb)
    1110: elapsed time 0
--> R (variables-37): {"type":"response","request_seq":37,"success":true,"command":"variables","body":{"variables":[{"name":"s","value":"0x7fffffffdba0 u\"test\"","type":"std::u16string","evaluateName":"s","variablesReference":1004},{"name":"argc","value":"1","type":"int","evaluateName":"argc","variablesReference":0,"memoryReference":"0x0000000000000001"},{"name":"argv","value":"0x7fffffffdcc8","type":"const char **","evaluateName":"argv","variablesReference":1005,"memoryReference":"0x00007fffffffdcc8"}]},"seq":1261}
<--   C (stepBack-38): {"command":"stepBack","arguments":{"threadId":1342},"type":"request","seq":38}
    <-1111-exec-next --reverse
    ->1111^error,msg="Target native does not support this command."
    ->(gdb)
    1111: elapsed time 1
    Send Event AD7MessageEvent
ERROR: Unexpected GDB output from command "-exec-next --reverse". Target native does not support this command.
--> R (stepBack-38): {"type":"response","request_seq":38,"success":false,"command":"stepBack","message":"Unable to step next. Operation failed with error code 0x80004004.","body":{},"seq":1276}
<--   C (next-39): {"command":"next","arguments":{"threadId":1342},"type":"request","seq":39}
    <-1112-exec-next
    ->1112^running
    ->*running,thread-id="all"
    1112: elapsed time 2
    ->(gdb)
--> R (next-39): {"type":"response","request_seq":39,"success":true,"command":"next","body":{},"seq":1289}
    ->*stopped,reason="end-stepping-range",frame={addr="0x00005555555552c5",func="main",args=[{name="argc",value="1"},{name="argv",value="0x7fffffffdcc8"}],file="test.cpp",fullname="/tmp/test.cpp",line="16",arch="i386:x86-64"},thread-id="1",stopped-threads="all",core="6"
    <-1113-stack-list-frames 0 1000
    ->1113^done,stack=[frame={level="0",addr="0x00005555555552c5",func="main",file="test.cpp",fullname="/tmp/test.cpp",line="16",arch="i386:x86-64"}]
    ->(gdb)
    1113: elapsed time 0
    <-1114-var-delete var22
    ->1114^done,ndeleted="1"
    ->(gdb)
    1114: elapsed time 0
    <-1115-var-delete var23
    ->1115^done,ndeleted="2"
    ->(gdb)
    1115: elapsed time 0
    <-1116-var-delete var24
    ->1116^done,ndeleted="7"
    ->(gdb)
    1116: elapsed time 0
    <-1117-var-delete var25
    ->1117^done,ndeleted="2"
    ->(gdb)
    1117: elapsed time 0
    <-1118-var-delete var26
    ->1118^done,ndeleted="2"
    ->(gdb)
    1118: elapsed time 0
    Send Event AD7StepCompleteEvent
--> E (capabilities): {"type":"event","event":"capabilities","body":{"capabilities":{"exceptionBreakpointFilters":[],"supportsStepBack":false,"completionTriggerCharacters":[],"additionalModuleColumns":[],"supportedChecksumAlgorithms":[]}},"seq":1343}
--> E (stopped): {"type":"event","event":"stopped","body":{"reason":"step","threadId":1342,"allThreadsStopped":true,"source":{"name":"test.cpp","path":"/tmp/test.cpp","sources":[],"checksums":[]},"line":16,"column":1},"seq":1345}
<--   C (threads-40): {"command":"threads","type":"request","seq":40}
--> R (threads-40): {"type":"response","request_seq":40,"success":true,"command":"threads","body":{"threads":[{"id":1342,"name":"test"}]},"seq":1348}
<--   C (stackTrace-41): {"command":"stackTrace","arguments":{"threadId":1342,"startFrame":0,"levels":20},"type":"request","seq":41}
    <-1119-stack-list-arguments 0 0 0
    ->1119^done,stack-args=[frame={level="0",args=[name="argc",name="argv"]}]
    ->(gdb)
    1119: elapsed time 1
    <-1120-var-create - * "argc"
    ->1120^done,name="var27",numchild="0",value="1",type="int",thread-id="1",has_more="0"
    ->(gdb)
    1120: elapsed time 0
    <-1121-var-delete var27
    ->1121^done,ndeleted="1"
    ->(gdb)
    1121: elapsed time 0
    <-1122-var-create - * "argv"
    ->1122^done,name="var28",numchild="1",value="0x7fffffffdcc8",type="const char **",thread-id="1",has_more="0"
    ->(gdb)
    1122: elapsed time 0
    <-1123-var-delete var28
    ->1123^done,ndeleted="1"
    ->(gdb)
    1123: elapsed time 0
--> R (stackTrace-41): {"type":"response","request_seq":41,"success":true,"command":"stackTrace","body":{"stackFrames":[{"id":1000,"name":"main(int argc, const char ** argv)","source":{"name":"test.cpp","path":"/tmp/test.cpp","sources":[],"checksums":[]},"line":16,"column":1}],"totalFrames":1},"seq":1391}
<--   C (stackTrace-42): {"command":"stackTrace","arguments":{"threadId":1342,"startFrame":0,"levels":20},"type":"request","seq":42}
--> R (stackTrace-42): {"type":"response","request_seq":42,"success":true,"command":"stackTrace","body":{"stackFrames":[{"id":1001,"name":"main(int argc, const char ** argv)","source":{"name":"test.cpp","path":"/tmp/test.cpp","sources":[],"checksums":[]},"line":16,"column":1}],"totalFrames":1},"seq":1394}
<--   C (scopes-43): {"command":"scopes","arguments":{"frameId":1000},"type":"request","seq":43}
    <-1124-stack-list-variables 0
    ->1124^done,variables=[{name="argc",arg="1"},{name="argv",arg="1"},{name="s"}]
    ->(gdb)
    1124: elapsed time 2
    <-1125-var-create - * "argc"
    ->1125^done,name="var29",numchild="0",value="1",type="int",thread-id="1",has_more="0"
    ->(gdb)
    1125: elapsed time 1
    <-1126-var-create - * "argv"
    ->1126^done,name="var30",numchild="1",value="0x7fffffffdcc8",type="const char **",thread-id="1",has_more="0"
    ->(gdb)
    1126: elapsed time 0
    <-1127-var-create - * "s"
    ->1127^done,name="var31",numchild="2",value="{...}",type="std::u16string",thread-id="1",has_more="0"
    ->(gdb)
    1127: elapsed time 0
--> R (scopes-43): {"type":"response","request_seq":43,"success":true,"command":"scopes","body":{"scopes":[{"name":"Locals","variablesReference":1000,"expensive":false}]},"seq":1429}
<--   C (variables-44): {"command":"variables","arguments":{"variablesReference":1000},"type":"request","seq":44}
    <-1128-var-list-children --simple-values "var31" 0 1000
    ->1128^done,numchild="2",children=[child={name="var31.public",exp="public",numchild="1",value="",thread-id="1"},child={name="var31.private",exp="private",numchild="3",value="",thread-id="1"}],has_more="0"
    ->(gdb)
    1128: elapsed time 1
    <-1129-var-list-children --simple-values "var31.public" 0 1000
    ->1129^done,numchild="1",children=[child={name="var31.public.npos",exp="npos",numchild="0",value="18446744073709551615",type="const std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::size_type",thread-id="1"}],has_more="0"
    ->(gdb)
    1129: elapsed time 1
    <-1130-var-list-children --simple-values "var31.private" 0 1000
    ->1130^done,numchild="3",children=[child={name="var31.private._M_dataplus",exp="_M_dataplus",numchild="2",type="std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::_Alloc_hider",thread-id="1"},child={name="var31.private._M_string_length",exp="_M_string_length",numchild="0",value="4",type="std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::size_type",thread-id="1"},child={name="var31.private.2_anonymous",exp="<anonymous union>",numchild="1",type="union {...}",thread-id="1"}],has_more="0"
    ->(gdb)
    1130: elapsed time 1
    <-1131-var-create - * "((s)._M_dataplus)._M_p"
    ->1131^done,name="var32",numchild="1",value="0x7fffffffdba0 u\"test\"",type="std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::pointer",thread-id="1",has_more="0"
    ->(gdb)
    1131: elapsed time 1
    <-1132-var-list-children --simple-values "var32" 0 1000
    ->1132^done,numchild="1",children=[child={name="var32.*((s)._M_dataplus)._M_p",exp="*((s)._M_dataplus)._M_p",numchild="0",value="116 u't'",type="char16_t",thread-id="1"}],has_more="0"
    ->(gdb)
    1132: elapsed time 0
    <-1133-var-list-children --simple-values "var30" 0 1000
    ->1133^done,numchild="1",children=[child={name="var30.*argv",exp="*argv",numchild="1",value="0x7fffffffdf65 \"/tmp/test\"",type="const char *",thread-id="1"}],has_more="0"
    ->(gdb)
    1133: elapsed time 1
--> R (variables-44): {"type":"response","request_seq":44,"success":true,"command":"variables","body":{"variables":[{"name":"s","value":"0x7fffffffdba0 u\"test\"","type":"std::u16string","evaluateName":"s","variablesReference":1001},{"name":"argc","value":"1","type":"int","evaluateName":"argc","variablesReference":0,"memoryReference":"0x0000000000000001"},{"name":"argv","value":"0x7fffffffdcc8","type":"const char **","evaluateName":"argv","variablesReference":1002,"memoryReference":"0x00007fffffffdcc8"}]},"seq":1480}
```
</details>

Trass3r:
It's also interesting that when trying to step back again after `record stop` (see above) the error is merely shown in the debug console and in the UI it seems like nothing happened while exceptions are shown right in the editor, also error messages aren't propagated consistently:

```js
<--   C (stepBack-38): {"command":"stepBack","arguments":{"threadId":1519},"type":"request","seq":38}
    <-1111-exec-next --reverse
    ->1111^error,msg="Target native does not support this command."
    ->(gdb)
    1111: elapsed time 0
    Send Event AD7MessageEvent
ERROR: Unexpected GDB output from command "-exec-next --reverse". Target native does not support this command.
--> R (stepBack-38): {"type":"response","request_seq":38,"success":false,"command":"stepBack","message":"Unable to step next. Operation failed with error code 0x80004004.","body":{},"seq":1276}
```

WardenGnaw:
That error message comes from:
https://github.com/microsoft/MIEngine/blob/7bde94d47101873c5649377b4cdacacd4d9aef70/src/MIDebugEngine/Natvis.Impl/Natvis.cs#L362

Something might have went wrong in the natvis you are referencing?

WardenGnaw:
> in the UI it seems like nothing happened while exceptions are shown right in the editor, also error messages aren't propagated consistently:

I don't VS code expects reverse stepping to fail if the capabilities are right.

Let me give this another thought to have it correctly set when we disable record.

Trass3r:
> Something might have went wrong in the natvis you are referencing?

Ah ok I guess _M_p is null then:
```xml
  <Type Name="std::u16string">
    <DisplayString>{_M_dataplus._M_p}</DisplayString>
  </Type>
```

WardenGnaw:
@Trass3r Sorry, I did some cleanup in the project so there are conflicts. We are removing any references to ClrDbg since that is a legacy engine now.

Also to solve the UI issue, you will want to check to see if the reverse state has changed in an evaluate request if it is an `isExecInConsole`, and if it has changed, send a CapabilitiesEvent. 
https://github.com/microsoft/MIEngine/blob/73db59d89b8f0fbfdaa3cee1423e674e092ffc0f/src/OpenDebugAD7/AD7DebugSession.cs#L2033

Trass3r:
> Also to solve the UI issue, you will want to check to see if the reverse state has changed in an evaluate request if it is an `isExecInConsole`, and if it has changed, send a CapabilitiesEvent.

I wouldn't do that. First of all the user could also use the MI version of the command even if it's less likely.
Secondly there may very well be other ways to trigger this, .gdbinit comes to mind as an example (but that should already be covered by the launch check).
Is there no way to notify OpenDebug? That would be the clean solution.

WardenGnaw:
We do not have an existing event for handling this case. 

>I wouldn't do that. First of all the user could also use the MI version of the command even if it's less likely

Then I think it would be always have a check such as:
```
bool canReverse = ((IDebugReversibleEngineProgram160)m_engine).CanReverse() == HRConstants.S_OK;
if (m_canReverseDebug != canReverse)
{
   m_canReverseDebug = canReverse;
   SendEvent(new CapabilitiesEvent() {
      SupportsReverseDebugging = true
   }
}
```

where IDebugReversibleEngineProgram160.CanReverse is
```
int IDebugReversibleEngineProgram160.CanReverse() {
   return m_canReverse ? S_OK : S_FALSE;
}
```

This check isn't expensive, but it is more of a poll based system than an event driven one.

Trass3r:
> Then I think it would be always have a check such as:
> 
> ```
> bool canReverse = ((IDebugReversibleEngineProgram160)m_engine).CanReverse() == HRConstants.S_OK;
> if (m_canReverseDebug != canReverse)
> {
>    m_canReverseDebug = canReverse;
>    SendEvent(new CapabilitiesEvent() {
>       SupportsReverseDebugging = true
>    }
> }
> ```

That's what I have.
https://github.com/microsoft/MIEngine/commit/3d220f7b2351b835012e6822a625b0d7f559c9e8#diff-af838dac0d9374c0d4b17ae4dabf4c46R657-R672

Trass3r:
> Something might have went wrong in the natvis you are referencing?

No it's actually an impl bug, it tries to evaluate the command as a variable:

```js
1: (7984) Natvis.FindType {"Name":"`record","Value":"","TypeName":null,"IsParameter":false,"Children":null,"Client":{"Id":1},"Error":false,"CountChildren":0,"IsChild":false,"Access":0,"IsVisualized":false,"PropertyInfoFlags":0,"IsPreformatted":false,"IsStringType":false,"ThreadContext":{"pc":93824992236155,"sp":0,"Function":"main","TextPosition":{"FileName":"\\tmp\\test.cpp","BeginPosition":{"dwLine":15,"dwColumn":0},"EndPosition":{"dwLine":15,"dwColumn":0}},"From":"","Level":0},"VariableNodeType":0}
1: (7984) ->(gdb)
1: (7984) 1040: elapsed time 1
1: (7985) natvis FormatDisplayString: Value cannot be null.
Parameter name: key
```

Trass3r:
ping

WardenGnaw (commented):
In the `-exec target record-full` scenario, stepping forwards a few times and then backwards.

I'm running into these lines.
https://github.com/microsoft/MIEngine/blob/728f403c563c2c50c46542e9b44271a7f8d5895b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs#L1286-L1290

Investigating.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2380,8 +2465,6 @@ public void HandleIDebugBreakEvent2(IDebugEngine2 pEngine, IDebugProcess2 pProce
 
         public void HandleIDebugExceptionEvent2(IDebugEngine2 pEngine, IDebugProcess2 pProcess, IDebugProgram2 pProgram, IDebugThread2 pThread, IDebugEvent2 pEvent)
         {
-            Stopped(pThread);
```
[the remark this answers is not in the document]
Trass3r: It's called in FireStoppedEvent below.

Trass3r:
> In the `-exec target record-full` scenario, stepping forwards a few times and then backwards.
> I'm running into these lines.
> 
> https://github.com/microsoft/MIEngine/blob/728f403c563c2c50c46542e9b44271a7f8d5895b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs#L1286-L1290

Same here.

Trass3r:
> In the `-exec target record-full` scenario, stepping forwards a few times and then backwards.
> I'm running into these lines.
> 
> https://github.com/microsoft/MIEngine/blob/728f403c563c2c50c46542e9b44271a7f8d5895b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs#L1286-L1290

Same here.
Didn't this come up in another PR as well?

Trass3r:
So the problem is that gdb only outputs `No more reverse-execution history` instead of giving a reason in the stopped event, not sure how to handle this properly:

```sh
1: (62155) <-1057-exec-next --reverse
1: (62156) ->1057^running
1: (62156) ->*running,thread-id="1"
1: (62156) ->(gdb)
1: (62163) ->~"\nNo more reverse-execution history.\n"
1: (62164) ->~"main (argc=1, argv=0x7fffffffdb78) at ../test.cpp:25\n"
1: (62164) ->~"25\t\tstd::u16string s = u\"test\";\n"
1: (62164) ->*stopped,frame={addr="0x000055555555734b",func="main",args=[{name="argc",value="1"},{name="argv",value="0x7fffffffdb78"}],file="../test.cpp",fullname="/tmp/test.cpp",line="25",arch="i386:x86-64"},thread-id="1",stopped-threads="all",core="2"
1: (62164) <-1058-stack-list-frames 0 1000
1: (62165) ->1058^done,stack=[frame={level="0",addr="0x000055555555734b",func="main",file="../test.cpp",fullname="/tmp/test.cpp",line="25",arch="i386:x86-64"}]
...
1: (62175) Send Event AD7ExceptionEvent
```

```cs
                if (breakRequest == BreakRequest.None)
                    _callback.OnStopComplete(thread);
```
prevents the exception but may "break" other cases ofc and also results in bad UX, you click the buttons multiple times and nothing happens.

But anyway that `callback.OnException` msg is displayed in a bad way.
"Exception occurred" makes you think of an exception in your program, not in the debug adapter.

pirpyn:
Ping for interest

Trass3r:
@WardenGnaw any feedback?

Trass3r:
@WardenGnaw I split off the step granularity and added TODOs and FIXMEs for the remaining issues.
Should be in good shape now.
Oh and that `CapabilitiesEvent` is still not in the DAP dll it seems.

WardenGnaw:
@Trass3r `microsoft.visualstudio.shared.vscodedebugprotocol v16.9.50204.1` should have the `CapabilitiesEvent`. 

This will be upgraded when `feature/dotnet` gets merged in.

Trass3r:
@WardenGnaw true, it's there now. But the problem remains that it seemingly initializes the `Capabilities` List properties like `ExceptionBreakpointFilters`, so they get included in the json message even though we don't want to override them, see my FIXME.
Same problem affects `InitializeResponse` which should actually reuse the new `Capabilities`  type I guess.

Trass3r:
@gregg-miskelly Could you also approve the workflow here? I don't know why it suddenly asks for this again for every PR.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1020,6 +1022,27 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
             responder.SetResponse(initializeResponse);
         }
 
+        private bool m_canReverse = false;
+        private void UpdateCapabilities()
+        {
+            bool canReverse = ((IDebugReversibleEngineProgram160)m_engine).CanReverse() == HRConstants.S_OK;
```
gregg-miskelly, line 1028: >((IDebugReversibleEngineProgram160)m_engine [](http://example.com/codeflow?start=30&length=43)

We should handle the case that the engine doesn't implement this, and we should use `m_program` instead of `m_engine`.

gregg-miskelly:
>                 Protocol.SendEvent(new InitializedEvent());

You want to call `UpdateCapabilities` here instead of after LaunchSuspended.

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:3516 in d84213d. [](commit_id = d84213d23462da942a1d5093d9973f2d1133b08e, deletion_comment = False)

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -830,6 +831,7 @@ private void StepInternal(int threadId, enum_STEPKIND stepKind, SteppingGranular
             }
             try
             {
+                ((IDebugReversibleEngineProgram160)m_engine).SetExecuteDirection(stepDirection);
```
gregg-miskelly, line 834: >((IDebugReversibleEngineProgram160)m_engine) [](http://example.com/codeflow?start=16&length=44)

Same comments here -- test m_program, and handle the engine not implementing it.

gregg-miskelly:
> @gregg-miskelly Could you also approve the workflow here? I don't know why it suddenly asks for this again for every PR.

It looks like someone did this for you already right? Or did I miss something?

WardenGnaw:
> > @gregg-miskelly Could you also approve the workflow here? I don't know why it suddenly asks for this again for every PR.
> 
> It looks like someone did this for you already right? Or did I miss something?

Sorry, that was me. I had a moment during my break to poke the "Approve button". I just approved it again.

gregg-miskelly:
@Trass3r I see that your PR still has some TODO comments. Would you consider this ready for a full review? Or are you still working on it.?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1020,6 +1023,30 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
             responder.SetResponse(initializeResponse);
         }
 
+        private bool m_canReverse = false;
+        private void UpdateCapabilities()
+        {
+            if (m_program !is IDebugReversibleEngineProgram160)
+                return;
+
+            bool canReverse = (m_program as IDebugReversibleEngineProgram160).CanReverse() == HRConstants.S_OK;
```
gregg-miskelly, line 1032: >(m_program as IDebugReversibleEngineProgram160) [](http://example.com/codeflow?start=30&length=47)

nit: don't use `as` unless you are going to null check. I would probably suggest:

```
var revProgram = m_program as IDebugReversibleEngineProgram160;
if (revProgram == null)
    return;

...
```

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -540,6 +556,12 @@ private async Task EnsureModulesLoaded()
             }
         }
 
+        private async Task UpdateTargetFeatures()
+        {
+            TargetFeatures = await MICommandFactory.GetTargetFeatures();
+            // TODO: notify OpenDebugAD7 to run AD7DebugSession.UpdateCapabilities
```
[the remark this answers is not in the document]
Trass3r: Wouldn't know how. Just an idea to make things more asynchronous rather than checking reversibility on every stop event.
But yeah, no major further development planned. Just want to get these old PRs merged.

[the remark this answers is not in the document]
gregg-miskelly: Got it. In that case I would probably suggest changing the comment to something like:

`// NOTE: Currently OpenDebugAD7 polls to decide if we can reverse execute. It might make sense to push an event. If we ever create such an event, probably makes sense to use it here also.`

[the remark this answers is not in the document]
Trass3r: Ok. Btw the events are RecordStartedEvent and RecordStoppedEvent for `gdb`. And we check it after starting up to support `rr`.

Trass3r:
Btw, for testing:
```cpp
int main()
{
	int i = 0;
	++i;
	++i;
	return i;
}
```

```sh
$ gdb --batch -ex 'b main' -ex r -ex n -ex 'p i' -ex 'record full' -ex n -ex 'p i' -ex n -ex 'p i' -ex n -ex 'p i' -ex 'reverse-next' -ex 'p i' -ex 'reverse-next' -ex 'p i' -ex 'reverse-next' -ex 'p i' test

3           int i = 0;
completed.
+p i
$1 = 17
+record full
+target record-full
+n
4           ++i;
completed.
+p i
$2 = 0
+n
5           ++i;
completed.
+p i
$3 = 1
+n
6               return i;
completed.
+p i
$4 = 2
+reverse-next
+next 
5           ++i;
completed.
+p i
$5 = 1
+reverse-next
+next 
4           ++i;
completed.
+p i
$6 = 0
+reverse-next
+next 

No more reverse-execution history.
main () at /tmp/test.cpp:3
3           int i = 0;
completed.
+p i
$7 = 17
```

Trass3r:
Any ideas for improving the UX when reaching the boundaries of the recorded data:

https://github.com/microsoft/MIEngine/pull/1037/files#diff-e6a6b2741c810dcffe4bf13d4c697e2a326b9aa13752825ee0544977797273b0R1409-R1416

image

gregg-miskelly:
> Any ideas for improving the UX when reaching the boundaries of the recorded data:

First, I would suggest trying a quick hack where you patch up the properties in `OpenDebugStoppedEvent` to see what it will take to make VS Code display something good -- do we just need to fix the `Description`? The `Description` and then `Reason`? etc.

If you can get that to a place you want, then it will give me a better idea in the MIEngine side.

Trass3r:
Created a gdb ticket for the missing reason: https://sourceware.org/bugzilla/show_bug.cgi?id=29260

Trass3r:
@gregg-miskelly but I guess that issue can be captured in a ticket and fixed separately.

Trass3r:
@gregg-miskelly but I guess that issue can be captured in a ticket and fixed separately?!

Trass3r:
> I also created a gdb ticket for the missing reason: https://sourceware.org/bugzilla/show_bug.cgi?id=29260

Finally fixed.
https://github.com/bminor/binutils-gdb/commit/37f54063ed0daaffbd52f69efbdb4c8e0a9c9cca

jeras:
Hi @Trass3r,

I would like support for reverse debugging for my project, do you have a fork I could test?

My project is a GDB stub that attaches to a HDL simulator written in SystemVerilog.
I will skip the details for now, but overall it would behave in a similar fashion to the [record/replay functionality in QEMU](https://qemu-project.gitlab.io/qemu/system/replay.html). So this does not rely on a specific CPU (x86, ARM) hardware tracing capability.

On the level of the GDB remote serial protocol, this capability is exposed by responding with `ReverseStep+;ReverseContinue+;` to the `qSupported` packet.

```
1: (375) ->&"  [remote] Sending packet: $qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;QThreadOptions+;no-resumed+;memory-tagging+#e7\n"
1: (375) ->&"  [remote] Received Ack\n"
1: (376) ->&"  [remote] Packet received: QStartNoAckMode+;ReverseContinue+;ReverseStep+;binary-upload-;error-message+;hwbreak+;multiprocess-;swbreak+\n"
1: (376) ->&"  [remote] packet_ok: Packet qSupported (supported-packets) is supported\n"
```

I got as far a checking whether I get the expected response when asking with `-list-target-features` over MI:
```
-exec -list-target-features
1: (163377) <-1034-list-target-features
1: (163377) ->1034^done,features=["async","reverse"]
1: (163377) ->(gdb)
1: (163377) 1034: elapsed time 0
result-class: done
features: [async,reverse]
```

I will also investigate another path, but it is ARM specific, so not really a good fit for my project which is centered around RISC-V but may support other CPU architectures with some generalization effort.

~I know reverse debugging is possible with the [`cortex-debug`](https://github.com/Marus/cortex-debug) extension, combined with a GDB from some ARM debug hardware vendor.~ Actually I em not sure what DAP implementation is used [here](https://www.justinmklam.com/posts/2017/10/vscode-debugger-setup/).
~But it does not seem to work with QEMU.~

I already have a [setup](https://github.com/jeras/zephyr-qemu-test-app) with QEMU ARM running a Zephyr OS application, I can use for testing.

PR #1038, opened 2020-08-27
avoid "Unable to open file" popups caused by missing source code

Trass3r:
Preliminary solution until full disassembly debugging is implemented.
@WardenGnaw


src/OpenDebugAD7/TextPositionTuple.cs
```diff
@@ -41,6 +41,12 @@ public static TextPositionTuple GetTextPositionOfFrame(PathConverter converter,
                     Path = convertedFilePath,
                     Name = Path.GetFileName(convertedFilePath)
                 };
+                if (!File.Exists(convertedFilePath))
+                {
+                    // mark the stack frame to be skipped by default
```
Trass3r, line 46: it can still be selected manually

WardenGnaw: Can you share an image of the dialog of it demphasized?

src/OpenDebugAD7/TextPositionTuple.cs
```diff
@@ -41,6 +41,12 @@ public static TextPositionTuple GetTextPositionOfFrame(PathConverter converter,
                     Path = convertedFilePath,
                     Name = Path.GetFileName(convertedFilePath)
                 };
+                if (!File.Exists(convertedFilePath))
```
gregg-miskelly, line 44: This seems like a reasonable approach to me, but I think you want to move this to HandleStackTraceRequestAsync. The other caller of this function (FireStoppedEvent) will not care about this.

It is probably also worth doing a parrellel for (or similar) over all the frames rather than doing this synchronously, and using a timeout since hitting random parts of the file system can hang indefinitely.

src/OpenDebugAD7/TextPositionTuple.cs
Trass3r: https://github.com/microsoft/vscode/issues/65025#issuecomment-453645164
stack

src/OpenDebugAD7/TextPositionTuple.cs
Trass3r: Well in theory it does but you are right, in practice vscode does not use the source property of the stopped event.

Trass3r: Suggestions welcome for how to do the timeout thing.

gregg-miskelly: I thought `Parrellel.ForEach` had that option, but I guess I was wrong. I think you would need to instead do:

```
        protected override async void HandleStackTraceRequestAsync(...)
        {
            try
            {
                ...

                CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
                try
                {
                    CancellationToken cancellationToken = cancellationTokenSource.Token;
                    Task[] frameTasks = inputs.Select(x => Task.Run(() =>
                    {
                        cancellationToken.ThrowIfCancellationRequested();
                        ...
                    })).ToArray();

                    Task allFramesComplete = Task.WhenAll(frameTasks);
                    Task timeoutTask = Task.Delay(20000, cancellationToken);
                    try
                    {
                        await Task.WhenAny(allFramesComplete, timeoutTask);
                    }
                    catch (Exception e)
                    {
                        Debug.Fail("Ignoring exception computing frame state: " + e.ToString());
                    }
                }
                finally
                {
                    cancellationTokenSource.Cancel();
                    cancellationTokenSource.Dispose();
                }
            }
            catch(Exception e)
            {
                responder.SetError(...);
            }
        }            
```

Trass3r: Not sure if it's worth it though since it's gonna be superseded by #1041.

gregg-miskelly: I am not sure I understand what you mean - I think in general if we are going to add File.Exists checks on each frame we should do something like I have.

Trass3r: Yeah true, so in #1041 we're mostly interested in the lowest frame but I check the other ones too to mark them.

Trass3r: Seems awfully complicated though just to handle an unlikely(?) event.

Trass3r: Please check my update.

Trass3r:
@gregg-miskelly maybe something like this?

Trass3r:
@gregg-miskelly maybe something like this?

Trass3r:
@gregg-miskelly maybe something like this?

src/OpenDebugAD7/TextPositionTuple.cs
```diff
@@ -41,6 +41,7 @@ public static TextPositionTuple GetTextPositionOfFrame(PathConverter converter,
                     Path = convertedFilePath,
                     Name = Path.GetFileName(convertedFilePath)
                 };
+
```
gregg-miskelly, line 44: nit: please remove this file from the review since it is just a whitespace change.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1476,6 +1476,19 @@ protected override void HandleStackTraceRequestAsync(IRequestResponder<StackTrac
                         });
                     }
 
+                    Parallel.ForEach(response.StackFrames, frame =>
```
gregg-miskelly, line 1479: We should probably put this under an `if (!IsClientVS)` block.

gregg-miskelly (approved):
Looks good to me aside from those small changes.

src/OpenDebugAD7/TextPositionTuple.cs
Trass3r: Ah a remnant from previous changes.

gregg-miskelly (approved):

gregg-miskelly:
@WardenGnaw do you want to review this?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1476,6 +1476,22 @@ protected override void HandleStackTraceRequestAsync(IRequestResponder<StackTrac
                         });
                     }
 
+                    if (!IsClientVS)
+                    {
+                        Parallel.ForEach(response.StackFrames, frame =>
```
chuckries, line 1481: my understanding is that Paralle.ForEach uses the default Task Schedular to spawn it's delegates, in which case we are spawning a task that spawns a task for each stack frame. Wouldn't we get the same level of parallelism by using a straight for loop to spawn tasks, collecting them, and rolling them up in a Task.WaitAll or something?

cc @gregg-miskelly

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1476,6 +1476,22 @@ protected override void HandleStackTraceRequestAsync(IRequestResponder<StackTrac
...
+                        {
+                            if (frame.Source == null)
+                                return;
+                            var task = new Task<bool>(() => File.Exists(frame.Source.Path));
```
chuckries, line 1485: if we do keep this, `Task.Run(...)` is preferrable to `new Task(...)` + `Task.Start()`

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: Yes, that was my suggested fix here: https://github.com/microsoft/MIEngine/pull/1038#discussion_r479437530

chuckries: ah my bad, sorry.

Trass3r: True, WaitAll should make things easier.
```cs
var frameTasks = response.StackFrames.Select(frame => Task.Run(() => {
    if (frame.Source == null || File.Exists(frame.Source.Path))
        return;

    // mark the stack frame to be skipped by default
    frame.Source.PresentationHint = Source.PresentationHintValue.Deemphasize;
})).ToArray();
try Task.WaitAll(frameTasks, 1000);
```

chuckries: @Trass3r this proposed change has the exact same issue @gregg-miskelly originally advocated against: if File.Exists hangs, one of your frameTasks will hang and Task.WaitAll will hang.

Trass3r: Then I don't get what the [WaitAll](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.waitall#System_Threading_Tasks_Task_WaitAll_System_Threading_Tasks_Task___System_Int32_) overload with timeout is for.

And wouldn't the await whenall solution leave zombie tasks behind in case of hangs?

chuckries: Ah my apologies, I didn't see a the timeout in `WaitAll`. That timeout will work as you expect. Sorry again.

One potential issue still exists here: if the the timeout is triggered, the code will move past the `Task.WaitAll`, but the delegate will still be running and may complete later and try to set `frame.Source.PresentationHint` at a unknown time in the futue.

A fix for this can be seen in gregg-miskelly's initial suggestion which cancels a `CancellationTokenSource` after getting past the `Task.WaitAll`. The delegate is gated by the cancellation and doesn't try to set `frame.Source.PresentationHint` if the TokenSource was canceled.

WardenGnaw:
This [commit (https://github.com/microsoft/MIEngine/commit/c9b7029cae2a6b40e78d73a64266fb6ae2778bd9)
converts it to be using `async/await` with Tasks instead of `Parallel.ForEach`

Still need to test with VS Code.

WardenGnaw:
@Trass3r I'm getting a remote rejected on pushing to the deemphasize branch.

Trass3r:
Fixed.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1305,198 +1305,224 @@ protected override void HandlePauseRequestAsync(IRequestResponder<PauseArguments
...
                             }
-                        });
-                    }
+                        }
 
-                    response.TotalFrames = (int)frameEnumInfo.TotalFrames;
+                        response.TotalFrames = (int)frameEnumInfo.TotalFrames;
+                    }
                 }
-            }
```
Trass3r, line 1481: This function is one of the largest and most deeply nested ones.
If you touch all lines anyway maybe it could be split up and/or the nesting reduced.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1305,182 +1305,224 @@ protected override void HandlePauseRequestAsync(IRequestResponder<PauseArguments
...
-                    response.TotalFrames = (int)frameEnumInfo.TotalFrames;
                 }
-            }
 
-            responder.SetResponse(response);
+                responder.SetResponse(response);
+            }
+            catch (Exception e)
+            {
+                responder.SetError(new ProtocolException(e.ToString()));
```
gregg-miskelly, line 1524: >.ToString() [](start = 58, length = 11)

`ToString()` will include the stack trace of the thrown exception. Do you want `Message` instead?

WardenGnaw:
Old Behavior when stepping into a function you do not have source:
**Popup**:
image

New Behavior:
Nothing but callstack shows there is no source for the top frame.
image

ntd252:
Hello, I hope my comment is in the right place.
I'm using MinGW-w64 (from MSYS2), debug and build settings are showed in video. I get the error:
`Unable to open 'crtexe.c': Unable to read file 'c:\_\M\mingw-w64-crt-git\src\mingw-w64\mingw-w64-crt\crt\crtexe.c' (Error: Unable to resolve non-existing file 'c:\_\M\mingw-w64-crt-git\src\mingw-w64\mingw-w64-crt\crt\crtexe.c').
`
I hope the final fix will come. Thank you!
https://user-images.githubusercontent.com/36019046/106465437-89076000-64cc-11eb-93d3-3ad49b80bf99.mp4

Trass3r:
@WardenGnaw wasn't this finished already (apart from Gregg's [comment](https://github.com/microsoft/MIEngine/pull/1038/files/d16d500968923cd1cfae6ab4033785d49e13d3ec#r562270771))?

daniel-brosche:
is this still in progress?

Trass3r:
It's been ready for a while, rebased it and addressed the last items..

daniel-brosche:
@WardenGnaw please review this feature, there are so many poeple waiting for this (since years) -> https://github.com/microsoft/vscode-cpptools/issues/811#issuecomment-880359537

evdenaxas:
Why is this still an issue? Do MS people not debug their stuff? 
This is still alive and well inthe latest vscode install in ubuntu

Trass3r:
Something seems to have changed though in vscode. For me when stepping over a `throw` I don't get the popup anymore.
It just opens the file showing `Could not load source './signal/../sysdeps/unix/sysv/linux/raise.c': 'SourceRequest' not supported..`

evdenaxas:
Which version @Trass3r ? Mine is 1.58.1
Version: 1.58.1
Commit: 2aeda6b18e13c4f4f9edf6667158a6b8d408874b
Date: 2021-07-13T06:20:02.397Z
Electron: 12.0.13
Chrome: 89.0.4389.128
Node.js: 14.16.0
V8: 8.9.255.25-electron.0
OS: Linux x64 5.8.0-59-generic

Trass3r:
Same version. On WSL2.

daniel-brosche:
@Trass3r + @WardenGnaw is there still something missing, what is needed to get this done?

diditforlulz273:
would you merge it at last? I'm aging waiting for it, and I'm not a wine to get better with years.

Greg-MV:
Guys we'd really like this feature.

PR #1039, opened 2020-08-28
implement register view

### LINKED ISSUE #1029: Display registers
As described in https://github.com/microsoft/MIEngine/issues/816#issuecomment-496785677.

Trass3r:
fixes #1029 @andrewcrawley
implemented as a second scope in the variables view
tested on x64
setting them does not work, suggestions welcome

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -66,6 +66,7 @@ namespace OpenDebugAD7
         private VariableManager m_variableManager;
 
         private static Guid s_guidFilterAllLocalsPlusArgs = new Guid("939729a8-4cb0-4647-9831-7ff465240d5f");
+        private static Guid s_guidFilterRegisters = new Guid("223ae797-bd09-4f28-8241-2763bdc5f713");
 
         private int m_nextModuleHandle = 1;
         private readonly Dictionary<IDebugModule2, int> m_moduleMap = new Dictionary<IDebugModule2, int>();
@@ -542,19 +543,33 @@ namespace OpenDebugAD7
             m_logger.WriteLine(category, prefixString + text);
         }
 
-        private VariablesResponse VariablesFromFrame(IDebugStackFrame2 frame, uint radix)
+        private VariablesResponse VariablesFromFrame(VariablesRef vref, uint radix)
         {
-            VariablesResponse response = new VariablesResponse();
+            var frame = vref.StackFrame;
+            var scope = vref.Scope;
+
+            var response = new VariablesResponse();
+
+            Guid filter = Guid.Empty;
+            switch (scope)
+            {
+            case VariablesScope.Locals:
+                filter = s_guidFilterAllLocalsPlusArgs;
+                break;
+            case VariablesScope.Registers:
+                filter = s_guidFilterRegisters;
+                break;
+            }
 
             uint n;
             IEnumDebugPropertyInfo2 varEnum;
-            if (frame.EnumProperties(enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_PROP, radix, ref s_guidFilterAllLocalsPlusArgs, 0, out n, out varEnum) == HRConstants.S_OK)
+            if (frame.EnumProperties(GetDefaultPropertyInfoFlags(), radix, ref filter, 0, out n, out varEnum) == HRConstants.S_OK)
             {
-                DEBUG_PROPERTY_INFO[] props = new DEBUG_PROPERTY_INFO[1];
+                var props = new DEBUG_PROPERTY_INFO[1];
                 uint nProps;
                 while (varEnum.Next(1, props, out nProps) == HRConstants.S_OK)
                 {
-                    response.Variables.Add(m_variableManager.CreateVariable(props[0].pProperty, GetDefaultPropertyInfoFlags()));
+                    response.Variables.Add(m_variableManager.CreateVariable(ref props[0], GetDefaultPropertyInfoFlags()));
                 }
             }
 
@@ -564,10 +579,7 @@ namespace OpenDebugAD7
         public enum_DEBUGPROP_INFO_FLAGS GetDefaultPropertyInfoFlags()
         {
             enum_DEBUGPROP_INFO_FLAGS flags =
-                enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_NAME |
-                enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_VALUE |
-                enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_TYPE |
-                enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_ATTRIB |
+                enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_STANDARD |
                 enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_PROP |
                 enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_FULLNAME |
                 (enum_DEBUGPROP_INFO_FLAGS)enum_DEBUGPROP_INFO_FLAGS110.DEBUGPROP110_INFO_FORCE_REAL_FUNCEVAL;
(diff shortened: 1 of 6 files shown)


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1318,11 +1330,22 @@ protected override void HandleScopesRequestAsync(IRequestResponder<ScopesArgumen
...
-                            VariablesReference = m_variableManager.Create(frame),
+                            VariablesReference = m_variableManager.Create(new VariablesRef() { StackFrame = frame, Scope = VariablesScope.Locals }),
+                            PresentationHint = Scope.PresentationHintValue.Locals,
                             Expensive = false
                         });
                     }
                 }
+
+                // registers should always be present
+                // and it's too expensive to read all values just to add the scope
```
Trass3r, line 1341: you could argue the Locals check should be avoided too

Trass3r:
@WardenGnaw ?

src/OpenDebugAD7/VariableManager.cs
```diff
@@ -14,9 +14,21 @@ internal class VariableEvaluationData
...
+    {
+        Locals,
+        Registers
+    }
+
+    internal class VariablesRef
+    {
+        internal IDebugStackFrame2 StackFrame;
+        internal VariablesScope Scope;
+    }
```
Trass3r, line 27: was a preliminary name/solution, feel free to suggest something better

Trass3r:
ping

WardenGnaw:
Fixed branch merge issues and and a fix from other changes (memory context) that came in.

See https://github.com/microsoft/MIEngine/tree/dev/waan/registers and the specific commit https://github.com/microsoft/MIEngine/commit/753da8203d7fd4c90d07ededd1661e10defd0d4d

Trass3r:
What about setting the register values?

WardenGnaw:
> What about setting the register values?

The request will reach [here](https://github.com/microsoft/MIEngine/blob/f3fafeb3dd1e7f50390d81e1348d1b29b31150b2/src/OpenDebugAD7/AD7DebugSession.cs#L1535) but it will call [AD7RegGroupProperty.SetValueAsString](https://github.com/microsoft/MIEngine/blob/f3fafeb3dd1e7f50390d81e1348d1b29b31150b2/src/MIDebugEngine/AD7.Impl/AD7RegProperty.cs#L134) which is not implemented yet.

src/OpenDebugAD7/VariableManager.cs
gregg-miskelly: I might suggest calling this 'Scope' and the enum 'ScopeKind'. If 'Scope' creates name conflicts, this could also be 'ScopeState'

---
In reply to: [493603479](https://github.com/microsoft/MIEngine/pull/1039#discussion_r493603479) [](ancestors = 493603479)

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1550,9 +1573,9 @@ protected override void HandleVariablesRequestAsync(IRequestResponder<VariablesA
             Object container;
             if (m_variableManager.TryGet(reference, out container))
             {
-                if (container is IDebugStackFrame2)
+                if (container is VariablesRef)
```
gregg-miskelly, line 1576: >container is VariablesRef [](start = 20, length = 25)

nit: should be `container is VariablesRef variablesRef` and then replace the `container as VariablesRef` with `variablesRef`.

gregg-miskelly (changes requested):
Aside from those smaller things, LGTM. I am assuming you haven't tested this with VS? If not, our team should.

Trass3r:
No only tested with VSCode.

src/OpenDebugAD7/AD7DebugSession.cs
Trass3r: @gregg-miskelly ?

gregg-miskelly: I wouldn't expect it to make too much of a difference because I would expect that if a client make a `scopes` request it is going to turn around and request children from each returned scope where `Expensive` is null or false. But I also don't know why this check was made -- it seems totally reasonable to always display locals and just have it be empty if there aren't any.

gregg-miskelly (approved):
LGTM, aside from that someone needs to test this with VS. @WardenGnaw do you want to give this PR a quick test?

WardenGnaw:
Yep. I can validate the newer commits.

Just one note is that there is a version of GDB that crashes when sending `-data-list-register-values` and we are unable to recover since GDB has died.

gregg-miskelly:
> Just one note is that there is a version of GDB that crashes when sending `-data-list-register-values` and we are unable to recover since GDB has died.

To make sure I understand, this should only happen in VS if one opens the registers window, and only in VS Code if one expands the registers scope. Correct?

WardenGnaw:
> > Just one note is that there is a version of GDB that crashes when sending `-data-list-register-values` and we are unable to recover since GDB has died.
> 
> To make sure I understand, this should only happen in VS if one opens the registers window, and only in VS Code if one expands the registers scope. Correct?

Yes. The fix is to use another version of GDB, it was an error on MinGW x64 running on 7.4.1. Higher versions fixed this issue.

WardenGnaw:
If we get to creating a register scope variable, it will need to have its presentation hint set to:
```
PresentationHint = new VariablePresentationHint() { Attributes = VariablePresentationHint.AttributesValue.ReadOnly }
```
Or else it will crash the debugger when someone tries to set a variable. 

https://github.com/microsoft/MIEngine/blob/2c8b1b7b04e1805045965356d51c8f29c7ab6f49/src/OpenDebugAD7/VariableManager.cs#L82-L90

gregg-miskelly:
RE: If we get to creating a register scope variable, it will need to have its presentation hint set to...

Note that I would NOT special case registers here. We can instead check for the readonly attribute value...

src/OpenDebugAD7/VariableManager.cs
```diff
@@ -88,6 +88,11 @@ internal Variable CreateVariable(ref DEBUG_PROPERTY_INFO propertyInfo, enum_DEBU
                 EvaluateName = propertyInfo.bstrFullName,
                 MemoryReference = memoryReference
             };
+
+            if (propertyInfo.dwAttrib.HasFlag(enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_VALUE_READONLY))
+                v.PresentationHint = new VariablePresentationHint() { Attributes = VariablePresentationHint.AttributesValue.ReadOnly };
```
Trass3r, line 93: The thing is it does get sent but VSCode ignores it.
```json
{"name":"rax","value":"0x0","presentationHint":{"attributes":["readOnly"]}
```

Trass3r: It's also interesting that this check doesn't trigger.
https://github.com/microsoft/MIEngine/blob/3c480786f26ec5340ff16dfeb8d7fd8cde077a9d/src/OpenDebugAD7/AD7DebugSession.cs#L1707-L1713

Instead I get `Stopping due to fatal error: NullReferenceException: Object reference not set to an instance of an object
`.

gregg-miskelly: > The thing is it does get sent but VSCode ignores it.

Yes, this was more to make VS happy.

> Instead I get Stopping due to fatal error: NullReferenceException: Object reference not set to an instance of an object .

The issue is that [this code](https://github.com/Microsoft/MIEngine/blob/896c72d633b39a0571b9599d15c5a8fd0f1fcf58/src/MIDebugEngine/AD7.Impl/AD7RegProperty.cs#L59-L79) doesn't provide a 'pProperty'. The easy fix is to delete line 1702 and 1703, and change the rest of the code in that block to use `props` instead of `propertyInfo` (or just rename `props`).

Trass3r: Doing that brings us to the `Error_VariableNotFound` part.
I guess because registers are grouped ("CPU", "SSE", "AVX", ...) and without 1702-3 it looks in the wrong level. But all that property code is Greek to me in general.

gregg-miskelly: It shouldn't change the 'VariableNotFound' code. What is going on is that for some reason this code was calling `GetPropertyInfo` when it should already have a property info from calling `Next`. So this just removes the refetch of the same data.

Trass3r: I meant that it reports VariableNotFound instead of variable is read-only.

WardenGnaw: The issue is `props` has no data so it can not match the register that was sent from the `HandleSetVariableRequestAsync`.
image

I believe this needs to also support the `readOnly` attribute for variables.

WardenGnaw:
Besides trying to set a value for the register. Looks good on VS Code and VS.

WardenGnaw:
This [commit](https://github.com/microsoft/MIEngine/commit/ef879b9a5b141daa17beec271d8734ddb04af7c6) fixes up the error NullReferenceException and Error_VariableNotFound

and will display this error:

image

Trass3r:
Btw I think I enabled collaboration on all PRs so you could also push to this branch.

WardenGnaw:
I must have mistyped something wrong on my end on another PR so I haven't tried pushing recently. Thanks for letting me know!

WardenGnaw:
VS:
image
VSCode:
image

@gregg-miskelly Any comments on https://github.com/microsoft/MIEngine/pull/1039/commits/8b32a3d5b50859798c9fbfcc37d3f101ebe3d512?

gregg-miskelly:
> @gregg-miskelly Any comments on [8b32a3d](https://github.com/microsoft/MIEngine/commit/8b32a3d5b50859798c9fbfcc37d3f101ebe3d512)?

Sorry, missed this earlier. LGTM

WardenGnaw:
Again, Thank you for the contribution @Trass3r!

PR #1040, opened 2020-08-28
add modules view support

Trass3r:
fixes microsoft/vscode-cpptools#5999
Unfortunately not testable for me since it's not yet implemented: microsoft/vscode#59329
@WardenGnaw

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -8,13 +8,13 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reflection;
-using System.Text;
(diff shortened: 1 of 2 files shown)


WardenGnaw:
@Trass3r We are validating this now, this should be testable with VS. 

 If this works andonce VS Code creates the UI and accepts the ModuleEvent, it should work there too.

WardenGnaw:
Linking https://github.com/microsoft/MIEngine/issues/1050 because TimeStamps are not set.

WardenGnaw:
This will be check in via https://github.com/microsoft/MIEngine/pull/1054.

@Trass3r Thank you again for the initial work! There will still be a need on the VS Code side to have a UI to display these modules.

Issue #816, opened 2018-12-14
[VS Code] Add memory view, assembly view and debugging for C/C++


Izikiel:
Referencing [this issue](https://github.com/Microsoft/vscode-cpptools/issues/206) in the cpptools repo, I would be willing to try and implement the feature if I could get a few examples or pointers of how the engine code interacts with gdb and vs code.

gregg-miskelly:
This is a fairly large amount of work to take on. So if you have the cycles, that would be very cool, but you might find this a bit scary. 

You wouldn't actually need to interact with GDB at all -- all that work is already done because this project supports disassembly and memory in full Visual Studio.

Here is the work that is needed:
1. The C++ extension (or VS Code itself) needs to provide a disassembly window. Assuming this is done in the extension, this would need to work by having the C++ extension provide HTML/CSS back to VS Code for VS Code to render. I know that VS Code has some support for this now, but I haven't looked into what exactly the API is, and from a quick search I couldn't find it. I recently have been looking at adding a 'set next statement' command for C# which is **slightly** similar (it adds a command to the context menu, which we would probably want as one of the gestures for bringing up the disassembly window -- https://github.com/gregg-miskelly/omnisharp-vscode/tree/SetNextStatement) that might get you slightly started. This is probably the single biggest work item.
2. A protocol needs to be worked out between the debug adapter and the UI. This doesn't necessarily need to be an official part of the protocol. But it probably does make sense to at least propose something to avoid breaking changes if VS Code later creates something of their own. The protocol is at https://github.com/Microsoft/debug-adapter-protocol/blob/gh-pages/debugAdapterProtocol.json.
3. The [OpenDebugAD7](https://github.com/Microsoft/MIEngine/tree/master/src/OpenDebugAD7) layer needs to implement the new request to obtain disassembly. To actually fetch diassembly, use IDebugProgram2.GetDisassemblyStream which the MIEngine already implements [here](https://github.com/Microsoft/MIEngine/blob/29c6569861b2042c43f04d8d78b2566db2e2ef1f/src/MIDebugEngine/AD7.Impl/AD7Engine.cs#L935).

Izikiel:
Ok, I will check how to create a webview and populate it

haneefdm:
Hi @Izikiel did you make any progress? I am interested in the same features but for me, a register view is very important. If you have already started, I want to help. I am an experienced C/C++/C# developer on desktop and embedded software. 

[...]

PR #1041, opened 2020-08-28
Implement disassembly debugging

Trass3r:
Proof of concept of disassembly debugging.
Will fix part of #816 and microsoft/vscode-cpptools#206.

Working:
- show disassembly if source is not available
- step through assembly code

Not working (in order of importance):
- support for `-data-disassemble -a` to get the entire current function: https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI-Data-Manipulation.html
- setting breakpoints
- goto
- show disassembly with source code when it is available
- support for different scopes

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/Disassembly.cs
+++ b/src/MIDebugEngine/Engine.Impl/Disassembly.cs
@@ -335,16 +335,22 @@ namespace Microsoft.MIDebugEngine
             return DecodeSourceAnnotatedDisassemblyInstructions(process, results.Find<ResultListValue>("asm_insns").FindAll<TupleValue>("src_and_asm_line"));
         }
 
+        // https://sourceware.org/gdb/current/onlinedocs/gdb/GDB_002fMI-Data-Manipulation.html#GDB_002fMI-Data-Manipulation
+        // NB: the spec is wrong, offset and func-name are optional
         private static DisasmInstruction[] DecodeDisassemblyInstructions(TupleValue[] items)
         {
-            DisasmInstruction[] instructions = new DisasmInstruction[items.Length];
+            var instructions = new DisasmInstruction[items.Length];
             for (int i = 0; i < items.Length; i++)
             {
-                DisasmInstruction inst = new DisasmInstruction();
+                var inst = new DisasmInstruction();
                 inst.Addr = items[i].FindAddr("address");
                 inst.AddressString = items[i].FindString("address");
-                inst.Symbol = items[i].TryFindString("func-name");
-                inst.Offset = items[i].Contains("offset") ? items[i].FindUint("offset") : 0;
+                int offset = items[i].Contains("offset") ? items[i].FindInt("offset") : 0;
+                if (offset > 0) // negative offset seen after throw
+                {
+                    inst.Offset = (uint)offset;
+                    inst.Symbol = items[i].TryFindString("func-name");
+                }
                 inst.Opcode = items[i].FindString("inst");
                 inst.CodeBytes = items[i].TryFindString("opcodes");
                 inst.Line = 0;
@@ -363,11 +369,15 @@ namespace Microsoft.MIDebugEngine
                 uint lineOffset = 0;
                 foreach (var asm_item in asm_items.Content)
                 {
-                    DisasmInstruction disassemblyData = new DisasmInstruction();
+                    var disassemblyData = new DisasmInstruction();
                     disassemblyData.Addr = asm_item.FindAddr("address");
                     disassemblyData.AddressString = asm_item.FindString("address");
-                    disassemblyData.Symbol = asm_item.TryFindString("func-name");
-                    disassemblyData.Offset = asm_item.Contains("offset") ? asm_item.FindUint("offset") : 0;
+                    int offset = asm_item.Contains("offset") ? asm_item.FindInt("offset") : 0;
+                    if (offset > 0) // negative offset seen after throw
+                    {
+                        disassemblyData.Offset = (uint)offset;
+                        disassemblyData.Symbol = asm_item.TryFindString("func-name");
+                    }
                     disassemblyData.Opcode = asm_item.FindString("inst");
                     disassemblyData.CodeBytes = asm_item.TryFindString("opcodes");
                     disassemblyData.Line = line;
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
@@ -539,6 +539,7 @@ namespace OpenDebugAD7
         }
 
         private CurrentLaunchState m_currentLaunchState;
+        private bool m_currentFrameHasSourceCode;
 
         private void SendMessageEvent(MessagePrefix prefix, string text)
         {
@@ -623,29 +624,32 @@ namespace OpenDebugAD7
         private void StepInternal(int threadId, enum_STEPKIND stepKind, enum_STEPUNIT stepUnit, string errorMessage)
         {
             // If we are already running ignore additional step requests
-            if (m_isStopped)
-            {
-                IDebugThread2 thread = null;
-                lock (m_threads)
-                {
-                    if (!m_threads.TryGetValue(threadId, out thread))
-                    {
-                        throw new AD7Exception(errorMessage);
-                    }
-                }
+            if (!m_isStopped)
+                return;
 
-                BeforeContinue();
-                ErrorBuilder builder = new ErrorBuilder(() => errorMessage);
-                m_isStepping = true;
-                try
+
+            IDebugThread2 thread = null;
+            lock (m_threads)
+            {
+                if (!m_threads.TryGetValue(threadId, out thread))
                 {
-                    builder.CheckHR(m_program.Step(thread, stepKind, stepUnit));
+                    throw new AD7Exception(errorMessage);
                 }
-                catch (AD7Exception)
-                {
-                    m_isStopped = true;
-                    throw;
-                }
+            }
+
+            BeforeContinue();
+            ErrorBuilder builder = new ErrorBuilder(() => errorMessage);
+            m_isStepping = true;
+            if (!m_currentFrameHasSourceCode)
+                stepUnit = enum_STEPUNIT.STEP_INSTRUCTION;
+            try
+            {
+                builder.CheckHR(m_program.Step(thread, stepKind, stepUnit));
+            }
+            catch (AD7Exception)
+            {
+                m_isStopped = true;
+                throw;
             }
         }
 
@@ -1422,7 +1426,7 @@ namespace OpenDebugAD7
             int startFrame = responder.Arguments.StartFrame.GetValueOrDefault(0);
             int levels = responder.Arguments.Levels.GetValueOrDefault(0);
 
-            StackTraceResponse response = new StackTraceResponse()
+            var response = new StackTraceResponse()
             {
                 TotalFrames = 0
             };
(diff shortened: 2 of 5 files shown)


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -2252,6 +2258,8 @@ private Task<string> ResetConsole()
 
         public bool IsChildProcessDebugging => _childProcessHandler != null;
 
+        public string[] Features { get; private set; }
```
gregg-miskelly, line 2261: nit: instead of exposing an array, use a `IReadOnlyCollection<string>`. You probably also want to initialize it to an empty collection since we will only have it for GDB.

src/MIDebugEngine/Engine.Impl/Disassembly.cs
```diff
@@ -334,16 +334,22 @@ internal async Task<IEnumerable<DisasmInstruction>> Disassemble(DebuggedProcess
             return DecodeSourceAnnotatedDisassemblyInstructions(process, results.Find<ResultListValue>("asm_insns").FindAll<TupleValue>("src_and_asm_line"));
         }
 
+        // https://sourceware.org/gdb/current/onlinedocs/gdb/GDB_002fMI-Data-Manipulation.html#GDB_002fMI-Data-Manipulation
+        // NB: the spec is wrong, offset and func-name are optional
         private static DisasmInstruction[] DecodeDisassemblyInstructions(TupleValue[] items)
         {
-            DisasmInstruction[] instructions = new DisasmInstruction[items.Length];
+            var instructions = new DisasmInstruction[items.Length];
```
gregg-miskelly, line 341: Why change to use `var`?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
Trass3r: lldb doesn't implement that part of the MI?

src/MIDebugEngine/Engine.Impl/Disassembly.cs
Trass3r: The type is redundant.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
gregg-miskelly: You put your code that populates under an 'if (gdb)' path. I don't know for sure if LLDB supports `-list-features` or not, but your code will not assign to it.

Trass3r: Yeah it's probably better as a virtual method in the factory like in the other PR.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -525,29 +527,32 @@ private void SetAllExceptions(enum_EXCEPTION_STATE state)
...
-                }
-                catch (AD7Exception)
-                {
-                    m_isStopped = true;
-                    throw;
-                }
+            BeforeContinue();
+            ErrorBuilder builder = new ErrorBuilder(() => errorMessage);
+            m_isStepping = true;
+            if (!m_currentFrameHasSourceCode)
```
gregg-miskelly, line 546: >m_currentFrameHasSourceCode [](start = 17, length = 27)

1. We need to add a client check here -- we don't want to do this in full Visual Studio (or other IDEs that have full disassembly support)
2. We should probably check if the top frame of the specified thread has source, since the user can switch threads.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1277,14 +1282,22 @@ protected override void HandleStackTraceRequestAsync(IRequestResponder<StackTrac
                             textPosition = TextPositionTuple.GetTextPositionOfFrame(m_pathConverter, frame) ?? TextPositionTuple.Nil;
                         }
 
-                        response.StackFrames.Add(new ProtocolMessages.StackFrame()
+                        // remember the state for the lowermost frame
+                        if (startFrame == 0 && i == 0)
+                            m_currentFrameHasSourceCode = textPosition.Source?.Path != null;
```
gregg-miskelly, line 1287: >m_currentFrameHasSourceCode [](start = 28, length = 27)

You will wind up setting this to true if the top frame of any thread doesn't have source. It needs to be changed to a Dictionary/HashSet. I would probably also suggest changing the name from 'current' to 'top' since 'current frame' has another meaning that can be confusing.

gregg-miskelly, line 1287: >textPosition.Source [](start = 58, length = 19)

I **think** your goal is to have VS Code make a source request to obtain the disassembly of the frame if there is no source. Is that correct? If so, don't you need to make a source object here that has a sourceReference?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1513,12 +1526,22 @@ protected override void HandleSetVariableRequestAsync(IRequestResponder<SetVaria
...
         protected override void HandleSourceRequestAsync(IRequestResponder<SourceArguments, SourceResponse> responder)
         {
-            responder.SetError(new ProtocolException("'SourceRequest' not supported."));
+            var source = responder.Arguments.Source;
+            var response = new SourceResponse();
+            if (source != null)
+            {
+                int sourceRef = (int)source.SourceReference;
+                string content = TextPositionTuple.GetSourceForRef(sourceRef);
+                response.MimeType = "text/x-lldb.disassembly";
```
gregg-miskelly, line 1540: >x-lldb [](start = 42, length = 6)

Does something in VS Code already understand this type?

src/OpenDebugAD7/AD7DebugSession.cs
Trass3r: Depends on the UI I guess. VSCode probably fetches only the current thread so it works.

src/OpenDebugAD7/AD7DebugSession.cs
Trass3r: It's borrowed from the CodeLLDB extension.
Asm CodeLens doesn't support this kind of disassembly yet.

src/OpenDebugAD7/TextPositionTuple.cs
```diff
@@ -24,30 +36,132 @@ private TextPositionTuple(Source source, int line, int column)
         public static TextPositionTuple GetTextPositionOfFrame(PathConverter converter, IDebugStackFrame2 frame)
```
gregg-miskelly, line 36: >GetTextPositionOfFrame [](start = 40, length = 22)

Similar to my suggestion in the other review - you don't want to do this here, but rather in HandleStackTraceRequestAsync because there is no need to do this stuff for stopping event processing.

src/OpenDebugAD7/TextPositionTuple.cs
```diff
@@ -24,30 +36,132 @@ private TextPositionTuple(Source source, int line, int column)
...
+                    int srcline = converter.ConvertDebuggerLineToClient((int)beginPosition[0].dwLine);
+                    int srccolumn = unchecked((int)(beginPosition[0].dwColumn + 1));
 
-                return new TextPositionTuple(source, line, column);
+                    return new TextPositionTuple(source, srcline, srccolumn);
+                }
             }
 
+            // this frame is lacking source code
+            source.Name = Path.ChangeExtension(source.Name, ".s") ?? "???";
```
gregg-miskelly, line 66: >Name [](start = 54, length = 4)

`source.Name` may be null.

src/OpenDebugAD7/TextPositionTuple.cs
Trass3r: Well here it's deliberate, sort of. VSCode didn't use the stopped event source but maybe other clients do?

src/OpenDebugAD7/TextPositionTuple.cs
```diff
@@ -24,30 +36,132 @@ private TextPositionTuple(Source source, int line, int column)
...
-                int line = converter.ConvertDebuggerLineToClient((int)beginPosition[0].dwLine);
-                int column = unchecked((int)(beginPosition[0].dwColumn + 1));
+                    source.Path = convertedFilePath;
+                    int srcline = converter.ConvertDebuggerLineToClient((int)beginPosition[0].dwLine);
+                    int srccolumn = unchecked((int)(beginPosition[0].dwColumn + 1));
 
-                return new TextPositionTuple(source, line, column);
+                    return new TextPositionTuple(source, srcline, srccolumn);
+                }
             }
```
gregg-miskelly, line 63: >} [](start = 12, length = 1)

A few comments for the rest of this:
1. We need a client check - we want to do this stuff with an IDE that has real disassembly support
2. I don't think you want to do the disassembly up front. Instead we should keep a collection of sourceReferences and do the disassembly when the IDE requests source.

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: No, VS Code fetches the stack of all threads very aggressively.

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: I don't know what VS Code uses the MimeType for. Does it matter what we put? Does it matter if CodeLLDB isn't installed?

src/OpenDebugAD7/TextPositionTuple.cs
gregg-miskelly: The source in the stopped event was actually an extension added long ago to support testing. It should be fine to just use the original path.

src/OpenDebugAD7/TextPositionTuple.cs
Trass3r: Which is fine according to the spec.

src/OpenDebugAD7/AD7DebugSession.cs
Trass3r: If it isn't installed you won't get syntax highlighting. If it is you will.

src/OpenDebugAD7/AD7DebugSession.cs
Trass3r: Ok. Haven't noticed any fallout yet though.
While without the if I had. All frames of a thread were marked.

src/OpenDebugAD7/TextPositionTuple.cs
Trass3r: Ha, something to add to the spec.

src/OpenDebugAD7/TextPositionTuple.cs
Trass3r: Makes sense.
The thing is there seems to be no way to find the start of the current function in gdb.
So different address requests belonging to the same function without actually disassembling seem to be impossible to relate.

src/OpenDebugAD7/AD7DebugSession.cs
Trass3r: This is used in VS?

gregg-miskelly: Yes.

src/OpenDebugAD7/TextPositionTuple.cs
Trass3r: Looks like VS uses DisassembleRequest though?
https://github.com/microsoft/MIEngine/commit/b2a5c7a891e949f129d8ded15aa320cc6409f910

gregg-miskelly: Yes, in the VS case, we want to return whatever source path we have in the symbols. VS is fine with source paths that don't exist and will request disassembly if that is what the user chooses to do.

Trass3r: But how to handle this? I guess there'd need to be a client support flag to do it properly.

gregg-miskelly: Do you mean how to check for VS? If yes, Gabrielle added support for this to AD7Session recently: https://github.com/gregg-miskelly/MIEngine/blob/896c72d633b39a0571b9599d15c5a8fd0f1fcf58/src/OpenDebugAD7/AD7DebugSession.cs#L632-L638

gregg-miskelly:
@Trass3r The VS Code team is currently working on a real disassembly window: https://github.com/microsoft/vscode/issues/124163 so I don't think we will need this anymore.

Trass3r:
Yeah I know. Closing this.

PR #1042, opened 2020-09-06
implement Jump to Line

Trass3r:


src/MICore/CommandFactories/gdb.cs
```diff
@@ -183,6 +183,25 @@ public override async Task<List<ulong>> StartAddressesForLine(string file, uint
...
+        private async Task JumpInternal(string target)
+        {
+            // temporary breakpoint + jump
+            await _debugger.CmdAsync("-break-insert -t " + target, ResultClass.done);
+            await _debugger.CmdAsync("-exec-jump " + target, ResultClass.running);
+        }
+
+        public override Task ExecJump(string filename, int line)
+        {
+            string target = "--source " + filename + " --line " + line;
```
WardenGnaw, line 195:
```suggestion
            string target = "--source " + filename + " --line " + line.ToString(CultureInfo.InvariantCulture);
```

src/MICore/CommandFactories/gdb.cs
```diff
@@ -183,6 +183,25 @@ public override async Task<List<ulong>> StartAddressesForLine(string file, uint
...
+            return JumpInternal(target);
+        }
+
+        public override Task ExecJump(ulong address)
+        {
+            string target = "*" + string.Format("0x{0:X}", address);
```
WardenGnaw, line 201:
```suggestion
            string target = "*" + string.Format(CultureInfo.InvariantCulture, "0x{0:X}", address);
```

src/MICore/CommandFactories/lldb.cs
```diff
@@ -177,6 +177,18 @@ public override Task Catch(string name, bool onlyOnce = false, ResultClass resul
             throw new NotImplementedException("lldb catch command");
         }
```
WardenGnaw, line 179: Would leave a comment here to update these commands if the jump command is implemented in [`lldb-mi`](https://github.com/lldb-tools/lldb-mi/tree/master/src)

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1193,6 +1197,88 @@ protected override void HandlePauseRequestAsync(IRequestResponder<PauseArguments
...
+            BeforeContinue();
+            var builder = new ErrorBuilder(() => AD7Resources.Error_UnableToSetNextStatement);
+            try
+            {
+                builder.CheckHR(thread.SetNextStatement(null, gotoTarget));
+            }
+            catch (AD7Exception)
+            {
+                m_isStopped = true;
+                throw;
```
WardenGnaw, line 1226: throwing here will stop the OpenDebugAD7 process, you will want to use `responder.SetError(new ProtocolException(e.Message));` so the error is sent to the client.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1193,6 +1197,88 @@ protected override void HandlePauseRequestAsync(IRequestResponder<PauseArguments
...
+            }
+            catch (Exception e)
+            {
+                e = Utilities.GetInnerMost(e);
+                if (Utilities.IsCorruptingException(e))
+                {
+                    Utilities.ReportException(e);
+                }
+
+                response.Targets = null;
```
WardenGnaw, line 1277: You will need to set `responder.SetError(e.Message)`.

WardenGnaw (changes requested):

Trass3r:
Done.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -42,6 +42,8 @@ internal class AD7DebugSession : DebugAdapterBase, IDebugPortNotify2, IDebugEven
 
         private readonly DebugEventLogger m_logger;
         private readonly Dictionary<string, Dictionary<int, IDebugPendingBreakpoint2>> m_breakpoints;
+        private readonly List<IDebugCodeContext2> m_gotoCodeContexts = new List<IDebugCodeContext2>();
```
gregg-miskelly, line 45: >List [](start = 25, length = 4)

I would suggest using `HandleCollection` instead. HandleCollection should really internally use a List instead of a Dictionary, but what you have here is exactly what HandleCollection want meant to do.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1193,6 +1197,87 @@ protected override void HandlePauseRequestAsync(IRequestResponder<PauseArguments
             m_program.CauseBreak();
             responder.SetResponse(new PauseResponse());
         }
+        
+        protected override void HandleGotoRequestAsync(IRequestResponder<GotoArguments> responder)
+        {
+            var response = new GotoResponse();
```
gregg-miskelly, line 1203: >var [](start = 12, length = 3)

@WardenGnaw would know better than I, but following the pattern, I think you want to wrap these in a `try`+`catch` where the catch will do a `responder.SetError(new ProtocolException(e.Message));`

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1193,6 +1197,87 @@ protected override void HandlePauseRequestAsync(IRequestResponder<PauseArguments
...
+
+            responder.SetResponse(response);
+        }
+        
+        protected override void HandleGotoTargetsRequestAsync(IRequestResponder<GotoTargetsArguments, GotoTargetsResponse> responder)
+        {
+            var response = new GotoTargetsResponse();
+
+            var source = responder.Arguments.Source;
+            // TODO: handle this for disassembly debugging
```
gregg-miskelly, line 1237: >// TODO: handle this for disassembly debugging [](start = 12, length = 46)

I feel like the right want to handle disassembly is to add a `gotoInstruction` request which would take a `instructionReference` and `offset` (like "InstructionBreakpoint")

src/MIDebugEngine/AD7.Impl/AD7Thread.cs
```diff
@@ -276,26 +276,19 @@ int IDebugThread2.Resume(out uint suspendCount)
         // Sets the next statement to the given stack frame and code context.
         int IDebugThread2.SetNextStatement(IDebugStackFrame2 stackFrame, IDebugCodeContext2 codeContext)
         {
-            // CLRDBG TODO: This implementation should be changed to call an MI command
-            ulong addr = ((AD7MemoryAddress)codeContext).Address;
-            AD7StackFrame frame = ((AD7StackFrame)stackFrame);
-            if (frame.ThreadContext.Level != 0 || frame.Thread != this || !frame.ThreadContext.pc.HasValue || _engine.DebuggedProcess.MICommandFactory.Mode == MIMode.Clrdbg)
```
gregg-miskelly, line 282: The first two parts of this `if` (checking that the level is zero, and that the frame is for the correct thread) still make sense if a non-null stack frame is provided.

src/MIDebugEngine/AD7.Impl/AD7Thread.cs
```diff
@@ -276,26 +276,19 @@ int IDebugThread2.Resume(out uint suspendCount)
...
-            {
+            var infos = new CONTEXT_INFO[1];
+            if (codeContext.GetInfo(enum_CONTEXT_INFO_FIELDS.CIF_ADDRESS, infos) != Constants.S_OK)
                 return Constants.S_FALSE;
```
gregg-miskelly, line 281: Internally code contexts (AD7MemoryAddress) have the address as a `ulong` already. Probably makes more sense to verify that the provided code context is an `AD7MemoryAddress` and use that.

src/MIDebugEngine/AD7.Impl/AD7MemoryAddress.cs
```diff
@@ -42,6 +39,7 @@ public void SetDocumentContext(IDebugDocumentContext2 docContext)
         // Adds a specified value to the current context's address to create a new context.
         public int Add(ulong dwCount, out IDebugMemoryContext2 newAddress)
         {
+            // FIXME: this is not correct for IDebugCodeContext2
```
gregg-miskelly, line 42: >// FIXME: this is not correct for IDebugCodeContext2 [](start = 12, length = 52)

Why do you think this is wrong? `dwCount` should be a byte count, not an instruction count.

src/MIDebugEngine/AD7.Impl/AD7MemoryAddress.cs
```diff
@@ -160,19 +158,15 @@ public int GetInfo(enum_CONTEXT_INFO_FIELDS dwFields, CONTEXT_INFO[] pinfo)
             {
                 pinfo[0].dwFields = 0;
 
-                if ((dwFields & enum_CONTEXT_INFO_FIELDS.CIF_ADDRESS) != 0)
+                if ((dwFields & enum_CONTEXT_INFO_FIELDS.CIF_ADDRESS) != 0 ||
+                    (dwFields & enum_CONTEXT_INFO_FIELDS.CIF_ADDRESSABSOLUTE) != 0)
                 {
                     pinfo[0].bstrAddress = EngineUtils.AsAddr(_address, _engine.DebuggedProcess.Is64BitArch);
-                    pinfo[0].dwFields |= enum_CONTEXT_INFO_FIELDS.CIF_ADDRESS;
+                    pinfo[0].bstrAddressAbsolute = pinfo[0].bstrAddress;
```
gregg-miskelly, line 165: >bstrAddressAbsolute [](start = 29, length = 19)

You don't want to assign to these fields unless they are requested. This can cause a memory leak in the Visual Studio case.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -187,6 +187,44 @@ public object GetMetric(string metric)
...
+            }
+            catch (Exception e)
+            {
+                _engineCallback.OnError(EngineUtils.GetExceptionDescription(e));
+                return Constants.E_ABORT;
+            }
+
+            return Constants.S_OK;
+        }
+
```
gregg-miskelly, line 208: Where do we use this overload?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1618,6 +1618,16 @@ public Task Continue(DebuggedThread thread)
             return Execute(thread);
         }
 
+        public async Task Jump(string filename, int line)
+        {
+            await MICommandFactory.ExecJump(filename, line);
+        }
```
gregg-miskelly, line 1624: Same

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -246,6 +246,12 @@ public async Task ExecNextInstruction(int threadId, ResultClass resultClass = Re
             await ThreadFrameCmdAsync(command, resultClass, threadId, 0);
         }
 
+        /// <summary>
+        /// Jumps to a specified target location
+        /// </summary>
+        abstract public Task ExecJump(string filename, int line);
```
gregg-miskelly, line 252: >abstract public Task ExecJump(string filename, int line); [](start = 8, length = 57)

Same

src/MIDebugEngine/AD7.Impl/AD7Thread.cs
Trass3r: According to the docs this is reserved and should be null.

src/OpenDebugAD7/AD7DebugSession.cs
Trass3r: Yes there are different patterns in the code.
Also something like extract inner exception etc.

src/MIDebugEngine/AD7.Impl/AD7MemoryAddress.cs
Trass3r: > IDebugCodeContext2 is incremented by instructions rather than by the basic storage size (typically byte).

https://docs.microsoft.com/en-us/visualstudio/extensibility/debugger/reference/idebugcodecontext2#remarks

Which is still not precise enough btw. Could mean Add has pointer semantics so +1 means next instruction or it's just aligned to the next instruction.

src/MIDebugEngine/AD7.Impl/AD7Thread.cs
gregg-miskelly: The docs are wrong. VS will always provide the frame. The frame is pretty useless, but it is provided.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
Trass3r: First I used this one, then added addresses as they are unique.
But we need to find a way to actually get all the addresses for a given source line from gdb.

src/MIDebugEngine/AD7.Impl/AD7MemoryAddress.cs
gregg-miskelly: Got it. That lines speaks to an an idea that I think the original designers of the IDebug* interfaces had. But in practice, I don't think anything in the VS UI will actually try to 'Add' to a code context except for the fact that a code context is also a memory context, so if you wind up getting one in the memory window. In such a memory window situation, the caller would expect Add to move by bytes.

The disassembly window does need to move by instructions, but it does so using [IDebugDisassemblyStream2.Seek](https://docs.microsoft.com/en-us/visualstudio/extensibility/debugger/reference/idebugdisassemblystream2-seek?view=vs-2019).

src/MIDebugEngine/AD7.Impl/AD7Thread.cs
Trass3r: Should be fixed then.
I wonder why they added that parameter.

gregg-miskelly: All of the '2' interfaces in AD7 (the name for the IDebug* family of interfaces) date back to the early 2000s and many are older than that (because they came from a previous product in the Visual Studio family and were redone a bit for VS 2002). So the reason has been lost to time. But I would **guess** either:
1. The designers wanted to support engines that could automatically unwind out of frames. This doesn't really make any sense for a native debuggger. But managed languages might know enough to be able to do that.
-or-
2. It was just an oversight

src/DebugEngineHost.VSCode/VSCode/HandleCollection.cs
```diff
@@ -2,82 +2,68 @@
...
 
-using System;
-using System.Linq;
 using System.Collections.Generic;
 
 namespace Microsoft.DebugEngineHost.VSCode
 {
     public sealed class HandleCollection<T>
     {
-        private const int START_HANDLE = 1000;
```
gregg-miskelly, line 13: >private const int START_HANDLE = 1000; [](start = 8, length = 38)

You still want `START_HANDLE`. Otherwise you will break all the tests.

src/DebugEngineHost.VSCode/VSCode/HandleCollection.cs
```diff
@@ -2,82 +2,68 @@
...
+            _collection.Clear();
         }
 
         public int Create(T value)
         {
-            var handle = _nextHandle++;
-            _handleMap[handle] = value;
-            return handle;
+            _collection.Add(value);
+            return _collection.Count - 1;
```
gregg-miskelly, line 21: >; [](start = 40, length = 1)

`+ START_HANDLE`

src/DebugEngineHost.VSCode/VSCode/HandleCollection.cs
```diff
@@ -2,82 +2,68 @@
...
         }
 
         public bool TryGet(int handle, out T value)
         {
-            if (_handleMap.TryGetValue(handle, out value))
+            if (handle >= _collection.Count)
```
gregg-miskelly, line 26: >handle [](start = 16, length = 6)

Need to subtract off START_HANDLE

src/DebugEngineHost.VSCode/VSCode/HandleCollection.cs
```diff
@@ -2,82 +2,68 @@
...
-                {
-                    throw new ArgumentOutOfRangeException("handle");
-                }
-
-                return value;
+                return _collection[handle];
             }
         }
 
         public bool Remove(int handle)
```
gregg-miskelly, line 56: Do we actually call this method? If so it isn't going to work with your new implementation.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
gregg-miskelly: Did you find that `-symbol-list-lines` doesn't work correctly for this case? (See (LinesForFile)[https://github.com/microsoft/MIEngine/blob/29c6569861b2042c43f04d8d78b2566db2e2ef1f/src/MIDebugEngine/Engine.Impl/SourceLine.cs#L93]).

Regardless, I don't think you want these overloads since we always convert the source file + line to instruction before we get this far.

Trass3r: This is what is used:
https://github.com/microsoft/MIEngine/blob/b2f1555b2f92e7e6a6090832a2145ac1658bc89a/src/MICore/CommandFactories/gdb.cs#L158-L160

I tested it with a template and only got 1 instantiation.

gregg-miskelly: Sorry if this is obvious: but did you double check that the two templates didn't get folded together by the compiler/linker.

Assuming that isn't it -- you could certainly try `-symbol-list-lines` to see if that does any better. If not, I am guessing we would need GDB support to make this feature work quite correctly in the template case.

Trass3r: Didn't check but didn't use any optimizations either.

Trass3r:
```cpp
#include <stdexcept>

template <typename T>
int foo(T t)
{
	return 0;
}

int main(int argc, const char* argv[])
{
	return foo(0.5f) + foo(5);
}
```
```
`-symbol-list-lines test.cpp
lines: [{pc=0x0000555555555129,line=10},{pc=0x000055555555513d,line=11},{pc=0x000055555555514c,line=11},{pc=0x0000555555555156,line=11},{pc=0x0000555555555158,line=12},{pc=0x000055555555515f,line=0},{pc=0x000055555555515f,line=4},{pc=0x000055555555516c,line=6},{pc=0x0000555555555171,line=7},{pc=0x0000555555555173,line=0},{pc=0x0000555555555173,line=4},{pc=0x000055555555517e,line=6},{pc=0x0000555555555183,line=7},{pc=0x0000555555555185,line=0}]

`i li test.cpp:6
Line 6 of "test.cpp" starts at address 0x55555555516c <foo<float>(float)+13> and ends at 0x555555555171 <foo<float>(float)+18>.
```
list-lines does cover everything, even single instructions per line.
But it would need to be cached.

src/DebugEngineHost.VSCode/VSCode/HandleCollection.cs
Trass3r: It's used in `HostMarshal.ReleaseCodeContextId`. You're right.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
gregg-miskelly: I am thinking we could detect when to ignore repeated records based on if the line number is the same between records. For example, in your first example output, save the first entry (line 10), and the second entry (line 11) but then ignore the next two since those are also line 11.

I would agree that `-symbol-list-lines` could be quite verbose, but, unless I am missing something, it seems to be what we use today for EnumCodeContexts.

Trass3r: > unless I am missing something, it seems to be what we use today for EnumCodeContexts.

Yes but it falls back to StartAddressesForLine in the end.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1193,6 +1196,86 @@ protected override void HandlePauseRequestAsync(IRequestResponder<PauseArguments
...
+                responder.SetResponse(response);
+                return;
+            }
+
+            var gotoTarget = m_gotoCodeContexts[responder.Arguments.TargetId];
+            IDebugThread2 thread = null;
+            lock (m_threads)
+            {
+                if (!m_threads.TryGetValue(responder.Arguments.ThreadId, out thread))
+                    throw new AD7Exception("Could not find thread!");
```
pieandcakes, line 1214: I wouldn't throw an exception here. As far as I remember, for protocol, we should just be sending failure through `responder.setError` in the callback handlers.

Trass3r: I think I copied that from another function.
Correct error handling should be clarified in general.

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
Trass3r: > you could certainly try `-symbol-list-lines` to see if that does any better. If not, I am guessing we would need GDB support to make this feature work quite correctly in the template case.

https://sourceware.org/bugzilla/show_bug.cgi?id=26594

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: Any exceptions that happen in AD7DebugSession that is in a Request method, it will need to be a `responder.SetError(ProtocolException)`.

Trass3r: I see I took it from StepInternal where the exception is caught in the callers.

src/MICore/CommandFactories/gdb.cs
```diff
@@ -183,6 +183,25 @@ public override async Task<List<ulong>> StartAddressesForLine(string file, uint
             return addresses;
         }
 
+        private async Task JumpInternal(string target)
+        {
+            // temporary breakpoint + jump
+            await _debugger.CmdAsync("-break-insert -t " + target, ResultClass.done);
```
WardenGnaw, line 189: With this temporary breakpoint not registered, when we get a stopping event it will be unknown and returned as an exception.

https://github.com/microsoft/MIEngine/blob/728f403c563c2c50c46542e9b44271a7f8d5895b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs#L1179

I think we need to treat this as an async break.

src/MICore/CommandFactories/gdb.cs
```diff
@@ -183,6 +183,25 @@ public override async Task<List<ulong>> StartAddressesForLine(string file, uint
...
+            await _debugger.CmdAsync("-exec-jump " + target, ResultClass.running);
```
WardenGnaw, line 190: `-exec-jump` will resume all threads, we may need to call `set scheduler-locking on` to ensure that it only runs the current thread. Then revert back to `set scheduler-locking replay`.

WardenGnaw (changes requested):
Made some fixes in https://github.com/microsoft/MIEngine/commit/96967704bca10f0ce107860d2ea723b62eb5deec

@gregg-miskelly 

We get some weirdness since GoToRequest assumes the process is always stopped but the way GDB handles `-exec-jump`, it causes it to resume the process, so a temporary breakpoint is used. 
However, in between the GoTo Request and before we get stopped at the temporary breakpoint, we get a `StackTraceRequest`. But since we are not stopped, we return no frames. 

Thoughts on how to address this issue and my other comment about freezing other threads in a comment below?

src/MICore/CommandFactories/gdb.cs
Trass3r: All stop mode shouldn't be assumed either.
https://sourceware.org/gdb/current/onlinedocs/gdb/Non_002dStop-Mode.html#Non_002dStop-Mode could be enabled, I'm currently looking into this as we have threads that need to keep running or the test will crash.

src/MICore/CommandFactories/gdb.cs
Trass3r: What do you mean by not registered?

src/MICore/CommandFactories/gdb.cs
WardenGnaw: There potentially can be a race where another thread hits a breakpoint before we hit this the temporary one created for `jump`.

src/MICore/CommandFactories/gdb.cs
WardenGnaw: On stopping and when it is a breakpoint, we go through a list of breakpoints:
https://github.com/microsoft/MIEngine/blob/728f403c563c2c50c46542e9b44271a7f8d5895b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs#L1132

If if there are none, we assume it is an entrypoint breakpoint, embedded, we hit a bp pending deletion, or it is an exception.

src/MICore/CommandFactories/gdb.cs
Trass3r: But is that an assumption or did you actually encounter that problem?
Cause the docs say "Resume execution at line linespec. Execution stops again immediately if there is a breakpoint there."

WardenGnaw: This is a potential issue. I think we can mark this with a note if users ever run into this problem if GDB starts other threads when it resumes from the jump.

This should be fine if it does "immediately" stop.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1321,11 +1326,16 @@ protected override void HandleGotoTargetsRequestAsync(IRequestResponder<GotoTarg
...
-                            if (documentContext.GetStatementRange(pos, null) == HRConstants.S_OK)
-                                line = m_pathConverter.ConvertDebuggerLineToClient((int)pos[0].dwLine);
+                            var startPos = new TEXT_POSITION[1];
+                            var endPos = new TEXT_POSITION[1];
+                            if (documentContext.GetStatementRange(startPos, endPos) == HRConstants.S_OK)
+                                line = m_pathConverter.ConvertDebuggerLineToClient((int)startPos[0].dwLine);
                         }
-                        targets.Add(new GotoTarget(m_gotoCodeContexts.Create(codeContext), contextName, line));
+
+                        int codeContextId = m_nextContextId++;
```
Trass3r, line 1335: Wouldn't this need to be atomic if multiple simultaneous requests are expected?

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1598,6 +1598,16 @@ public Task Continue(DebuggedThread thread)
...
```
gregg-miskelly, line 1604: I think we want to remove all the overloads that use filename and line number since we always seem to use the overload that take the address.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1598,6 +1598,16 @@ public Task Continue(DebuggedThread thread)
...
+
+        public async Task Jump(ulong address)
+        {
```
gregg-miskelly, line 1607: >{ [](start = 8, length = 1)

As Andrew talked about, I believe we need to also:
- Take an exclusive lock to make sure that no other operation is happening
- Block until we have reentered break state.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1304,6 +1311,97 @@ protected override void HandlePauseRequestAsync(IRequestResponder<PauseArguments
...
+                {
+                    IDebugThread2 thread = null;
+                    lock (m_threads)
+                    {
+                        if (!m_threads.TryGetValue(responder.Arguments.ThreadId, out thread))
+                            throw new AD7Exception("Unknown thread id: " + responder.Arguments.ThreadId.ToString(CultureInfo.InvariantCulture));
+                    }
+                    BeforeContinue();
+                    builder.CheckHR(thread.SetNextStatement(null, gotoTarget));
+                }
```
gregg-miskelly, line 1337: >} [](start = 16, length = 1)

I believe we also need to send a Stopped event with reason 'Goto'.

Trass3r: True.
> The debug adapter first sends the response and then a ‘stopped’ event with reason ‘goto’.

src/MICore/CommandFactories/gdb.cs
Trass3r: Indeed. It does receive a `breakpoint-modified` event and `disp` is `"del"` fwiw.
```sh
<-1067-break-insert -t *0x5555555573B1
->1067^done,bkpt={number="3",type="breakpoint",disp="del",enabled="y",addr="0x00005555555573b1",func="main(int, "...
<-1068-exec-jump *0x5555555573B1
->1068^running
->*running,thread-id="all"
->=breakpoint-modified,bkpt={number="3",type="breakpoint",disp="del",enabled="y",addr="0x00005555555573b1",func="main(int, "...
->*stopped,reason="breakpoint-hit",disp="del",bkptno="3",frame={addr="0x00005555555573b1",func="main",args=[{name="argc",value="1"},{name="argv",value="0x7fffffffdb68"}],file="../test.cpp",fullname="/tmp/test.cpp",line="28",arch="i386:x86-64"},thread-id="1",stopped-threads="all",core="6"
->=breakpoint-deleted,id="3"

```

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1305,6 +1312,99 @@ protected override void HandlePauseRequestAsync(IRequestResponder<PauseArguments
...
+                if (m_gotoCodeContexts.TryGetValue(responder.Arguments.TargetId, out IDebugCodeContext2 gotoTarget))
+                {
+                    lock (m_threads)
+                    {
+                        if (!m_threads.TryGetValue(responder.Arguments.ThreadId, out thread))
+                            throw new AD7Exception("Unknown thread id: " + responder.Arguments.ThreadId.ToString(CultureInfo.InvariantCulture));
+                    }
+                    BeforeContinue();
+                    builder.CheckHR(thread.SetNextStatement(null, gotoTarget));
+                }
```
gregg-miskelly, line 1337: >} [](start = 16, length = 1)

We need to throw if this if is false.

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: > But what if something goes wrong, do you still send an event?
I don't think you should need to.

WardenGnaw:
We are looking into splitting GoToTargets and the GoTo Request.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
Trass3r: So maybe setting $pc would be better after all to avoid the jump command problems.

gregg-miskelly: It could be. I am not allowed to look at GDB code -- only documentation. So I don't know if the jump commands handle additional scenarios that would be good to support, or if they are essentially equivalent. If you have the cycles and the interest and want to take a look at the code and see if the jump commands do more, we would be happy to utilize your advice.

Trass3r:
Squashed and updated.
I also added a test commit using `set $pc = ` instead of the jump command (see https://sourceware.org/gdb/current/onlinedocs/gdb/Jumping.html).
Maybe this could be a way out of the temporary breakpoint problems discussed previously.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
Trass3r: Well the [jump cmd](https://github.com/bminor/binutils-gdb/blob/4258df85f13dbaf9a1cb5b523bf2b9d390b25e69/gdb/infcmd.c#L1126-L1137) is equal to `set $pc` + `continue` [it seems](https://github.com/bminor/binutils-gdb/blob/c651f0a6147e4de96326f6eecdff5520b9159816/gdb/infrun.c#L2898-L2922).

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -216,6 +216,7 @@ public int Jump(ulong address)
                 _engineCallback.OnError(EngineUtils.GetExceptionDescription(e));
                 return Constants.E_ABORT;
             }
+            DebuggedProcess.ThreadCache.MarkDirty();
```
Trass3r, line 219: This was needed to update the current line in the UI.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1344,7 +1344,7 @@ protected override void HandleGotoRequestAsync(IRequestResponder<GotoArguments>
                         if (!m_threads.TryGetValue(responder.Arguments.ThreadId, out thread))
                             throw new AD7Exception("Unknown thread id: " + responder.Arguments.ThreadId.ToString(CultureInfo.InvariantCulture));
                     }
-                    BeforeContinue();
+                    //BeforeContinue();
                     builder.CheckHR(thread.SetNextStatement(null, gotoTarget));
```
Trass3r, line 1348: Is that ok?

src/MICore/CommandFactories/gdb.cs
```diff
@@ -201,6 +201,8 @@ public override Task ExecJump(string filename, int line)
 
         public override Task ExecJump(ulong address)
         {
+            return _debugger.ConsoleCmdAsync("set $pc = " + string.Format(CultureInfo.InvariantCulture, "0x{0:X}", address), false);
```
Trass3r, line 204: The question is if some kind of error handling is needed.

gregg-miskelly: Ideally yes, but since this is a non-MI command, there aren't wonderful options for knowing it is succeeded. I am not sure if GDB supports assigning to `$pc` through MI commands instead (`-var-assign`?)

Trass3r: Looks like it but requires a variable object incl. creation/deletion (possibly just once): https://sourceware.org/gdb/current/onlinedocs/gdb/GDB_002fMI-Variable-Objects.html

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: You *probably* want BeforeContinue. The other debug adapter that the VS debugger team owns currently doesn't clear our handle collections on SetNextStatement, but you are probably correct that it should. But we should check VS Code and make sure it is happy.

Trass3r: Yeah idk technically it doesn't run the program but on the other hand the current line of code needs to be updated. The thread cache invalidation worked in VSCode.

FabianoGK:
If I may ask, what is the state of this pull request? Is there additional things to discuss?
I'm really excited to see this ready for vscode-cppdbg!

Trass3r:
See all the comments after my last commit.

FabianoGK:
Eclipse GDB debugger has two commands around _-exec-jump_: Move to Line, and Resume at Line.

In the Move to Line, it does insert a breakpoint and then jumps:
```
480,319 100-break-insert -t -p 1 -f /projects/test/test.cpp:160
480,326 100^done,bkpt={number="7",type="breakpoint",disp="del",enabled="y",addr="0x0004fb78",func="main(int, char**)",file="../src/test.cpp",fullname="/projects/test/test.cpp",line="160",thread-groups=["i1"],thread="1",thread="1",times="0",original-location="/projects/test/test.cpp:160"}
480,326 (gdb) 
480,327 101-exec-jump --thread 1 /projects/test/test.cpp:160
480,333 101^running
480,333 *running,thread-id="all"
480,333 (gdb) 
480,345 =breakpoint-modified,bkpt={number="7",type="breakpoint",disp="del",enabled="y",addr="0x0004fb78",func="main(int, char**)",file="../src/test.cpp",fullname="/projects/test/test.cpp",line="160",thread-groups=["i1"],thread="1",thread="1",times="1",original-location="/projects/test/test.cpp:160"}
480,365 ~"\nTemporary breakpoint "
480,366 ~"7, main (argc=1, argv=0xbefffd94) at ../src/test.cpp:160\n"
480,367 ~"160\t\tTest();\n"
480,367 *stopped,reason="breakpoint-hit",disp="del",bkptno="7",frame={addr="0x0004fb78",func="main",args=[{name="argc",value="1"},{name="argv",value="0xbefffd94"}],file="../src/test.cpp",fullname="/projects/test/test.cpp",line="160"},thread-id="1",stopped-threads="all",core="0"
480,367 =breakpoint-deleted,id="7"
```

In the Resume at Line, it just does the jump:
```
688,000 171-exec-jump --thread 1 /projects/test/test.cpp:131
688,009 171^running
688,009 *running,thread-id="all"
688,009 (gdb)
```

Visual Studio _Set Next Statement_ behavior is the same as Eclipse's _Move to Line_. In other words, it does not continue the program execution.

Trass3r:
> ```
> 480,327 101-exec-jump --thread 1 /projects/test/test.cpp:160
> 480,333 101^running
> 480,333 *running,thread-id="all"
> ```
That `--thread` parameter doesn't seem to exist:
https://sourceware.org/gdb/current/onlinedocs/gdb/GDB_002fMI-Program-Execution.html

FabianoGK:
The Context Management talks about the ‘--thread’ parameter:
> each MI command should explicitly specify which thread and frame to operate on. To make it possible, each MI command accepts the ‘--thread’ and ‘--frame’ options, the value to each is GDB global identifier for thread and frame to operate on.

https://sourceware.org/gdb/current/onlinedocs/gdb/Context-management.html#Context-management

Trass3r:
Interesting. But it doesn't make a difference here as it still runs [all threads](https://sourceware.org/gdb/current/onlinedocs/gdb/Non_002dStop-Mode.html):
```
*running,thread-id="all"
```

So it could still theoretically break in another thread first.
But if Eclipse is using the same temp bp method it probably doesn't happen in practice.

FabianoGK:
> Interesting. But it doesn't make a difference here as it still runs [all threads](https://sourceware.org/gdb/current/onlinedocs/gdb/Non_002dStop-Mode.html):
> 
> ```
> *running,thread-id="all"
> ```
> 
> So it could still theoretically break in another thread first.
> But if Eclipse is using the same temp bp method it probably doesn't happen in practice.

Interesting observation indeed. I had to turn on [non-stop mode](https://sourceware.org/gdb/current/onlinedocs/gdb/Non_002dStop-Mode.html) in order to behave as expected:

```
1: (53481) <-1151-exec-jump --thread 2 test.cpp:24
1: (53483) ->1151^running
1: (53483) ->*running,thread-id="2"
```

Trass3r:
And there's also [scheduler-locking mode](https://sourceware.org/gdb/current/onlinedocs/gdb/All_002dStop-Mode.html), very convoluted.

aallrd:
Hello,
What is still required to merge this development?

Trass3r:
See all the open discussions. There were some points without final resolution.

PR #1043, opened 2020-09-11
support debug console auto-completion

Trass3r:
It basically works but sometimes vscode only offers the history suggestions even though it receives a proper completions response:
```json
{
  "type":"response",
  "request_seq":15,
  "success":true,
  "command":"completions",
  "body":{
    "targets":[
      {
        "label":"`passcount",
        "start":0,
        "length":1
      },
      {
        "label":"`path",
        "start":0,
        "length":1
      },
      {
        "label":"`pipe",
        "start":0,
        "length":1
      },
      {
        "label":"`print",
        "start":0,
        "length":1
      },
......
    ]
  },
  "seq":565
}
```
May be a vscode bug but I can't see anything in
https://github.com/microsoft/vscode/blob/e0d5fae35ff101b37da0c8bee61551ceeb6a2e1c/src/vs/workbench/contrib/debug/browser/repl.ts#L151-L190


Trass3r:
ping

WardenGnaw:
> Proof of concept.

I was ignoring it due to this comment. Is this testable in VS Code?

Trass3r:
I see, yes should still be.

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -659,6 +659,23 @@ public virtual bool CanDetach()
 
         abstract public Task<List<ulong>> StartAddressesForLine(string file, uint line);
 
+        public virtual async Task<string[]> AutoComplete(string command, int threadId, uint frameLevel)
+        {
+            command = "-complete \"" + command + "\"";
```
gregg-miskelly, line 664: Looking at [the documentation](https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI-Miscellaneous-Commands.html#GDB_002fMI-Miscellaneous-Commands) for the `-complete` command, and a quick example, it looks like this just tells us about GDB commands, correct?

If so, don't we need to restrict this to when the command buffer starts with `-exec`

My quick example:
```
-complete "f"
^done,completion="f",matches=["faas","file","find","finish","flash-erase","flushregs","focus","forward-search","frame","fs","ftrace","function"],max_completions_reached="0"
```

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2007,6 +2008,46 @@ protected override void HandleSetFunctionBreakpointsRequestAsync(IRequestRespond
...
+
+            IDebugStackFrame2 frame = null;
+            int? frameId = responder.Arguments.FrameId;
+            bool success = false;
+            if (frameId != null)
+                success = m_frameHandles.TryGet(frameId.Value, out frame);
+
+            try
+            {
+                MethodInfo methodInfo = m_program.GetType().GetMethod("AutoComplete");
```
gregg-miskelly, line 2027: Rather than invoking through reflection, we should add a new interface for this to src\Microsoft.VisualStudio.Debugger.Interop.DAP\Microsoft.VisualStudio.Debugger.Interop.DAP.cs

src/MICore/CommandFactories/MICommandFactory.cs
Trass3r: I did handle the short form ` since I can't be bothered to type -exec all the time but made it optional since I'm in the microsoft/vscode-cpptools#106 camp.
Gdb could be tricked into expression completion via the p command I think but for gdb command mode should really be the default.

src/OpenDebugAD7/AD7DebugSession.cs
Trass3r: Please provide the interface.

WardenGnaw: I think you can extend `IDebugProgramDAP`.

https://github.com/microsoft/MIEngine/blob/728f403c563c2c50c46542e9b44271a7f8d5895b/src/Microsoft.VisualStudio.Debugger.Interop.DAP/Microsoft.VisualStudio.Debugger.Interop.DAP.cs#L46

You will create `IDebugProgramDAP.AutoComplete(string command, IDebugStackFrame2 stackFrame, out string[] result)`.

Then in AD7Engine, you will implement `AutoComplete(string command, IDebugStackFrame2 stackFrame, out string[] result)` and it can call `_debuggedProcess.MICommandFactory.AutoComplete(..)`

https://github.com/microsoft/MIEngine/blob/73db59d89b8f0fbfdaa3cee1423e674e092ffc0f/src/MIDebugEngine/AD7.Impl/AD7Engine.cs#L38


You can use `IDebugProgramDAP.GetPointerSize` as an example.

@gregg-miskelly Is this correct or did you have another idea?

gregg-miskelly: Yup, that is my suggestion.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2007,6 +2008,50 @@ protected override void HandleSetFunctionBreakpointsRequestAsync(IRequestRespond
...
+            IDebugStackFrame2 frame = null;
+            int? frameId = responder.Arguments.FrameId;
+            if (frameId != null)
+                _ = m_frameHandles.TryGet(frameId.Value, out frame);
+
+            string command = responder.Arguments.Text;
+            if (command[0] == '`')
+                command = command.Substring(1);
+            else if (command.StartsWith("-exec", StringComparison.Ordinal))
+                command = command.Substring(6);
```
gregg-miskelly, line 2028: If the command doesn't start with ``` or `-exec` don't you want to not call AutoComplete?

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: Though if your goal is just to auto complete on commands, probably we should rename this to `AutoCompleteCommand` so that we could later have an `AutoCompleteExpression` if we ever wanted to support that.

src/OpenDebugAD7/AD7DebugSession.cs
Trass3r: Well that's the question. As I said if we talk about the vscode debug console and gdb nobody expects expression completion. If the completion is also used for Watches (spec doesn't talk about that) then that's something else.
And I guess for VS Immediate Window you'd expect expressions.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2359,9 +2359,8 @@ protected override void HandleCompletionsRequestAsync(IRequestResponder<Completi
                 command = command.Substring(6);
             try
             {
-                MethodInfo methodInfo = m_program.GetType().GetMethod("AutoComplete");
-                string[] results = (string[])methodInfo.Invoke(m_program, new object[] { command, frame });
-                if (results == null)
+                var debugProgram = (IDebugProgramDAP)m_engine;
```
Trass3r, line 2362: For some reason this fails for me locally while the following works
```cs
var methodInfo = m_engine.GetType().GetInterface("IDebugProgramDAP").GetMethod("AutoCompleteCommand");
```

gregg-miskelly: What OS are you on?

Trass3r: WSL. The DAP dll isn't part of the vscode plugin distribution so I tried copying it but to no avail.

WardenGnaw: It looks like an issue with the mono we ship, it does not handle [`EmbedInteropTypes` correctly](https://github.com/microsoft/MIEngine/blob/fbbc9938f8a75e3ca5fb7573a331f61ff484676f/src/OpenDebugAD7/OpenDebugAD7.csproj#L93).

I'll push out a fix to deploy it if we are in the XPlat scenario.

Trass3r: @WardenGnaw did that happen already?

Trass3r: Apparently not, so let's hope for the .net 5 change to fix it.

gregg-miskelly: The work to switch to running on .NET 5 is almost complete.

Trass3r: Ok it still fails.
```cs
m_engine is IDebugProgramDAP
error CS0433: The type 'IDebugProgramDAP' exists in both 'Microsoft.MIDebugEngine, Version=14.0.0.0, Culture=neutral, PublicKeyToken=cfb98189dc466aca' and 'OpenDebugAD7, Version=14.0.0.0, Culture=neutral, PublicKeyToken=cfb98189dc466aca'
```
Needed to disable Embed and copy the DAP dll manually.

PublishOpenDebugAD7.sh
```diff
@@ -132,4 +132,5 @@ fi
 log_and_exec_cmd "dotnet build ${__ScriptDirectory}/src/MIDebugEngine-Unix.sln -c ${__Configuration}"
 log_and_exec_cmd "dotnet publish ${__ScriptDirectory}/src/OpenDebugAD7/OpenDebugAD7.csproj ${__DotnetPublishArgs}"
 log_and_exec_cmd "cp ${__ScriptDirectory}/bin/${__Configuration}/Microsoft.MIDebugEngine.dll ${__OutputFolder}/."
-log_and_exec_cmd "cp ${__ScriptDirectory}/bin/${__Configuration}/Microsoft.MICore.dll ${__OutputFolder}/."
\ No newline at end of file
+log_and_exec_cmd "cp ${__ScriptDirectory}/bin/${__Configuration}/Microsoft.MICore.dll ${__OutputFolder}/."
+log_and_exec_cmd "cp ${__ScriptDirectory}/bin/${__Configuration}/Microsoft.VisualStudio.Debugger.Interop.DAP.dll ${__OutputFolder}/."
```
WardenGnaw, line 136: Since `Microsoft.VisualStudio.Debugger.Interop.DAP` is already a reference in `OpenDebugAD7.csproj`, you can have it copied if you add `<Private>true</Private>` in the `<ProjectReference>...</ProjectReference>` instead of forcing a copy in the publish script. 

Reference:
image

https://docs.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-items?view=vs-2019#projectreference

gregg-miskelly: You shouldn't need to copy it - it is embedded

src/MIDebugEngine/MIDebugEngine.csproj
```diff
@@ -60,7 +60,6 @@
     <ProjectReference Include="..\Microsoft.VisualStudio.Debugger.Interop.DAP\Microsoft.VisualStudio.Debugger.Interop.DAP.csproj">
       <Project>{23DE943C-1F84-4E23-9490-50FE1DEDE858}</Project>
       <Name>Microsoft.VisualStudio.Debugger.Interop.DAP</Name>
-      <EmbedInteropTypes>True</EmbedInteropTypes>
```
gregg-miskelly, line 63: ><EmbedInteropTypes>True</EmbedInteropTypes> [](http://example.com/codeflow?start=6&length=43)

Put this back

src/OpenDebugAD7/OpenDebugAD7.csproj
```diff
@@ -59,7 +59,6 @@
     <ProjectReference Include="..\Microsoft.VisualStudio.Debugger.Interop.DAP\Microsoft.VisualStudio.Debugger.Interop.DAP.csproj">
       <Project>{23DE943C-1F84-4E23-9490-50FE1DEDE858}</Project>
       <Name>Microsoft.VisualStudio.Debugger.Interop.DAP</Name>
-      <EmbedInteropTypes>True</EmbedInteropTypes>
```
gregg-miskelly, line 62: ><EmbedInteropTypes>True</EmbedInteropTypes> [](http://example.com/codeflow?start=6&length=43)

Same

PublishOpenDebugAD7.sh
gregg-miskelly: I saw your comment about getting the build error with ambiguous types. I don't see it when I build MIEngine on my own box. Where are you seeing the problem? I rebased your changes to make it easier to view: [gregg-miskelly/completion](https://github.com/gregg-miskelly/MIEngine/tree/completion).

src/Microsoft.VisualStudio.Debugger.Interop.DAP/Microsoft.VisualStudio.Debugger.Interop.DAP.cs
```diff
@@ -51,6 +51,15 @@ public interface IDebugProgramDAP
         /// <param name="pResult">Number of bits in a pointer.</param>
         /// [PreserveSig]
         int GetPointerSize([Out] out int pResult);
+
+        /// <summary>
```
gregg-miskelly, line 55: I would suggest putting this is another interface since an engine might want to be hosted by OpenDebugAD7, but not support this feature.

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: If it doesn't start with ``` or `-exec` then it is an expression. Since GDB doesn't have expression auto complete, I think we need to just return an empty match list without calling the engine.

PublishOpenDebugAD7.sh
Trass3r: When trying to evaluate `m_engine is IDebugProgramDAP` in the debug console, as it is done in the code.

gregg-miskelly: Are you saying that everything actually works, but the debugger's expression evaluator gets confused?

gregg-miskelly: (Like at runtime, the capability gets correctly assigned, but if you try to evaluate `m_engine is IDebugProgramDAP` in the debugger, the debugger is confused which definition of `IDebugProgramDAP` to use)

Trass3r: No the expression is also false at runtime so the whole feature gets disabled.

gregg-miskelly: What .NET Runtime are you using? (.NET 5, Mono, .NET Framework)
What OS?

Trass3r:
```sh
$ tail ~/.vscode-server/extensions/ms-vscode.cpptools-*/debugAdapters/OpenDebugAD7
#ARGS+=("--pauseForDebugger")
dotnet "${BIN_DIR}/bin/OpenDebugAD7.dll" "${ARGS[@]}" $*
$ dotnet --version
5.0.301
```
WSL2, Ubuntu 21.04

gregg-miskelly: Okay, I don't know why you are seeing that. I will set it up and see if I can repro.

WardenGnaw: I can reproduce it. Currently investigating why the [embedded type equivalence](https://docs.microsoft.com/en-us/dotnet/framework/interop/type-equivalence-and-embedded-interop-types#marking-com-types-for-type-equivalence) is not working in this scenario.

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -666,6 +666,23 @@ public virtual bool CanDetach()
 
         abstract public Task<List<ulong>> StartAddressesForLine(string file, uint line);
 
+        public virtual async Task<string[]> AutoComplete(string command, int threadId, uint frameLevel)
```
WardenGnaw, line 669: This will need to be moved to `gdb.cs` and throw `throw new NotImplementedException();` here, `-complete` in lldb-mi responds with:
```
^error,msg="Driver. Received command '-complete b'. It was not handled. Command 'complete' not in Command Factory"
```

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -811,6 +811,7 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
             InitializeResponse initializeResponse = new InitializeResponse()
             {
                 SupportsConfigurationDoneRequest = true,
+                SupportsCompletionsRequest = m_engine is IDebugProgramDAP,
```
WardenGnaw, line 814: This is going to be weird since we need to know which debug adapter we are targeting, for now we could always send `true` but it will not work in `lldb`.

WardenGnaw (changes requested):
Sorry for talking long to get back to this. We were finalizing getting testing tp be OpenSourced and moving MIEngine to .NET 5.

With these changes we request that any changes to OpenDebugAD7/DAP is also covered in the testing scenario.

Here is a commit with changes for testing AutoComplete: https://github.com/microsoft/MIEngine/commit/3a1d8d48ed35daf1e92c4a18a0356fd483e11d0a

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: Is VS Code happy if we just return an empty array of completions?

WardenGnaw: Yes. There will just be nothing for autocomplete, but the debug adapter does support the request, which will fail with Not Implemented exception

WardenGnaw:
If you want to merge in my branch [dev/waan/completions](https://github.com/microsoft/MIEngine/tree/dev/waan/completions), it addresses the issues above.

Trass3r:
@WardenGnaw "First-time contributors need a maintainer to approve running workflows"

gregg-miskelly:
> @WardenGnaw "First-time contributors need a maintainer to approve running workflows"

Done

WardenGnaw:
Sorry, one more thing to address that is fixed in https://github.com/microsoft/MIEngine/commit/f23ba3cced39df26b8d64683c9ad2533b2e2ebbb

Heres a gif of it working:
AutoComplete

Trass3r:
Well my opinion is still that debug console inputs should just be treated as commands in general when you're on Linux. That's what you expect as a gdb user.
But let's just get this thing merged asap.

WardenGnaw (approved):

PR #1046, opened 2020-09-15
Fixed ps call for Docker Linux containers

gc46:
Fix for Bug 1152117: Attach to process doesn't work on alpine based Linux containers. Verified on images listed [here](https://hub.docker.com/search?q=&type=image&operating_system=linux&category=base).

Before:
image

After:
image

### DIFF
--- a/src/SSHDebugPS/PSOutputParser.cs
+++ b/src/SSHDebugPS/PSOutputParser.cs
@@ -65,7 +65,7 @@ namespace Microsoft.SSHDebugPS
         private ColumnDef _ruserCol;
         private ColumnDef _argsCol;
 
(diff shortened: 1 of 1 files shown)


WardenGnaw:
Modern version of ps have this in the man pages.
```
This version of ps accepts several kinds of options:

       1   UNIX options, which may be grouped and must be preceded by a dash.
       2   BSD options, which may be grouped and must not be used with a dash.
       3   GNU long options, which are preceded by two dashes.
```

It looks like all the Linux based docker versions accept BSD options way, too. However `mcr.microsoft.com/dotnet/core/aspnet:3.0-alpine3.9` does not allow the UNIX options way.

gregg-miskelly (approved):

PR #1048, opened 2020-09-22
Removing ClrDBG references

WardenGnaw:
We are not shipping clrdbg anymore and the later version of the C# debugger does not support MI commands. 

Deleting references of it in MIEngine to clean up the repository.


src/MICore/ExampleLaunchOptions.xml
```diff
@@ -66,8 +66,4 @@
     TargetArchitecture="x86"
     AdditionalSOLibSearchPath=""/>
 
-  <SSHLaunchOptions xmlns="http://schemas.microsoft.com/vstudio/MDDDebuggerOptions/2014"
```
gregg-miskelly, line 69: >SSHLaunchOptions [](start = 3, length = 16)

nit: this was the only example of SSHLaunchOptions, so might want to instead make it a GDB one...

gregg-miskelly (approved):
Otherwise LGTM

PR #1051, opened 2020-09-24
Add SymbolLoadInfo for LaunchOptionJson

WardenGnaw:
To keep the JSON parsing in OpenDebugAD7 and MIEngine in sync, this PR adds the JSON parsing for SymbolLoadInfo if launch options come in as JSON to MIEngine.

The current support for SymbolLoadInfo is in OpenDebugAD7, but it converts it to XML.

Support for reading JSON for SymbolLoadInfo : 
https://github.com/microsoft/MIEngine/blob/ec52d600d5d4a5d9a2517aa392e0c1767ad49e55/src/OpenDebugAD7/MILaunchOptions.cs#L625-L636

We set WaitDynamicLibLoad here:
https://github.com/microsoft/MIEngine/blob/ec52d600d5d4a5d9a2517aa392e0c1767ad49e55/src/OpenDebugAD7/MILaunchOptions.cs#L532-L540

Here is the XML parsing implementation for SymbolLoadInfo:
https://github.com/microsoft/MIEngine/blob/9be1a0f01c8a0183be22708bbefbfc0ad32427bc/src/MICore/LaunchOptions.cs#L1868-L1884

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -1795,21 +1795,21 @@ namespace MICore
 
             if (options.SymbolLoadInfo != null)
             {
-                this.WaitDynamicLibLoad = !String.IsNullOrWhiteSpace(options.SymbolLoadInfo.ExceptionList) || options.SymbolLoadInfo.LoadAll.HasValue;
-
                 SymbolInfoLoadAll = options.SymbolLoadInfo.LoadAll.GetValueOrDefault(true);
 
-                if (DebuggerMIMode == MIMode.Lldb && !string.IsNullOrWhiteSpace(options.SymbolLoadInfo.ExceptionList))
+                if (!string.IsNullOrWhiteSpace(options.SymbolLoadInfo.ExceptionList))
                 {
-                    throw new InvalidLaunchOptionsException(String.Format(CultureInfo.InvariantCulture, MICoreResources.Error_OptionNotSupported, nameof(options.SymbolLoadInfo.ExceptionList), nameof(MIMode.Lldb)));
+                    if (DebuggerMIMode == MIMode.Lldb)
+                    {
+                        throw new InvalidLaunchOptionsException(String.Format(CultureInfo.InvariantCulture, MICoreResources.Error_OptionNotSupported, nameof(options.SymbolLoadInfo.ExceptionList), nameof(MIMode.Lldb)));
+                    }
+
+                    this.WaitDynamicLibLoad = true;
+                    SymbolInfoExceptionList.SetTo(options.SymbolLoadInfo.ExceptionList.Split(';'));
                 }
-
-                SymbolInfoExceptionList.SetTo(options.SymbolLoadInfo.ExceptionList == null ? new string[0] : options.SymbolLoadInfo.ExceptionList.Split(';'));
-
-                // Ensure that symbol loading options are consistent
-                if (!WaitDynamicLibLoad && !SymbolInfoExceptionList.IsEmpty)
+                else
                 {
-                    throw new InvalidLaunchOptionsException(MICoreResources.Error_InvalidSymbolInfo);
+                    SymbolInfoExceptionList.SetTo(new string[0]);
                 }
             }
 
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -1792,6 +1792,27 @@ namespace MICore
             this.WaitDynamicLibLoad = false;
 
             this.SourceMap = SourceMapEntry.CreateCollection(options.SourceFileMap);
+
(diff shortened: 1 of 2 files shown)


src/MICore/LaunchOptions.cs
```diff
@@ -1792,6 +1792,27 @@ protected void InitializeCommonOptions(Json.LaunchOptions.BaseOptions options)
...
+                    throw new InvalidLaunchOptionsException(String.Format(CultureInfo.InvariantCulture, MICoreResources.Error_OptionNotSupported, nameof(options.SymbolLoadInfo.ExceptionList), nameof(MIMode.Lldb)));
+                }
+
+                SymbolInfoExceptionList.SetTo(options.SymbolLoadInfo.ExceptionList == null ? new string[0] : options.SymbolLoadInfo.ExceptionList.Split(';'));
+
+                // Ensure that symbol loading options are consistent
+                if (!WaitDynamicLibLoad && !SymbolInfoExceptionList.IsEmpty)
+                {
+                    throw new InvalidLaunchOptionsException(MICoreResources.Error_InvalidSymbolInfo);
+                }
```
gregg-miskelly, line 1813: Isn't your code on line 1798 + 1792 designed to make this impossible?

src/MICore/LaunchOptions.cs
```diff
@@ -1792,6 +1792,27 @@ protected void InitializeCommonOptions(Json.LaunchOptions.BaseOptions options)
             this.WaitDynamicLibLoad = false;
 
             this.SourceMap = SourceMapEntry.CreateCollection(options.SourceFileMap);
+
+            if (options.SymbolLoadInfo != null)
+            {
+                this.WaitDynamicLibLoad = !String.IsNullOrWhiteSpace(options.SymbolLoadInfo.ExceptionList) || options.SymbolLoadInfo.LoadAll.HasValue;
```
gregg-miskelly, line 1798: >options.SymbolLoadInfo.LoadAll.HasValue [](start = 110, length = 39)

This part doesn't seem to make sense: Why should we wait if LoadAll is set to `true`, which is the same as the default.

src/MICore/LaunchOptions.cs
WardenGnaw: 🤦‍♂️

src/MICore/LaunchOptions.cs
```diff
@@ -1792,6 +1792,27 @@ protected void InitializeCommonOptions(Json.LaunchOptions.BaseOptions options)
...
+
+                SymbolInfoLoadAll = options.SymbolLoadInfo.LoadAll.GetValueOrDefault(true);
+
+                if (DebuggerMIMode == MIMode.Lldb && !string.IsNullOrWhiteSpace(options.SymbolLoadInfo.ExceptionList))
```
gregg-miskelly, line 1802: This code feels like it would be a lot cleaner as:

```
if (!string.IsNullOrWhiteSpace(options.SymbolLoadInfo.ExceptionList))
{
    if (DebuggerMIMode == MIMode.Lldb)
        throw...

    this.WaitDynamicLibLoad = true;
    SymbolInfoExceptionList.SetTo(options.SymbolLoadInfo.ExceptionList.Split(';'));
}
else
{
    SymbolInfoExceptionList.SetTo(new string[0]);
}
```

gregg-miskelly (commented):
Otherwise LGTM

src/MICore/LaunchOptions.cs
WardenGnaw: So if we have an exception list or we are not loading all modules, we should wait on dynamic library load.
```suggestion
                this.WaitDynamicLibLoad = !String.IsNullOrWhiteSpace(options.SymbolLoadInfo.ExceptionList) || !options.SymbolLoadInfo.LoadAll.GetValueOrDefault(true);
```

gregg-miskelly: I *think* it only needs to be set if there is an exception list. If you don't have an exception list and load all is disabled -- we will load no symbols, but it doesn't look like we should need to somehow intervene.

gregg-miskelly (approved):

src/MICore/LaunchOptions.cs
```diff
@@ -1792,6 +1792,27 @@ protected void InitializeCommonOptions(Json.LaunchOptions.BaseOptions options)
...
+
+                    this.WaitDynamicLibLoad = true;
+                    SymbolInfoExceptionList.SetTo(options.SymbolLoadInfo.ExceptionList.Split(';'));
+                }
+                else
+                {
+                    SymbolInfoExceptionList.SetTo(new string[0]);
+                }
+            }
+
```
pieandcakes, line 1815: @WardenGnaw why the extra space?

PR #1054, opened 2020-10-07
[GDB] Support Modules window for OpenDebugAD7 scenarios

gc46:
- Adds support for MIF_TIMESTAMP in IDebugModule2.GetInfo.
- Handles SupportsModulesRequest for DAP. 
  - Thank you @Trass3r for the intial work!
- Sends ModuleEvents when a `IDebugModuleLoadEvent2` is recieved.


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -610,6 +611,14 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
                 });
             }
 
+            List<ColumnDescriptor> additionalModuleColumns = new List<ColumnDescriptor>();
```
WardenGnaw, line 614: You will want to only do this if the client is VS.
You can check this by:
`Host.GetHostUIIdentifier() == HostUIIdentifier.VSIDE`

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1546,6 +1557,56 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
             responder.SetResponse(response);
         }
 
+        private static long FileTimeToPosix(FILETIME ft)
```
WardenGnaw, line 1560: nit: This is in the `DebugAdapterBase` region. Move to `Utility` region (Line 266)

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1546,6 +1557,56 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
...
+                var debugModules = new IDebugModule2[1];
+
+                uint numReturned = 0;
+                while (enumDebugModules.Next(1, debugModules, ref numReturned) == HRConstants.S_OK && numReturned == 1)
+                {
+                    var debugModuleInfos = new MODULE_INFO[1];
+                    if (debugModules[0].GetInfo(enum_MODULE_INFO_FIELDS.MIF_ALLFIELDS, debugModuleInfos) == HRConstants.S_OK)
+                    {
+                        var debugModuleInfo = debugModuleInfos[0];
+                        var mod = new ProtocolMessages.Module(debugModuleInfo.m_bstrName, debugModuleInfo.m_bstrName)
```
WardenGnaw, line 1587: Using the name (`debugModuleInfo.m_bstrName`) of the module is not reccomended for the Module ID. 

Create a static int for the count of modules and use that as the ID.

```
private static uint s_nextModuleId = 0;

static public uint GetNextModuleId()
{
   return ++s_nextModuleId;
}
```

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2359,8 +2420,30 @@ public void HandleIDebugModuleLoadEvent2(IDebugEngine2 pEngine, IDebugProcess2 p
...
             ((IDebugModuleLoadEvent2)pEvent).GetModule(out module, ref moduleLoadMessage, ref isLoad);
-
+            
             m_logger.WriteLine(LoggingCategory.Module, moduleLoadMessage);
+
+            var debugModuleInfos = new MODULE_INFO[1];
+            if (module.GetInfo(enum_MODULE_INFO_FIELDS.MIF_ALLFIELDS, debugModuleInfos) == HRConstants.S_OK)
+            {
+                var debugModuleInfo = debugModuleInfos[0];
+                var mod = new ProtocolMessages.Module(debugModuleInfo.m_bstrName, debugModuleInfo.m_bstrName)
```
WardenGnaw, line 2430: Same as other comment on ModuleIDs

WardenGnaw (changes requested):

src/MIDebugEngine/AD7.Impl/AD7Module.cs
```diff
@@ -10,6 +10,8 @@
 using MICore;
 using System.Globalization;
 using System.Threading.Tasks;
+using System.IO;
```
pieandcakes, line 13: Sort using statements

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -24,6 +24,7 @@
 using OpenDebug.CustomProtocolObjects;
 using OpenDebugAD7.AD7Impl;
 using ProtocolMessages = Microsoft.VisualStudio.Shared.VSCodeDebugProtocol.Messages;
+using static System.FormattableString;
```
pieandcakes, line 27: why is this here?

src/OpenDebugAD7/AD7DebugSession.cs
pieandcakes: Does specifying `AdditionalModuleColumns` as blank do something unintended?

src/OpenDebugAD7/AD7DebugSession.cs
pieandcakes: might want to start something larger than 0 (1000?)

src/OpenDebugAD7/OpenDebugAD7.csproj
```diff
@@ -82,6 +82,7 @@
   <ItemGroup>
     <Reference Include="System" />
     <Reference Include="Microsoft.CSharp" />
+    <Reference Include="Microsoft.VisualStudio.OLE.Interop" />
```
pieandcakes, line 85: why doesn't this have the same definition as the one in MIDebugEngine? either stick to not having the version and public key or include it for both

src/MIDebugEngine/MIDebugEngine.csproj
```diff
@@ -69,6 +69,7 @@
     <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.VisualStudio.OLE.Interop, Version=7.1.40304.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
```
pieandcakes, line 72: You are missing the nuget change to get this package pulled in.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: Blank is fine.

WardenGnaw:
Please address this build warning and similar ones:
```
##[warning]src\OpenDebugAD7\AD7DebugSession.cs(620,41): Warning CS0618: 'ColumnDescriptor.ColumnDescriptor(string, string, string, ColumnDescriptor.TypeValue?, int?)' is obsolete: 'This constructor contains optional parameters and is no longer supported.  Please use the constructor which includes only required parameters.'
AD7DebugSession.cs(620,41): warning CS0618: 'ColumnDescriptor.ColumnDescriptor(string, string, string, ColumnDescriptor.TypeValue?, int?)' is obsolete: 'This constructor contains optional parameters and is no longer supported.  Please use the constructor which includes only required parameters.' [D:\a\1\s\src\OpenDebugAD7\OpenDebugAD7.csproj]
```

src/MIDebugEngine/MIDebugEngine.csproj
WardenGnaw, line 72:
```suggestion
<Reference Include="Microsoft.VisualStudio.OLE.Interop">
      <HintPath>$(NuGetPackagesDirectory)/Microsoft.VisualStudio.OLE.Interop.16.7.30328.74/lib/net45/Microsoft.VisualStudio.OLE.Interop.dll</HintPath>
      <ForceIncludeInVSIX>true</ForceIncludeInVSIX>
    </Reference>
```

src/OpenDebugAD7/OpenDebugAD7.csproj
WardenGnaw, line 85:
```suggestion
<Reference Include="Microsoft.VisualStudio.OLE.Interop">
      <HintPath>$(NuGetPackagesDirectory)/Microsoft.VisualStudio.OLE.Interop.16.7.30328.74/lib/net45/Microsoft.VisualStudio.OLE.Interop.dll</HintPath>
      <ForceIncludeInVSIX>true</ForceIncludeInVSIX>
    </Reference>
```

WardenGnaw (commented):
Add the following to `src\OpenDebugAD7\packages.config`
`<package id="Microsoft.VisualStudio.OLE.Interop" version="16.7.30328.74" />`

Also add `<DropUnsignedFile Include="$(OutputPath)\Microsoft.VisualStudio.OLE.Interop.dll" />` to:
https://github.com/microsoft/MIEngine/blob/4fe7e9ae44103bf723d2f4cd9ee9fee3505ff935/src/OpenDebugAD7/OpenDebugAD7.csproj#L177-L184

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2359,8 +2420,30 @@ public void HandleIDebugModuleLoadEvent2(IDebugEngine2 pEngine, IDebugProcess2 p
             string moduleLoadMessage = null;
             int isLoad = 0;
             ((IDebugModuleLoadEvent2)pEvent).GetModule(out module, ref moduleLoadMessage, ref isLoad);
-
+            
             m_logger.WriteLine(LoggingCategory.Module, moduleLoadMessage);
+
+            var debugModuleInfos = new MODULE_INFO[1];
+            if (module.GetInfo(enum_MODULE_INFO_FIELDS.MIF_ALLFIELDS, debugModuleInfos) == HRConstants.S_OK)
```
WardenGnaw, line 2427: Create a helper method to convert a MODULE_INFO into a ProtocolMessages.Module to remove duplicated code.

src/OpenDebugAD7/AD7DebugSession.cs
Trass3r: Probably debugging remnant.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1546,6 +1548,56 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
...
+                            Version = debugModuleInfo.m_bstrVersion,
+                            VsLoadAddress = debugModuleInfo.m_addrLoadAddress.ToString(CultureInfo.InvariantCulture),
+                            VsPreferredLoadAddress = debugModuleInfo.m_addrPreferredLoadAddress.ToString(CultureInfo.InvariantCulture),
+                            VsModuleSize = (int)debugModuleInfo.m_dwSize,
+                            VsLoadOrder = (int)debugModuleInfo.m_dwLoadOrder,
+                            SymbolFilePath = debugModuleInfo.m_bstrUrlSymbolLocation,
+                            SymbolStatus = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_SYMBOLS) == 0 ? "✓" : "✗",
+                            VsIs64Bit = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_64BIT) == 0,
+                            IsOptimized = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_OPTIMIZED) == 0, // not set by gdb
+                            IsUserCode = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_SYSTEM) != 0, // not set by gdb
```
Trass3r, line 1600: 3 of these seem to be inverted.

src/OpenDebugAD7/OpenDebugAD7.csproj
Trass3r: I vote for the former.

Trass3r (changes requested):

src/MIDebugEngine/AD7.Impl/AD7Module.cs
```diff
@@ -71,6 +73,17 @@ int IDebugModule2.GetInfo(enum_MODULE_INFO_FIELDS dwFields, MODULE_INFO[] infoAr
                     info.m_dwLoadOrder = this.DebuggedModule.GetLoadOrder();
                     info.dwValidFields |= enum_MODULE_INFO_FIELDS.MIF_LOADORDER;
                 }
+                if ((dwFields & enum_MODULE_INFO_FIELDS.MIF_TIMESTAMP) != 0)
+                {   
+                    long ft = File.GetLastWriteTimeUtc(this.DebuggedModule.SymbolPath).ToFileTime();
```
gregg-miskelly, line 78: >this.DebuggedModule.SymbolPath [](start = 55, length = 30)

I think you are missing a null check against `this.DebuggedModule.SymbolPath`

gregg-miskelly, line 78: >long [](start = 20, length = 4)

You are also missig a try/catch around this to not return a value if GetLastWriteTimeUtc throws.

gregg-miskelly: Actually, it seems like this should probably be looking at `Name` rather than SymbolPath.

---
In reply to: [501252420](https://github.com/microsoft/MIEngine/pull/1054#discussion_r501252420) [](ancestors = 501252420)

src/MIDebugEngine/AD7.Impl/AD7Module.cs
```diff
@@ -71,6 +73,17 @@ int IDebugModule2.GetInfo(enum_MODULE_INFO_FIELDS dwFields, MODULE_INFO[] infoAr
                     info.m_dwLoadOrder = this.DebuggedModule.GetLoadOrder();
                     info.dwValidFields |= enum_MODULE_INFO_FIELDS.MIF_LOADORDER;
                 }
+                if ((dwFields & enum_MODULE_INFO_FIELDS.MIF_TIMESTAMP) != 0)
```
gregg-miskelly, line 76: >) [](start = 75, length = 1)

I don't remember the best way to check for this anymore, but you probably only want to do this if we are local debugging. Since otherwise the path that you have may be from another computer. Something like:

Look at the launch options and see if they are LocalLaunchOptions, then do `string.IsNullOrWhiteSpace(localLaunchOptions.MIDebuggerServerAddress) && string.IsNullOrWhiteSpace(localLaunchOptions.DebugServer)`

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1546,6 +1557,56 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
...
+                            VsTimestampUTC = FileTimeToPosix(debugModuleInfo.m_TimeStamp).ToString(CultureInfo.InvariantCulture),
+                            Version = debugModuleInfo.m_bstrVersion,
+                            VsLoadAddress = debugModuleInfo.m_addrLoadAddress.ToString(CultureInfo.InvariantCulture),
+                            VsPreferredLoadAddress = debugModuleInfo.m_addrPreferredLoadAddress.ToString(CultureInfo.InvariantCulture),
+                            VsModuleSize = (int)debugModuleInfo.m_dwSize,
+                            VsLoadOrder = (int)debugModuleInfo.m_dwLoadOrder,
+                            SymbolFilePath = debugModuleInfo.m_bstrUrlSymbolLocation,
+                            SymbolStatus = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_SYMBOLS) == 0 ? "✓" : "✗",
+                            VsIs64Bit = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_64BIT) == 0,
+                            IsOptimized = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_OPTIMIZED) == 0, // not set by gdb
```
gregg-miskelly, line 1599: >IsOptimized [](start = 28, length = 11)

This is more complicated than just checking a single flag because `null` is also a possible value (and what you should wind up with in the GDB case). This is true for both `IsOptimized` and `IsUserCode`.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1546,6 +1557,56 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
...
+                while (enumDebugModules.Next(1, debugModules, ref numReturned) == HRConstants.S_OK && numReturned == 1)
+                {
+                    var debugModuleInfos = new MODULE_INFO[1];
+                    if (debugModules[0].GetInfo(enum_MODULE_INFO_FIELDS.MIF_ALLFIELDS, debugModuleInfos) == HRConstants.S_OK)
+                    {
+                        var debugModuleInfo = debugModuleInfos[0];
+                        var mod = new ProtocolMessages.Module(debugModuleInfo.m_bstrName, debugModuleInfo.m_bstrName)
+                        {
+                            Path = debugModuleInfo.m_bstrUrl,
+                            VsTimestampUTC = FileTimeToPosix(debugModuleInfo.m_TimeStamp).ToString(CultureInfo.InvariantCulture),
```
gregg-miskelly, line 1590: >VsTimestampUTC = FileTimeToPosix(debugModuleInfo.m_TimeStamp).ToString(CultureInfo.InvariantCulture), [](start = 28, length = 101)

Only set it if you have a timestamp.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: I lied earlier, I have updated the comment to do the correct check.

src/MIDebugEngine/AD7.Impl/AD7Module.cs
gc46: I have tried "Remove and Sort Usings" to no avail, unfortunately.

src/OpenDebugAD7/OpenDebugAD7.csproj
```diff
@@ -178,6 +178,7 @@
     <DropUnsignedFile Condition="'$(IsXPlat)' == 'true'" Include="$(OutputPath)\OpenDebugAD7.exe.mdb" />
   </ItemGroup>
   <ItemGroup Condition="'$(IsXPlat)' == 'false'">
+	<DropUnsignedFile Include="$(OutputPath)\Microsoft.VisualStudio.OLE.Interop.dll" />
```
WardenGnaw, line 181: nit: 4 spaces instead of tab
```suggestion
    <DropUnsignedFile Include="$(OutputPath)\Microsoft.VisualStudio.OLE.Interop.dll" />
```

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2359,8 +2443,17 @@ public void HandleIDebugModuleLoadEvent2(IDebugEngine2 pEngine, IDebugProcess2 p
...
             ((IDebugModuleLoadEvent2)pEvent).GetModule(out module, ref moduleLoadMessage, ref isLoad);
-
+            
             m_logger.WriteLine(LoggingCategory.Module, moduleLoadMessage);
+
+            var debugModuleInfos = new MODULE_INFO[1];
+            if (module.GetInfo(enum_MODULE_INFO_FIELDS.MIF_ALLFIELDS, debugModuleInfos) == HRConstants.S_OK)
+            {
+                var mod = ConvertToModule(debugModuleInfos[0]);
+                (module as IDebugModule3).IsUserCode(out int isUserCode);
```
WardenGnaw, line 2453: I would move this into `ConvertToModule`

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1577,31 +1577,28 @@ public static uint GetNextModuleId()
...
                 VsPreferredLoadAddress = debugModuleInfo.m_addrPreferredLoadAddress.ToString(CultureInfo.InvariantCulture),
                 VsModuleSize = (int)debugModuleInfo.m_dwSize,
                 VsLoadOrder = (int)debugModuleInfo.m_dwLoadOrder,
                 SymbolFilePath = debugModuleInfo.m_bstrUrlSymbolLocation,
                 SymbolStatus = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_SYMBOLS) != 0 ? "✓" : "✗",
-                VsIs64Bit = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_64BIT) != 0
-                // IsOptimized = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_OPTIMIZED) != 0 ? true : ((debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_UNOPTIMIZED) != 0 ? false : null) // not set by gdb
+                VsIs64Bit = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_64BIT) != 0,
+                IsOptimized = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_OPTIMIZED) != 0 ? true : ((debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_UNOPTIMIZED) != 0 ? false : null) // not set by gdb
                 // test -- need to delete?
```
WardenGnaw, line 1595: Delete comment

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1577,31 +1577,28 @@ public static uint GetNextModuleId()
...
             };
-            if((debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_OPTIMIZED) != 0)
-            {
-                mod.IsOptimized = true;
-            } else if ((debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_UNOPTIMIZED) != 0)
-            {
-                mod.IsOptimized = false;
-            }
+
+            var status = (module as IDebugModule3).IsUserCode(out int isUserCode);
```
WardenGnaw, line 1599: If somehow `module` is not a `IDebugModule3`, the result would be `null` and you would be calling `IsUserCode` on `null`.

Create a new variable and null check it.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1577,31 +1577,28 @@ public static uint GetNextModuleId()
...
                 VsPreferredLoadAddress = debugModuleInfo.m_addrPreferredLoadAddress.ToString(CultureInfo.InvariantCulture),
                 VsModuleSize = (int)debugModuleInfo.m_dwSize,
                 VsLoadOrder = (int)debugModuleInfo.m_dwLoadOrder,
                 SymbolFilePath = debugModuleInfo.m_bstrUrlSymbolLocation,
                 SymbolStatus = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_SYMBOLS) != 0 ? "✓" : "✗",
-                VsIs64Bit = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_64BIT) != 0
-                // IsOptimized = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_OPTIMIZED) != 0 ? true : ((debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_UNOPTIMIZED) != 0 ? false : null) // not set by gdb
+                VsIs64Bit = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_64BIT) != 0,
+                IsOptimized = (debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_OPTIMIZED) != 0 ? true : ((debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_UNOPTIMIZED) != 0 ? false : null) // not set by gdb
```
WardenGnaw, line 1594: Can this not be inlined and calculated before assigning it?

WardenGnaw (changes requested):
Please undo the `*.csproj` changes that added `<LangVersion>9.0</LangVersion>`

src/DebugEngineHost/DebugEngineHost.csproj
```diff
@@ -37,13 +37,15 @@
     <DefineConstants>$(DefineConstants);DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <LangVersion>9.0</LangVersion>
```
gregg-miskelly, line 40: ><LangVersion>9.0</LangVersion> [](start = 4, length = 30)

Did the IDE puke on each one of our project files? Probably worth reverting all of these changes an instead sticking it in /build/miengine.settings.targets

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: You can fix this with: 

```C#
if (module is IDebugModule3 module3 &&
    module3.IsUserCode(out int isUserCode) == VSConstants.S_OK)
{
    ...
}
```

---
In reply to: [503486210](https://github.com/microsoft/MIEngine/pull/1054#discussion_r503486210) [](ancestors = 503486210)

src/DebugEngineHost/DebugEngineHost.csproj
gc46: This was required for nested ternary operator to work.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1546,6 +1570,62 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
...
         }
 
+        // test -- need to delete; replace module ID here, first param
+        private static uint s_nextModuleId = 0; // change this?
+        public static uint GetNextModuleId()
+        {
+            return ++s_nextModuleId;
+        }
+
+        public static ProtocolMessages.Module ConvertToModule(MODULE_INFO debugModuleInfo, IDebugModule2 module)
```
gregg-miskelly, line 1580: >MODULE_INFO [](start = 62, length = 11)

nit: `in`

gregg-miskelly:
>         public void HandleIDebugModuleLoadEvent2(IDebugEngine2 pEngine, IDebugProcess2 pProcess, IDebugProgram2 pProgram, IDebugThread2 pThread, IDebugEvent2 pEvent)

We also need to handle module unload.

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:2436 in 710658d. [](commit_id = 710658d5102f7392430d126fe4bc0dd775157426, deletion_comment = False)

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1546,6 +1570,62 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
             responder.SetResponse(response);
         }
 
+        // test -- need to delete; replace module ID here, first param
+        private static uint s_nextModuleId = 0; // change this?
+        public static uint GetNextModuleId()
+        {
+            return ++s_nextModuleId;
+        }
```
gregg-miskelly, line 1578: I would replace this with `HandleCollection<IDebugModule2>` instead. Right now you are always creating a new module id, which isn't correct.

gregg-miskelly:
>                         {

I am not sure if you want to fix this now, but you can now fill out the 'moduleId' here: [https://github.com/microsoft/debug-adapter-protocol/blob/fdaa68760ff4c22dab619b1dbb8cfa8f80b6966f/debugAdapterProtocol.json#L3215](https://github.com/microsoft/debug-adapter-protocol/blob/fdaa68760ff4c22dab619b1dbb8cfa8f80b6966f/debugAdapterProtocol.json#L3215)

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:1306 in 710658d. [](commit_id = 710658d5102f7392430d126fe4bc0dd775157426, deletion_comment = False)

src/DebugEngineHost/DebugEngineHost.csproj
WardenGnaw: I don't think we can use 9. This night break the Mono runtime that we use. We use Framework 4.7

See https://www.mono-project.com/docs/about-mono/compatibility/

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: For example, take a look at `s_codeContexts` in https://github.com/microsoft/MIEngine/blob/1ccfd123289100f2415450d68b6011766941cc3d/src/DebugEngineHost.VSCode/HostMarshal.cs#L14

The the id now is the IntPointer.ToInt32()

src/DebugEngineHost/DebugEngineHost.csproj
gregg-miskelly: Maybe I am just missing what you are pointing at in the docs, but it would surprise me slightly if an entire language version didn't work in mono. Most language features work in all frameworks.

WardenGnaw: I wasn't sure if we would end up with IL that it has no idea what to do and it would break. However, it does not seem necessary to update the language version for just one thing.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2354,13 +2458,38 @@ public void HandleIDebugThreadDestroyEvent2(IDebugEngine2 pEngine, IDebugProcess
...
+
+            var debugModuleInfos = new MODULE_INFO[1];
+            if (module.GetInfo(enum_MODULE_INFO_FIELDS.MIF_ALLFIELDS, debugModuleInfos) == HRConstants.S_OK)
+            {
+                var mod = ConvertToModule(debugModuleInfos[0], module);
+                Protocol.SendEvent(new ModuleEvent(ModuleEvent.ReasonValue.New, mod));
+            }
+        }
+
+        public void HandleIDebugModuleUnloadEvent2(IDebugEngine2 pEngine, IDebugProcess2 pProcess, IDebugProgram2 pProgram, IDebugThread2 pThread, IDebugEvent2 pEvent)
```
WardenGnaw, line 2477: This callback never happens since a module unload is a `IDebugModuleLoad2Event` with `isLoad = false`.

You want to move this block:
```
            var debugModuleInfos = new MODULE_INFO[1];
            if (module.GetInfo(enum_MODULE_INFO_FIELDS.MIF_ALLFIELDS, debugModuleInfos) == HRConstants.S_OK)
            {
                var mod = ConvertToModule(debugModuleInfos[0], module);
                ReleaseDebugModuleId((IntPtr)mod.Id);
                Protocol.SendEvent(new ModuleEvent(ModuleEvent.ReasonValue.Removed, mod));
            }
```
into `HandleIDebugModuleLoadEvent2` when isLoad is `false`.

WardenGnaw: This is where the call back is registered.
https://github.com/microsoft/MIEngine/blob/61702cd8f53d20dedd8c694a6fc00a51b0ad3f17/src/OpenDebugAD7/AD7DebugSession.cs#L2252

There is no interface for `IDebugModuleUnloadEvent2`.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1546,6 +1594,62 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
             responder.SetResponse(response);
         }
 
+        public static ProtocolMessages.Module ConvertToModule(in MODULE_INFO debugModuleInfo, IDebugModule2 module)
+        {
+            var mod = new ProtocolMessages.Module(RegisterDebugModule(module), debugModuleInfo.m_bstrName)
```
WardenGnaw, line 1599: This will be registering the module for both `load` and `unload`.
You only want to register for `load`.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -610,6 +645,17 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
                 });
             }
 
+            List<ColumnDescriptor> additionalModuleColumns = new List<ColumnDescriptor>();
```
WardenGnaw, line 648: nit: I would set this to null and then initialize it within the `if`.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1546,6 +1594,62 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
...
+            if((debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_OPTIMIZED) != 0)
+            {
+                mod.IsOptimized = true;
+            } else if ((debugModuleInfo.m_dwModuleFlags & enum_MODULE_FLAGS.MODULE_FLAG_UNOPTIMIZED) != 0)
+            {
+                mod.IsOptimized = false;
+            }
+            if (module is IDebugModule3 module3 && module3.IsUserCode(out int isUserCode) == HRConstants.S_OK)
+            {
+                if (isUserCode == 0) mod.IsUserCode = false;
```
WardenGnaw, line 1623: nit: not a fan of single line `if`/`else` without braces, causes problems if someone decides to add extra code underneath expecting it to run only in the `else` case.

WardenGnaw (changes requested):

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -67,6 +67,30 @@ internal class AD7DebugSession : DebugAdapterBase, IDebugPortNotify2, IDebugEven
 
         private static Guid s_guidFilterAllLocalsPlusArgs = new Guid("939729a8-4cb0-4647-9831-7ff465240d5f");
 
+        private static readonly HandleCollection<IDebugModule2> s_debugModules = new HandleCollection<IDebugModule2>();
```
gregg-miskelly, line 70: >static [](start = 16, length = 6)

Why `static`?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -67,6 +67,30 @@ internal class AD7DebugSession : DebugAdapterBase, IDebugPortNotify2, IDebugEven
...
+        public static IntPtr RegisterDebugModule(IDebugModule2 debugModule)
```
gregg-miskelly, line 71: >IntPtr [](start = 22, length = 6)

I would suggest `int` instead of `IntPtr`.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -610,6 +645,18 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
                 });
             }
 
+            List<ColumnDescriptor> additionalModuleColumns = null;
+            if (responder.Arguments.ClientID == "visualstudio")
```
gregg-miskelly, line 649: >") [](start = 61, length = 2)

You want to also do it for `liveshare-server-host`

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1546,6 +1595,81 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
             responder.SetResponse(response);
         }
 
+        public static ProtocolMessages.Module ConvertToModule(in MODULE_INFO debugModuleInfo, IDebugModule2 module, bool isLoad)
```
gregg-miskelly, line 1598: >bool isLoad [](start = 116, length = 11)

You actually need three states here - load, list, and unload. Probably better to pass in the module id as an argument.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1546,6 +1595,81 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
...
+            ProtocolMessages.Module mod = null;
+            if (isLoad)
+            {
+                mod = new ProtocolMessages.Module(RegisterDebugModule(module), debugModuleInfo.m_bstrName)
+                {
+                    Path = path, VsTimestampUTC = vsTimestampUTC, Version = version, VsLoadAddress = vsLoadAddress, VsPreferredLoadAddress = vsPreferredLoadAddress,
+                    VsModuleSize = vsModuleSize, VsLoadOrder = vsLoadOrder, SymbolFilePath = symbolFilePath, SymbolStatus = symbolStatus, VsIs64Bit = vsIs64Bit
+                };
+            } else {
+                mod = new ProtocolMessages.Module(module.GetHashCode(), debugModuleInfo.m_bstrName)
```
gregg-miskelly, line 1620: >module.GetHashCode() [](start = 50, length = 20)

No, you need the id from the handle collection.

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly, line 71: >public [](start = 8, length = 6)

No reason for your new methods to be public

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly, line 70: >HandleCollection [](start = 32, length = 16)

I apologize, but actually HandleCollection doesn't do a great job for your scenario since you need to convert IDebugModule2->`id`, but, currently at least, we don't have any operations that go the other way.

Probably a better fit it:
```
private int m_nextModuleHandle = 1;
private readonly Dictionary<IDebugModule2, int> m_moduleMap = new Dictionary<IDebugModule2, int>();

object RegisterDebugModule(IDebugModule2 debugModule)
{
    Debug.Assert(!m_moduleMap.Contains(debugModule));
    lock (m_moduleMap)
    {
        int moduleHandle = m_nextModuleHandle;
        m_moduleMap[debugModule] = moduleHandle;
        m_nextModuleHandle++;
        return moduleHandle;
    }
}
```

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: BTW: If we have any other code which is checking for a ClientID of 'visualstudio', it needs the same fix.

---
In reply to: [504884955](https://github.com/microsoft/MIEngine/pull/1054#discussion_r504884955) [](ancestors = 504884955)

src/OpenDebugAD7/OpenDebugAD7.csproj
```diff
@@ -82,6 +82,10 @@
   <ItemGroup>
     <Reference Include="System" />
     <Reference Include="Microsoft.CSharp" />
+	<Reference Include="Microsoft.VisualStudio.OLE.Interop">
+		<HintPath>$(NuGetPackagesDirectory)/Microsoft.VisualStudio.OLE.Interop.16.7.30328.74/lib/net45/Microsoft.VisualStudio.OLE.Interop.dll</HintPath>
```
WardenGnaw, line 86: You are missing the package reference for us to pull this in.
```
<package id="Microsoft.VisualStudio.OLE.Interop" version="16.7.30328.74" />
```
in 
https://github.com/microsoft/MIEngine/blob/1ccfd123289100f2415450d68b6011766941cc3d/src/OpenDebugAD7/packages.config#L9

The build machine is using the one from MSBuild rather than a specific version:
```
Copying file from "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\PublicAssemblies\Microsoft.VisualStudio.OLE.Interop.dll" to "D:\a\1\s\bin\Desktop.Debug\Microsoft.VisualStudio.OLE.Interop.dll".
```

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -610,6 +645,18 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
                 });
             }
 
+            List<ColumnDescriptor> additionalModuleColumns = null;
+            if (responder.Arguments.ClientID == "visualstudio")
+            {
+                additionalModuleColumns = new List<ColumnDescriptor>();
+                additionalModuleColumns.Add(new ColumnDescriptor("vsLoadAddress", "Load Address", "string", ColumnDescriptor.TypeValue.String));
```
WardenGnaw, line 652: The builds are outputting warnings for ColumnDescriptor
```
##[warning]src\OpenDebugAD7\AD7DebugSession.cs(652,45): Warning CS0618: 'ColumnDescriptor.ColumnDescriptor(string, string, string, ColumnDescriptor.TypeValue?, int?)' is obsolete: 'This constructor contains optional parameters and is no longer supported.  Please use the constructor which includes only required parameters.'
```
Please address the other warnings too:
https://dev.azure.com/ms/MIEngine/_build/results?buildId=115779&view=results

You can ignore
```
Warning MSB3277: Found conflicts between different versions of "Newtonsoft.Json" that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed.
```

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: this should be the first time ClientID is used. 

Other instances try to differentiate VS/VSCode with DebugEngineHost, but now that DAP works for VS Code/VS/LiveShare we will need to keep track of ClientID.

gc46: Correct - this is the first time ClientID is used in MIEngine.

src/OpenDebugAD7/AD7DebugSession.cs
gc46: Is there a reason we need list as well?

gregg-miskelly: 'list' (or better yet pulling out the code that figures out the module id from this function) is required for 'HandleModulesRequestAsync' -- right now your code is treating that as an add, which is wrong.

---
In reply to: [505074678](https://github.com/microsoft/MIEngine/pull/1054#discussion_r505074678) [](ancestors = 505074678)

src/OpenDebugAD7/AD7DebugSession.cs
gc46: As you mentioned below, using HandleCollection only allows us to retrieve an IDebugModule2 given an int but not the other way around.

src/OpenDebugAD7/AD7DebugSession.cs
gc46: The return type for RegisterDebugModule() is an object and not an int because of the Debug.Assert(), correct?

gregg-miskelly: The 'id' field of a Protocol module is an object, so I made it an object so that we could change `m_nextModuleHandle` to something else if we wanted to without changing anything else. But you could certainly leave it as an int if you like.

src/OpenDebugAD7/AD7DebugSession.cs
gc46: The ColumnDescriptor warning is unavoidable for Timestamp.

WardenGnaw: The full parameterized constructor is obsolete/depreciated. You will want to use either the default constructor or the one with two required parameters.

I prefer just using the default constructor and providing the parameters.

```
additionalModuleColumns.Add(new ColumnDescriptor()
{
   AttributeName = "vsLoadAddress",
   Label = "Load Address",
   Format = "string",
   Type = ColumnDescriptor.TypeValue.String
});
```

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1595,7 +1597,7 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
             responder.SetResponse(response);
         }
 
-        public static ProtocolMessages.Module ConvertToModule(in MODULE_INFO debugModuleInfo, IDebugModule2 module, bool isLoad)
+        private ProtocolMessages.Module ConvertToModule(in MODULE_INFO debugModuleInfo, IDebugModule2 module, bool isLoad)
```
WardenGnaw, line 1600: I would convert this method to:
```suggestion
        private ProtocolMessages.Module ConvertToModule(in MODULE_INFO debugModuleInfo, int moduleId)
```

and in `HandleIDebugModuleLoadEvent2`
```
int? moduleId = null;
ModuleEvent.ReasonValue reason = ModuleEvent.ReasonValue.Unknown;
if (isLoad)
{
   moduleId = RegisterDebugModule(module);
   reason = ModuleEvent.ReasonValue.New;
}
else
{
    moduleId = ReleaseDebugModule(module);
    reason = ModuleEvent.ReasonValue.Removed;
}

if (moduleId != null)
{
   mod = ConvertToModule(debugModuleInfos[0], moduleId);
   Protocol.SendEvent(new ModuleEvent(reason, mod));
}
```

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -67,26 +67,27 @@ internal class AD7DebugSession : DebugAdapterBase, IDebugPortNotify2, IDebugEven
...
             {
-                return new IntPtr(s_debugModules.Create(debugModule));
+                int moduleHandle = m_nextModuleHandle;
+                m_moduleMap[debugModule] = moduleHandle;
+                m_nextModuleHandle++;
+                return moduleHandle;
             }
         }
-        public static void ReleaseDebugModuleId(IntPtr debugModuleId)
+        private void ReleaseDebugModule(IDebugModule2 debugModule)
```
WardenGnaw, line 84:
```suggestion
        private int? ReleaseDebugModule(IDebugModule2 debugModule)
        {
           lock(m_moduleMap)
           {
              if (m_moduleMap.TryGetValue(module, out int moduleId))
              {
                 m_moduleMap.Remove(debugModule);
                 return moduleId;
              }
              else
              {
                 Debug.Fail("Trying to unload a module that has not been registered.");
                 return null;
              }
           }
        }
```

WardenGnaw (commented):
Let me know if you have any questions.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: VSCodeDebugAdapter hosts sends a `ModulesRequest` on the start of an `attach`. 

This is in `RequestInitialModules`, so these modules should be `new`.

src/OpenDebugAD7/AD7DebugSession.cs
gc46: I will move the IsUserCode assignment to HandleIDebugModuleLoadEvent2().

WardenGnaw: You could modify this to take a IDebugModule2 and just call IDebugModule2.GetInfo in here.

gc46: I am confused on how to handle HandleModulesRequestAsync().

WardenGnaw: Any modules enumerated from `IEnumDebugModules2.EnumModules` would be new modules. Treat them like a load event.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1546,6 +1631,79 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
...
+            IEnumDebugModules2 enumDebugModules;
+            if (m_program.EnumModules(out enumDebugModules) == HRConstants.S_OK)
+            {
+                var debugModules = new IDebugModule2[1];
+                uint numReturned = 0;
+                while (enumDebugModules.Next(1, debugModules, ref numReturned) == HRConstants.S_OK && numReturned == 1)
+                {
+                    IDebugModule2 module = debugModules[0];
+                    var moduleId = RegisterDebugModule(module);
+                    if (moduleId != null)
```
gregg-miskelly, line 1696: You don't want `RegisterDebugModule` here -- the module should already be in the collection.

```suggestion
                    int moduleId;
                    lock (m_moduleMap)
                    {
                        if (!m_moduleMap.TryGetValue(debugModule, out moduleId))
                        {
                            Debug.Fail("Missing ModuleLoadEvent?");
                            continue;
                        }
                    }
                    
                    var mod = ConvertToModule(module, moduleId);
                    response.Modules.Add(mod);
```

gregg-miskelly (approved):
Otherwise looks good to me.

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: Sorry, blame Andrew. He gave you bad advice :).

gregg-miskelly: Andrew, is there any case I am not thinking of where these module could be new? If so, instead of asserting and ignoring, you could call `RegisterModule` there.

WardenGnaw (approved):
LGTM once you address Gregg's comment. 

Please make sure to use squash and merge, and clean up the description before completing it. 
image

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: Yep. I misunderstood how IDebugProgram's module's list and IDebugModuleEvent worked. I got it clarified.

They should all be existing, so asserting is fine.

WardenGnaw: @gregg-miskelly Do we want to support ModulesArguments's `startModule` and `moduleCount`? Or do we want to always return the full module list.

`startModule` is going to be weird with the Dictionary implementation of keeping track of the modules.

gregg-miskelly: In looks like VS will always request all modules. I am not sure why any IDE would want a subset. Probably fine to just return all of them until we have a reason to implement it.

gregg-miskelly (approved):

WardenGnaw:
> > ```
> >                     {
> > ```
> 
> I am not sure if you want to fix this now, but you can now fill out the 'moduleId' here: https://github.com/microsoft/debug-adapter-protocol/blob/fdaa68760ff4c22dab619b1dbb8cfa8f80b6966f/debugAdapterProtocol.json#L3215
> 
> Refers to: src/OpenDebugAD7/AD7DebugSession.cs:1306 in [710658d](https://github.com/microsoft/MIEngine/commit/710658d5102f7392430d126fe4bc0dd775157426). [](commit_id = [710658d](https://github.com/microsoft/MIEngine/commit/710658d5102f7392430d126fe4bc0dd775157426), deletion_comment = False)

Created https://github.com/microsoft/MIEngine/issues/1058 to track.

PR #1055, opened 2020-10-09
Fix SourceFileMap in MICore

WardenGnaw:
Added PathNormalization methods in PlatformUtilities that is both used
in MICore and MIDebugEngine.

This change will ensure all paths created for sourceFileMap are
normalized to the client. E.g. Windows '\' if on Windows and '/' if on
macOS or Unix.

When submitting for breakpoints in MIEngine, it will normalize compiler
sources to use Unix styled paths for GDB.

This also fixes the issue that we do not correctly normalize the compilerSrcPath before checking it with sourceFileMappings for MITextPositions.

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -379,20 +379,27 @@ namespace MICore
                 {
                     editorPath = value;
                 }
-                else if (item.Value is JObject)
+                else if (item.Value is JObject jObject)
                 {
-                    Json.LaunchOptions.SourceFileMapOptions sourceMapItem =
-                        ((JObject)item.Value).ToObject<Json.LaunchOptions.SourceFileMapOptions>();
+                    try
+                    {
+                        Json.LaunchOptions.SourceFileMapOptions sourceMapItem =
+                        jObject.ToObject<Json.LaunchOptions.SourceFileMapOptions>();
 
-                    editorPath = sourceMapItem.EditorPath;
-                    useForBreakpoints = sourceMapItem.UseForBreakpoints.GetValueOrDefault(true);
+                        editorPath = sourceMapItem.EditorPath;
+                        useForBreakpoints = sourceMapItem.UseForBreakpoints.GetValueOrDefault(true);
+                    }
+                    catch (JsonReaderException)
+                    {
+                        throw new InvalidLaunchOptionsException(String.Format(CultureInfo.CurrentCulture, MICoreResources.Error_SourceFileMapFormat, compileTimePath));
+                    }
                 }
                 else
                 {
-                    throw new InvalidLaunchOptionsException(String.Format(CultureInfo.CurrentCulture, MICoreResources.Error_SourceFileMapFormat, item.Key));
+                    throw new InvalidLaunchOptionsException(String.Format(CultureInfo.CurrentCulture, MICoreResources.Error_SourceFileMapFormat, compileTimePath));
                 }
 
-                if (!string.IsNullOrEmpty(compileTimePath) && !string.IsNullOrEmpty(editorPath))
+                if (!string.IsNullOrEmpty(editorPath))
                 {
                     sourceMaps.Add(new SourceMapEntry()
                     {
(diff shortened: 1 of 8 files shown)


src/MICore/PlatformUtilities.cs
```diff
@@ -131,6 +131,33 @@ public static string GetEnvironmentVariable(this ProcessStartInfo processStartIn
...
+        }
+
+        public static string NormalizeClientPath(string path)
+        {
+            if (string.IsNullOrWhiteSpace(path))
+            {
+                return path;
+            }
+
+            if (IsWindows())
```
gregg-miskelly, line 152: I think this is the wrong check -- if you are trying to normalize to what we should be returning to the client then you want to reverse this (on Windows we should be calling `UnixPathToWindowsPath`). If you are trying to normalize to what you should be sending to GDB then you want to check the OS of the target rather than the OS of MI Engine.

src/MIDebugEngine/Engine.Impl/MITextPosition.cs
```diff
@@ -35,10 +35,6 @@ public MITextPosition(string filename, uint line)
         public static MITextPosition TryParse(DebuggedProcess process, TupleValue miTuple)
         {
             string filename = process.GetMappedFileFromTuple(miTuple);
```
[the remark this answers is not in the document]
WardenGnaw: I decided to migrate this to the code that actually does the map, so not every caller has to do normalization before calling `MapCompileTimeSrcToCurrentSrc`

https://github.com/microsoft/MIEngine/blob/ee2b090f343f88104c6e7f1c9d0b2b7059531113/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs#L2217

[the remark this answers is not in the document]
gregg-miskelly: Got it.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -2218,6 +2213,9 @@ public bool MapCompileTimeSrcToCurrentSrc(string compilerSrc, out string current
         {
             if (_launchOptions.SourceMap != null)
             {
+                // Convert to Client source paths
+                compilerSrc = PlatformUtilities.PathToHostOSPath(compilerSrc);
```
gregg-miskelly, line 2217: >compilerSrc [](start = 16, length = 11)

This is going to change the return value even in the case that we don't find the path (see line 2249), but only if there is a source map. You probably instead want to assign to a new local, and change the uses of `compilerSrc` on line 2219 to 2247 to use your new local variable.

src/MICore/LaunchOptions.cs
```diff
@@ -370,30 +371,36 @@ public static ReadOnlyCollection<SourceMapEntry> CreateCollection(Dictionary<str
...
-                        CompileTimePath = item.Key,
-                        EditorPath = (String)item.Value,
-                        UseForBreakpoints = true
-                    });
+                    editorPath = value;
                 }
                 else if (item.Value is JObject)
                 {
                     Json.LaunchOptions.SourceFileMapOptions sourceMapItem =
                         ((JObject)item.Value).ToObject<Json.LaunchOptions.SourceFileMapOptions>();
```
pieandcakes, line 385: try.catch this

src/MICore/PlatformUtilities.cs
```diff
@@ -131,6 +131,33 @@ public static string GetEnvironmentVariable(this ProcessStartInfo processStartIn
...
         }
+
+        public static string UnixPathToWindowsPath(string unixPath)
+        {
+            return unixPath.Replace('/', '\\');
+        }
+
+        public static string WindowsPathToUnixPath(string windowsPath)
+        {
+            return windowsPath.Replace('\\', '/');
```
pieandcakes, line 142: this might not work properly if they are using network shares.

pieandcakes (approved):

src/MICore/PlatformUtilities.cs
WardenGnaw: I think this might be broken if there are mounted paths in Windows and you are locally on a unix machine. 

E.g. the Windows network shares will have a drive `Z:\` or UNC path `\\path\to\UNC` but then you will need to map it to a Unix mnt path. E.g. `/mnt/UNCPath`

gregg-miskelly (approved):

PR #1059, opened 2020-10-19
Support running Setup.csx on non-Windows

WardenGnaw:


tools/Setup.sh
```diff
@@ -0,0 +1,16 @@
+if ! hash dotnet 2>/dev/null; then
```
gregg-miskelly, line 1: Add `#!/bin/bash` (or whatever shell you want to require)

tools/Setup.csx
```diff
@@ -58,7 +58,7 @@ class Setup {
 
         foreach (string arg in args)
```
gregg-miskelly, line 59: Add a line to PrintHelp to point at the script files?

gregg-miskelly (approved):
Otherwise LGTM

PR #1061, opened 2020-10-20
Include MS.VS.OLE.Interop.dll for XPLAT

WardenGnaw:
We also need Microsoft.VisualStudio.OLE.Interop.dll for VS Code since it will be sending ModuleEvents and parsing FILETIME.


gregg-miskelly:
I don't understand why we are dropping any of these files in non-XPlat. Do you know?

WardenGnaw:
I thought that it was dropped for `<VS_ROOT>\Common7\IDE\CommonExtensions\Microsoft\MDD\Debugger\vscode`.
The VS -> VSCodeDebugAdapterHost -> OpenDebugAD7.exe

gregg-miskelly:
Got it. We should probably fix App.VS.config to reference these files from PrivateAssemblies (or add a AssemblyResolve hook for that) so we don't need to drop a bunch of extra copies of dlls that VS will already drop. But no need to fix this now I supposed.

gregg-miskelly (approved):

WardenGnaw:
Tracking Fixing App.VS.config in https://github.com/microsoft/MIEngine/issues/1062

PR #1067, opened 2020-10-28
Removing old Checksum Code for ClrDbg

WardenGnaw:
GDB does not implement checksums for their files/symbols so removing the
code for this and lowering the OpenDebugAD7 version to 4.6.2. Keeping
the split between Debug and Desktop.Debug since we may move away from
mono and use netstandard and netcoreapp.


src/OpenDebugAD7/OpenDebugAD7.csproj
```diff
@@ -23,7 +23,7 @@
     <OutputType>Exe</OutputType>
     <DropSubDir>vscode</DropSubDir>
     <TargetFrameworkVersion Condition="'$(IsXPlat)' == 'true'">v4.6.2</TargetFrameworkVersion>
-    <TargetFrameworkVersion Condition="'$(IsXPlat)' == 'false'">v4.7.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion Condition="'$(IsXPlat)' == 'false'">v4.6.2</TargetFrameworkVersion>
```
gregg-miskelly, line 26: nit: remove the condition

gregg-miskelly (approved):
Otherwise LGTM

PR #1069, opened 2020-10-28
Removing old Checksum Code for ClrDbg (#1067)

lancesun:
* Removing old Checksum Code for ClrDbg

GDB does not implement checksums for their files/symbols so removing the
code for this and lowering the OpenDebugAD7 version to 4.6.2. Keeping
the split between Debug and Desktop.Debug since we may move away from
mono and use netstandard and netcoreapp.


WardenGnaw:
@lancesun Do you have a scenario which is using the checksum paths?

PR #1070, opened 2020-10-29
Fix DebugServerPath for JSON parsing

WardenGnaw:
The XML creation for InitializeServerOptions checks for DebugServerPath
being null or whitespace. Making JSON's InitializeServerOptions the
same.

See: 
https://github.com/microsoft/MIEngine/blob/728f403c563c2c50c46542e9b44271a7f8d5895b/src/MICore/LaunchOptions.cs#L461-L476

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -443,7 +443,7 @@ namespace MICore
 
         private void InitializeServerOptions(Json.LaunchOptions.LaunchOptions launchOptions)
         {
-            if (!String.IsNullOrWhiteSpace(launchOptions.MiDebuggerServerAddress))
+            if (!String.IsNullOrWhiteSpace(launchOptions.DebugServerPath))
             {
                 this.DebugServer = launchOptions.DebugServerPath;
                 this.DebugServerArgs = launchOptions.DebugServerArgs;


src/MICore/LaunchOptions.cs
```diff
@@ -443,7 +443,7 @@ public LocalLaunchOptions(string MIDebuggerPath, string MIDebuggerServerAddress)
 
         private void InitializeServerOptions(Json.LaunchOptions.LaunchOptions launchOptions)
         {
-            if (!String.IsNullOrWhiteSpace(launchOptions.MiDebuggerServerAddress))
+            if (!String.IsNullOrWhiteSpace(launchOptions.DebugServerPath))
```
pieandcakes, line 446: I think the original code was if the miDebuggerServerAddress was set, then use debug server. You are changing the behavior here slightly with this change so that it doesn't require miDebuggerServerAddress to be set anymore. You sure that is what you want to change here?

WardenGnaw: This was what we had in the XML case and is now breaking a customer in the C++ extension. Because XML didnt care but JSON does, we now don't run DebugServerPath. 

See Issue: https://github.com/microsoft/vscode-cpptools/issues/6411

pieandcakes: Right, but I'm saying I think the right behavior is that we need miDebuggerServerAddress and the other issue he is complaining about is the actual reason which is he wants to specify the miDebuggerPath which he can't.

pieandcakes: You should also fix that problem. 

This change makes it so i can always use a debug server regardless if an midebuggerServeraddress is specified which is what i think is wrong.

gregg-miskelly: @pieandcakes In case it isn't clear, what Andrew is trying to say is that right now there is a difference between the JSON and XML path. In XML, the check is for `DebugServerPath` instead of `MiDebuggerServerAddress`. See [this line](https://github.com/microsoft/MIEngine/blob/728f403c563c2c50c46542e9b44271a7f8d5895b/src/MICore/LaunchOptions.cs#L463).

I would think that one would generally want to set `MiDebuggerServerAddress` when using client/server transport. I guess the case you don't is if you are also using a setup step that already defaults to a port?

gregg-miskelly (approved):

src/MICore/LaunchOptions.cs
WardenGnaw: Looks like there is already a tool that generates a `launch.json` that sets `DebugServerPath` and the associated customSetupCommand for `-target-select remote` command. 

Reference: https://www.elastetic.com/wp/2018/05/19/visual-studio-code-as-an-ide-for-mbed/

I think we can allow users to use `DebugServerPath` without `MiDebuggerServerAddress`, but update the description on `DebugServerPath` to say:

 ` "Optional full path to debug server to launch. Defaults to null. It is used in conjuction of either "miDebugServerAddress" or your own server".`

PR #1072, opened 2020-10-30
Set SetupCommands to be common for JSON parsing

WardenGnaw:
SetupCommands is used for both launch and attach. For JSON, it was only
set in launch. This PR fixes it so it does it for both launch and attach
by setting it as a common option.

https://github.com/microsoft/vscode-cpptools/issues/6417


andrewcrawley (approved):
LGTM!

PR #1074, opened 2020-11-03
[GDB] Support Disassembly window for OpenDebugAD7 scenarios

gc46:


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -303,6 +303,31 @@ private static long FileTimeToPosix(FILETIME ft)
             return date / 10000000;
         }
 
+        private IDebugMemoryContext2 GetMemoryContext(string memoryReference, int? offset, out ulong address, out int hr)
```
WardenGnaw, line 306: nit: Normally we return the `hr` as the result of the function and make the object we want an `out` param.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -701,7 +726,8 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
                 SupportsLogPoints = true,
                 SupportsReadMemoryRequest = true,
                 SupportsModulesRequest = true,
-                AdditionalModuleColumns = additionalModuleColumns
+                AdditionalModuleColumns = additionalModuleColumns,
+                SupportsDisassembleRequest = true
```
WardenGnaw, line 730: We need to double check if this is valid for `lldb-mi`.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1667,6 +1693,49 @@ protected override void HandleModulesRequestAsync(IRequestResponder<ModulesArgum
             responder.SetResponse(response);
         }
 
+        // test -- need to delete
```
gregg-miskelly, line 1696: >// test -- need to delete [](start = 8, length = 25)

Do you still want this comment here?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1667,6 +1693,49 @@ protected override void HandleModulesRequestAsync(IRequestResponder<ModulesArgum
...
+                                    EndLine = (int?)data.posEnd.dwLine,
+                                    EndColumn = (int?)data.posEnd.dwColumn
+                                };
+                                response.Instructions.Add(instruction);
+                            }
+                        }
+                    }
+
+                }
+            } catch { }
```
gregg-miskelly, line 1734: >{ [](start = 20, length = 1)

You want a `responder.SetError` call in the catch and move `responder.SetResponse(response);` to the end of the try block.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1667,6 +1693,49 @@ protected override void HandleModulesRequestAsync(IRequestResponder<ModulesArgum
...
+        protected override void HandleDisassembleRequestAsync(IRequestResponder<DisassembleArguments, DisassembleResponse> responder)
+        {
+            DisassembleResponse response = new DisassembleResponse();
+
+            int hr;
+            DisassembleArguments disassembleArguments = responder.Arguments;
+            Debug.Assert(!string.IsNullOrEmpty(disassembleArguments.MemoryReference));
+            try
+            {
+                IDebugMemoryContext2 memoryContext = GetMemoryContext(disassembleArguments.MemoryReference, disassembleArguments.Offset, out ulong address, out hr);
```
gregg-miskelly, line 1706: >hr [](start = 160, length = 2)

Throw if hr is failed?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1667,6 +1693,49 @@ protected override void HandleModulesRequestAsync(IRequestResponder<ModulesArgum
...
+                if (m_program.GetDisassemblyStream(enum_DISASSEMBLY_STREAM_SCOPE.DSS_ALL, (IDebugCodeContext2)memoryContext , out IDebugDisassemblyStream2 disassemblyStream) == HRConstants.S_OK)
```
gregg-miskelly, line 1707: >(IDebugCodeContext2)memoryContext [](start = 90, length = 33)

We should probably do an `as` here and throw if the memory context isn't a code context.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1667,6 +1693,49 @@ protected override void HandleModulesRequestAsync(IRequestResponder<ModulesArgum
...
+                {
+                    if (disassemblyStream.Seek(enum_SEEK_START.SEEK_START_BEGIN, (IDebugCodeContext2)memoryContext, address, (long)disassembleArguments.InstructionOffset) == HRConstants.S_OK)
```
gregg-miskelly, line 1709: >== HRConstants.S_OK [](start = 171, length = 19)

Throw if these operations fail instead of returning nothing?

gregg-miskelly, line 1709: >disassembleArguments.InstructionOffset [](start = 131, length = 38)

nit: We only need to call `Seek` if `InstructionOffset` is non-zero

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1667,6 +1693,49 @@ protected override void HandleModulesRequestAsync(IRequestResponder<ModulesArgum
...
+                                DisassembledInstruction instruction = new DisassembledInstruction() {
+                                    Address = data.bstrAddress,
+                                    InstructionBytes = data.bstrCodeBytes,
+                                    Instruction = data.bstrOpcode,
+                                    Symbol = data.bstrSymbol,
+                                    // Location = data.uCodeLocationId,
+                                    Line = (int?)data.posBeg.dwLine,
+                                    Column = (int?)data.posBeg.dwColumn,
+                                    EndLine = (int?)data.posEnd.dwLine,
+                                    EndColumn = (int?)data.posEnd.dwColumn
```
gregg-miskelly, line 1726: As discussed, you don't want to assign to these fields, and add an Assert that the engine doesn't return source info

src/OpenDebugAD7/AD7DebugSession.cs
gc46: This will be removed once all changes are completed.

src/OpenDebugAD7/AD7DebugSession.cs
gc46: I assume I will need to create a new error string in AD7Resources. Is this correct?

gregg-miskelly: Yes

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: We should be ok:
https://github.com/lldb-tools/lldb-mi/blob/a9faab69fc30ac2bba0521bbcf9e4931071272cf/src/MICmdCmdData.h#L94

src/OpenDebugAD7/AD7Resources.Designer.cs
```diff
@@ -524,5 +524,55 @@ internal class AD7Resources {
                 return ResourceManager.GetString("Warning_Scenario_TerminateProcess", resourceCulture);
             }
         }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Failed to get memory context.
+        ///.
+        /// </summary>
+        internal static string Error_GetMemoryContext {
```
gregg-miskelly, line 532: >internal [](start = 8, length = 8)

I think you must have hand edited this file and the .resx file, because you wound up putting your .resx changes in the wrong .resx file.

gc46: Yes, I did. Thank you for catching that.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1667,6 +1693,69 @@ protected override void HandleModulesRequestAsync(IRequestResponder<ModulesArgum
...
 
+        // test -- need to delete
+        protected override void HandleDisassembleRequestAsync(IRequestResponder<DisassembleArguments, DisassembleResponse> responder)
+        {
+            DisassembleResponse response = new DisassembleResponse();
+            DisassembleArguments disassembleArguments = responder.Arguments;
+            Debug.Assert(!string.IsNullOrEmpty(disassembleArguments.MemoryReference));
+            try
+            {
+                if (GetMemoryContext(disassembleArguments.MemoryReference, disassembleArguments.Offset, out IDebugMemoryContext2 memoryContext, out ulong address) == HRConstants.S_OK)
```
gregg-miskelly, line 1704: >if (GetMemoryContext(disassembleArguments.MemoryReference, disassembleArguments.Offset, out IDebugMemoryContext2 memoryContext, out ulong address) == HRConstants.S_OK) [](start = 16, length = 167)

Rather than adding a different resource string for each failure and ignoring the HRESULTs that you get, I would instead suggest defining one new resource string: `Error_Scenario_Disassemble = "Unable to disassemble. {0}"` and you can do:

```C#
            var eb = new ErrorBuilder(() => AD7Resources.Error_Scenario_Disassemble);

            eb.CheckHR(GetMemoryContext(disassembleArguments.MemoryReference, disassembleArguments.Offset, out IDebugMemoryContext2 memoryContext, out ulong address));
            
            IDebugCodeContext2 codeContext = memoryContext as IDebugCodeContext2;
            if (codeContext == null)
            {
                eb.CheckHR(HRConstants.E_NOTIMPL);
            }

            eb.CheckOutput(m_program.GetDisassemblyStream(enum_DISASSEMBLY_STREAM_SCOPE.DSS_ALL, codeContext, out IDebugDisassemblyStream2 disassemblyStream));
```

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1667,6 +1693,69 @@ protected override void HandleModulesRequestAsync(IRequestResponder<ModulesArgum
...
+                    {
+                        throw new AD7Exception(AD7Resources.Error_GetDisassemblyStream);
+                    }
+                }
+                else
+                {
+                    throw new AD7Exception(AD7Resources.Error_GetMemoryContext);
+                }
+                responder.SetResponse(response);
+            } catch (AD7Exception e) {
```
gregg-miskelly, line 1754: >AD7Exception [](start = 21, length = 12)

Probably want to catch all exceptions

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -303,6 +303,31 @@ private static long FileTimeToPosix(FILETIME ft)
...
+        private int GetMemoryContext(string memoryReference, int? offset, out IDebugMemoryContext2 memoryContext, out ulong address)
+        {
+            if (memoryReference.StartsWith("0x", StringComparison.Ordinal))
+            {
+                address = Convert.ToUInt64(memoryReference.Substring(2), 16);
+            }
+            else
+            {
+                address = Convert.ToUInt64(memoryReference, 10);
+            }
```
pieandcakes, line 315: nit: newline after this `if/else` block will make it more readable.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1667,6 +1693,69 @@ protected override void HandleModulesRequestAsync(IRequestResponder<ModulesArgum
...
+                        }
+                        DisassemblyData[] prgDisassembly = new DisassemblyData[disassembleArguments.InstructionCount];
+                        if (disassemblyStream.Read((uint)disassembleArguments.InstructionCount, enum_DISASSEMBLY_STREAM_FIELDS.DSF_ALL, out uint pdwInstructionsRead, prgDisassembly) == HRConstants.S_OK)
+                        {
+                            Debug.Assert(disassembleArguments.InstructionCount == pdwInstructionsRead);
+                            foreach (DisassemblyData data in prgDisassembly)
+                            {
+                                if (data.dwFlags.HasFlag(enum_DISASSEMBLY_FLAGS.DF_HASSOURCE))
+                                {
+                                    Debug.Fail("Warning: engine supports mixed instruction/source disassembly, but OpenDebugAD7 does not.");
```
pieandcakes: Would this be better off as a logged warning instead of a Debug.Fail? The end user won't see this message and you as the developer of MIEngine/OpenDebugAD7 will only see this with debug builds.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1667,6 +1693,69 @@ protected override void HandleModulesRequestAsync(IRequestResponder<ModulesArgum
...
+        {
+            DisassembleResponse response = new DisassembleResponse();
+            DisassembleArguments disassembleArguments = responder.Arguments;
+            Debug.Assert(!string.IsNullOrEmpty(disassembleArguments.MemoryReference));
+            try
+            {
+                if (GetMemoryContext(disassembleArguments.MemoryReference, disassembleArguments.Offset, out IDebugMemoryContext2 memoryContext, out ulong address) == HRConstants.S_OK)
+                {
+                    IDebugCodeContext2 codeContext = memoryContext as IDebugCodeContext2;
+                    if (!(codeContext is IDebugCodeContext2))
```
pieandcakes: when will this actually be false? Your line above is the cast. I think your check here should be if `codeContext == null` indicating a cast failure instead of a type check.

gc46: See latest commit.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1667,6 +1693,69 @@ protected override void HandleModulesRequestAsync(IRequestResponder<ModulesArgum
...
+                    {
+                        throw new AD7Exception(AD7Resources.Error_ConvertMemoryContext);
+                    }
+                    if (m_program.GetDisassemblyStream(enum_DISASSEMBLY_STREAM_SCOPE.DSS_ALL, codeContext, out IDebugDisassemblyStream2 disassemblyStream) == HRConstants.S_OK)
+                    {
+                        if (disassembleArguments.InstructionOffset != 0)
```
pieandcakes: According to the spec (https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Disassemble) InstructionOffset is a nullable int. This should probably be `disassembleArguments.InstructionOffset.GetValueOrDefault(0) != 0`

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: That seems fine to me -- this is here so that if we add support in the MIEngine (or any other engine we might use in the future), and are wondering why it doesn't work in OpenDebugAD7, we will know right away.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: I would use `disassembleArguments.InstructionOffset.HasValue` in the conditional and then use `disassembleArguments.InstructionOffset.Value` when using it.

WardenGnaw (approved):

Issue #1081, opened 2020-11-19
[VS Code] ReadMemory request causes cast exception on macOS


gareth-davies-img:
I'm issuing a ReadMemory request from a VSCode extension using:

`const readMemoryResp = await vscode.debug.activeDebugSession.customRequest('readMemory', { memoryReference: memoryReference, count: count, offset: offset });`

It works absolutely fine on Windows. However on macOS (Catalina 10.15.7) the following exception occurs:

"Specified cast is not valid."

I have traced it to the following line of code:

`int hr = ((IDebugMemoryBytesDAP)m_engine).CreateMemoryContext(address, out memoryContext);`

in AD7DebugSession.cs in the function GetMemoryContext().

WardenGnaw:
@gareth-davies-img Thank you for finding this bug and creating this issue.

One question, are you using your own custom engine and loading it with a different configuration or are you using [cppdbg.ad7Engine.json](https://github.com/microsoft/MIEngine/blob/main/src/OpenDebugAD7/cppdbg.ad7Engine.json) which uses MIEngine's [AD7Engine](https://github.com/microsoft/MIEngine/blob/9470e98d946772ed1ace9ab68e9fbd4d88090b42/src/MIDebugEngine/AD7.Impl/AD7Engine.cs#L38)? 

If you are using another engine, we will fix this type cast to check to see if `IDebugMemoryBytesDAP` is implemented before calling `CreateMemoryContext`. But if the cast fails, then the `readMemory` request will result in an error.

My first thought with this issue it may be a mono framework issue since it is working on Windows, which would use .NET Framework. One good way to check is to see if this call also fails in Linux.

We are planning to move to .NET Core which may fix this casting issue, if not we will need to file an issue on the runtime.

gareth-davies-img:
@WardenGnaw Many thanks for looking at this issue.

We are using VSCode 1.51.1 and cpptools 1.1.2 out the box, so are using the standard cppdbg.ad7Engine.json.

And I can confirm that the same cast exception happens on Linux.

WardenGnaw:
Yep. This most likely is a bug with mono and how it handles embedded interfaces.

We have not had the chance to test `readMemory` requests on macOS/Linux. 

https://github.com/microsoft/MIEngine/pull/1082 should fix the cast exception, but the cast may still fail in mono.
[...]

PR #1082, opened 2020-11-19
Support non-IDebugMemoryBytesDAP engines

WardenGnaw:
There can be cases where the engine loaded is not the latest AD7Engine or the Microsoft.MIDebugEngine. 
In GetMemoryContext, we do a hard cast to IDebugMemoryBytesDAP. This
will error with "Specified cast is not valid" if the engine does not
implement IDebugMemoryBytesDAP and the engine is not the latest
Microsoft.MIDebugEngine.

This PR converts the hard cast to be a `is` check. If the engine is not
IDebugMemoryBytesDAP, it will return E_NOTIMPL.

Related: https://github.com/microsoft/MIEngine/issues/1081

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -305,6 +305,8 @@ namespace OpenDebugAD7
 
         private int GetMemoryContext(string memoryReference, int? offset, out IDebugMemoryContext2 memoryContext, out ulong address)
         {
+            memoryContext = null;
+
             if (memoryReference.StartsWith("0x", StringComparison.Ordinal))
             {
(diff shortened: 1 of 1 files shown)


WardenGnaw:
Should we send `false` for `SupportsReadMemoryRequest` if `m_engine is not IDebugMemoryBytesDAP`, too?

https://github.com/microsoft/MIEngine/blob/f3fafeb3dd1e7f50390d81e1348d1b29b31150b2/src/OpenDebugAD7/AD7DebugSession.cs#L729

gregg-miskelly:
> Should we send `false` for `SupportsReadMemoryRequest` if `m_engine is not IDebugMemoryBytesDAP`, too?

Ideally, yes

gregg-miskelly (approved):
Aside from the capabilities, looks good to me.

PR #1084, opened 2020-12-01
Support "format" field on "stackTrace" request

andrewcrawley:
Adds support for the "value formatting" feature of DAP, which allows flags to be passed to control formatting of stack frames, as well as variables and evaluation results.

Note that while this commit wires up the DAP formatting flags to the relevant AD7 calls, MIEngine currently doesn't make use of the radix value passed, so the flag to control hex vs decimal formatting doesn't work.

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -39,6 +39,7 @@ namespace OpenDebugAD7
         private IDebugEngine2 m_engine;
         private EngineConfiguration m_engineConfiguration;
         private AD7Port m_port;
+        private DebugSettingsCallback m_settingsCallback;
 
         private readonly DebugEventLogger m_logger;
         private readonly Dictionary<string, Dictionary<int, IDebugPendingBreakpoint2>> m_breakpoints;
@@ -642,6 +643,15 @@ namespace OpenDebugAD7
             m_port = new AD7Port(this);
             m_disconnectedOrTerminated = new ManualResetEvent(false);
             m_firstStoppingEvent = 0;
+
+            if (m_engine is IDebugEngine110 engine110)
+            {
+                // MIEngine generally gets the radix from IDebugSettingsCallback110 rather than using the radix passed to individual
+                //  APIs.  To support this mechanism outside of VS, provide a fake settings callback here that we can use to control
+                //  the radix.
+                m_settingsCallback = new DebugSettingsCallback();
+                engine110.SetMainThreadSettingsCallback110(m_settingsCallback);
+            }
 
             m_pathConverter.ClientLinesStartAt1 = arguments.LinesStartAt1.GetValueOrDefault(true);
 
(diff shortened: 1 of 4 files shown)


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1274,31 +1272,77 @@ protected override void HandleStackTraceRequestAsync(IRequestResponder<StackTrac
...
-                            frameEnumInfo = new ThreadFrameEnumInfo(frameEnum, totalFrames);
-                            m_threadFrameEnumInfos.Add(threadReference, frameEnumInfo);
+                            // No formatting flags provided in the request - use the default format, which includes the module name and argument names / types
+                            flags |= enum_FRAMEINFO_FLAGS.FIF_FUNCNAME_MODULE |
+                                     enum_FRAMEINFO_FLAGS.FIF_FUNCNAME_ARGS |
+                                     enum_FRAMEINFO_FLAGS.FIF_FUNCNAME_ARGS_TYPES |
+                                     enum_FRAMEINFO_FLAGS.FIF_FUNCNAME_ARGS_NAMES;
                         }
+
+                        thread.EnumFrameInfo(flags, radix, out IEnumDebugFrameInfo2 frameEnum);
```
gregg-miskelly, line 1336: >thread [](start = 24, length = 6)

Looks like we had this problem before also, but: check hr?

gregg-miskelly:
>             property.GetPropertyInfo(propertyInfoFlags, Constants.EvaluationRadix, Constants.EvaluationTimeout, null, 0, propertyInfo);

Should we wire up this one too, or at least leave a TODO comment?

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:2270 in 7375184. [](commit_id = 73751849ddfec619d50f5af2cda445f503c22800, deletion_comment = False)

gregg-miskelly:
Sorry if you already know this, but if we want this to really work, we will need to have OpenDebugAD7 start calling SetMainThreadSettingsCallback110, and implement IDebugSettingsCallback110.GetDisplayRadix.

andrewcrawley:
> 
> 
> Sorry if you already know this, but if we want the radix stuff to work, we will need to have OpenDebugAD7 start calling SetMainThreadSettingsCallback110, and implement IDebugSettingsCallback110.GetDisplayRadix.

I considered this, but I was concerned about race conditions when (temporarily?) changing global state in response to an async DAP request.  It's common to simultaneously issue several `stackTrace` requests with different formatting options, for instance.  Is there a reason MIEngine always refers to the global setting instead of using the radix passed via the individual APIs?

gregg-miskelly:
> Is there a reason MIEngine always refers to the global setting instead of using the radix passed via the individual APIs?
I believe it was done this way because:
- This is a global setting in GDB, so the race would be there no matter what
- The poor design of the original AD7 evaluation APIs where display radix isn't passed until evaluate sync
- In VS, there isn't much of a reason not to do it that way since there really is a global setting

I suspect we could get away with using the global setting if we just were careful to only update the global setting in cases where the request actually needs it. For example, a request to obtain stack frames that doesn't want values doesn't need to know the radix.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1274,31 +1282,85 @@ protected override void HandleStackTraceRequestAsync(IRequestResponder<StackTrac
             // Make sure we are stopped and receiving valid input or else return an empty stack trace
```
gregg-miskelly, line 1282: I think this method needs a try/catch to turn exceptions into error responses.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2211,7 +2291,25 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
...
+                if (format.Hex == true)
+                {
+                    radix = 16;
+                }
+            }
+
+            if (m_settingsCallback != null)
+            {
+                // MIEngine generally gets the radix from IDebugSettingsCallback110 rather than using the radix passed
+                m_settingsCallback.Radix = radix;
```
gregg-miskelly, line 2309: This needs to be done earlier. I would probably do it before making any calls into the engine, but definitely before EvaluateSync.

gregg-miskelly (approved):
Otherwise LGTM

gregg-miskelly (approved):

PR #1085, opened 2020-12-02
Bug 1249829: Added externalConsole check for Visual Studio scenarios

gc46:
externalConsole must be true for Visual Studio scenarios. If externalConsole is false, however, this is the dialog customers would have seen before the fix:
image

Here is what they should see after the fix:
image

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -39,6 +39,7 @@ namespace OpenDebugAD7
         private IDebugEngine2 m_engine;
         private EngineConfiguration m_engineConfiguration;
         private AD7Port m_port;
+        private string clientId;
 
         private readonly DebugEventLogger m_logger;
         private readonly Dictionary<string, Dictionary<int, IDebugPendingBreakpoint2>> m_breakpoints;
@@ -682,7 +683,7 @@ namespace OpenDebugAD7
             }
 
             List<ColumnDescriptor> additionalModuleColumns = null;
-            string clientId = responder.Arguments.ClientID;
+            clientId = responder.Arguments.ClientID;
             if (clientId == "visualstudio" || clientId == "liveshare-server-host")
             {
                 additionalModuleColumns = new List<ColumnDescriptor>();
@@ -745,6 +746,12 @@ namespace OpenDebugAD7
 
         protected override void HandleLaunchRequestAsync(IRequestResponder<LaunchArguments> responder)
         {
+            if (clientId == "visualstudio" && !responder.Arguments.ConfigurationProperties.GetValueAsBool("externalConsole").GetValueOrDefault(false))
+            {
+                responder.SetError(new ProtocolException("Integrated terminal is not supported in Visual Studio. To fix, set external console to true."));
+                return;
+            }
+
             const string telemetryEventName = DebuggerTelemetry.TelemetryLaunchEventName;
 
             int hr;
@@ -934,6 +941,12 @@ namespace OpenDebugAD7
 
         protected override void HandleAttachRequestAsync(IRequestResponder<AttachArguments> responder)
         {
+            if (clientId == "visualstudio" && !responder.Arguments.ConfigurationProperties.GetValueAsBool("externalConsole").GetValueOrDefault(false))
+            {
+                responder.SetError(new ProtocolException("Integrated terminal is not supported in Visual Studio. To fix, set external console to true."));
+                return;
+            }
+
             const string telemetryEventName = DebuggerTelemetry.TelemetryAttachEventName;
 
             // ProcessId can be either a string or an int. We attempt to parse as int, if that does not exist we attempt to parse as a string.
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -625,10 +625,10 @@ namespace OpenDebugAD7
 
         private enum ClientId
         {
+            Unknown,
             VisualStudio,
             VsCode,
-            LiveshareServerHost,
-            Unknown
+            LiveshareServerHost
         };
 
         private bool IsClientVS
@@ -668,15 +668,39 @@ namespace OpenDebugAD7
             // Default is that they are URIs
             m_pathConverter.ClientPathsAreURI = !(arguments.PathFormat.GetValueOrDefault(InitializeArguments.PathFormatValue.Unknown) == InitializeArguments.PathFormatValue.Path);
 
+            string clientId = responder.Arguments.ClientID;
+            if (clientId == "visualstudio")
+            {
+                m_clientId = ClientId.VisualStudio;
+            }
+            else if (clientId == "vscode")
+            {
+                m_clientId = ClientId.VsCode;
+            }
+            else if (clientId == "liveshare-server-host")
+            {
+                m_clientId = ClientId.LiveshareServerHost;
+            }
+            else
+            {
+                m_clientId = ClientId.Unknown;
+            }
+
             // If the UI supports RunInTerminal, then register the callback.
             if (arguments.SupportsRunInTerminalRequest.GetValueOrDefault(false))
             {
                 HostRunInTerminal.RegisterRunInTerminalCallback((title, cwd, useExternalConsole, commandArgs, env, success, error) =>
                 {
+                    RunInTerminalArguments.KindValue kind = useExternalConsole ? RunInTerminalArguments.KindValue.External : RunInTerminalArguments.KindValue.Integrated;
+                    if (IsClientVS)
+                    {
+                        kind = RunInTerminalArguments.KindValue.External;
+                    }
+
                     RunInTerminalRequest request = new RunInTerminalRequest()
                     {
                         Arguments = commandArgs.ToList<string>(),
-                        Kind = useExternalConsole ? RunInTerminalArguments.KindValue.External : RunInTerminalArguments.KindValue.Integrated,
+                        Kind = kind,
                         Title = title,
                         Cwd = cwd,
                         Env = env
@@ -699,21 +723,6 @@ namespace OpenDebugAD7
             }
 
             List<ColumnDescriptor> additionalModuleColumns = null;
-            string clientId = responder.Arguments.ClientID;
-            if (clientId == "visualstudio")
-            {
-                m_clientId = ClientId.VisualStudio;
-            } else if (clientId == "vscode")
-            {
-                m_clientId = ClientId.VsCode;
-            } else if (clientId == "liveshare-server-host")
-            {
-                m_clientId = ClientId.LiveshareServerHost;
-            }
-            else
-            {
-                m_clientId = ClientId.Unknown;
-            }
 
             if (IsClientVS)
             {
@@ -777,11 +786,13 @@ namespace OpenDebugAD7
 
         protected override void HandleLaunchRequestAsync(IRequestResponder<LaunchArguments> responder)
         {
+            /*
             if (IsClientVS && !responder.Arguments.ConfigurationProperties.GetValueAsBool("externalConsole").GetValueOrDefault(false))
             {
                 responder.SetError(new ProtocolException("Integrated terminal is not supported in Visual Studio. To fix, set external console to true."));
                 return;
             }
+            */
 
             const string telemetryEventName = DebuggerTelemetry.TelemetryLaunchEventName;
 
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -694,6 +694,8 @@ namespace OpenDebugAD7
             }
 
             // If the UI supports RunInTerminal, then register the callback.
+            // NOTE: Currently we don't support using the RunInTerminal request with VS or Windows Codespaces.
+            //       This is because: (1) they don't support 'Integrated' terminal, and (2) for MIEngine, we don't ship WindowsDebugLauncher.exe.
             if (!IsClientVS && arguments.SupportsRunInTerminalRequest.GetValueOrDefault(false))
             {
                 HostRunInTerminal.RegisterRunInTerminalCallback((title, cwd, useExternalConsole, commandArgs, env, success, error) =>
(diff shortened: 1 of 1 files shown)


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -934,6 +941,12 @@ protected override void HandleLaunchRequestAsync(IRequestResponder<LaunchArgumen
 
         protected override void HandleAttachRequestAsync(IRequestResponder<AttachArguments> responder)
         {
+            if (clientId == "visualstudio" && !responder.Arguments.ConfigurationProperties.GetValueAsBool("externalConsole").GetValueOrDefault(false))
+            {
+                responder.SetError(new ProtocolException("Integrated terminal is not supported in Visual Studio. To fix, set external console to true."));
+                return;
+            }
```
gregg-miskelly, line 948: I don't think attach requests have this property.

gregg-miskelly:
>                         Kind = useExternalConsole ? RunInTerminalArguments.KindValue.External : RunInTerminalArguments.KindValue.Integrated,

Instead of failing, I wonder if we should instead ignore `useExternalConsole` if the client is Visual Studio?

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:663 in 4e30bc0. [](commit_id = 4e30bc0e60c4513f45202bf51746bfe445516bcf, deletion_comment = False)

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -745,6 +746,12 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
 
         protected override void HandleLaunchRequestAsync(IRequestResponder<LaunchArguments> responder)
         {
+            if (clientId == "visualstudio" && !responder.Arguments.ConfigurationProperties.GetValueAsBool("externalConsole").GetValueOrDefault(false))
+            {
+                responder.SetError(new ProtocolException("Integrated terminal is not supported in Visual Studio. To fix, set external console to true."));
```
gregg-miskelly, line 751: >"Integrated terminal is not supported in Visual Studio. To fix, set external console to true." [](start = 57, length = 94)

If we decide we want to fail, this should be localized.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -39,6 +39,7 @@ internal class AD7DebugSession : DebugAdapterBase, IDebugPortNotify2, IDebugEven
         private IDebugEngine2 m_engine;
         private EngineConfiguration m_engineConfiguration;
         private AD7Port m_port;
+        private string clientId;
```
gregg-miskelly, line 42: >clientId [](start = 23, length = 8)

Private fields should begin with `m_` or `_` in at least most of our code bases.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -682,7 +683,7 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
             }
 
             List<ColumnDescriptor> additionalModuleColumns = null;
-            string clientId = responder.Arguments.ClientID;
+            clientId = responder.Arguments.ClientID;
```
gregg-miskelly, line 686: >clientId [](start = 12, length = 8)

Rather than storing 'clientId' you might want to instead have a 'm_isClientVS` which would be set to true in either of these cases. We will might someday have a case where we want to differentiate "visualstudio" and "liveshare-server-host", but so far all of the cases would want to treat both the same (including your new check).

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: Yep. My mistake of searching `externalConsole` and got 4 hits in the cpptools launch.json because of cppvsdbg. 

It is only in launch for cppdbg
https://github.com/microsoft/vscode-cpptools/blob/94294dd3728e034fbe2be6db0437baabcde4ef8b/Extension/package.json#L1452-L1456

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: I think its about time we have m_clientId enum.
```
 enum ClientID
{
   VisualStudio,
   VsCode,
   LiveShareServerHost,
};
```

Then we can have a property `IsClientVS` as
```
bool IsClientVS {
   get {
      return m_clientId == ClientID.VisualStudio || m_clientId == ClientID.LiveShareServerHost;
   }
}
```

src/OpenDebugAD7/AD7DebugSession.cs
gc46: Noted. Thank you for bringing this up.

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: Sounds fine to me

gregg-miskelly: None that you need an 'Unknown' value to for ClientID.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -623,6 +623,22 @@ private void StepInternal(int threadId, enum_STEPKIND stepKind, enum_STEPUNIT st
             }
         }
 
+        private enum ClientId
+        {
```
gregg-miskelly, line 627: I would suggest declaring 'Unknown' as the first value. This will mean that if anyone checks `m_clientId` before the Initialize request, they will get 'Unknown' instead of 'VisualStudio'.

src/OpenDebugAD7/AD7DebugSession.cs
gc46: Correct me if I am wrong, but the main difference between the two options - one to fail and the other to ignore useExternalConsole - is the user awareness of this issue. Perhaps it would be better from the customer's perspective if this error dialog does not appear at all.

What are your thoughts, @gregg-miskelly  and @WardenGnaw?

gregg-miskelly: I sent you an email about this last night, but I would suggest ignoring `useExternalConsole` if the host IDE is Visual Studio. Three main reasons for this:

1. If we were in a world where one is trying to share launch.json content between VS and VS Code, it seems better for VS to ignore the option rather than failing.
2. The fact that the default value of this option is `false` means that we would otherwise require it to be specified, which really seems wrong. We could definitely fix this instead (turn this into a `bool?` throughout the API), but the effort doesn't seem worth it.
3. This is violates layering -- OpenDebugAD7 now has hard coded knowledge of an MIEngine-specific option

gc46: I just saw the email. Thank you for bringing that up. One question - by "if the host IDE is Visual Studio", you are referring to both when ClientId = VisualStudio and ClientId = LiveshareServerHost, correct?

gregg-miskelly: Correct.

gc46: Thank you!

gc46: I have spent quite some time adjusting my approach to ignore useExternalConsole when IsClientVS is true. However, the below error dialog continues to appear when externalConsole is set to false in the test launch.json file:
image

Here is the corresponding Debug Adapter Host Log output:
```
 1> [DebugAdapter] --> C (launch-2): {"type":"request","command":"launch","arguments":{"name":"g++.exe - Build and debug active file","type":"cppdbg","request":"launch","program":"C:\\Users\\gabchen\\Downloads\\projects\\helloworld\\helloworld.exe","args":[],"stopAtEntry":false,"cwd":"C:\\Users\\gabchen\\Downloads\\projects\\helloworld","environment":[],"externalConsole":false,"MIMode":"gdb","miDebuggerPath":"D:\\Program Files (x86)\\mingw-w64\\i686-8.1.0-posix-dwarf-rt_v6-rev0\\mingw32\\bin\\gdb.exe","setupCommands":[{"description":"Enable pretty-printing for gdb","text":"-enable-pretty-printing","ignoreFailures":true}],"preLaunchTask":"C/C++: g++.exe build active file"},"seq":2}
 1> [DebugAdapter] <--   E (output): {"type":"event","event":"output","body":{"category":"stderr","output":"Internal file \"c:\\program files (x86)\\microsoft visual studio\\2019\\main\\common7\\ide\\commonextensions\\microsoft\\mdd\\debugger\\WindowsDebugLauncher.exe\" could not be found.\r\n\r\n"},"seq":2}
 1> [DebugAdapter] <--   R (launch-2) [246 ms]: {"type":"response","request_seq":2,"success":false,"command":"launch","message":"Unable to start debugging. Unable to establish a connection to GDB. The following message was written to stderr:\r\n\r\nInternal file \"c:\\program files (x86)\\microsoft visual studio\\2019\\main\\common7\\ide\\commonextensions\\microsoft\\mdd\\debugger\\WindowsDebugLauncher.exe\" could not be found.\r\n\r\nSee Output Window for details.","body":{"error":{"id":1005,"format":"Unable to start debugging. Unable to establish a connection to GDB. The following message was written to stderr:\r\n\r\nInternal file \"c:\\program files (x86)\\microsoft visual studio\\2019\\main\\common7\\ide\\commonextensions\\microsoft\\mdd\\debugger\\WindowsDebugLauncher.exe\" could not be found.\r\n\r\nSee Output Window for details."}},"seq":3}
 1> ERROR: Unexpected error

AggregateException: One or more errors occurred.

Aggregate exception: 
    DebugAdapterLaunchException: Failed to launch debug adapter.  Additional information may be available in the output window.

    Failure Location: LaunchRequest
    Inner Exception: 
        ProtocolException: Unable to start debugging. Unable to establish a connection to GDB. The following message was written to stderr:
        
        Internal file "c:\program files (x86)\microsoft visual studio\2019\main\common7\ide\commonextensions\microsoft\mdd\debugger\WindowsDebugLauncher.exe" could not be found.
        
        See Output Window for details.
Inner Exception: 
    DebugAdapterLaunchException: Failed to launch debug adapter.  Additional information may be available in the output window.

Microsoft.VisualStudio.Debugger.VSCodeDebuggerHost.Engine.Implementation.DebuggedProcess.<StartDebugAdapter>b__114_3(Exception ex)
Microsoft.VisualStudio.Debugger.VSCodeDebuggerHost.Utilities.TaskExtensions.<>c__DisplayClass11_0`1.<Catch>b__0(TException ex)
Microsoft.VisualStudio.Debugger.VSCodeDebuggerHost.Utilities.TaskExtensions.<>c__DisplayClass10_0`1.<Catch>b__0(AggregateException ex)

    Failure Location: LaunchRequest
    Inner Exception: 
        ProtocolException: Unable to start debugging. Unable to establish a connection to GDB. The following message was written to stderr:
        
        Internal file "c:\program files (x86)\microsoft visual studio\2019\main\common7\ide\commonextensions\microsoft\mdd\debugger\WindowsDebugLauncher.exe" could not be found.
        
        See Output Window for details.

 1> ERROR: One or more errors occurred.

Failed to launch debug adapter.  Additional information may be available in the output window.

Unable to start debugging. Unable to establish a connection to GDB. The following message was written to stderr:

Internal file "c:\program files (x86)\microsoft visual studio\2019\main\common7\ide\commonextensions\microsoft\mdd\debugger\WindowsDebugLauncher.exe" could not be found.

See Output Window for details.
 1> Debug adapter process exited.
```

gc46: Note: the externalConsole = true in launch.json scenario works as expected; the modules show in the Modules window.

Some extra information that may be helpful:
- clientId = "visualstudio"
- m_clientId = VisualStudio
- IsClientVS = true

gregg-miskelly: This is because we don't install WindowsDebugLauncher.exe with VS.

@WardenGnaw Is the scenario in this bug (launching a mingw program) actually a Codespaces scenario? Or is this really just a test scenario?

If yes - the right way to fix this is to add WindowsDebugLauncher.exe to setup.
If no - the right way to fix this is to stop calling `RegisterRunInTerminalCallback` so that (IsRunInTerminalAvailable)(https://github.com/gregg-miskelly/MIEngine/blob/c9dd86298463a60df043148e3cab49f4ad08994b/src/DebugEngineHost.VSCode/HostRunInTerminal.cs#L19-L22) will return false.

gregg-miskelly: BTW: If we do need to support mingw with Codespaces, then [this code](https://github.com/gregg-miskelly/MIEngine/blob/c9dd86298463a60df043148e3cab49f4ad08994b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs#L766-L772) is wrong if `UseExternalConsole` can ever return `true` -- all Codespaces launches of mingw apps need to use RunInTerminal. So we might need a new `HostRunInTerminal` method that we could use to tell MIEngine not to use `-gdb-set new-console on`.

WardenGnaw: I'm actually not sure. @andrewcrawley Added OpenDebugAD7 to Codespaces initially, was there not a need for adding this binary?

We ran into this issue because CTI created this `launch.json` with `externalConsole: false` and I'm assuming they are also doing the local launch scenario with DebugAdapterHost.Launch instead of testing in a CodeSpace.

gc46: Should be good to go now. 

Here is what the Modules window looks like when externalConsole is false:
image

And here is what the Modules window looks like when externalConsole is false:
image

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -658,8 +675,26 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
...
+            else if (clientId == "liveshare-server-host")
+            {
+                m_clientId = ClientId.LiveshareServerHost;
+            }
+            else
+            {
+                m_clientId = ClientId.Unknown;
+            }
+
             // If the UI supports RunInTerminal, then register the callback.
```
gregg-miskelly, line 696: >callback [](start = 67, length = 8)

Can you add more info to this comment. Something like:
```
// NOTE: Currently we don't support using the RunInTerminal request with VS or Windows Codespaces. This is because: (1) they don't support 'Integrated' terminal, and (2) for MIEngine, we don't ship WindowsDebugLauncher.exe.
```

gregg-miskelly (approved):
Aside from that nit, LGTM.

src/OpenDebugAD7/AD7DebugSession.cs
gc46: Done! Thank you so much!

PR #1098, opened 2021-01-22
Ensure filename is formatted for the host OS.

paulmaybee:
In cross complie scenarios I have seen filename symbols in the .out file partially formatted in unix style and partially in Windows:

    <11>   DW_AT_name        : (indirect string, offset: 0x1709): ../../../hello_world.cpp
    <15>   DW_AT_comp_dir    : (indirect string, offset: 0x1922): C:\temp\raspi-toolchain\hello_world\out\build\x64-Debug

Which results in gdb returning file names like this:

    "C:\\temp\\raspi-toolchain\\hello_world\\out\\build\\x64-Debug/../../../hello_world.cpp"

VS doesn't understand names in this format so cannot find the source; replace the '/'s with '\\'.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/MITextPosition.cs
+++ b/src/MIDebugEngine/Engine.Impl/MITextPosition.cs
@@ -39,6 +39,8 @@ namespace Microsoft.MIDebugEngine
             if (string.IsNullOrWhiteSpace(filename))
                 return null;
 
+            filename = PlatformUtilities.PathToHostOSPath(filename);
+
             uint? line = miTuple.TryFindUint("line");
             if (!line.HasValue || line.Value == 0)
                 return null;


WardenGnaw (approved):
Ah, sorry I deleted that line in this [commit](https://github.com/microsoft/MIEngine/commit/1ccfd123289100f2415450d68b6011766941cc3d).
image since I thought paths changes would only occur in sourceFileMapping and didn't know we could get a mixed source path with `/` and `\`.

gregg-miskelly (approved):

paulmaybee:
@WardenGnaw Any chance I can get this into preview 4?

WardenGnaw:
Yep. I'll add you to the insertion PR.

PR #1102, opened 2021-01-29
Support GoToTargets Request 

WardenGnaw:
Thank you @Trass3r for the initial work.

This PR will add in the GoToTargets request which can be used by the client to get `instructionPointerReference` from a file and line.

https://github.com/microsoft/MIEngine/pull/1042 will focus on implement GoToRequest (Jump to Line)

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7Engine.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Engine.cs
@@ -182,44 +182,6 @@ namespace Microsoft.MIDebugEngine
             return _configStore.GetEngineMetric(metric);
         }
 
-        public int Jump(string filename, int line)
-        {
-            try
-            {
-                _debuggedProcess.WorkerThread.RunOperation(() => _debuggedProcess.Jump(filename, line));
-            }
-            catch (InvalidCoreDumpOperationException)
-            {
-                return AD7_HRESULT.E_CRASHDUMP_UNSUPPORTED;
-            }
-            catch (Exception e)
-            {
-                _engineCallback.OnError(EngineUtils.GetExceptionDescription(e));
-                return Constants.E_ABORT;
-            }
-
-            return Constants.S_OK;
-        }
-
-        public int Jump(ulong address)
-        {
-            try
-            {
-                _debuggedProcess.WorkerThread.RunOperation(() => _debuggedProcess.Jump(address));
-            }
-            catch (InvalidCoreDumpOperationException)
-            {
-                return AD7_HRESULT.E_CRASHDUMP_UNSUPPORTED;
-            }
-            catch (Exception e)
-            {
-                _engineCallback.OnError(EngineUtils.GetExceptionDescription(e));
-                return Constants.E_ABORT;
-            }
-
-            return Constants.S_OK;
-        }
-
         #region IDebugEngine2 Members
 
         // Attach the debug engine to a program.
@@ -877,7 +839,7 @@ namespace Microsoft.MIDebugEngine
                 {
                     var codeCxt = new AD7MemoryAddress(this, a, null);
                     TEXT_POSITION pos;
-                    pos.dwLine = line;
+                    pos.dwLine = line - 1;
                     pos.dwColumn = 0;
                     MITextPosition textPosition = new MITextPosition(documentName, pos, pos);
                     codeCxt.SetDocumentContext(new AD7DocumentContext(textPosition, codeCxt));
--- a/src/MIDebugEngine/AD7.Impl/AD7Engine.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Engine.cs
@@ -182,6 +182,44 @@ namespace Microsoft.MIDebugEngine
             return _configStore.GetEngineMetric(metric);
         }
 
+        public int Jump(string filename, int line)
+        {
+            try
+            {
+                _debuggedProcess.WorkerThread.RunOperation(() => _debuggedProcess.Jump(filename, line));
+            }
+            catch (InvalidCoreDumpOperationException)
+            {
+                return AD7_HRESULT.E_CRASHDUMP_UNSUPPORTED;
+            }
+            catch (Exception e)
+            {
+                _engineCallback.OnError(EngineUtils.GetExceptionDescription(e));
+                return Constants.E_ABORT;
+            }
+
+            return Constants.S_OK;
+        }
+
+        public int Jump(ulong address)
+        {
+            try
+            {
+                _debuggedProcess.WorkerThread.RunOperation(() => _debuggedProcess.Jump(address));
+            }
+            catch (InvalidCoreDumpOperationException)
+            {
+                return AD7_HRESULT.E_CRASHDUMP_UNSUPPORTED;
+            }
+            catch (Exception e)
+            {
+                _engineCallback.OnError(EngineUtils.GetExceptionDescription(e));
+                return Constants.E_ABORT;
+            }
+
+            return Constants.S_OK;
+        }
+
         #region IDebugEngine2 Members
 
         // Attach the debug engine to a program.
--- a/src/MIDebugEngine/AD7.Impl/AD7Thread.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Thread.cs
@@ -282,23 +282,27 @@ namespace Microsoft.MIDebugEngine
         }
 
         // Sets the next statement to the given stack frame and code context.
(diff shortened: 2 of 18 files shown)


src/MIDebugEngine/AD7.Impl/AD7Engine.cs
```diff
@@ -877,7 +839,7 @@ public int EnumCodeContexts(IDebugDocumentPosition2 docPosition, out IEnumDebugC
                 {
                     var codeCxt = new AD7MemoryAddress(this, a, null);
                     TEXT_POSITION pos;
-                    pos.dwLine = line;
+                    pos.dwLine = line - 1;
```
Trass3r, line 842: Does this need to go through line conversion functions?

src/MIDebugEngine/AD7.Impl/AD7Thread.cs
```diff
@@ -282,23 +282,27 @@ int IDebugThread2.Resume(out uint suspendCount)
...
+                return Constants.S_FALSE;
             }
-            catch (Exception)
+            string toFunc = EngineUtils.GetAddressDescription(_engine.DebuggedProcess, addr);
+            string fromFunc = EngineUtils.GetAddressDescription(_engine.DebuggedProcess, frame.ThreadContext.pc.Value);
+            if (toFunc != fromFunc)
             {
                 return Constants.S_FALSE;
             }
+            string result = frame.EvaluateExpression("$pc=" + EngineUtils.AsAddr(addr, _engine.DebuggedProcess.Is64BitArch));
```
Trass3r: This function is now only used by VS, right?
It can handle this expression?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -350,6 +355,7 @@ public void BeforeContinue()
             m_isStopped = false;
             m_variableManager.Reset();
             m_frameHandles.Reset();
+			m_gotoCodeContexts.Clear();
```
gregg-miskelly, line 358: >			 [](start = 0, length = 3)

nit: tabs

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1317,6 +1324,84 @@ protected override void HandlePauseRequestAsync(IRequestResponder<PauseArguments
             responder.SetResponse(new PauseResponse());
         }
 
+        protected override void HandleGotoRequestAsync(IRequestResponder<GotoArguments> responder)
+        {
+            responder.SetError(new ProtocolException("Not implemented exception."));
```
gregg-miskelly, line 1329: >"Not implemented exception. [](start = 53, length = 27)

I might suggest instead: "Set next statement is not supported by the current debugger." since this error message might actually be show to the user.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1317,6 +1324,84 @@ protected override void HandlePauseRequestAsync(IRequestResponder<PauseArguments
...
+        }
+
+        protected override void HandleGotoTargetsRequestAsync(IRequestResponder<GotoTargetsArguments, GotoTargetsResponse> responder)
+        {
+            var response = new GotoTargetsResponse();
+
+            var source = responder.Arguments.Source;
+            // TODO: handle this for disassembly debugging
```
gregg-miskelly, line 1337: >// TODO: handle this for disassembly debugging [](start = 12, length = 46)

Remove?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1317,6 +1324,84 @@ protected override void HandlePauseRequestAsync(IRequestResponder<PauseArguments
...
+            if (source.Path == null)
```
gregg-miskelly, line 1338: >source.Path [](start = 16, length = 11)

I am not sure why this would ever be null, but if we are going to check it, seems like we should do: `string.IsNullOrEmpty`?

Trass3r: It's null for virtual documents, which I used for implementing the disassembly debugging PoC.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1317,6 +1324,84 @@ protected override void HandlePauseRequestAsync(IRequestResponder<PauseArguments
...
+                        {
+                            var startPos = new TEXT_POSITION[1];
+                            var endPos = new TEXT_POSITION[1];
+                            if (documentContext.GetStatementRange(startPos, endPos) == HRConstants.S_OK)
+                                line = m_pathConverter.ConvertDebuggerLineToClient((int)startPos[0].dwLine);
+                        }
+
+                        string instructionPointerReference = null;
+                        CONTEXT_INFO[] contextInfo = new CONTEXT_INFO[1];
+                        if (codeContext.GetInfo(enum_CONTEXT_INFO_FIELDS.CIF_ADDRESS, contextInfo) == HRConstants.S_OK)
```
gregg-miskelly, line 1376: >) [](start = 118, length = 1)

nit: `&& contextInfo[0].dwFields.HasFlag(enum_CONTEXT_INFO_FIELDS.CIF_ADDRESS))`

gregg-miskelly:
>                         response.StackFrames.Add(new ProtocolMessages.StackFrame()

nit: While you are here, do you want to add the instruction reference here as well. That should finish off the feature.

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:1565 in f37fe48. [](commit_id = f37fe48f6417c1525b29e30b8c4f238616bee8e6, deletion_comment = False)

src/MIDebugEngine/AD7.Impl/AD7Engine.cs
gregg-miskelly: No. AD7-based line numbers are always zero based. GDB line numbers are always one-based. `line` is to send to GDB, so it was a 1-based number. But MITextPosition uses AD7-based text positions, so there should have been a subtraction here.

---
In reply to: [566701203](https://github.com/microsoft/MIEngine/pull/1102#discussion_r566701203) [](ancestors = 566701203)

gregg-miskelly (approved):
Otherwise LGTM

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: Ah.

gregg-miskelly: Probably worth adding a comment `// Virtual documents don't have paths`.

src/MIDebugEngine/AD7.Impl/AD7Thread.cs
gregg-miskelly: > This function is now only used by VS, right?

Correct. Currently this code isn't reachable in DAP scenarios.

> It can handle this expression?

This is just an internal call within MIEngine, so it will not matter what IDE is used. It worked with whatever scenario Set Next Statement was originally tested (Android?). I don't know if it has been tested more broadly. Certainly using the actual GDB command is better. But this PR just seeks to finish off disassembly support. Actually implementing set next statement is still left for your PR. Someone needs to finish off handling the work of handling that stopping event...

Trass3r: Ah yeah my bad, of course this actually goes down to gdb and not up. Setting `$pc` is actually different from the jump command in that it `does not start your program running; it only changes the address of where it will run when you continue`: https://sourceware.org/gdb/current/onlinedocs/gdb/Jumping.html
So I'm not sure how this worked out regarding UI/UX.

gregg-miskelly: Yup, we would need to hide this difference -- the new set next statement would need to save some state in the process object to indicate that it is waiting for that stop event, and then SetNextStatement will need to wait for the stop event to be raised (or for the target process to exit).

WardenGnaw:
> > ```
> >                     response.StackFrames.Add(new ProtocolMessages.StackFrame()
> > ```
> 
> nit: While you are here, do you want to add the instruction reference here as well. That should finish off the feature.
> 
> Refers to: src/OpenDebugAD7/AD7DebugSession.cs:1565 in [f37fe48](https://github.com/microsoft/MIEngine/commit/f37fe48f6417c1525b29e30b8c4f238616bee8e6). [](commit_id = [f37fe48](https://github.com/microsoft/MIEngine/commit/f37fe48f6417c1525b29e30b8c4f238616bee8e6), deletion_comment = False)

This will be handled in another PR.

gregg-miskelly (approved):

PR #1107, opened 2021-02-23
Task 1274960: Add support for IDebugProperty160.GetDataBreakpointInfo160 in MIEngine

gc46:
Before:
image

After:
image
image

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7Property.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Property.cs
@@ -77,6 +77,9 @@ namespace Microsoft.MIDebugEngine
                 if (variable.Error)
                 {
                     propertyInfo.dwAttrib |= enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_VALUE_ERROR;
+                } else
+                {
+                    propertyInfo.dwAttrib |= enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_DATA;
                 }
 
                 if (variable.IsStringType)
@@ -85,8 +88,6 @@ namespace Microsoft.MIDebugEngine
                 }
 
                 propertyInfo.dwAttrib |= variable.Access;
-
-                propertyInfo.dwAttrib |= enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_DATA;
             }
 
             // If the debugger has asked for the property, or the property has children (meaning it is a pointer in the sample)
--- a/src/MIDebugEngine/AD7.Impl/AD7Property.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Property.cs
@@ -14,7 +14,7 @@ namespace Microsoft.MIDebugEngine
     // The property is usually the result of an expression evaluation. 
     //
     // The sample engine only supports locals and parameters for functions that have symbols loaded.
-    internal class AD7Property : IDebugProperty3
+    internal class AD7Property : IDebugProperty3, IDebugProperty160
     {
         private static uint s_maxChars = 1000000;
         private byte[] _bytes;
@@ -413,6 +413,26 @@ namespace Microsoft.MIDebugEngine
                 }
             }
             return Constants.E_FAIL;
+        }
+
+        /*
+         * test -- need to delete; insert something here
+         * See https://devdiv.visualstudio.com/DevDiv/_git/Concord?path=%2Fsrc%2Fimpl%2Fad7%2FProperty.cpp, Line 1413
+         * Maybe this as well: https://devdiv.visualstudio.com/DevDiv/_git/VSCodeDebugAdapterHost?path=%2Fsrc%2Fproduct%2FVsCodeDebuggerHost%2FAD7%2FImplementation%2FAD7Property.cs&_a=contents&version=GBmain, Line 226?
+         */
+        public int GetDataBreakpointInfo160(out string pbstrAddress, out uint pSize, out string pbstrDisplayName, out string pbstrError)
+        {
+            int hr = Constants.S_OK;
+
+            pbstrAddress = "";
+            pSize = 0;
+            pbstrDisplayName = "";
+            pbstrError = "";
+
+            // to be continued...
+            // trying to access _variableInformation.Address() and Size() here...
+
+            return hr;
         }
     }
 
--- a/src/MIDebugEngine/AD7.Impl/AD7Property.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Property.cs
@@ -86,12 +86,7 @@ namespace Microsoft.MIDebugEngine
 
                 propertyInfo.dwAttrib |= variable.Access;
 
-                /*
-                if (!string.IsNullOrEmpty(variable.Address()) && variable.Size() > 0)
-                {
-                    propertyInfo.dwAttrib |= enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_DATA;
-                }
-                */
+                propertyInfo.dwAttrib |= enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_DATA;
             }
 
             // If the debugger has asked for the property, or the property has children (meaning it is a pointer in the sample)
@@ -421,17 +416,10 @@ namespace Microsoft.MIDebugEngine
             return Constants.E_FAIL;
         }
 
-        /*
-         * test -- need to delete; insert something here
-         * See https://devdiv.visualstudio.com/DevDiv/_git/Concord?path=%2Fsrc%2Fimpl%2Fad7%2FProperty.cpp, Line 1413
-         * Maybe this as well: https://devdiv.visualstudio.com/DevDiv/_git/VSCodeDebugAdapterHost?path=%2Fsrc%2Fproduct%2FVsCodeDebuggerHost%2FAD7%2FImplementation%2FAD7Property.cs&_a=contents&version=GBmain, Line 226?
-         */
         public int GetDataBreakpointInfo160(out string pbstrAddress, out uint pSize, out string pbstrDisplayName, out string pbstrError)
         {
             int hr = Constants.S_OK;
 
-            // to be continued...
-            // trying to access _variableInformation.Address() and Size() here...
             pbstrAddress = _variableInformation.Address();
             pSize = _variableInformation.Size();
             pbstrDisplayName = _variableInformation.Name; // or do we want FullName here?
--- a/src/MIDebugEngine/AD7.Impl/AD7Property.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Property.cs
@@ -2,8 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Collections.Generic;
-using System.Text;
 using Microsoft.VisualStudio.Debugger.Interop;
 using System.Diagnostics;
 
@@ -424,13 +422,12 @@ namespace Microsoft.MIDebugEngine
         {
             int hr = Constants.S_OK;
 
-            pbstrAddress = "";
-            pSize = 0;
-            pbstrDisplayName = "";
-            pbstrError = "";
-
             // to be continued...
             // trying to access _variableInformation.Address() and Size() here...
+            pbstrAddress = _variableInformation.Address();
+            pSize = _variableInformation.Size();
+            pbstrDisplayName = _variableInformation.Name; // or do we want FullName here?
+            pbstrError = "";
 
             return hr;
         }
--- a/src/MIDebugEngine/AD7.Impl/AD7Property.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Property.cs
@@ -436,12 +436,20 @@ namespace Microsoft.MIDebugEngine
                 pSize = _variableInformation.Size();
                 pbstrDisplayName = _variableInformation.Name;
                 pbstrError = "";
-                // test -- need to delete; add variable to DataBreakpointVariables
+                // test -- need to delete; add or remove variable in DataBreakpointVariables
                 lock (_engine.DebuggedProcess.DataBreakpointVariables)
                 {
-                    _engine.DebuggedProcess.DataBreakpointVariables.Add(_variableInformation.FullName());
+                    if (_engine.DebuggedProcess.DataBreakpointVariables.Contains(_variableInformation.FullName()))
+                    {
+                        _engine.DebuggedProcess.DataBreakpointVariables.Remove(_variableInformation.FullName());
+                        // _engine.DebuggedProcess.VariableNameAddressMap.Remove(_variableInformation.FullName() + "," + pbstrAddress);
+                    }
+                    else
+                    {
+                        _engine.DebuggedProcess.DataBreakpointVariables.Add(_variableInformation.FullName());
+                        _engine.DebuggedProcess.VariableNameAddressMap.Add(_variableInformation.FullName() + "," + pbstrAddress);
+                    }
                 }
-                _engine.DebuggedProcess.VariableNameAddressMap.Add(_variableInformation.FullName() + "," + pbstrAddress);
                 return Constants.S_OK;
             }
             catch (Exception e)
--- a/src/MIDebugEngine/AD7.Impl/AD7Property.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Property.cs
@@ -82,9 +82,10 @@ namespace Microsoft.MIDebugEngine
                 } else
                 {
                     propertyInfo.dwAttrib |= enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_DATA;
-                    if (_engine.DebuggedProcess.DataBreakpointVariables.Contains(variable.FullName()))
+                    string fullName = variable.FullName();
+                    if (_engine.DebuggedProcess.DataBreakpointVariables.Contains(fullName))
                     {
-                        if (_engine.DebuggedProcess.VariableNameAddressMap.Contains(variable.FullName() + "," + variable.Address()))
+                        if (_engine.DebuggedProcess.VariableNameAddressMap.Contains(fullName + "," + variable.Address()))
                         {
                             propertyInfo.dwAttrib |= (enum_DBG_ATTRIB_FLAGS)DBG_ATTRIB_HAS_DATA_BREAKPOINT;
                         }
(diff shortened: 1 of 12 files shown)


gc46:
Some comments:
- A check when setting `propertyInfo.dwAttrib |= enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_DATA;` may be needed
- `GetDataBreakpointInfo160()` needs to be cleaned up as error cases are unaccounted for

src/MIDebugEngine/MIDebugEngine.csproj
```diff
@@ -70,6 +70,7 @@
   <ItemGroup Label="NuGet Packages">
     <PackageReference Include="Microsoft.VisualStudio.Debugger.Interop.Portable" Version="$(Microsoft_VisualStudio_Debugger_Interop_Portable_Version)" />
     <PackageReference Include="Microsoft.VisualStudio.Debugger.Interop.15.0" Version="$(Microsoft_VisualStudio_Debugger_Interop_15_0_Version)" />
+    <PackageReference Include="Microsoft.VisualStudio.Debugger.Interop.16.0" Version="$(Microsoft_VisualStudio_Debugger_Interop_16_0_Version)" />
```
WardenGnaw, line 73: You will want to add this reference in [MIDebugPackage.csproj](https://github.com/microsoft/MIEngine/blob/74f810009496b87a0f6a7d592d691a303e099656/src/MIDebugPackage/MIDebugPackage.csproj#L78) and [OpenDebugAD7.csproj](https://github.com/microsoft/MIEngine/blob/74f810009496b87a0f6a7d592d691a303e099656/src/OpenDebugAD7/OpenDebugAD7.csproj#L110).

So it is packaged for MDD for VS and XPlat for VS Code.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -90,6 +90,23 @@ internal sealed class VariableInformation : IVariableInformation
         private string DisplayHint { get; set; }
         public bool IsPreformatted { get; set; }
 
+        public string Address()
+        {
+            // ask GDB to evaluate "&expression"
+            string command = "&("+_strippedName+")";
```
gregg-miskelly, line 96: >_strippedName [](start = 34, length = 13)

You need the `FullName()` instead of `_strippedName`

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -90,6 +90,23 @@ internal sealed class VariableInformation : IVariableInformation
...
+            IVariableInformation execVariable = new VariableInformation(command, this);
+            execVariable.SyncEval();
+            return _engine.DebuggedProcess.Natvis.FormatDisplayString(execVariable);
```
gregg-miskelly, line 99: I think you want to use MI Commands directly instead of going through the VariableInformation classes. The variable information classes do extra work, like applying natvis, setting the output radix that you don't want, as that does extra work, and also may reformat the pointer in a way that might break things.

Looking at [MI documentation](https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI-Data-Manipulation.html#GDB_002fMI-Data-Manipulation) I think you can use the `-data-evaluate-expression` command instead of the VarCreate stuff to avoid creating variables for no reason.

gregg-miskelly, line 99: >return [](start = 12, length = 6)

You probably want to verify that the result you get back looks like a pointer.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -90,6 +90,23 @@ internal sealed class VariableInformation : IVariableInformation
...
+        }
+
+        public uint Size()
```
gregg-miskelly, line 102: >public [](start = 8, length = 6)

Same notes as with 'Address'.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -90,6 +90,23 @@ internal sealed class VariableInformation : IVariableInformation
...
+        {
+            // ask GDB to evaluate "sizeof(expression)"
+            string command = "sizeof("+_strippedName+")";
+            IVariableInformation execVariable = new VariableInformation(command, this);
+            execVariable.SyncEval();
+            return Convert.ToUInt32(_engine.DebuggedProcess.Natvis.FormatDisplayString(execVariable), CultureInfo.InvariantCulture);
```
gregg-miskelly, line 108: >ToUInt32 [](start = 27, length = 8)

Does this produce the right thing if GDB is in hex mode?

src/MIDebugEngine/AD7.Impl/AD7Property.cs
```diff
@@ -414,6 +415,18 @@ public int SetValueAsStringWithError(string pszValue, uint dwRadix, uint dwTimeo
...
         }
+
+        public int GetDataBreakpointInfo160(out string pbstrAddress, out uint pSize, out string pbstrDisplayName, out string pbstrError)
+        {
+            int hr = Constants.S_OK;
+
+            pbstrAddress = _variableInformation.Address();
+            pSize = _variableInformation.Size();
+            pbstrDisplayName = _variableInformation.Name; // or do we want FullName here?
+            pbstrError = "";
```
gregg-miskelly, line 426: try/catch and fill out the error string?

gregg-miskelly (changes requested):
See comments for list of requested changes.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -61,6 +62,15 @@ public void Dispose()
         {
         }
         public bool IsPreformatted { get { return Parent.IsPreformatted; } set { } }
+
+        public string Address()
+        {
+            throw new NotImplementedException();
```
gregg-miskelly, line 68: >throw new NotImplementedException(); [](start = 12, length = 36)

I think you instead want to call the Parent for these two.

src/MIDebugEngine/Engine.Impl/Variables.cs
gregg-miskelly: A test case to try BTW: stick a pointer/reference in the watch window, expand it, use the feature on one of the child fields.

---
In reply to: [581388881](https://github.com/microsoft/MIEngine/pull/1107#discussion_r581388881) [](ancestors = 581388881)

gc46: Break When Value Changes works when setting a data breakpoint from the Watch window
image

WardenGnaw:
I just noticed, we are missing the red circle on the variable when we set a data bp.
image

It should look like
image

You will need to keep track if the variable has been called on for a data bp with
```
dwAttrib |= DBG_ATTRIB_HAS_DATA_BREAKPOINT;
```

You may want to keep track of the `FullName+Address` that have been set before and check to see if any new VariableInformation matches that name, 

If the `FullName` matches, do an `Address()` call on it to see if it matches the one you saved. If it matches, then mark it having a data breakpoint. 

Any other case, dont add `DBG_ATTRIB_HAS_DATA_BREAKPOINT`

src/MIDebugEngine/Engine.Impl/Variables.cs
gc46: Changed to use EvalDependentExpression()

gc46: Changed to use EvalDependentExpression() as mentioned above. `&(*test)` returns `"0x7ffcb35f89cc"` for example.

src/MIDebugEngine/Engine.Impl/Variables.cs
gc46: I am a bit confused as to what is meant by hex mode. Is it simply having Hexadecimal Display checked as is shown below?
image

Or do you mean running said command with the format letter x as stated here: https://ftp.gnu.org/old-gnu/Manuals/gdb/html_node/gdb_54.html? Note that I am now using the MI command directly.

gregg-miskelly: Yes, it should be triggered by checking that checkbox and doing an evaluation. That will cause a call to [SetRadix](https://github.com/Microsoft/MIEngine/blob/0e48a887feb19cf525025fa8bd5788464b667065/src/MICore/CommandFactories/MICommandFactory.cs#L300-L300) with 16, which then changes the default output radix.

src/MIDebugEngine/AD7.Impl/AD7Property.cs
gc46: For the catch clause, are the following assignments okay?
- pbstrAddress = string.Empty
- pSize = 0
- pbstrDisplayName = string.Empty
- pbstrError = e.Message where e is a generic Exception

src/MIDebugEngine/Natvis.Impl/Natvis.cs
gc46: `public string Address() { return Parent.Address(); }` and `public string Size() { return Parent.Size(); }` break Break When Value Changes. I can look into why this occurs.

src/MIDebugEngine/Engine.Impl/Variables.cs
gc46: Edit: Unresolving since Break When Value Changes is breaking in the Watch window...

gc46: Update: Changing from _strippedName to FullName() breaks Break When Value Changes in the Watch window. Would it be okay to leave this as is then, @gregg-miskelly?

gregg-miskelly: Do you mean root items in the watch window? If so, what are you seeing for `FullName()` and `_strippedName`?

src/MIDebugEngine/AD7.Impl/AD7Property.cs
gregg-miskelly: You might want to use null instead of string.Empty. But otherwise looks correct.

src/MIDebugEngine/Engine.Impl/Variables.cs
gc46: `FullName(): "((int *)0x7fff93033e0c)->*&testint"`
`_strippedName: "*&testint"`

gregg-miskelly: So that is a bug in our full name generation in this case, though I am not sure how easy it would be to fix. We can't just use _strippedName because that, in the general case, isn't an expression.

Ex:
struct S
{
   int myField;
};
S s = {};

Stick 's' in the watch window and expand it, try to set a data breakpoint on 'myField'. `_strippedName` will be `myField`.

gregg-miskelly: I *think* what we will need to do is special case children than start with `*` here: https://github.com/microsoft/MIEngine/blob/0e48a887feb19cf525025fa8bd5788464b667065/src/MIDebugEngine/Engine.Impl/Variables.cs#L272. We probably need a new NodeType of 'Dereference'.

I would test something like:

```
struct S2
{
     int *p;
};
struct S1
{
    S2 s2;
};
S1 s1 = {};
int myVar = 42;
s1.s2.p = &myVar;
```

Now put s1 in the watch window and expand all the way to '*p'. Also do the same but starting with 's1.s2' in the watch window. Assuming we wind up with a stripped name of '*p' in both cases then I think the way to fix this is to change the full name to be `*(<parent_name_here)`.

gc46: Okay. Looking into this. Thanks.

gc46: Update: I have verified the following:
- `_strippedName = "myField"` when sticking 's' in the Watch window, expanding it, and setting a data breakpoint on 'myField'
- `_strippedName = "*p"` when sticking both 's1' and 's1.s2' in the Watch window, expanding them, and setting data breakpoints on '*p'

gc46: Hi @gregg-miskelly, I found this comment in the NodeType.Field case in FullName():
https://github.com/microsoft/MIEngine/blob/0e48a887feb19cf525025fa8bd5788464b667065/src/MIDebugEngine/Engine.Impl/Variables.cs#L120-L121

From what I understand, this may be the bug you are referring to as we need an expression for the commands in Address() and Size(). That said, I have a few questions:
1. Is it correct to say that this is why I was encountering the `Unable to set data breakpoint. One or more errors occurred.` error when switching from _strippedName to FullName()?
2. Should I be modeling NodeType.Dereference after NodeType.Field? It seems the new NodeType should be similar but not exactly the same since NodeType.Field has a conditional in `FullName()` that cuts off the expression and replaces it with an address instead.

Thanks!

gregg-miskelly: Its related, but not really the problem -- without that code, in the example I have above, you would see a full name for '*p' of `s1.s2.p.*p` instead of `*(s1.s2.p)`. The full name stuff is really a bit of a heuristic to guess the names that GDB gives us and try to turn that into what the full expression is. But that code doesn't handle deferencing a pointer.

src/MIDebugEngine/AD7.Impl/AD7Property.cs
```diff
@@ -85,7 +83,10 @@ public DEBUG_PROPERTY_INFO ConstructDebugPropertyInfo(enum_DEBUGPROP_INFO_FLAGS
                 {
                     propertyInfo.dwAttrib |= enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_VALUE_RAW_STRING;
                 }
+
                 propertyInfo.dwAttrib |= variable.Access;
+
+                propertyInfo.dwAttrib |= enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_DATA;
```
gregg-miskelly, line 89: >propertyInfo.dwAttrib |= enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_DATA; [](start = 16, length = 63)

This should probably be under `if (!variable.Error)`

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -90,6 +90,19 @@ internal sealed class VariableInformation : IVariableInformation
         private string DisplayHint { get; set; }
         public bool IsPreformatted { get; set; }
 
+        public string Address()
+        {
+            // ask GDB to evaluate "&expression"
+            string command = "&("+FullName()+")";
+            return EvalDependentExpression(command);
```
gregg-miskelly, line 97: >return [](start = 12, length = 6)

We should probably verify the result we gets back looks like an address.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -269,6 +285,10 @@ private VariableInformation(TupleValue results, VariableInformation parent, stri
             {
                 VariableNodeType = NodeType.AnonymousUnion;
             }
+            else if (Name[0] == '*')
```
gregg-miskelly, line 288: >Name [](start = 21, length = 4)

I think you need `Name.Length >= 1 &&`

src/MIDebugEngine/Engine.Impl/Variables.cs
gc46: You mean have a conditional that checks for "0x" at the beginning? If so, ohh. I think I misunderstood it before and physically checked for it.

src/MIDebugEngine/Engine.Impl/Variables.cs
gc46: Will be using `Name.Length **>** 1 &&` instead as the first character in `Name `must be an asterisk. Let me know if there are any cases where Name can just be "*" though. Thanks for pointing this out!

gregg-miskelly: I don't know if GDB would ever return a name of just '*'. I guess your way is better.

src/MIDebugEngine/Engine.Impl/Variables.cs
gregg-miskelly: Yeah, I would think we would want to verify the result is either:
* `^0x[0-9a-fA-F]+$`
-or-
* `^(0x[0-9a-fA-F]+)\b` and return the captured string

src/OpenDebugAD7/OpenDebugAD7.csproj
```diff
@@ -108,6 +108,7 @@
     <DropUnsignedFile Include="$(OutputPath)\Microsoft.VisualStudio.Debugger.Interop.11.0.dll" />
     <DropUnsignedFile Include="$(OutputPath)\Microsoft.VisualStudio.Debugger.Interop.12.0.dll" />
     <DropUnsignedFile Include="$(OutputPath)\Microsoft.VisualStudio.Debugger.Interop.15.0.dll" />
+    <DropUnsignedFile Include="$(OutputPath)\Microsoft.VisualStudio.Debugger.Interop.16.0.dll" />
```
WardenGnaw, line 111: You also need to add it as a package reference so its in the output path for OpenDebugAD7
```
<PackageReference Include="Microsoft.VisualStudio.Debugger.Interop.16.0" Version="$(Microsoft_VisualStudio_Debugger_Interop_16_0_Version)" />
```

WardenGnaw:
You will want to define `Microsoft_VisualStudio_Debugger_Interop_16_0_Version` at https://github.com/microsoft/MIEngine/blob/74f810009496b87a0f6a7d592d691a303e099656/build/package_versions.settings.targets#L4

gregg-miskelly:
Your build is failing because you need to edit this: https://github.com/Microsoft/MIEngine/blob/0e48a887feb19cf525025fa8bd5788464b667065/src/MakePIAPortable/MakePortableLegacyPIAs.cmd#L30-L30

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -90,6 +90,32 @@ internal sealed class VariableInformation : IVariableInformation
...
+            string command = "&("+FullName()+")";
+            var result = EvalDependentExpression(command);
+            for (int i = 0; i < s_validAddressFormats.Length; i++)
+            {
+                if (Regex.IsMatch(result, s_validAddressFormats[i]))
+                {
+                    return result;
+                }
+            }
+            return null;
```
gregg-miskelly, line 110: I believe you want to throw an exception here. Something with a message like: $"Unexpected result '${result}' when evaluating '${command}".

gregg-miskelly (approved):
Otherwise LGTM.

gc46:
@gregg-miskelly, I found inconsistent behavior between the EE and Breakpoints windows when adding and removing data breakpoints while testing my code changes. The testing scenarios and their respective behaviors are summarized below:
image

Questions:
1. I looked through the VS repository and it seems `GetDataBreakpointInfo160()` is called for the EE window but not the Breakpoints window. Is this correct?
2. Why might GDB return 0x0000000000000000 when setting a data breakpoint at *test from the Breakpoints window?
3. How can I force data breakpoint changes made in the Breakpoints window to reflect in the EE window?
4. [Coding standards] At what point do I create a new variable (string, for instance) to minimize the number of calls to the same function? For example, I make two calls to `variable.FullName()` in `ConstructDebugPropertyInfo()` and three calls to `_variableInformation.FullName` in `GetDataBreakpointInfo160()`.

Relevant section of sample test code is as follows:
```
int testint = 100;
int* test = &testint;
```

gregg-miskelly:
> 1. I looked through the VS repository and it seems `GetDataBreakpointInfo160()` is called for the EE window but not the Breakpoints window. Is this correct?

Yes

> 2. Why might GDB return 0x0000000000000000 when setting a data breakpoint at *test from the Breakpoints window?

Is GDB returning 0 for '&<Expression>'? If yes, what value for 'Expression' are you using? If not, what command?

> 3. How can I force data breakpoint changes made in the Breakpoints window to reflect in the EE window?

After removing the breakpoint from the breakpoints window, are you seeing the EE re-request property info? If not, that sounds like a bug in the UI.

> 4. [Coding standards] At what point do I create a new variable (string, for instance) to minimize the number of calls to the same function? For example, I make two calls to `variable.FullName()` in `ConstructDebugPropertyInfo()` and three calls to `_variableInformation.FullName` in `GetDataBreakpointInfo160()`.

I would say that you usually would want to introduce a new local variable to remove redundant calls unless you know for sure that the call is super cheap (which it certainly isn't in this case).

PR #1108, opened 2021-02-26
Remove entry breakpoint after stopping without reason

benmcmorran:
Attempting to launch a debug configuration in VS Code using OpenOCD targeting an ARM microcontroller throws a spurious exception on entry.

What's happening is MIEngine adds a breakpoint on `main` with `-break-insert -f main`, then connects to the remote target. The target immediately responds with `*stopped,` which triggers the code path edited in this PR and sets `EntrypointHit` to `true`. However, the breakpoint on `main` is still set. Shortly, OpenOCD sends a `breakpoint-hit` event, but the entry breakpoint does not exist in the `_breakpointManager`, only in `_entryPointBreakpoint`, so it's treated as an exception instead.

The fix calls `OnEntrypointHit()` after changing the value of `EntrypointHit` to ensure that the breakpoint is removed before continuing execution.

cc @aleun

image

Relevant portion of the log:

```
1: (500) <-1010-break-insert -f main
1: (500) ->&"\n"
1: (500) ->^done
1: (500) ->(gdb)
1: (500) ->1010^done,bkpt={number="1",type="breakpoint",disp="keep",enabled="y",addr="0x08000e66",func="main",file="../main.c",fullname="D:\\repos\\embedded\\examples\\blinky\\AZ3166\\main.c",line="7",thread-groups=["i1"],times="0",original-location="main"}
1: (500) ->(gdb)
1: (500) ->&"\n"
1: (500) ->^done
1: (500) ->(gdb)
1: (504) 1010: elapsed time 4
1: (505) <-1011-target-select remote localhost:3333
1: (2606) ->=thread-group-started,id="i1",pid="42000"
1: (2606) ->=thread-created,id="1",group-id="i1"
1: (2613) <-1012-thread-info 1
1: (2615) ->~"Reset_Handler () at D:/repos/embedded/examples/official-azure-rtos-samples/MXChip/AZ3166/lib/stm32cubef4/Drivers/CMSIS/Device/ST/STM32F4xx/Source/Templates/gcc\\startup_stm32f412rx.s:62\n"
1: (2615) ->~"62\t  ldr   sp, =_estack       /* set stack pointer */\n"
[...]

PR #1109, opened 2021-03-01
Add stopAtConnect option and improve entrypoint breakpoint handling

benmcmorran:
I was running into issues debugging an embedded device with OpenOCD where the `stopAtEntry` setting in `launch.json` was ignored and the debugger always threw an exception with "Unknown breakpoint" when entering main.

The problem was that `_callback.OnEntryPoint(thread)` wasn't called after setting `EntrypointHit`, so none of the `stopAtEntry` logic in OpenDebugAD7 had a chance to run, and that the entrypoint breakpoint inserted on `main` by MIEngine was never removed by `this.OnEntrypointHit()`. However, even with that added, there was another problem because `CmdContinueAsync()` runs unconditionally.

The fix adds a new option, `stopAtConnect` that will always treat unexpected stopped events when connecting as if an entrypoint has been hit and stop. The fix also makes sure that the entrypoint breakpoint inserted on `main` is always cleaned up to avoid "Unknown breakpoint" exceptions.

With these changes, an embedded device debugging session will correctly respond to `stopAtEntry` and `stopAtConnect`, and will no longer throw spurious exceptions.

cc @aleun

### DIFF
--- a/src/MICore/JsonLaunchOptions.cs
+++ b/src/MICore/JsonLaunchOptions.cs
@@ -95,6 +95,12 @@ namespace MICore.Json.LaunchOptions
         /// </summary>
         [JsonProperty("setupCommands", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public List<SetupCommand> SetupCommands { get; protected set; }
+
+        /// <summary>
+        /// Optional parameter. If true, the debugger should stop after connecting to the target.
+        /// </summary>
+        [JsonProperty("stopAtConnect", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public bool? StopAtConnect { get; set; }
     }
 
     public partial class AttachOptions : BaseOptions
@@ -127,7 +133,8 @@ namespace MICore.Json.LaunchOptions
             string miDebuggerServerAddress = null,
             Dictionary<string, object> sourceFileMap = null,
             PipeTransport pipeTransport = null,
-            SymbolLoadInfo symbolLoadInfo = null)
+            SymbolLoadInfo symbolLoadInfo = null,
+            bool? stopAtConnect = null)
         {
             this.Program = program;
             this.Type = type;
@@ -143,6 +150,7 @@ namespace MICore.Json.LaunchOptions
             this.SourceFileMap = sourceFileMap;
             this.PipeTransport = pipeTransport;
             this.SymbolLoadInfo = symbolLoadInfo;
+            this.StopAtConnect = stopAtConnect;
         }
 
         #endregion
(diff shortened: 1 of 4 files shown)


benmcmorran:
FYI this will have (easily resolved) merge conflicts after #1108 goes in.

src/MICore/JsonLaunchOptions.cs
```diff
@@ -95,6 +95,12 @@ public abstract partial class BaseOptions
         /// </summary>
         [JsonProperty("setupCommands", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public List<SetupCommand> SetupCommands { get; protected set; }
+
+        /// <summary>
+        /// Optional parameter. If true, the debugger should stop after connecting to the target.
+        /// </summary>
+        [JsonProperty("stopAtConnect", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public bool? StopAtConnect { get; set; }
```
gregg-miskelly, line 103: We should add this to the openFolder schema: src/MIDebugPackage/OpenFolderSchema.json

src/MICore/JsonLaunchOptions.cs
```diff
@@ -95,6 +95,12 @@ public abstract partial class BaseOptions
         /// </summary>
         [JsonProperty("setupCommands", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public List<SetupCommand> SetupCommands { get; protected set; }
+
+        /// <summary>
+        /// Optional parameter. If true, the debugger should stop after connecting to the target.
+        /// </summary>
+        [JsonProperty("stopAtConnect", DefaultValueHandling = DefaultValueHandling.Ignore)]
```
gregg-miskelly, line 102: You have this in `BaseOptions` so that it is common between attach and launch, but does it actually apply to attach?

gregg-miskelly (approved):
Otherwise LGTM

src/MICore/JsonLaunchOptions.cs
benmcmorran: [Added](https://github.com/microsoft/MIEngine/pull/1109/files#diff-98a456ea0533a77610e519a56b3866fac24085519f4d42522a09ffbb4fa593ebR176).

src/MICore/JsonLaunchOptions.cs
benmcmorran: Moved to `LaunchOptions`.

gregg-miskelly:
@benmcmorran Can you update the description to match what your PR does now?
@paulmaybee @WardenGnaw Do either of you want to have a look before this is merged?

benmcmorran:
Updated title and description.

PR #1111, opened 2021-03-09
Set lldb mode on the attach initialization path. Enable pipeCmd args.

paulmaybee:
Ran across a few issues in trying to attach to process on a mac. 
1. The json launch configuration didn't allow specification of the pipeCMd string.
2. MIMode was not propagated to the launch options when attaching.

### DIFF
--- a/src/MICore/JsonLaunchOptions.cs
+++ b/src/MICore/JsonLaunchOptions.cs
@@ -487,7 +487,7 @@ namespace MICore.Json.LaunchOptions
         public List<string> PipeArgs { get; private set; }
 
         /// <summary>
-        /// Command line arguments passed to the pipe program to execute a reote command.
+        /// Command line arguments passed to the pipe program to execute a remote command.
         /// </summary>
         [JsonProperty("pipeCmd", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public List<string> PipeCmd { get; private set; }
--- a/src/MICore/JsonLaunchOptions.cs
+++ b/src/MICore/JsonLaunchOptions.cs
@@ -487,6 +487,12 @@ namespace MICore.Json.LaunchOptions
         public List<string> PipeArgs { get; private set; }
 
         /// <summary>
+        /// Command line arguments passed to the pipe program to execute a reote command.
+        /// </summary>
+        [JsonProperty("pipeCmd", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public List<string> PipeCmd { get; private set; }
+
+        /// <summary>
         /// The full path to the debugger on the target machine, for example /usr/bin/gdb.
         /// </summary>
         [JsonProperty("debuggerPath", DefaultValueHandling = DefaultValueHandling.Ignore)]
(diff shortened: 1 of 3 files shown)


src/MICore/JsonLaunchOptions.cs
```diff
@@ -486,6 +486,12 @@ public partial class PipeTransportOptions
         [JsonProperty("pipeArgs", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public List<string> PipeArgs { get; private set; }
 
+        /// <summary>
+        /// Command line arguments passed to the pipe program to execute a reote command.
```
WardenGnaw, line 490: Spelling: `reote` -> `remote`

```suggestion
        /// Command line arguments passed to the pipe program to execute a remote command.
```

WardenGnaw (changes requested):
Is this for VS Code or OpenFolder in VS?

If so, we will also need to update the [package.json](https://github.com/microsoft/vscode-cpptools/blob/16f5f68ff899b6b11c1c6a34df89f0e54f75e3c7/Extension/tools/OptionsSchema.json#L8-L57) in cpptools or the OpenFolderSchema.json:
https://github.com/microsoft/MIEngine/blob/e27243b596c0dff8f33a264cf57266d8f9286076/src/MIDebugPackage/OpenFolderSchema.json#L189-L224

src/MIDebugPackage/OpenFolderSchema.json
```diff
@@ -205,6 +205,13 @@
                       "type": "string"
                     }
                   },
+                  "pipeCmd": {
+                    "type": "array",
+                    "description": "Command line arguments passed to the pipe program to execute commands.",
```
gregg-miskelly, line 210: >Command line arguments passed to the pipe program to execute commands. [](start = 36, length = 70)

I think the description we have in the .xsd is better: `Arguments passed to the pipe executable to run a remote Unix (ex: bash or similar) shell command. This is used for running 'kill' on the remote system, or other commands that the debugger may need. If not specified, the debugger will do its best without the shell. But some features, such as setting breakpoints in run mode, may not work. This string should contain the string '{0}' which will be replaced with the command to execute.`

src/MICore/JsonLaunchOptions.cs
gregg-miskelly: Use the description from XSD instead: `Arguments passed to the pipe executable to run a remote Unix (ex: bash or similar) shell command. This is used for running 'kill' on the remote system, or other commands that the debugger may need. If not specified, the debugger will do its best without the shell. But some features, such as setting breakpoints in run mode, may not work. This string should contain the string '{0}' which will be replaced with the command to execute.`

---
In reply to: [590604833](https://github.com/microsoft/MIEngine/pull/1111#discussion_r590604833) [](ancestors = 590604833)

paulmaybee: pipeCmd is to be used in VS. Updated OpenFolderSchema.json.
Fixed comment

gregg-miskelly (approved):
Otherwise looks good to me.

src/MIDebugPackage/OpenFolderSchema.json
paulmaybee: Fixed.

WardenGnaw (approved):

paulmaybee:
@WardenGnaw When do you think will be the next push to VS?

PR #1115, opened 2021-03-10
Extended remote

mbittarelli6river:
This enables the usage of the GDB client "extended-remote" feature. The schema for enabling this feature is totally up for discussion, but I made some decisions just to get the ball rolling.

One big downside to this PR (and the design of the `vscode-cpptools` plugin in general) is that it only anticipates connecting to a single remote process at a time, though GDB allows for connecting to several running processes at once.  I had some ideas to expand the plugin functionality to support this, but I figured this small PR to enable the functionality for a single process was worthwhile on its own.

This PR is in part a response to a requested feature: https://github.com/microsoft/vscode-cpptools/issues/2118
After getting approval for the design / code change in this PR I will reflect those decisions in the `launch.json` schema change for the https://github.com/microsoft/vscode-cpptools repo


WardenGnaw:
Thank you for the initial PR. 

We will be supporting `extended-remote` with a flag `useExtendedRemote` as defined in https://github.com/microsoft/MIEngine/pull/1260

PR #1116, opened 2021-03-11
Don't require the pipeCmd.

paulmaybee:
Fixes null ref exception when pipeCmd is NOT present. Falls back to previous behavior - use the pipeArgs instead.


WardenGnaw (approved):

paulmaybee:
@WardenGnaw I am really sorry about that. That last checkin broke Linux CMake debugging. I suppose we should run the linux tests on MIEngine insertion.

PR #1117, opened 2021-03-11
Task 1274960: Add support for IDebugProperty160.GetDataBreakpointInfo160 in MIEngine (cont.)

gc46:
I addressed the following PR comments:
1.	You don’t want to add/remove from DataBreakpointVariables/VariableNameAddressMap when GetDataBreakpointInfo160 is called, as this is before the UI has actually us to set the data breakpoint. Instead this should be done when the data breakpoint is created/disabled/deleted.
2.	To support [1], I believe you will need to change the ‘pbstrAddress’ to be some sort of encoding of the full name and the address. You can then parse the two back apart when handling the create/delete/disable data breakpoint. Maybe: ‘<address>,<full-name>’
3.	On AD7Property.cs line 88, you need a try/catch around variable.Address() since that can throw
4.	It’s possible to add the same variable multiple times with different addresses, so I think I would suggest changing DataBreakpointVariables and VariableNameAddressMap to a siengle list with both the pair that contains both the variable name and its address

Remaining PR comments to address:
1.	I should have noticed this earlier, but on AD7Proces.cs line 435, I think you actually want the display name to be the full name.
2.	In Variable.cs line 106 – if the regex your checking is the first regex in your list: returning the result is fine. But if it is your second Regex then you need to return the value of the capture rather than ‘result’. I guess really you could just check the second regex since anything that matches the first regex would also match the second.

@gregg-miskelly, I have questions about the remaining PR comments:
1.	What should the display name be - Name or FullName(), and why? (This is a good learning opportunity for me to understand why one over the other and gives me insight as to the decision process.)
2.	What is meant by "the value of the capture"? Is this `result` with the opening and closing parentheses removed (in the case of the second regex)?

Thank you!

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
@@ -4,6 +4,7 @@
 using Microsoft.DebugEngineHost;
 using Microsoft.VisualStudio.Debugger.Interop;
 using System;
+using System.Diagnostics;
 
 namespace Microsoft.MIDebugEngine
 {
@@ -53,14 +54,15 @@ namespace Microsoft.MIDebugEngine
         {
             if (Delete() == Constants.S_OK)
             {
-                if ((enum_BP_LOCATION_TYPE)PendingBreakpoint.BpRequestInfo.bpLocation.bpLocationType == enum_BP_LOCATION_TYPE.BPLT_DATA_STRING)
+                if (this.IsDataBreakpoint)
                 {
                     lock (_engine.DebuggedProcess.DataBreakpointVariables)
                     {
-                        string addressName = HostMarshal.GetDataBreakpointStringForIntPtr(PendingBreakpoint.BpRequestInfo.bpLocation.unionmember3);
-                        if (_engine.DebuggedProcess.DataBreakpointVariables.Contains(addressName)) // might need to expand condition
+                        string addressId = _pendingBreakpoint.AddressId;
+                        if (addressId != null)
                         {
-                            _engine.DebuggedProcess.DataBreakpointVariables.Remove(addressName);
+                            Debug.Assert(_engine.DebuggedProcess.DataBreakpointVariables.Contains(addressId));
+                            _engine.DebuggedProcess.DataBreakpointVariables.Remove(addressId);
                         }
                     }
                 }
@@ -84,19 +86,19 @@ namespace Microsoft.MIDebugEngine
         int IDebugBoundBreakpoint2.Enable(int fEnable)
         {
             Enabled = fEnable == 0 ? false : true;
-            if ((enum_BP_LOCATION_TYPE)PendingBreakpoint.BpRequestInfo.bpLocation.bpLocationType == enum_BP_LOCATION_TYPE.BPLT_DATA_STRING)
+            if (this.IsDataBreakpoint)
             {
                 lock (_engine.DebuggedProcess.DataBreakpointVariables)
                 {
-                    string addressName = HostMarshal.GetDataBreakpointStringForIntPtr(PendingBreakpoint.BpRequestInfo.bpLocation.unionmember3);
-                    bool InDataBreakpointVariables = _engine.DebuggedProcess.DataBreakpointVariables.Contains(addressName);
-                    if (Enabled && !InDataBreakpointVariables) // might need to expand condition
+                    string addressId = _pendingBreakpoint.AddressId;
+                    bool InDataBreakpointVariables = _engine.DebuggedProcess.DataBreakpointVariables.Contains(addressId);
+                    if (Enabled && !InDataBreakpointVariables)
                     {
-                        _engine.DebuggedProcess.DataBreakpointVariables.Add(addressName);
+                        _engine.DebuggedProcess.DataBreakpointVariables.Add(addressId);
                     }
                     else if (!Enabled && InDataBreakpointVariables)
                     {
-                        _engine.DebuggedProcess.DataBreakpointVariables.Remove(addressName);
+                        _engine.DebuggedProcess.DataBreakpointVariables.Remove(addressId);
                     }
                 }
             }
--- a/src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
@@ -91,14 +91,17 @@ namespace Microsoft.MIDebugEngine
                 lock (_engine.DebuggedProcess.DataBreakpointVariables)
                 {
                     string addressId = _pendingBreakpoint.AddressId;
-                    bool InDataBreakpointVariables = _engine.DebuggedProcess.DataBreakpointVariables.Contains(addressId);
-                    if (Enabled && !InDataBreakpointVariables)
+                    if (addressId != null)
                     {
-                        _engine.DebuggedProcess.DataBreakpointVariables.Add(addressId);
-                    }
-                    else if (!Enabled && InDataBreakpointVariables)
-                    {
-                        _engine.DebuggedProcess.DataBreakpointVariables.Remove(addressId);
+                        bool InDataBreakpointVariables = _engine.DebuggedProcess.DataBreakpointVariables.Contains(addressId);
+                        if (Enabled && !InDataBreakpointVariables)
+                        {
+                            _engine.DebuggedProcess.DataBreakpointVariables.Add(addressId);
+                        }
+                        else if (!Enabled && InDataBreakpointVariables)
+                        {
+                            _engine.DebuggedProcess.DataBreakpointVariables.Remove(addressId);
+                        }
                     }
                 }
             }
--- a/src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
@@ -1,12 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
-using System.Collections.Generic;
-using System.Text;
 using Microsoft.VisualStudio.Debugger.Interop;
-using System.Diagnostics;
-using System.Threading.Tasks;
+using System;
 
 namespace Microsoft.MIDebugEngine
 {
@@ -54,7 +50,20 @@ namespace Microsoft.MIDebugEngine
         // Called when the breakpoint is being deleted by the user.
         int IDebugBoundBreakpoint2.Delete()
         {
-            return Delete();
+            if (Delete() == Constants.S_OK)
+            {
+                // test -- need to delete; insert something here
+                lock (_engine.DebuggedProcess.DataBreakpointVariables)
+                {
+                    string address = PendingBreakpoint.Address;
+                    if (_engine.DebuggedProcess.DataBreakpointVariables.Contains(address)) // might need to expand condition
+                    {
+                        _engine.DebuggedProcess.DataBreakpointVariables.Remove(address);
+                    }
+                }
+                return Constants.S_OK;
+            }
+            return Constants.E_FAIL;
         }
 
         //called by the AD7 Entry Point and by the Detach code path to clean up breakpoints on detach
@@ -72,6 +81,20 @@ namespace Microsoft.MIDebugEngine
         int IDebugBoundBreakpoint2.Enable(int fEnable)
         {
             Enabled = fEnable == 0 ? false : true;
+            // test -- need to delete; insert something here
+            lock (_engine.DebuggedProcess.DataBreakpointVariables)
+            {
+                string address = PendingBreakpoint.Address;
+                bool InDataBreakpointVariables = _engine.DebuggedProcess.DataBreakpointVariables.Contains(address);
+                if (Enabled && !InDataBreakpointVariables) // might need to expand condition
+                {
+                    _engine.DebuggedProcess.DataBreakpointVariables.Add(address);
+                }
+                else if (!Enabled && InDataBreakpointVariables)
+                {
+                    _engine.DebuggedProcess.DataBreakpointVariables.Remove(address);
+                }
+            }
             return Constants.S_OK;
         }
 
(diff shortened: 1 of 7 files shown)


src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
```diff
@@ -54,7 +51,22 @@ public AD7BoundBreakpoint(AD7Engine engine, AD7PendingBreakpoint pendingBreakpoi
         // Called when the breakpoint is being deleted by the user.
         int IDebugBoundBreakpoint2.Delete()
         {
-            return Delete();
+            if (Delete() == Constants.S_OK)
+            {
+                if ((enum_BP_LOCATION_TYPE)PendingBreakpoint.BpRequestInfo.bpLocation.bpLocationType == enum_BP_LOCATION_TYPE.BPLT_DATA_STRING)
```
WardenGnaw, line 56:
```suggestion
                if (this.IsDataBreakpoint)
```

src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
```diff
@@ -72,6 +84,22 @@ public int Delete()
         int IDebugBoundBreakpoint2.Enable(int fEnable)
         {
             Enabled = fEnable == 0 ? false : true;
+            if ((enum_BP_LOCATION_TYPE)PendingBreakpoint.BpRequestInfo.bpLocation.bpLocationType == enum_BP_LOCATION_TYPE.BPLT_DATA_STRING)
```
WardenGnaw, line 87:
```suggestion
            if (this.IsDataBreakpoint)
```

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -45,6 +42,7 @@ internal class AD7PendingBreakpoint : IDebugPendingBreakpoint2
         private IEnumerable<Checksum> _checksums = null;
 
         public DebuggedProcess DebuggedProcess { get { return _engine.DebuggedProcess; } }
+        public BP_REQUEST_INFO BpRequestInfo { get { return _bpRequestInfo; } }
```
WardenGnaw, line 45: Remove when you switch to `IsDataBreakpoint`

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -241,7 +239,7 @@ int IDebugPendingBreakpoint2.Bind()
 
                                     break;
                                 case enum_BP_LOCATION_TYPE.BPLT_DATA_STRING:
-                                    _address = HostMarshal.GetDataBreakpointStringForIntPtr(_bpRequestInfo.bpLocation.unionmember3);
+                                    _address = HostMarshal.GetDataBreakpointStringForIntPtr(_bpRequestInfo.bpLocation.unionmember3).Split(',')[0];
```
gregg-miskelly, line 242: Rather than repeatedly calling `HostMarshal.GetDataBreakpointStringForIntPtr(_bpRequestInfo.bpLocation.unionmember3)` I would suggest doing:

1. Add a new internal property: `internal string AddressId { get; }`
2. Here do:

```C#
string address = HostMarshal.GetDataBreakpointStringForIntPtr(_bpRequestInfo.bpLocation.unionmember3);
if (address.Contains(","))
{
    this.AddressId = address;
    _address = address.Split(',')[0];
}
else
{
    this.AddressId = null;
    _address = address;
}
```

3. In AD7BoundBreakpoint.cs, use _pendingBreakpoint.AddressId

src/MIDebugEngine/AD7.Impl/AD7Property.cs
```diff
@@ -82,13 +82,15 @@ public DEBUG_PROPERTY_INFO ConstructDebugPropertyInfo(enum_DEBUGPROP_INFO_FLAGS
...
+                    try
                     {
-                        if (_engine.DebuggedProcess.VariableNameAddressMap.Contains(fullName + "," + variable.Address()))
+                        if (_engine.DebuggedProcess.DataBreakpointVariables.Contains(variable.Address() + "," + variable.FullName()))
                         {
                             propertyInfo.dwAttrib |= (enum_DBG_ATTRIB_FLAGS)DBG_ATTRIB_HAS_DATA_BREAKPOINT;
                         }
+                    } catch (Exception e)
+                    {
+                        // do something here
```
gregg-miskelly, line 93: I don't think there is anything you can do.

src/MIDebugEngine/AD7.Impl/AD7Property.cs
```diff
@@ -430,25 +432,10 @@ public int GetDataBreakpointInfo160(out string pbstrAddress, out uint pSize, out
         {
             try
             {
-                pbstrAddress = _variableInformation.Address();
+                pbstrAddress = _variableInformation.Address() + "," + _variableInformation.Name;
                 pSize = _variableInformation.Size();
                 pbstrDisplayName = _variableInformation.Name;
```
gregg-miskelly, line 437: = fullName;

src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
```diff
@@ -54,7 +51,22 @@ public AD7BoundBreakpoint(AD7Engine engine, AD7PendingBreakpoint pendingBreakpoi
...
         int IDebugBoundBreakpoint2.Delete()
         {
-            return Delete();
+            if (Delete() == Constants.S_OK)
+            {
+                if ((enum_BP_LOCATION_TYPE)PendingBreakpoint.BpRequestInfo.bpLocation.bpLocationType == enum_BP_LOCATION_TYPE.BPLT_DATA_STRING)
+                {
+                    lock (_engine.DebuggedProcess.DataBreakpointVariables)
+                    {
+                        string addressName = HostMarshal.GetDataBreakpointStringForIntPtr(PendingBreakpoint.BpRequestInfo.bpLocation.unionmember3);
```
gregg-miskelly, line 60: 1. See my other note about using `AddressId` instead of `GetDataBreakpointStringForIntPtr`
2. You want to null check that the AddressId before adding it.

src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
```diff
@@ -54,7 +51,22 @@ public AD7BoundBreakpoint(AD7Engine engine, AD7PendingBreakpoint pendingBreakpoi
...
+                        if (_engine.DebuggedProcess.DataBreakpointVariables.Contains(addressName)) // might need to expand condition
```
gregg-miskelly, line 61: You shouldn't need to call .Contains. Firstly because removing a non-existing element doesn't throw, it just returns false. Secondly because it is probably a bug if the element isn't in the list, so we should probably be asserting.

src/MIDebugEngine/AD7.Impl/AD7Property.cs
```diff
@@ -82,13 +82,15 @@ public DEBUG_PROPERTY_INFO ConstructDebugPropertyInfo(enum_DEBUGPROP_INFO_FLAGS
                 } else
                 {
                     propertyInfo.dwAttrib |= enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_DATA;
-                    string fullName = variable.FullName();
-                    if (_engine.DebuggedProcess.DataBreakpointVariables.Contains(fullName))
+                    try
```
gregg-miskelly, line 85: You need an optimization here where we only call `variable.Address()` if DataBreakpointVariables has an entry that might match:

1. Move the call to `variable.FullName()` outside of ifs and assign it to a temporary.
2. Here add:

```C#
                
                    if (!string.IsNullOrEmpty(fullName))
                    {
                        lock (_engine.DebuggedProcess.DataBreakpointVariables)
                        {
                            if (_engine.DebuggedProcess.DataBreakpointVariables.Any(candidate =>
                                candidate.Length > fullName.Length
                                && candidate.EndsWith(fullName, StringComparison.Ordinal)
                                && candidate[candidate.Length - 1 - fullName.Length] == ','))
                            {
<your-code-here>
                            }
                        }
                    }
```

gregg-miskelly (changes requested):

src/MIDebugEngine/AD7.Impl/AD7Property.cs
gregg-miskelly: FYI This is why you were seeing refresh being so slow.

src/MIDebugEngine/AD7.Impl/AD7Property.cs
gregg-miskelly, line 437: >pbstrDisplayName [](start = 16, length = 16)

To address:

> 1. What should the display name be - Name or FullName(), and why? (This is a good learning opportunity for me to understand why one over the other and gives me insight as to the decision process.)

The purpose of the display name is to be the name for the data breakpoint in the breakpoints window. You don't want it to be just 'Name' because, in general, 'Name' isn't very descriptive. Ex: I I drill down into 'a.b.c' the name for 'c' is just 'c' rather than 'a.b.c'. We could probably do even better than FullName() since MIEngine does some tricks with FullName to make it easier to evaluate (ex: converting pointers). But it should still be better than name.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -1,15 +1,15 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using MICore;
```
gregg-miskelly, line 4: To address:

> 2. What is meant by "the value of the capture"? Is this result with the opening and closing parentheses removed (in the case of the second regex)?

Replace s_validAddressFormats with:
`static readonly Lazy<Regex> s_addressPattern = new Lazy<Regex>(new Regex(@"^(0x[0-9a-fA-F]+)\b"));`

Replace:
```
                if (Regex.IsMatch(result, s_validAddressFormats[i]))
                {
                    return result;
                }
```

With:
```
            Match m = s_addressPattern.Value.Match(result);
            if (m.Success)
            {
                return m.Captures[0]
            }
```

src/MIDebugEngine/AD7.Impl/AD7Property.cs
gregg-miskelly: Sorry if this is obvious, but you will need to add `using System.Linq;` for `.Any` to work.

src/MIDebugEngine/Engine.Impl/Variables.cs
gregg-miskelly: To give a bit more of an explanation - in regex language '(<something-here>)' is a 'capture group'. Whatever you put there is available in the 'Captures' collection of a successful regex match.

---
In reply to: [592817678](https://github.com/microsoft/MIEngine/pull/1117#discussion_r592817678) [](ancestors = 592817678)

src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
gc46: Would calling `.Contains` in the assert be okay?

src/MIDebugEngine/AD7.Impl/AD7Property.cs
gc46: Thanks!

src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
gregg-miskelly: Sure

src/MIDebugEngine/AD7.Impl/AD7BoundBreakpoint.cs
```diff
@@ -72,6 +86,22 @@ public int Delete()
         int IDebugBoundBreakpoint2.Enable(int fEnable)
         {
             Enabled = fEnable == 0 ? false : true;
+            if (this.IsDataBreakpoint)
+            {
+                lock (_engine.DebuggedProcess.DataBreakpointVariables)
+                {
+                    string addressId = _pendingBreakpoint.AddressId;
+                    bool InDataBreakpointVariables = _engine.DebuggedProcess.DataBreakpointVariables.Contains(addressId);
```
gregg-miskelly, line 94: You need an `if (addressId != null)` here

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -89,22 +89,17 @@ internal sealed class VariableInformation : IVariableInformation
         private string DisplayHint { get; set; }
         public bool IsPreformatted { get; set; }
 
-        private static readonly string[] s_validAddressFormats = new string[] {
-            @"^0x[0-9a-fA-F]+$",
-            @"^(0x[0-9a-fA-F]+)\b"
-        };
+        static readonly Lazy<Regex> s_addressPattern = new Lazy<Regex>(() => new Regex(@"^(0x[0-9a-fA-F]+)\b"));
```
WardenGnaw, line 92:
```suggestion
        static readonly Lazy<Regex> s_addressPattern = new Lazy<Regex>(() => new Regex(@"^(0x[0-9a-fA-F]+)\b", RegexOptions.Compiled));
```
Compile Regex for performance. https://docs.microsoft.com/en-us/dotnet/standard/base-types/compilation-and-reuse-in-regular-expressions

src/MIDebugEngine/AD7.Impl/AD7Property.cs
```diff
@@ -430,25 +444,10 @@ public int GetDataBreakpointInfo160(out string pbstrAddress, out uint pSize, out
         {
             try
             {
-                pbstrAddress = _variableInformation.Address();
+                pbstrAddress = _variableInformation.Address() + "," + _variableInformation.FullName();
                 pSize = _variableInformation.Size();
                 pbstrDisplayName = _variableInformation.Name;
```
gregg-miskelly, line 449: `pbstrDisplayName` still needs to be changed to fullName.

src/MIDebugEngine/Engine.Impl/Variables.cs
gregg-miskelly: This feels like it is too rarely used to be worth using.

---
In reply to: [594648798](https://github.com/microsoft/MIEngine/pull/1117#discussion_r594648798) [](ancestors = 594648798)

WardenGnaw (approved):

gregg-miskelly (approved):

PR #1123, opened 2021-03-18
Add hardware breakpoint support

aleun:
This PR adds a `hardwareBreakpoints` setting for toggling hardware breakpoint behavior.

Example: `"hardwareBreakpoints": { "require": true, "limit": 5 }`

1. `require` forces all breakpoints to be hardware breakpoints via the `-break-insert -h` option.
2. `limit` restricts the number of remote hardware breakpoints via GDB's `set remote hardware-breakpoint-limit` command. It is optional. Any breakpoints past the limit do not bind and instead show an error message (screenshots below).

## In VSCode

Screenshot:
vscode_screen

Note that the error message comes directly from GDB.

The relevant excerpt of the debug console log:

```
1: (47954) <-1024-break-insert -f -h main.c:18
1: (47957) ->1024^done,bkpt={number="3",type="hw breakpoint",disp="keep",enabled="y",addr="0x08000fd8",func="main",file="../main.c",fullname="C:\\proj\\embedded\\examples\\blinky\\STM32L4\\main.c",line="18",thread-groups=["i1"],times="0",original-location="main.c:18"}
1: (47957) ->(gdb)
1: (47957) 1024: elapsed time 3
1: (47957) Send Event AD7BreakpointBoundEvent
1: (47957) ->&"\n"
1: (47958) ->^done
1: (47958) ->(gdb)
1: (49596) <-1025-break-insert -f -h main.c:19
1: (49601) ->1025^error,msg="Hardware breakpoints used exceeds limit."
1: (49601) ->(gdb)
1: (49601) 1025: elapsed time 4
1: (49602) ->&"\n"
1: (49602) ->^done
1: (49602) ->(gdb)
1: (49602) Send Event AD7BreakpointErrorEvent
ERROR: Hardware breakpoints used exceeds limit.
```

## In VS 2019

Screenshot:
vs_screen

Relevant log excerpt:

```
1: (247977) <-1020-break-insert -f -h main.c:99
1: (247986) ->1020^done,bkpt={number="3",type="hw breakpoint",disp="keep",enabled="y",addr="0x0800017a",func="main",file="../Core/Src/main.c",fullname="C:\\Users\\alleu\\STM32CubeIDE\\workspace_1.5.0\\nucleo-f103rb-blinky\\Core\\Src\\main.c",line="99",thread-groups=["i1"],times="0",original-location="main.c:99"}
1: (247989) ->(gdb)
1: (247989) 1020: elapsed time 11
1: (247995) Send Event AD7BreakpointBoundEvent
1: (247998) ->&"\n"
1: (248009) ->^done
1: (248011) ->(gdb)
1: (251233) <-1021-break-insert -f -h main.c:88
1: (251237) ->1021^error,msg="Hardware breakpoints used exceeds limit."
1: (251239) ->(gdb)
1: (251239) 1021: elapsed time 6
1: (251247) ->&"\n"
1: (251250) ->^done
1: (251250) Send Event AD7BreakpointErrorEvent
1: (251253) ->(gdb)
```

## Implementation Notes

- Between VS and VSCode, there is a difference in behavior when enabling/disabling breakpoints past the limit. Illustrative scenario: suppose the user has just gone above the breakpoint limit, causing their latest breakpoint to remain unbound. They then disable some previous breakpoints, and attempt rebinding the offending breakpoint. The user needs to do slightly different things in VS and VSCode. This is due to the difference between the two in the way user gestures for enabling/disabling breakpoints are interpreted:
  - In VSCode, toggling enables always translates into `-break-insert` and `-break-delete` GDB commands, even if the breakpoints are unbound, so toggling the enable state can "revive" a previously unbound breakpoint.
  - In VS, toggling enables translate into `-break-enable` and `-break-disable` GDB commands, but they are not fired in the case of an unbound breakpoint (because there is nothing for GDB to enable). This means the only way to "revive" the breakpoint in the UI is by deleting and adding it again.
- Tested with OpenOCD on Windows with VS and VSCode.

### DIFF
--- a/src/MICore/JsonLaunchOptions.cs
+++ b/src/MICore/JsonLaunchOptions.cs
@@ -97,7 +97,7 @@ namespace MICore.Json.LaunchOptions
         public List<SetupCommand> SetupCommands { get; protected set; }
 
         /// <summary>
-        /// Force all breakpoints to be hardware breakpoints. Enforce an optional limit if specified. Example: "hardwareBreakpoint": { "require": true, "limit": 5 }.
+        /// Explicitly control whether hardware breakpoints are used. If an optional limit is provided, additionally restrict the number of hardware breakpoints for remote targets. Example: "hardwareBreakpoints": { "require": true, "limit": 5 }.
         /// </summary>
         [JsonProperty("hardwareBreakpoints", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public HardwareBreakpointInfo HardwareBreakpointInfo { get; set; }
@@ -229,7 +229,7 @@ namespace MICore.Json.LaunchOptions
         public bool Require { get; set; }
 
         /// <summary>
-        /// If provided, limit the number of hardware breakpoints to the given number. Default is 0, in which case there is no limit. This setting is only enforced with remote GDB targets.
+        /// When <see cref="Require"/> is true, restrict the number of available hardware breakpoints. Default is 0, in which case there is no limit. This setting is only enforced with remote GDB targets.
         /// </summary>
         [JsonProperty("limit", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public int? Limit { get; set; }
--- a/src/MICore/JsonLaunchOptions.cs
+++ b/src/MICore/JsonLaunchOptions.cs
@@ -97,7 +97,7 @@ namespace MICore.Json.LaunchOptions
         public List<SetupCommand> SetupCommands { get; protected set; }
 
         /// <summary>
-        /// Force all breakpoints to be hardware breakpoints. Enforce an optional limit if specified. Example: "hardwareBreakpoint": { "enable": true, "limit": 5 }.
+        /// Force all breakpoints to be hardware breakpoints. Enforce an optional limit if specified. Example: "hardwareBreakpoint": { "require": true, "limit": 5 }.
         /// </summary>
         [JsonProperty("hardwareBreakpoints", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public HardwareBreakpointInfo HardwareBreakpointInfo { get; set; }
@@ -225,8 +225,8 @@ namespace MICore.Json.LaunchOptions
         /// <summary>
         /// If true, always use hardware breakpoints. Default value is false.
         /// </summary>
-        [JsonProperty("enable")]
-        public bool Enable { get; set; }
+        [JsonProperty("require")]
+        public bool Require { get; set; }
 
         /// <summary>
         /// If provided, limit the number of hardware breakpoints to the given number. Default is 0, in which case there is no limit. This setting is only enforced with remote GDB targets.
@@ -242,9 +242,9 @@ namespace MICore.Json.LaunchOptions
         {
         }
 
-        public HardwareBreakpointInfo(bool enable = false, int? limit = null)
+        public HardwareBreakpointInfo(bool require = false, int? limit = null)
         {
-            this.Enable = enable;
+            this.Require = require;
             this.Limit = limit;
         }
 
--- a/src/MICore/JsonLaunchOptions.cs
+++ b/src/MICore/JsonLaunchOptions.cs
@@ -95,6 +95,12 @@ namespace MICore.Json.LaunchOptions
         /// </summary>
         [JsonProperty("setupCommands", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public List<SetupCommand> SetupCommands { get; protected set; }
+
+        /// <summary>
+        /// If true, always use hardware breakpoints.
+        /// </summary>
+        [JsonProperty("requireHardwareBreakpoints", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public bool? RequireHardwareBreakpoints { get; protected set; }
     }
 
     public partial class AttachOptions : BaseOptions
@@ -125,6 +131,7 @@ namespace MICore.Json.LaunchOptions
             string miDebuggerPath = null,
             string miDebuggerArgs = null,
             string miDebuggerServerAddress = null,
+            bool? requireHardwareBreakpoints = null,
             Dictionary<string, object> sourceFileMap = null,
             PipeTransport pipeTransport = null,
             SymbolLoadInfo symbolLoadInfo = null)
@@ -140,6 +147,7 @@ namespace MICore.Json.LaunchOptions
             this.MiDebuggerArgs = miDebuggerArgs;
             this.MiDebuggerServerAddress = miDebuggerServerAddress;
             this.ProcessId = processId;
+            this.RequireHardwareBreakpoints = requireHardwareBreakpoints;
             this.SourceFileMap = sourceFileMap;
             this.PipeTransport = pipeTransport;
             this.SymbolLoadInfo = symbolLoadInfo;
@@ -344,6 +352,7 @@ namespace MICore.Json.LaunchOptions
             int? serverLaunchTimeout = null,
             string coreDumpPath = null,
             bool? externalConsole = null,
+            bool? requireHardwareBreakpoints = null,
             Dictionary<string, object> sourceFileMap = null,
             PipeTransport pipeTransport = null)
         {
@@ -372,6 +381,7 @@ namespace MICore.Json.LaunchOptions
             this.ServerLaunchTimeout = serverLaunchTimeout;
             this.CoreDumpPath = coreDumpPath;
             this.ExternalConsole = externalConsole;
+            this.RequireHardwareBreakpoints = requireHardwareBreakpoints;
             this.SourceFileMap = sourceFileMap;
             this.PipeTransport = pipeTransport;
         }
--- a/src/MICore/JsonLaunchOptions.cs
+++ b/src/MICore/JsonLaunchOptions.cs
@@ -97,10 +97,10 @@ namespace MICore.Json.LaunchOptions
         public List<SetupCommand> SetupCommands { get; protected set; }
 
         /// <summary>
-        /// If true, always use hardware breakpoints.
+        /// Force all breakpoints to be hardware breakpoints. Enforce an optional limit if specified. Example: "hardwareBreakpoint": { "enable": true, "limit": 5 }.
         /// </summary>
-        [JsonProperty("requireHardwareBreakpoints", DefaultValueHandling = DefaultValueHandling.Ignore)]
-        public bool? RequireHardwareBreakpoints { get; protected set; }
+        [JsonProperty("hardwareBreakpoints", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public HardwareBreakpointInfo HardwareBreakpointInfo { get; set; }
     }
 
     public partial class AttachOptions : BaseOptions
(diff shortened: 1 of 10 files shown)


src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -468,7 +485,15 @@ int IDebugPendingBreakpoint2.Enable(int fEnable)
...
                     {
                         _engine.DebuggedProcess.AddInternalBreakAction(
-                            () => bp.EnableAsync(_enabled, _engine.DebuggedProcess)
+                            async () => {
+                                try
+                                {
+                                    await bp.EnableAsync(_enabled, _engine.DebuggedProcess);
+                                } catch (UnexpectedMIResultException exc)
+                                {
+                                    this.SetError(new AD7ErrorBreakpoint(this, exc.MIError, enum_BP_ERROR_TYPE.BPET_GENERAL_ERROR), true);
```
aleun, line 494: `-break-enable` can fail if this would exceed the breakpoint limit. The exception is caught and turned into a breakpoint error event, as you can see in this log:

```
1: (262805) <-1024-break-enable 3
1: (262813) ->1024^error,msg="Hardware breakpoints used exceeds limit."
1: (262816) 1024: elapsed time 11
1: (262816) ->(gdb)
1: (262852) ->&"\n"
1: (262877) ->^done
1: (262880) ->(gdb)
1: (262981) Send Event AD7BreakpointErrorEvent
```

However, the VS debugger does not reflect the error. Is this a UI issue?
vs_screen_enable

WardenGnaw:
FYI: For the json to appear in VS Code, you will need to add it to [OptionsSchema.json](https://github.com/microsoft/vscode-cpptools/blob/main/Extension/tools/OptionsSchema.json) and run `yarn run generateOptionsSchema` and that will update the `package.json`

src/MICore/JsonLaunchOptions.cs
```diff
@@ -210,6 +218,39 @@ public SymbolLoadInfo(bool? loadAll = null, string exceptionList = null)
         #endregion
     }
 
+    public partial class HardwareBreakpointInfo
```
gregg-miskelly, line 221: >HardwareBreakpointInfo [](start = 25, length = 22)

Need to add this to the json schema

gregg-miskelly: Andrew already left this comment.

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -428,7 +435,17 @@ int IDebugPendingBreakpoint2.Delete()
                 _deleted = true;
                 if (_engine.DebuggedProcess.ProcessState != ProcessState.Stopped && !_engine.DebuggedProcess.MICommandFactory.AllowCommandsWhileRunning())
                 {
-                    _pendingDelete = true;
+                    if (!PendingDeletesEnabled() && _bp != null)
```
gregg-miskelly, line 438: >PendingDeletesEnabled [](start = 25, length = 21)

I am not sure I understand why this change is being made. Is this because we want to make sure we delete the breakpoint before we set another? If so, would it instead make sense to call DeleteBreakpointsPendingDeletion before setting one when `RequireHardwareBreakpoints` is set? Or are there other reasons why it is important to block waiting on the delete to happen?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2923,10 +2927,13 @@ public void HandleIDebugBreakpointErrorEvent2(IDebugEngine2 pEngine, IDebugProce
                             };
 
                             // TODO: currently VSCode will ignore the error message from "breakpoint" event, the workaround is to log the error to output window
-                            string outputMsg = string.Format(CultureInfo.CurrentCulture, AD7Resources.Error_FunctionBreakpoint, ad7BPRequest.FunctionPosition.Name, errorMsg);
-                            m_logger.WriteLine(LoggingCategory.DebuggerError, outputMsg);
+                            outputMessage = string.Format(CultureInfo.CurrentCulture, AD7Resources.Error_FunctionBreakpoint, ad7BPRequest.FunctionPosition.Name, errorMsg);
                         }
 
+                        if (!string.IsNullOrEmpty(outputMessage))
```
gregg-miskelly, line 2933: >outputMessage [](start = 50, length = 13)

You probably want to pay attention to severity here (and maybe we need to tweak which severity we use on line 357) -- if this is just because of a long bind, or because GDB couldn't find a record for the specified source file in the PDB, we wouldn't want to dump it to the console.

gregg-miskelly (changes requested):
Otherwise LGTM

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
aleun: Yes in answer to your first question: we need to make sure it's deleted before setting another so that GDB knows how many hardware breakpoint registers are actually occupied (as opposed to just being squatted on by a breakpoint pending delete). There is no other reason to block waiting for the delete. I reverted the logic here and instead placed a call to `_bpManager.DeleteBreakpointsPendingDeletion()` at the top of `BindAsync()`. Is this what you had in mind?

src/OpenDebugAD7/AD7DebugSession.cs
aleun: I added an additional check for severity greater than `BPET_SEV_LOW`. Line 357 of this file doesn't have anything related to severity. Did you mean to point to another location?

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
aleun: Ping on this one @gregg-miskelly. Is this something that can be addressed in this PR, or does it need investigation in VS debugger UI?

aleun:
> FYI: For the json to appear in VS Code, you will need to add it to [OptionsSchema.json](https://github.com/microsoft/vscode-cpptools/blob/main/Extension/tools/OptionsSchema.json) and run `yarn run generateOptionsSchema` and that will update the `package.json`
> 
> Add your description strings in [package.nls.json](https://github.com/microsoft/vscode-cpptools/blob/main/Extension/package.nls.json)

Thanks for the heads up. I'll open a PR over there once this one is cleared.

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
gregg-miskelly: Yup

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
gregg-miskelly: Sorry, I missed your question. It isn't obvious to me from reading the code why that wouldn't work. But I do see several suspect things that MIEngine does with error breakpoints:

* IDebugPendingBreakpoint2.CanBind would still return S_OK/no error after this failure (there is no check for existing errors)
* `_enabled` is left true after this failure.
* The `&& ((bpErrorType & enum_BP_ERROR_TYPE.BPET_TYPE_ERROR) != 0)` check in IDebugPendingBreakpoint2.EnumErrorBreakpoints looks wrong to me. I don't know if that has anything to do with this, but I don't think it should exist.


---
In reply to: [602621718](https://github.com/microsoft/MIEngine/pull/1123#discussion_r602621718) [](ancestors = 602621718)

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: Sorry, I meant [main branch AD7PendingBreakpoint.cs line 357](https://github.com/microsoft/MIEngine/blob/8456c49b4731971b885c2909f9feaac4bdcd494d/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs#L357). It looks like that uses BPET_GENERAL_WARNING.

I think what you probably want is to check if this is an error rather than a warning: `((errorType & enum_BP_ERROR_TYPE.BPET_TYPE_ERROR) != 0)`

---
In reply to: [602621170](https://github.com/microsoft/MIEngine/pull/1123#discussion_r602621170) [](ancestors = 602621170)

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
gregg-miskelly: Also, we are not failing this call, which I *think* we want to.

aleun: Thanks for you help. Using your pointers, I was able to get a better experience working. Trying to enable too many breakpoints will pop up an error message and leave the breakpoint unbound.
image
This required setting `_enabled` to false and making sure to return `E_FAIL` in `Enable`. It also required adding timeout logic similar to what's used in `Bind` to cope with the fact that the `IDebugPendingBreakpoint2` API is synchronous.  I did not end up modifying `CanBind` or `EnumErrorBreakpoints` because neither code path is exercised by the enable logic. Do you still want me to delete that check for the breakpoint error type even though it's unrelated?

src/OpenDebugAD7/AD7DebugSession.cs
aleun: I changed the error to `BPET_GENERAL_ERROR` and swapped in your check for error type.

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -354,7 +360,7 @@ internal async Task BindAsync()
                     }
                     if (bindResult.BoundBreakpoints == null || bindResult.BoundBreakpoints.Count == 0)
                     {
-                        this.SetError(new AD7ErrorBreakpoint(this, bindResult.ErrorMessage), true);
+                        this.SetError(new AD7ErrorBreakpoint(this, bindResult.ErrorMessage, enum_BP_ERROR_TYPE.BPET_GENERAL_ERROR), true);
```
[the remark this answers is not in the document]
gregg-miskelly: If this is where you are seeing hardware breakpoint failures: we may need some sort of hack here where we look at words in the message to guess the importance (yucky, but I can't think of anything better). Otherwise, I think you want to just take out the change.

---
In reply to: [605125034](https://github.com/microsoft/MIEngine/pull/1123#discussion_r605125034) [](ancestors = 605125034)

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -465,12 +471,35 @@ int IDebugPendingBreakpoint2.Enable(int fEnable)
...
-                        _engine.DebuggedProcess.AddInternalBreakAction(
+                        enableTask = _engine.DebuggedProcess.AddInternalBreakAction(
                             () => bp.EnableAsync(_enabled, _engine.DebuggedProcess)
                         );
                     });
+                    try
+                    {
+                        // Using similar timeout logic to what's used in Bind so we don't block too long
+                        enableTask.Wait(_engine.GetBPLongEnableTimeout());
+                        if (!enableTask.IsCompleted)
```
gregg-miskelly, line 485: >if (!enableTask.IsCompleted) [](start = 24, length = 28)

Are you running into this being very slow? Being async is problematic for this API because there aren't events that an engine can send to update the UI's breakpoint enable state. So the UI will show one state even though the underlying breakpoint may be in a different state. Not the end of the world, but not ideal either.

aleun:
Based on offline discussion, closing this PR in favor of #1128.

Issue #1126, opened 2021-04-12
Docker: command results exceeding BUFMAX (=4096) characters cannot be processed


Valtaroth:
While attaching the Visual Studio 2019 debugger to a process running in a Ubuntu 18.04 Linux Docker container, VS throws multiple `MIResultFormatException` exceptions and fails to attach correctly. After debugging the source code I tracked the issue down to the `RawLocalCommandRunner` used for local Docker connections which reads command results in chunks of `BUFMAX` characters (equaling 4096 characters at the time of writing) and sends the chunk off to processing. In my case though, multiple messages (i.e. `symbol-list-lines` and `threads`) are exceeding this limit and are therefore incomplete, leading to the format exception and subsequent failures and making it impossible to attach the debugger to the process using a Docker connection.

To verify this issue only occurs using the Docker connection type, I installed an SSH server in my container and attached the debugger using that instead. The debugger was able to successfully attach to the process and debugging worked flawlessly, so I assume that it is mainly the Docker connection type which is affected by the length limitation.

From my tests I assume this is fixable by delaying processing of messages hitting `BUFMAX` in `RawLocalCommanderRunner`'s  `ReadLoop` method and only starting processing if the stream returns a result of a lower length or no result at all. I've created a pull request with the changes that enabled me to attach to the process in my particular test case: #1127

PR #1127, opened 2021-04-12
Delay processing of raw command results hitting the buffer limit (#1126)

Valtaroth:
This pull request attempts to fix issue #1126 by delaying processing of messages hitting the BUFMAX length limit until a shorter message is received to ensure lines are complete before parsing them. I've tested it using a Docker connection into a Linux container and was able to attach to and debug a process which fails without the fix because messages are too long and cannot be parsed.


src/SSHDebugPS/CommandRunner.cs
```diff
@@ -302,10 +303,20 @@ protected override void ReadLoop(StreamReader reader, CancellationToken token, b
...
+                        result += new string(buffer, 0, task.Result);
+                        if (task.Result < BUFMAX)
+                        {
+                            action(result);
+                            result = string.Empty;
+                        }
                     }
                     else
                     {
+                        if (string.IsNullOrEmpty(result))
```
gregg-miskelly, line 315: >string [](http://example.com/codeflow?start=28&length=6)

You are missing a `!` here.

gregg-miskelly (changes requested):
Otherwise LGTM

src/SSHDebugPS/CommandRunner.cs
Valtaroth: Of course, thanks for pointing it out!

gregg-miskelly (approved):
Thanks so much for taking the time to submit this PR!

PR #1128, opened 2021-04-13
Add hardware breakpoints support -- unbind on disable

aleun:
This PR supercedes the abandoned PR #1123. It has the same goal, but improves the experience by ensuring that error messages propagate to the UI in VS on enable/disable by unbinding breakpoints on disable.

GIF of behavior in VS 2019:
Hardware Breakpoints


src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -264,36 +266,7 @@ int IDebugPendingBreakpoint2.Bind()
...
-                                string addressName = HostMarshal.GetDataBreakpointStringForIntPtr(_bpRequestInfo.bpLocation.unionmember3);
-                                if (!_engine.DebuggedProcess.DataBreakpointVariables.Contains(addressName)) // might need to expand condition
-                                {
-                                    _engine.DebuggedProcess.DataBreakpointVariables.Add(addressName);
-                                }
-                            }
-                        }
-                        return Constants.S_OK;
-                    }
+                    return BindWithTimeout();
```
gregg-miskelly, line 269: >return [](start = 20, length = 6)

I think we want: `if (!_enabled) { return S_OK; }`

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -448,15 +460,50 @@ internal async Task DeletePendingDelete()
...
+                foreach (var boundBp in _boundBreakpoints)
+                {
+                    _engine.Callback.OnBreakpointUnbound(boundBp, enum_BP_UNBOUND_REASON.BPUR_UNKNOWN);
+                }
+                (this as IDebugPendingBreakpoint2).Delete();
+                _boundBreakpoints.Clear();
+                _BPError = null;
+                 // this pending breakpoint is not actually deleted, just disabled, so override these flags
+                _deleted = false;
+                _pendingDelete = false;
```
gregg-miskelly, line 495: >_pendingDelete = false; [](start = 16, length = 23)

Does this mean that if you disable a breakpoint from run mode it will never get disabled?

gregg-miskelly: I **think** you want to leave _pendingDelete as true (if it was true in the first place), but with the following changes: 
* In [DeleteBreakpointsPendingDeletion](https://github.com/Microsoft/MIEngine/blob/d078b660535b851ab49491c05eb854e10e4b563b/src/MIDebugEngine/Engine.Impl/BreakpointManager.cs#L268-L268) change the condition to removed deleted breakpoints instead of PendingDelete breakpoints: (`b.Deleted`). It looks like this will also fix a reference leak MIEngine has since we don't otherwise remove pending breakpoints.
* Add a comment to PendingDelete to indicate that it returns true if either the breakpoint is deleted or hardware breakpoints are being used and the breakpoint is disabled.

---
In reply to: [612786532](https://github.com/microsoft/MIEngine/pull/1128#discussion_r612786532) [](ancestors = 612786532)

gregg-miskelly (changes requested):
Otherwise LGTM

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
aleun: Yes, you're right. Otherwise we're requesting a pointless disabled breakpoint from GDB. With this change, I don't need to do the somewhat inelegant "always unbind on enable" to deal with initially disabled breakpoints, [so I've removed it](https://github.com/microsoft/MIEngine/pull/1128/commits/5514dd680be7dd755d3ef35e6fec48bd4e1f9452#diff-ceef2a8399bc4087d455776c21f6948685c24b4967202803d844289332003856L486-L500).

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
aleun: Thanks for the catch. Yes, it's my mistake. Your suggestions work as intended -- I did more manual testing in run mode and issued custom `info break` commands to check that there were no GDB breakpoints being left behind anymore.

gregg-miskelly (approved):

WardenGnaw (approved):

PR #1130, opened 2021-04-15
Fix .Designer.cs generation for .resx files

gregg-miskelly:
The conversion to SDK style projects broke having Visual Studio automatically generate the .Designer.cs file when editing a .resx file. This fixes it and regenerates all of our .Designer.cs files.


WardenGnaw (approved):

WardenGnaw:
Does the SDK style proj not auto compile the *.Designer.cs files after they have been generated? Also does it not complain you are pointing to files that should have been captured the compile group?

gregg-miskelly:
> Does the SDK style proj not auto compile the *.Designer.cs files after they have been generated? Also does it not complain you are pointing to files that should have been captured the compile group?

There is a subtle part of this change that I missed when I first saw the work around too -- these are doing an `Update` on items in the item group rather than a `Include`. This is just adding metadata to these items so that the generator works.

PR #1132, opened 2021-04-27
Support GetProcessArchitecture and read ps flags

WardenGnaw:
This PR adds GetProcessArchitecture in IDebugUnixProcess for SSHDebugPS
and obtains the flags in ps.

For most processes, it will be the same as the system architecture from
'uname -m'.

However, on Apple M1 Machines. we need to determine if the process is being
emulated or not. We determine this by checking if P_TRANSLATED (0x20000) on OSX ARM64
is enabled on the process, it is under x86_64 emulation.

### DIFF
--- a/src/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.cs
+++ b/src/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.cs
@@ -164,7 +164,7 @@ namespace Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier
         /// <summary>
         /// Gets the architecture unix process.
         /// </summary>
-        /// <returns>the string representation of the process's architecture</returns>
+        /// <returns>The string representation of the process's architecture. The format of this is architecture-specific and will match that returned by uname -m. Example: 'x86_64'</returns>
         string GetProcessArchitecture();
     };
 }
--- a/src/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.cs
+++ b/src/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.cs
@@ -151,4 +151,20 @@ namespace Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier
         /// <param name="exitCode">The exit code of the command, assuming one was printed.</param>
         void OnExit(string exitCode);
     };
+
+    /// <summary>
+    /// Interface to query unix process architecture information.
+    /// </summary>
+    [ComImport()]
+    [ComVisible(true)]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    [Guid("61D4C2C4-7CA8-4472-8446-7B1DF5C9D16D")]
+    public interface IDebugUnixProcess
+    {
+        /// <summary>
+        /// Gets the architecture unix process.
+        /// </summary>
+        /// <returns>the string representation of the process's architecture</returns>
+        string GetProcessArchitecture();
+    };
 }
--- a/src/SSHDebugPS/IConnection.cs
+++ b/src/SSHDebugPS/IConnection.cs
@@ -53,16 +53,32 @@ namespace Microsoft.SSHDebugPS
     public class Process
     {
         public uint Id { get; private set; }
+        public uint Flags { get; private set; }
+        public string SystemArch { get; private set; }
         public string CommandLine { get; private set; }
         public string UserName { get; private set; }
         public bool IsSameUser { get; private set; }
 
-        public Process(uint id, string userName, string commandLine, bool isSameUser)
+        public Process(uint id, string arch, uint flags, string userName, string commandLine, bool isSameUser)
         {
             this.Id = id;
+            this.Flags = flags;
             this.UserName = userName;
             this.CommandLine = commandLine;
             this.IsSameUser = isSameUser;
+            this.SystemArch = arch;
+        }
+    }
+
+    public class SystemInformation
+    {
+        public string UserName { get; private set; }
+        public string Architecture { get; private set; }
+
+        public SystemInformation(string username, string architecture)
+        {
+            this.UserName = username;
+            this.Architecture = architecture;
         }
     }
 }
--- a/src/SSHDebugPS/IConnection.cs
+++ b/src/SSHDebugPS/IConnection.cs
@@ -53,6 +53,9 @@ namespace Microsoft.SSHDebugPS
     public class Process
     {
         public uint Id { get; private set; }
+        /// <summary>
+        /// Only used by the PSOutputParser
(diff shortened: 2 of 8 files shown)


src/SSHDebugPS/ProcFSOutputParser.cs
```diff
@@ -100,9 +100,12 @@ private void ProcessLine(string line, string username)
...
 
             // If the passed in username is empty, then treat all processes as the same user
-            bool isSameUser = string.IsNullOrWhiteSpace(username) ? true : username.Equals(procUsername, StringComparison.Ordinal);
+            bool isSameUser = string.IsNullOrWhiteSpace(systemInformation.UserName) ? true : systemInformation.UserName.Equals(procUsername, StringComparison.Ordinal);
 
-            Process process = new Process(processId, procUsername, commandLine, isSameUser);
+            // Note:
+            //    For Flags, will need to parse /proc/[pid]/stat flags.
+            //    But at this commit, flags are unused for Linux scenarios.
+            Process process = new Process(processId, systemInformation.Architecture, /* TODO: Flags */ 0, procUsername, commandLine, isSameUser);
```
gregg-miskelly, line 108: >TODO [](http://example.com/codeflow?start=88&length=4)

I don't know that this really needs a TODO comment since we don't plan to use it. Might make more sense to just update the documentation to indicate that this is absent in this case.

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -108,35 +108,50 @@ public override bool IsLinux()
...
 
                 username = commandOutput;
-                return true;
             }
 
-            return false;
+            string architecture = string.Empty;
+            command = "uname -m";
+            if (ExecuteCommand(command, Timeout.Infinite, throwOnFailure: false, commandOutput: out commandOutput, errorMessage: out errorMessage))
+            {
```
gregg-miskelly, line 128: >{ [](http://example.com/codeflow?start=12&length=1)

nit: extra new line

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -108,35 +108,50 @@ public override bool IsLinux()
...
+
+                architecture = commandOutput;
+            }
+
+            if (string.IsNullOrWhiteSpace(username) || string.IsNullOrWhiteSpace(architecture))
```
gregg-miskelly, line 133: >|| [](http://example.com/codeflow?start=52&length=2)

If we only get one, should we still return it?

src/SSHDebugPS/PSOutputParser.cs
```diff
@@ -160,23 +170,28 @@ private Process SplitPSLine(string line)
...
+                return null;
+
             string ruser = _ruserCol.Extract(line);
             string commandLine = _argsCol.Extract(line);
             if (string.IsNullOrEmpty(commandLine))
                 return null;
 
             bool isSameUser;
-            if (!string.IsNullOrEmpty(_currentUserName))
+            if (!string.IsNullOrEmpty(_currentSystemInformation.UserName))
```
gregg-miskelly, line 184: >_currentSystemInformation [](http://example.com/codeflow?start=38&length=25)

This can be null

src/SSHDebugPS/IConnection.cs
```diff
@@ -53,16 +53,32 @@ internal abstract class Connection : IConnection, IDebugUnixShellPort
...
+            this.Flags = flags;
             this.UserName = userName;
             this.CommandLine = commandLine;
             this.IsSameUser = isSameUser;
+            this.SystemArch = arch;
         }
     }
-}
+
+    public class SystemInformation
```
gregg-miskelly, line 73: >public [](http://example.com/codeflow?start=4&length=6)

Does this need to be public?

src/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.cs
```diff
@@ -151,4 +151,20 @@ public interface IDebugUnixShellCommandCallback
...
+    [ComImport()]
+    [ComVisible(true)]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    [Guid("61D4C2C4-7CA8-4472-8446-7B1DF5C9D16D")]
+    public interface IDebugUnixProcess
+    {
+        /// <summary>
+        /// Gets the architecture unix process.
+        /// </summary>
+        /// <returns>the string representation of the process's architecture</returns>
```
gregg-miskelly, line 167: >architecture [](http://example.com/codeflow?start=64&length=12)

Might want to also add: `The format of this is architecture-specific and will match that returned by `uname -m`. Example: x86_64.`

src/SSHDebugTests/PSOutputParserTests.cs
```diff
@@ -62,13 +64,14 @@ public void PSOutputParser_NoUserName()
...
             const string username = "";
+            const string architecture = "";
             const string input =
                 "pppppppppp rrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr COMMAND\n" +
                 "         1 root                             /sbin/init\n" +
                 "       720                                  dbus-daemon --system --fork\n" +
                 "      2389 greggm                           -bash\n";
 
-            List<Process> r = PSOutputParser.Parse(input, username);
+            List<Process> r = PSOutputParser.Parse(input, new SystemInformation(username, architecture));
```
gregg-miskelly, line 74: >new SystemInformation(username, architecture) [](http://example.com/codeflow?start=58&length=45)

If I read the code correctly, `null` would actually be passed in this case.

gregg-miskelly: Oh, I see. In SSH, you have it right. In Pipe, it would be null. We should probably pick one.

src/SSHDebugPS/PipeConnection.cs
```diff
@@ -108,35 +108,50 @@ public override bool IsLinux()
...
+            systemInformation = new SystemInformation(username, architecture);
+            return true;
         }
 
         public override List<Process> ListProcesses()
         {
-            string username;
-            TryGetUsername(out username);
+            SystemInformation systemInformation;
+            TryGetSystemInformation(out systemInformation);
```
gregg-miskelly, line 145: >TryGetSystemInformation [](http://example.com/codeflow?start=12&length=23)

Suggestion: `if (!...) { systemInformation = new SystemInformation(string.Empty, string.Empty); }`

gregg-miskelly (approved):
Otherwise LGTM

src/SSHDebugTests/PSOutputParserTests.cs
gregg-miskelly: Added a suggestion for how to fix this.

src/SSHDebugPS/PipeConnection.cs
WardenGnaw: Changed it to always return a system information, however it can return `new SystemInformation(string.Empty, string.Empty);`

src/SSHDebugTests/PSOutputParserTests.cs
WardenGnaw: Kept it to be similar to `new SystemInformation(string.Empty, string.Empty);`

gregg-miskelly (approved):

PR #1075, opened 2020-11-04
Convert MIDebugEngine to use SDK csproj

WardenGnaw:
This PR converts all the C# projects to use the new styled [sdk projects](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/overview).

Removed prebuild.csproj. The IL used for each specific project is added and is generated by the target which will run before resolving References.

| Status | Task |
| --- | :-- |
| ✔️ | Analyzers |
| ✔️ | Validate MicroBuild.Core |
| ✔️ | Fix Build Pipelines | 
| ✔️| Fix ` The literal or constant value <...> should be passed as the  'expected' argument in the call to 'Assert.Equal(expected, actual)' ` in tests |
| ✔️ | Fix MIDebugPackage.pkgdef Generation | 
| ✔️|  VSIX generation for SSHDebugPS |

TODO List:
| Status | Task |
| --- | :-- |
| 😖 | Fix MIDebugPackagePackage Threading Issues |
| :x: |   Remove `<PropertyGroup Condition = $(Configuration) \| $(Platform) == Desktop.Debug\|AnyCPU>` |

PR #1140, opened 2021-05-11
Fix Symbol Settings

WardenGnaw:
In https://github.com/microsoft/MIEngine/pull/1075, MIDebugPackage and SSHDebugPS lost some symbol PropertySettings and defaulted their PDBs to be portable. This fixes that issue.

Also, removed the per project settings and just define `DebugSymbols`, `DebugSymbols`, `Optimize` and `DEBUG;TRACE` constants in `miengine.settings.targets`.


src/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.DesignTime.csproj
```diff
@@ -18,44 +18,33 @@
     <DocumentationFile>$(DropRootDir)\ReferenceAssemblies\$(AssemblyName).xml</DocumentationFile>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <DebugSymbols>false</DebugSymbols>
-    <DebugType>pdbonly</DebugType>
```
WardenGnaw, line 21: Does it matter if UnixPortSupplier is not always `pdbonly`?

gregg-miskelly: There is no real code in this module, so I believe that is why this is done.

PR #1141, opened 2021-05-12
Fix output buffering with IDebugUnixShellPort.BeginExecuteAsyncCommand

gregg-miskelly:
### Summary

This PR addresses a fairly wide ranging issue that this code base had with [IDebugUnixShellPort.BeginExecuteAsyncCommand](https://github.com/microsoft/MIEngine/blob/70bf82b1f42094e4f7d534a1f4cf8b65beb3ab82/src/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.cs#L43) and output buffering. The way things were supposed to work was that if `runInShell` was set to `true` then output would always be delivered one line at a time and without newline characters and if it was false then output characters would be delivered whenever the pipe decided to flush and raw characters would be given. However, this isn't how things really were -- we had some places specify `false` for `runInShell` even though they didn't actually handle non-line-based output, and we also had code paths were the raw output would be used even though the caller specified `true` for `runInShell`.

### Fixes along the way

* UnixShellPortTransport.cs: 
    * Stop dumping the output from `kill` into the stdout stream and instead of just adding it to the log
    * `kill` isn't available as a command without starting a shell in Docker
* CommandRunner.cs:
    * Stop using 'raw' output reading for stderr
    * Stop abusing the thread pool by using Task.Run for a very long running operation
    * Stop using Task.Wait to wait for I/O to be ready
    * Dispose the process
    * Don't null out `_outputReadLoopTask`/`_errorReadLoopTask` as that causes null derefs in `OnProcessExited`
    * Fix hang from calling `Process.HasExited` under a lock. This is not a trivial method and could deadlock with the reader thread
    * Wait for stderr to complete as well (we were just waiting for stdout)
* DockerAsyncCommand.cs: Fix crash that I saw when Close was called more than once

### Testing

- [X] WSL: .NET debugging: tested attach, break, inspection, detach, reattach and run-to-completion
- [X] WSL: GDB debugging: tested attach, break, inspection, detach, reattach and run-to-completion
- [x] Docker Local: .NET debugging: tested attach, break, inspection, detach, reattach and run-to-completion
- [x] Docker Local: GDB debugging: tested attach, break, inspection, detach, reattach and run-to-completion
- [x] Docker Remote: .NET debugging: tested attach, break, inspection, detach, reattach and run-to-completion
- [x] Docker Remote: GDB debugging: tested attach, break, inspection, detach, reattach and run-to-completion

### DIFF
--- a/src/SSHDebugPS/CommandRunner.cs
+++ b/src/SSHDebugPS/CommandRunner.cs
@@ -50,9 +50,9 @@ namespace Microsoft.SSHDebugPS
 
         protected ProcessStartInfo ProcessStartInfo { get; }
         private System.Diagnostics.Process _process;
-        private System.Threading.Tasks.Task _outputReadLoopTask;
-        private System.Threading.Tasks.Task _errorReadLoopTask;
-        private bool _hasExited = false;
+
+        private byte _readerCompleteCount = 0;
+        private bool _processExited = false;
 
         private StreamWriter _stdInWriter;
         private StreamReader _stdOutReader;
@@ -117,8 +117,8 @@ namespace Microsoft.SSHDebugPS
                 _stdOutReader = new StreamReader(_process.StandardOutput.BaseStream, new UTF8Encoding(encoderShouldEmitUTF8Identifier: false), false, BUFMAX);
                 _stdErrReader = new StreamReader(_process.StandardError.BaseStream, new UTF8Encoding(encoderShouldEmitUTF8Identifier: false), false, BUFMAX);
 
-                _outputReadLoopTask = ReadStdOutAsync(_stdOutReader);
-                _errorReadLoopTask = ReadLoopAsync(_stdErrReader, false, OnErrorReceived);
+                _ = ReadStdOutAsync(_stdOutReader);
+                _ = ReadLoopAsync(_stdErrReader, OnErrorReceived);
             }
         }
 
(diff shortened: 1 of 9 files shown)


src/SSHDebugPS/CommandRunner.cs
```diff
@@ -177,13 +201,16 @@ protected virtual void ReadLoop(StreamReader reader, CancellationToken token, bo
...
+                    action(line);
                 }
             }
             catch (Exception e)
             {
-                ReportException(e);
-                Dispose();
+                if (this.IsDisposeStarted) // ignore exceptions if we are disposing
+                {
+                    ReportException(e);
```
WardenGnaw, line 211: Is the `ReportException` supported to be in the `else` clause? I'm confused since the comment says `// ignore exceptions if we are disposing`.

WardenGnaw (approved):

src/SSHDebugPS/CommandRunner.cs
gregg-miskelly: I had my `if` condition backwards.

src/SSHDebugPS/CommandRunner.cs
```diff
@@ -206,31 +207,29 @@ private async Task ReadLoopAsync(StreamReader reader, bool checkForExitedProcess
...
             bool shouldClose = false;
+
             lock (_lock)
             {
-                if (!_hasExited && _process.HasExited)
-                {
-                    _hasExited = true;
-                    shouldClose = true;
-                }
+                Debug.Assert(_readerCompleteCount < 2, "OnReaderComplete not be called more than twice");
```
WardenGnaw, line 224: nit
```suggestion
                Debug.Assert(_readerCompleteCount < 2, "OnReaderComplete should not be called more than twice");
```

WardenGnaw (approved):

WardenGnaw:
🚢

PR #1146, opened 2021-05-24
Bug 1289718: Adding data breakpoint via Breakpoints Window results in..

gc46:
.."When '0x0000000000000000' changes (4 bytes)" when input is an expression

Before:
image

After:
image

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
@@ -9,6 +9,8 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
+using System.Text.RegularExpressions;
+using System.Globalization;
 
 namespace Microsoft.MIDebugEngine
 {
@@ -159,6 +161,33 @@ namespace Microsoft.MIDebugEngine
 
         #region IDebugPendingBreakpoint2 Members
 
+        public string EvalDependentExpression(string expr)
+        {
+            string val = null;
+            Task eval = Task.Run(async () =>
+            {
+                val = await _engine.DebuggedProcess.MICommandFactory.DataEvaluateExpression(expr, 0, 0);
+            });
+            eval.Wait();
+            return val;
+        }
+
+        static readonly Lazy<Regex> s_addressPattern = new Lazy<Regex>(() => new Regex(@"^(0x[0-9a-fA-F]+)\b"));
+
+        public string Address(string fullName)
+        {
+            // ask GDB to evaluate "&expression"
+            string command = "&(" + fullName + ")";
+            var result = EvalDependentExpression(command);
+            Match m = s_addressPattern.Value.Match(result);
+            if (m.Success)
+            {
+                return m.Captures[0].ToString();
+            }
+            string errorMessage = String.Format(CultureInfo.InvariantCulture, "Unexpected result {0} from evaluating {1}", result, command);
+            throw new UnexpectedMIResultException(_engine.DebuggedProcess.MICommandFactory.Name, "-data-evaluate-expression", errorMessage);
+        }
+
         // Binds this pending breakpoint to one or more code locations.
         int IDebugPendingBreakpoint2.Bind()
         {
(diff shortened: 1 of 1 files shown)


gc46:
Will clean up extraneous comments

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -239,16 +278,25 @@ int IDebugPendingBreakpoint2.Bind()
 
                                     break;
                                 case enum_BP_LOCATION_TYPE.BPLT_DATA_STRING:
-                                    string address = HostMarshal.GetDataBreakpointStringForIntPtr(_bpRequestInfo.bpLocation.unionmember3);
-                                    if (address.Contains(","))
+                                    string address = HostMarshal.GetDataBreakpointStringForIntPtr(_bpRequestInfo.bpLocation.unionmember3); // this returns "*test" and "&test" (expression, broken) or "0x7ffc7e9513f0" (address, working)
+                                    // test -- need to delete; insert something here
+                                    // need to check for 0x => if no 0x, get the address
+                                    if (address.Contains("0x"))
```
gregg-miskelly, line 284: >Contains [](http://example.com/codeflow?start=48&length=8)

You need to be much more strict about these checks.

I think what you want is something like:

```C#
// In AD7Property.cs:
static readonly Lazy<Regex> s_propertyDataBreakpointPattern = new Lazy<Regex>(() => new Regex(@"^(0x[0-9a-fA-F]+),.+"));

internal bool TryParsePropertyDataBreakpoint(string input, out string address)
{
     Match match = s_propertyDataBreakpointPattern.Value.Match(input);
     if (!match.Success)
     {
         address = null;
         return false;
     }
     
     address = match.Captures[0].ToString();
     return true;
}
```

```C#
// Top of the class
static readonly Lazy<Regex> s_addressPattern = new Lazy<Regex>(() => new Regex(@"^0x[0-9a-fA-F]+$"));

// Here
string dataBpExpression = HostMarshal.GetDataBreakpointStringForIntPtr(_bpRequestInfo.bpLocation.unionmember3);

if (AD7Property.TryParsePropertyDataBreakpoint(dataBpExpression, out string addressString))
{
    this.AddressId = dataBpExpression;
    _address = addressString;
}
else 
{
    this.AddressId = null;
    dataBpExpression = dataBpExpression.Trim();    

    if (s_addressPattern.Value.Match(dataBpExpression).Success)
    {
       _address = dataBpExpression;
    }
    else
    {
       _address = Address(dataBpExpression);
    }
}
```

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -159,6 +162,42 @@ public void SetError(AD7ErrorBreakpoint bpError, bool sendEvent = false)
 
         #region IDebugPendingBreakpoint2 Members
 
+        public string EvalDependentExpression(string expr)
+        {
+            string val = null;
+            Task eval = Task.Run(async () =>
+            {
```
gregg-miskelly, line 169: Why are you Task.Run`ing and then waiting?

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
gregg-miskelly: Ignore this for the second. I think our approach on this is wrong.

gregg-miskelly (changes requested):
First, let me make sure I understand the problem you are trying to fix.

Is the problem that if you do this:
1. Start debugging and get into break mode
2. Open the breakpoints window
3. New->Data Breakpoint
4. Type in something which isn't an address

Is that right?
And what you are seeing is that whatever the user typed in during the last step is being given to MIEngine as the return value from HostMarshal.GetDataBreakpointStringForIntPtr?

If so, while it probably would be good to some version of the change I suggested so that we do validate the dataBpExpression that we get from AD7, the bigger problem is that we should be getting an address here.

This is important because the engine doesn't know the current frame. An engine can kind of guess the current frame like you are doing here, but it shouldn't need to. So we should understand why the UI isn't translating the address expression. This could be a UI bug, or this could be a problem with some other part of MIEngine (ex: our memory contexts aren't sufficient).

gc46:
@gregg-miskelly, your understanding of the problem is correct. If the user inputs an expression other than an address, the corresponding address displayed is 0x0000000000000000. However, if the user inputs an address, the correct behavior is observed.

To address the return value from HostMarshal.GetDataBreakpointStringForIntPtr(), again, you are correct; if the user inputs "*test", the function returns "*test".

Thus, my current approach is to retrieve the address of the user-input expression if said expression is not an address; I use the return value of HostMarshal.GetDataBreakpointStringForIntPtr() to determine this.

Please let me know if you have any further questions. Thanks!

gc46:
In addition to the above, here is what I messaged Charlie on May 20, 2021:
image

gregg-miskelly:
@gc46 Thanks! I think this is a bug in the VS UI. I sent you an email yesterday with where I think the bug is.

gc46:
@gregg-miskelly - yep, I saw your email. Will look into it. Thank you again!

gc46:
@gregg-miskelly, to follow up on your email regarding SetDataLocation() in dbgbp.cpp, I set a breakpoint at the line you highlighted and found the following:

> location
> - DataFullName: null
> - DataExpression: 0x143bb73c L"test"
> - DataCount: 4
> 
> bstrDataExpression: 0x143bb73c L"test"
> 
> GetDebugger()->DoesCurrentStackFrameSupportMetric(metricDataBPAsString): true
> 
> currentLocation
> - DataFullName: null
> - DataExpression: null
> - DataCount: 4
> 
> updatedLocation
> - DataFullName: null
> - DataExpression: 0x13ee0944 L"test"
> - DataCount: 4

Notice that updatedLocation.DataFullName is null; I will investigate this. Furthermore, there is no need to modify the if-condition as DataExpression is already populated and correct. Thanks!

PR #1152, opened 2021-06-03
Merge VS 2022 changes to GitHub

WardenGnaw:
This PR changes MIEngine so that it will only run on VS 2022 or newer. This includes referencing the 17.0 preview 1 (`17.0.0-previews-1-31410-258` for most packages) version of Visual Studio reference assemblies, and updating the file version to 17.0.


metadata.json
```diff
@@ -1,5 +1,5 @@
 {
   "correlationId": "595D6107-E618-4BD9-85A0-C323EFC400D1",
   "owner":"miengine",
-  "version": "16.5"
+  "version": "17.5"
```
gregg-miskelly, line 4: >17.5 [](http://example.com/codeflow?start=14&length=4)

nit: This file could probably be deleted, but this version number is wrong. Should be '17.0'.

eng/pipelines/templates/VS-release.template.yml
```diff
@@ -4,6 +4,10 @@ steps:
 
 - template: ../tasks/NuGetToolInstaller.yml
 
+- task: NuGetAuthenticate@0
+  inputs:
+    forceReinstallCredentialProvider: true
```
gregg-miskelly, line 9: Do we want to keep this?

src/DebugEngineHost.Stub/DebugEngineHost.Stub.csproj
```diff
@@ -20,7 +20,7 @@
     of DebugEngineHost, so we don't want to conflict with the real version. -->
     <OutputPath>bin\$(Configuration)</OutputPath>
     <DocumentationFile>$(DropRootDir)\ReferenceAssemblies\$(AssemblyName).xml</DocumentationFile>
-    <TargetFramework>net462</TargetFramework>
+    <TargetFramework>net472</TargetFramework>
```
gregg-miskelly, line 23: >net472 [](http://example.com/codeflow?start=21&length=6)

You mentioned that it was difficult to stay on net462 because of the threading analyzer. Is that being used on any of the projects that we want to use with VS Code?

eng/pipelines/templates/VS-release.template.yml
WardenGnaw: Yep. These are running on Microbuild ScaleSet machines.

src/DebugEngineHost.Stub/DebugEngineHost.Stub.csproj
WardenGnaw: Ah, the issue might be that we are loading Analyzers for all projects via https://github.com/microsoft/MIEngine/blob/main/build/Analyzers.targets

eng/pipelines/templates/VS-release.template.yml
gregg-miskelly: Right, but our nuget feed is nuget.org, so we don't need authentication

WardenGnaw: Ah, good point.

src/MIDebugEngine.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28315.86
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31427.69
```
gregg-miskelly, line 4: nit: change to `17.65535.65535.65535`

src/MakePIAPortable/MakePIAPortable.cs
```diff
@@ -37,6 +37,47 @@ class Program
             { "System.ValueType", System_Runtime },
             { "System.Collections.Generic.IEnumerable", System_Runtime },
             { "System.Reflection.DefaultMemberAttribute", System_Runtime },
+            { "System.Reflection.AssemblyDelaySignAttribute", System_Runtime},
+            { "System.Runtime.CompilerServices.CompilationRelaxationsAttribute", System_Runtime },
+            { "System.Runtime.CompilerServices.RuntimeCompatibilityAttribute", System_Runtime },
+            { "System.Diagnostics.DebuggableAttribute", System_Runtime },
+            { "System.Reflection.AssemblyTitleAttribute", System_Runtime },
+            { "System.Runtime.Versioning.TargetFrameworkAttribute", System_Runtime },
+            { "System.Resources.SatelliteContractVersionAttribute", System_Runtime },
```
gregg-miskelly, line 46: This isn't in netstandard1.3 system.runtime.

gregg-miskelly: It should be in System.Resources.ResourceManager

src/MakePIAPortable/MakePIAPortable.cs
```diff
@@ -37,6 +37,47 @@ class Program
...
+            { "System.Reflection.AssemblyCompanyAttribute", System_Runtime },
+            { "System.Reflection.AssemblyConfigurationAttribute", System_Runtime },
+            { "System.Reflection.AssemblyCopyrightAttribute", System_Runtime },
+            { "System.Reflection.AssemblyFileVersionAttribute", System_Runtime },
+            { "System.Reflection.AssemblyInformationalVersionAttribute", System_Runtime },
+            { "System.Reflection.AssemblyProductAttribute", System_Runtime },
+            { "System.Runtime.InteropServices.ComVisibleAttribute", System_Runtime },
+            { "System.Runtime.InteropServices.ComSourceInterfacesAttribute", System_Runtime },
+            { "System.Runtime.InteropServices.ComEventInterfaceAttribute", System_Runtime },
```
gregg-miskelly, line 55: Move to System.Runtime.InteropServices

src/MakePIAPortable/MakePIAPortable.cs
```diff
@@ -37,6 +37,47 @@ class Program
...
+            { "System.Runtime.CompilerServices.CompilerGeneratedAttribute", System_Runtime },
+            { "System.MulticastDelegate", System_Runtime },
+            { "System.IAsyncResult", System_Runtime },
+            { "System.AsyncCallback", System_Runtime },
+            { "System.IDisposable", System_Runtime },
+            { "System.Runtime.InteropServices.ComTypes.IConnectionPointContainer", System_Runtime },
```
gregg-miskelly, line 61: Move to System.Runtime.InteropServices

src/MakePIAPortable/MakePIAPortable.cs
```diff
@@ -37,6 +37,47 @@ class Program
...
+            { "System.Collections.ArrayList", System_Runtime },
```
gregg-miskelly, line 62: >{ "System.Collections.ArrayList", System_Runtime }, [](http://example.com/codeflow?start=12&length=51)

This isn't in .net standard 1.3. Do we need it?

src/MakePIAPortable/MakePIAPortable.cs
```diff
@@ -37,6 +37,47 @@ class Program
...
+            { "System.Runtime.InteropServices.ComTypes.IConnectionPoint", System_Runtime },
```
gregg-miskelly, line 63: >{ "System.Runtime.InteropServices.ComTypes.IConnectionPoint", System_Runtime }, [](http://example.com/codeflow?start=12&length=79)

Move to System.Runtime.InteropServices

src/MakePIAPortable/MakePIAPortable.cs
```diff
@@ -37,6 +37,47 @@ class Program
...
+            { "System.Runtime.CompilerServices.RuntimeHelpers", System_Runtime },
+            { "System.Array", System_Runtime },
+            { "System.RuntimeFieldHandle", System_Runtime },
+            { "System.Threading.Monitor", System_Runtime },
```
gregg-miskelly, line 67: >{ "System.Threading.Monitor", System_Runtime }, [](http://example.com/codeflow?start=12&length=47)

This is in System.Threading

src/MakePIAPortable/MakePIAPortable.cs
```diff
@@ -37,6 +37,47 @@ class Program
...
+            { "System.Runtime.InteropServices.Marshal", System_Runtime },
```
gregg-miskelly, line 68: >{ "System.Runtime.InteropServices.Marshal", System_Runtime }, [](http://example.com/codeflow?start=12&length=61)

Move to System.Runtime.InteropServices

src/MakePIAPortable/MakePIAPortable.cs
```diff
@@ -37,6 +37,47 @@ class Program
...
+            { "System.GC", System_Runtime },
+            { "System.Exception", System_Runtime },
+            { "System.Collections.IEnumerable", System_Runtime },
+            { "System.Collections.IEnumerator", System_Runtime },
+            { "System.DateTime", System_Runtime },
+            { "System.ObsoleteAttribute", System_Runtime },
+            { "System.Runtime.InteropServices.InAttribute", System_Runtime },
+            { "System.Runtime.InteropServices.ComTypes.ITypeLib", System_Runtime },
+            { "System.Runtime.InteropServices.TypeIdentifierAttribute", System_Runtime },
```
gregg-miskelly, line 77: Move to System.Runtime.InteropServices

src/MakePIAPortable/MakePIAPortable.cs
```diff
@@ -37,6 +37,47 @@ class Program
...
+            { "System.Runtime.InteropServices.TypeLibFuncAttribute", System_Runtime },
```
gregg-miskelly, line 78: This isn't in .net standard 1.3. Do we need it?

src/MakePIAPortable/MakePIAPortable.cs
```diff
@@ -37,6 +37,47 @@ class Program
...
+            { "System.Runtime.InteropServices.TypeLibFuncFlags", System_Runtime },
```
gregg-miskelly, line 79: >{ "System.Runtime.InteropServices.TypeLibFuncFlags", System_Runtime }, [](http://example.com/codeflow?start=12&length=70)

Same

src/MakePIAPortable/MakePIAPortable.cs
gregg-miskelly: My guess is we would just want to stip out these two.

src/MakePIAPortable/MakePIAPortable.cs
gregg-miskelly: I think we may need to strip out any `_EventProvider` classes to avoid this.

gregg-miskelly: Or maybe we could not process System.VisualStudio.Interop? I don't think we target anything against .net standard 1.3 anymore. So maybe everything would work if we just used the net45 version of the assembly? I am not sure if mono gets upset about COM interop stuff the same way that CoreCLR does.

gregg-miskelly: Actually, since our goal is to eventually move to CoreCLR, and there is probably a non-trivial amount of work to skip processing System.VisualStudio.Interop, unless it proves difficult to strip out the classes that use ArrayList that is probably the best option.

src/MakePIAPortable/MakePIAPortable.cs
gregg-miskelly: You could probably also just drop this attribute on the floor since we will never have a satellite dll.

src/MakePIAPortable/MakePIAPortable.cs
WardenGnaw: Wrote code in the MakePIAPortable to skip the whole class if we see it contains `_EventProvider`. Let me know if this was what you were thinking.

WardenGnaw: Or do we want to explicitly state which classes to skip.

src/MakePIAPortable/MakePIAPortable.cs
```diff
@@ -291,9 +299,54 @@ private static bool IsIgnoredCustomAttributeLine(string line)
...
+        {
+            int startLineNumber = inputFile.LineNumber;
+
+            // Find first open brace for class.
+            while (true)
+            {
+                string nextLine = inputFile.ReadLine();
+                if (nextLine == null)
+                {
+                    Error.Emit(Error.Code.BadCustomAttribute, inputFile, "Unexpected end-of-file while looking for start of class brace starting on line {0}", startLineNumber);
```
gregg-miskelly, line 318: break;

src/MakePIAPortable/MakePIAPortable.cs
```diff
@@ -247,11 +296,68 @@ private static bool IsIgnoredCustomAttributeLine(string line)
...
+            int braceCount = 1;
+
+            // Look for end class brace.
+            while (true)
+            {
+                string nextLine = inputFile.ReadLine();
+                if (nextLine == null)
+                {
+                    Error.Emit(Error.Code.BadCustomAttribute, inputFile, "Unexpected end-of-file while looking for end of class starting on line {0}", startLineNumber);
+                }
```
gregg-miskelly, line 334: >} [](http://example.com/codeflow?start=16&length=1)

Here too

gregg-miskelly (approved):
Otherwise LGTM

src/MakePIAPortable/MakePIAPortable.cs
gregg-miskelly: LGTM

PR #1153, opened 2021-06-03
Add Telemetry for FrameworkVersion

WardenGnaw:
This PR will add Telemetry to see which version of .NET customers are using to help determine if a majority of customers are still on v4.6.2.

### DIFF
--- a/src/OpenDebugAD7/Telemetry/DebuggerTelemetry.cs
+++ b/src/OpenDebugAD7/Telemetry/DebuggerTelemetry.cs
@@ -42,12 +42,10 @@ namespace OpenDebugAD7
         private const string TelemetryEngineVersion = "EngineVersion";
         private const string TelemetryHostVersion = "HostVersion";
         private const string TelemetryAdapterId = "AdapterId";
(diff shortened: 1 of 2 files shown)


src/OpenDebugAD7/Telemetry/DebuggerTelemetry.cs
```diff
@@ -210,6 +213,7 @@ public void SendTelemetryEvent(string eventName, Dictionary<string, object> even
             properties[TelemetryEngineVersion] = _engineVersion;
             properties[TelemetryHostVersion] = _hostVersion;
             properties[TelemetryAdapterId] = _adapterId;
+            properties[TelemetryFrameworkVersion] = _frameworkVersion;
```
gregg-miskelly, line 216: >properties[TelemetryFrameworkVersion] = _frameworkVersion; [](http://example.com/codeflow?start=12&length=58)

We might want to be a bit more choosy on what events we add this property -- seems unnecessary to have this on all events.

gregg-miskelly (approved):
Otherwise LGTM

gregg-miskelly (approved):

PR #1157, opened 2021-06-30
Add WSL port supplier

gregg-miskelly:
This PR adds a port supplier for WSL. This enables attaching to processes running in a WSL distro for both .NET (vsdbg) and GDB debugging.

I also cleaned up the SSHDebugPS code a bit:
* Connection now has two overloads of ExecuteCommand that do one of the following. Fixed most call sites to use whichever one they were intending:
     * Throw if the exit code is non-zero
     -or-
     * Return the exit code
* Connection no longer needlessly implements IDebugUnixShellPort
* Added a much cleaner local command runner

### Testing

Verified both native and .NET attach worked to an Ubuntu 20.04 WSL image.


WardenGnaw (approved):
:shipit:

PR #1169, opened 2021-07-16
implement step granularity

Trass3r:
https://microsoft.github.io/debug-adapter-protocol/specification#Types_SteppingGranularity

Also see https://github.com/microsoft/vscode/issues/102236.


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -1302,12 +1304,24 @@ protected override void HandleContinueRequestAsync(IRequestResponder<ContinueArg
...
         {
-            StepInResponse response = new StepInResponse();
+            try
+            {
+                var granularity = responder.Arguments.Granularity.GetValueOrDefault();
+                StepInternal(responder.Arguments.ThreadId, enum_STEPKIND.STEP_INTO, granularity, AD7Resources.Error_Scenario_Step_In);
+                responder.SetResponse(new StepInResponse());
+            }
+            catch (AD7Exception e)
+            {
```
gregg-miskelly, line 1314: >{ [](http://example.com/codeflow?start=12&length=1)

Shouldn't this have `responder.SetError(new ProtocolException(e.Message));`?

gregg-miskelly (changes requested):
Otherwise LGTM

src/OpenDebugAD7/AD7DebugSession.cs
Trass3r: Thx done!

gregg-miskelly (approved):

WardenGnaw (approved):

WardenGnaw:
Not sure why the checks are not running. Investigating.

Trass3r:
Indeed they haven't been run for a while.

WardenGnaw:
It seems like AzureDevOps is not correctly working with branches not in the Microsoft/MIEngine repo. I'll see if we can migrate to GitHub Actions which seems to be working correctly.

PR #1174, opened 2021-07-19
Fix all Build Warnings

WardenGnaw:
- [X] Validate SSH Attach To Process
- [X] Validate Docker Attach to Process
- [X] Validate WSL Attach to Process
- [X] Validate Output Window Logs

### DIFF
--- a/src/MIDebugPackage/MIDebugPackagePackage.cs
+++ b/src/MIDebugPackage/MIDebugPackagePackage.cs
@@ -15,6 +15,7 @@ using System.IO;
 using Microsoft.MIDebugEngine;
 using System.Collections.Generic;
 using MICore;
+using System.Threading.Tasks;
 
 namespace Microsoft.MIDebugPackage
 {
@@ -66,14 +67,17 @@ namespace Microsoft.MIDebugPackage
         /// </summary>
         protected override void Initialize()
         {
+            ThreadHelper.ThrowIfNotOnUIThread();
             base.Initialize();
 
             _packageCommandTarget = GetService(typeof(IOleCommandTarget)) as IOleCommandTarget;
+            Assumes.Present(_packageCommandTarget);
         }
         #endregion
 
         int IOleCommandTarget.Exec(ref Guid cmdGroup, uint nCmdID, uint nCmdExecOpt, IntPtr pvaIn, IntPtr pvaOut)
         {
+            ThreadHelper.ThrowIfNotOnUIThread();
             if (cmdGroup == GuidList.guidMIDebugPackageCmdSet)
             {
                 switch (nCmdID)
@@ -98,6 +102,7 @@ namespace Microsoft.MIDebugPackage
 
         int IOleCommandTarget.QueryStatus(ref Guid cmdGroup, uint cCmds, OLECMD[] prgCmds, IntPtr pCmdText)
         {
+            ThreadHelper.ThrowIfNotOnUIThread();
             if (cmdGroup == GuidList.guidMIDebugPackageCmdSet)
             {
                 switch (prgCmds[0].cmdID)
@@ -134,6 +139,7 @@ namespace Microsoft.MIDebugPackage
 
         private int LaunchMIDebug(uint nCmdExecOpt, IntPtr pvaIn, IntPtr pvaOut)
         {
+            ThreadHelper.ThrowIfNotOnUIThread("Microsoft.MIDebugPackage.LaunchMIDebug");
             int hr;
 
             if (IsQueryParameterList(pvaIn, pvaOut, nCmdExecOpt))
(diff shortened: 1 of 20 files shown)


src/MIDebugPackage/MIDebugPackagePackage.cs
```diff
@@ -306,9 +321,14 @@ private int MIDebugLog(uint nCmdLogOpt, IntPtr pvaIn, IntPtr pvaOut)
             return 0;
         }
 
-        private async void MIDebugExecAsync(string command)
+        private async Task MIDebugExecAsync(string command)
```
gregg-miskelly, line 324: >Task [](http://example.com/codeflow?start=22&length=4)

Is there a reason you did this instead of just throwing if called off of the UI thread?

WardenGnaw: This was due to
```
##[warning]src\MIDebugPackage\MIDebugPackagePackage.cs(309,28): Warning VSTHRD100: Avoid "async void" methods, because any exceptions not handled by the method will crash the process.
C:\a\1\s\src\MIDebugPackage\MIDebugPackagePackage.cs(309,28): warning VSTHRD100: Avoid "async void" methods, because any exceptions not handled by the method will crash the process. [C:\a\1\s\src\MIDebugPackage\MIDebugPackage.csproj]

```

src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -31,6 +31,7 @@ public static DockerConnection GetDockerConnection(string name, bool supportSSHC
             DockerContainerTransportSettings settings;
             Connection remoteConnection;
 
+            ThreadHelper.ThrowIfNotOnUIThread();
```
gregg-miskelly, line 34: >ThreadHelper.ThrowIfNotOnUIThread(); [](http://example.com/codeflow?start=12&length=36)

I am not sure this will always be called on the UI thread. Maybe it is true, but it is difficult to prove because the result of this is cached in the port object.

What was the thing complaining that caused you to add it?

src/MIDebugPackage/MIDebugPackagePackage.cs
gregg-miskelly: Ah. I missed that this was async before.

gregg-miskelly (approved):

src/SSHDebugPS/ConnectionManager.cs
WardenGnaw:
```
##[warning]src\SSHDebugPS\ConnectionManager.cs(80,111): Warning VSTHRD010: Accessing "ServiceProvider" should only be done on the main thread. Call Microsoft.VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread() first.
C:\a\1\s\src\SSHDebugPS\ConnectionManager.cs(80,111): warning VSTHRD010: Accessing "ServiceProvider" should only be done on the main thread. Call Microsoft.VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread() first. [C:\a\1\s\src\SSHDebugPS\SSHDebugPS.csproj]
##[warning]src\SSHDebugPS\ConnectionManager.cs(80,96): Warning VSTHRD010: Accessing "ServiceProvider" should only be done on the main thread. Call Microsoft.VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread() first.
C:\a\1\s\src\SSHDebugPS\ConnectionManager.cs(80,96): warning VSTHRD010: Accessing "ServiceProvider" should only be done on the main thread. Call Microsoft.VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread() first. [C:\a\1\s\src\SSHDebugPS\SSHDebugPS.csproj]
```

WardenGnaw: Ah, should I just suppress it then with a comment?

gregg-miskelly: I think your fix is better than what we had before, at least if you aren't seeing any issues when you test. It does seem better to throw here instead of potentially asking the service provider for the connection manager on a background thread.

We don't have the ideal design here, but hopefully your fix is good enough.

WardenGnaw:
@gregg-miskelly Any other scenarios to double check to see if it throws if its not on UI thread?

gregg-miskelly:
> @gregg-miskelly Any other scenarios to double check to see if it throws if its not on UI thread?

Do you mean besides SSH/Docker attaches? If so: nope, that is the only thing I was worried about.

WardenGnaw:
Yep. C++ and C# SSH/Docker/WSL don't throw any errors.

Tested via F5 MIDebugPackage with the MIDebugPackage.dll, SSHDebugPS.dll modules loaded and hit breakpoints to double verify.

gregg-miskelly:
:shipit:

PR #1180, opened 2021-07-22
Enable InvariantGlobalization for OpenDebugAD7

WardenGnaw:
Starting OpenDebugAD7 on Alpine will cause the error message:
> Process terminated. Couldn't find a valid ICU package installed on the system.

OpenDebugAD7 does not support Localization so we will enable InvariantGlobalization.

### DIFF
--- a/src/OpenDebugAD7/OpenDebugAD7.csproj
+++ b/src/OpenDebugAD7/OpenDebugAD7.csproj
@@ -20,8 +20,10 @@
     <!--
       For non-Windows platforms, .NET Core depends on libicu for data about locales and international settings.
       GDB does not support localization so neither does OpenDebugAD7 or MICore.
+
+      Condition: If we are building on linux, set invariant globalization. If we are publishing for linux (non-Windows and non-macOS), also set invariant globalization.
     -->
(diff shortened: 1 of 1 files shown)


src/OpenDebugAD7/OpenDebugAD7.csproj
```diff
@@ -17,6 +17,12 @@
     <TargetFramework>net5.0</TargetFramework>
     <Prefer32Bit>false</Prefer32Bit>
 
+    <!--
+      For non-Windows platforms, .NET Core depends on libicu for data about locales and international settings.
+      GDB does not support localization so neither does OpenDebugAD7 or MICore.
+    -->
+    <InvariantGlobalization Condition="!$([MSBuild]::IsOSPlatform('windows'))">true</InvariantGlobalization>
```
gregg-miskelly, line 24: AFAIK this is just a Linux problem (it would work on macOS). So up to you if you want to make this conditional to `!Windows` or `Linux`

gregg-miskelly (approved):
Otherwise LGTM

gregg-miskelly (approved):

PR #1181, opened 2021-07-27
Added PostRemoteConnectCommands

gcampbell-msft:
Added these in order to allow for commands that could be executed after
the connection to the debugger (i.e to gdb) has been made. For example, flashing the
board and/or resetting the board (at the correct time).


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -824,7 +826,7 @@ private async Task<List<LaunchCommand>> GetInitializeCommands()
                         if (!string.IsNullOrWhiteSpace(destination))
                         {
                             commands.Add(new LaunchCommand("-target-select remote " + destination, string.Format(CultureInfo.CurrentCulture, ResourceStrings.ConnectingMessage, destination)));
-
+                            commands.AddRange(_launchOptions.PostRemoteConnectCommands);
```
paulmaybee, line 829: Looks like you have attach and local launch covered, but aren't you missing a case, the non-local (e.g. pipe) launch options?

gcampbell-msft: I think you're right, I relocated a commands.AddRange call to correct this. Let me know if that doesn't fix what you're referring to.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -824,13 +826,14 @@ private async Task<List<LaunchCommand>> GetInitializeCommands()
                         if (!string.IsNullOrWhiteSpace(destination))
                         {
                             commands.Add(new LaunchCommand("-target-select remote " + destination, string.Format(CultureInfo.CurrentCulture, ResourceStrings.ConnectingMessage, destination)));
-
+
```
benmcmorran, line 829: If I'm reading the diff right, this line has a lot of trailing whitespace.

gregg-miskelly (approved):

PR #1198, opened 2021-08-13
Publish win-x86 and remove prefer32bit

WardenGnaw:
Remove win7-x86 and win-x64 publish and only publish win-x86 since
OpenDebugAD7 will just run as 32-bit.

### DIFF
--- a/eng/pipelines/steps/PublishOpenDebugAD7.yml
+++ b/eng/pipelines/steps/PublishOpenDebugAD7.yml
@@ -6,15 +6,17 @@ parameters:
 steps:
 - script: |
     dotnet publish $(Build.SourcesDirectory)\src\OpenDebugAD7\OpenDebugAD7.csproj -c ${{ parameters.Configuration }} -r ${{ parameters.RuntimeID }} --self-contained -o $(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin
+    copy ${{ parameters.SignedBinariesFolder }}\Release\vscode\OpenDebugAD7.dll "$(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin\.
     copy ${{ parameters.SignedBinariesFolder }}\Release\vscode\Microsoft.DebugEngineHost.dll $(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin\.
     copy ${{ parameters.SignedBinariesFolder }}\Release\Microsoft.MIDebugEngine.dll $(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin\.
     copy ${{ parameters.SignedBinariesFolder }}\Release\Microsoft.MICore.dll $(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin\.
   displayName: "Publish OpenDebugAD7 ${{ parameters.RuntimeID }}"
 
-# Windows Steps for signing OpenDebugAD7.dll and OpenDebugAD7.exe, copying over WindowsDebugLauncher.exe, and verify the windows binaries.
+# Windows Steps for signing OpenDebugAD7.exe, copying over WindowsDebugLauncher.exe, and verify the windows binaries.
 - ${{ if startsWith(parameters.RuntimeID, 'win') }}:
   - template: ../tasks/MSBuild.yml
     parameters:
+      DisplayName: "Sign OpenDebugAD7.exe"
       solution: $(Build.SourcesDirectory)\src\OpenDebugAD7\OpenDebugAD7.csproj
       configuration: ${{ parameters.Configuration }}
       msbuildArguments: '/t:SignPublishedFiles /p:PublishPath=$(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin'
@@ -24,7 +26,7 @@ steps:
 
   - script: |
       copy ${{ parameters.SignedBinariesFolder }}\Release\vscode\WindowsDebugLauncher.exe $(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin\.
-    displayName: "Copy OpenDebugAD7.exe and WindowsDebugLauncher.exe"
+    displayName: "Copy WindowsDebugLauncher.exe"
 
   - template: ../tasks/SignVerify.yml
     parameters:
--- a/eng/pipelines/steps/PublishOpenDebugAD7.yml
+++ b/eng/pipelines/steps/PublishOpenDebugAD7.yml
@@ -6,16 +6,23 @@ parameters:
 steps:
 - script: |
     dotnet publish $(Build.SourcesDirectory)\src\OpenDebugAD7\OpenDebugAD7.csproj -c ${{ parameters.Configuration }} -r ${{ parameters.RuntimeID }} --self-contained -o $(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin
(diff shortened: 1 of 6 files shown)


gregg-miskelly:
FYI: I don't think `Prefer32Bit` does anything.

eng/pipelines/templates/VSCode-release.template.yml
```diff
@@ -1,6 +1,6 @@
 ---
 parameters:
-  rids: ["win7-x86", "win-x64", "win10-arm64", "osx-x64", "linux-x64", "linux-arm", "linux-arm64", "linux-musl-x64" ]
+  rids: ["win-x86", "win10-arm64", "osx-x64", "linux-x64", "linux-arm", "linux-arm64", "linux-musl-x64" ]
```
gregg-miskelly, line 3: Out of curiosity: is there actually a difference between 'win7-x86' and 'win-x86'? I would have thought that would result in exactly the same bits.

gregg-miskelly (approved):

eng/pipelines/templates/VSCode-release.template.yml
WardenGnaw: I was trying to figure out what the difference was from: https://docs.microsoft.com/en-us/dotnet/core/rid-catalog#windows-rids

But I'm guessing by marking it more specific, the runtime does some assumptions based on how specifc you get on the graph.
```
    win7-x64    win7-x86
       |   \   /    |
       |   win7     |
       |     |      |
    win-x64  |  win-x86
          \  |  /
            win
             |
            any
```

gregg-miskelly: If you diff the files you get from publishing the two, are there any differences?

WardenGnaw: There is no difference. However, this might be an issue during the release when it copies OpenDebugAD7.exe

https://github.com/microsoft/MIEngine/blob/d1eced214596b22613b0f90f09c67a3f0c1a457e/eng/pipelines/steps/PublishOpenDebugAD7.yml#L18-L19

My thoughts are that its copying over a 64-bit exe on top the of the 32-bit one and its expecting a different runtime?

gregg-miskelly: Ah yes, that definitely seems like it is going to be a problem. I wonder if there is a project setting to control which host exe you get?

gregg-miskelly: Maybe set `AppHostRuntimeIdentifier` or just `RuntimeIdentifiers` to `win-x86` when building OpenDebugAD7 on Windows and the runtime id isn't specified?

WardenGnaw: I'm working on a msbuild target that will sign `OpenDebugAD7.exe` since it will also be an issue for `win10-arm64`.

gregg-miskelly: Makes sense.

eng/pipelines/steps/PublishOpenDebugAD7.yml
```diff
@@ -6,16 +6,23 @@ parameters:
...
     dotnet publish $(Build.SourcesDirectory)\src\OpenDebugAD7\OpenDebugAD7.csproj -c ${{ parameters.Configuration }} -r ${{ parameters.RuntimeID }} --self-contained -o $(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin
-    copy ${{ parameters.SignedBinariesFolder }}\Release\vscode\OpenDebugAD7.dll "$(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin\.
     copy ${{ parameters.SignedBinariesFolder }}\Release\vscode\Microsoft.DebugEngineHost.dll $(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin\.
     copy ${{ parameters.SignedBinariesFolder }}\Release\Microsoft.MIDebugEngine.dll $(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin\.
     copy ${{ parameters.SignedBinariesFolder }}\Release\Microsoft.MICore.dll $(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin\.
   displayName: "Publish OpenDebugAD7 ${{ parameters.RuntimeID }}"
 
-# Windows Steps for copying over OpenDebugAD7.exe, WindowsDebugLauncher.exe, and verify the windows binaries.
+# Windows Steps for signing OpenDebugAD7.dll and OpenDebugAD7.exe, copying over WindowsDebugLauncher.exe, and verify the windows binaries.
```
gregg-miskelly, line 14: Is there a reason to sign OpenDebugAD7.**dll**? I would think we could still do that during the build

gregg-miskelly:
>     <DropSignedFile Condition="$([MSBuild]::IsOSPlatform('windows')) AND ('$(RuntimeIdentifier)' == '' OR $(RuntimeIdentifier.Contains('win')))" Include="$(OutputPath)\OpenDebugAD7.exe" />

Remove

---
Refers to: src/OpenDebugAD7/OpenDebugAD7.csproj:77 in 8e0f0b2. [](commit_id = 8e0f0b23ac4a2f38185da2f2487fef5edcadcfdd, deletion_comment = False)

eng/pipelines/steps/PublishOpenDebugAD7.yml
```diff
@@ -6,16 +6,23 @@ parameters:
...
+      configuration: ${{ parameters.Configuration }}
+      msbuildArguments: '/t:SignPublishedFiles /p:PublishPath=$(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin'
+      env: { 
+        "SIGN_TYPE": "$(SignType)" 
+      } 
+
   - script: |
-      copy ${{ parameters.SignedBinariesFolder }}\Release\vscode\OpenDebugAD7.exe $(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin\.
       copy ${{ parameters.SignedBinariesFolder }}\Release\vscode\WindowsDebugLauncher.exe $(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin\.
     displayName: "Copy OpenDebugAD7.exe and WindowsDebugLauncher.exe"
```
gregg-miskelly: >OpenDebugAD7.exe and [](http://example.com/codeflow?start=23&length=20)

Remove

eng/pipelines/steps/PublishOpenDebugAD7.yml
```diff
@@ -6,15 +6,17 @@ parameters:
 steps:
 - script: |
     dotnet publish $(Build.SourcesDirectory)\src\OpenDebugAD7\OpenDebugAD7.csproj -c ${{ parameters.Configuration }} -r ${{ parameters.RuntimeID }} --self-contained -o $(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin
+    copy ${{ parameters.SignedBinariesFolder }}\Release\vscode\OpenDebugAD7.dll "$(Build.StagingDirectory)\${{ parameters.RuntimeID }}\debugAdapters\bin\.
```
gregg-miskelly: Do you need this? I would have thought the publishing process would do this automatically.

gregg-miskelly (approved):

eng/pipelines/steps/PublishOpenDebugAD7.yml
WardenGnaw: Yes. The `dotnet publish` overwrites the signed `OpenDebugAD7.dll` that was built by msbuild and drops a new copy.

gregg-miskelly (approved):

PR #1208, opened 2021-08-28
Fix LogPoint Issue + Add IgnoreResponseOrder in DAR

WardenGnaw:
Customer reported in v1.6.0 that LogPoints stopped working and are treated like regular breakpoints.
This was due to IDebugBreakpointEvent2.EnumBreakpoints returning the
same IEnumDebugBoundBreakpoints2.

The fix was to save IDebugBoundBreakpoint2[] in the event and return a
new IEnumDebugBoundBreakpoints2 object per call.

Added a test for logpoints and fixed up ConsoleEvent to be OutputEvent
since thats what it should actually be.

Added flag `IgnoreResponseOrder` to allow looking at historical responses to find out if the expected response occurred before.


test/DebuggerTesting/OpenDebug/Events/OutputEvent.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public Body body = new Body();
+    }
+
+    public sealed class OutputEvent : Event<OutputEventValue>
+    {
+        public OutputEvent(string text, CategoryValue category) : base("output")
+        {
+            this.ExpectedResponse.body.category = GeCategory(category);
+            this.ExpectedResponse.body.output = text;
```
WardenGnaw, line 34: Is there a good way to make this a `Contains` check instead of a strict `string.equals`?

test/CppTests/Tests/BreakpointTests.cs
```diff
@@ -154,6 +154,48 @@ public void FunctionBreakpointsBasic(ITestSettings settings)
...
+
+                this.Comment("Set initial breakpoints");
+                runner.SetBreakpoints(callingBreakpoints);
+
+                this.Comment("Launch and run until first breakpoint");
+                runner.Expects.HitBreakpointEvent(SinkHelper.Calling, 48)
+                              .AfterConfigurationDone();
+
+                this.Comment("Set a breakpoint while in break mode");
+                callingBreakpoints.Add(52, null, "Log Message");
```
WardenGnaw, line 183: nit: move to variable and use for both lines 183 and 187.

test/DebuggerTesting/OpenDebug/Events/OutputEvent.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        private static string GeCategory(CategoryValue category)
```
WardenGnaw, line 37:
```suggestion
        private static string GetCategory(CategoryValue category)
```

test/DebuggerTesting/OpenDebug/Events/OutputEvent.cs
WardenGnaw: Another thing is, do I need a way to check the next `n` OutputEvents since the first one won't be the LogMessage?

gregg-miskelly: I am not sure if the Cpp tests stuff works differently, but I think this is just using the DAR stuff for comparing events. If so, all strings are treated as regular expressions.

gregg-miskelly (approved):

test/DebuggerTesting/OpenDebug/Events/OutputEvent.cs
WardenGnaw:
```
 Expected = {"body":{"output":"Log Message(\r\n|\r|\n)","category":"console"},"event":"output"}
 Actual Responses =
 1. {"event":"breakpoint","body":{"reason":"changed","breakpoint":{"id":2,"verified":true,"line":52,"BoundBreakpoints":[]}},"seq":40,"type":"event"}
 2. {"event":"output","body":{"category":"stdout","output":"Calling recursive function.\n"},"seq":42,"type":"event"}
 3. {"event":"output","body":{"category":"stdout","output":"\n"},"seq":43,"type":"event"}
 4. {"event":"output","body":{"category":"stdout","output":"Breakpoint 3, Calling::CoreRun (this=0x7fffffffda50) at /home/runner/work/MIEngine/MIEngine/bin/DebugAdapterProtocolTests/Debug/CppTests/debuggees/instance/kitchensink-3/src/calling.cpp:52\n"},"seq":44,"type":"event"}
 5. {"event":"output","body":{"category":"stdout","output":"52\t    this->Log(\"Calling variable arg function.\");\n"},"seq":45,"type":"event"}
 6. {"event":"output","body":{"category":"console","output":"Log Message\n"},"seq":46,"type":"event"}
 ```

I'm a bit confused why DAR is unhappy on linux even though the event is being shown. My thoughts were "do I need a way to check the next n OutputEvents since the first one won't be the LogMessage?" but it seems to work correctly on windows and macOS.

Comment (gregg-miskelly) [reply test/DebuggerTesting/OpenDebug/Events/OutputEvent.cs:34]: Does `"\\bLog Message\\b"` work? I am not sure why `\n` doesn't work. The only guess I have is that it is connected with the fact that we aren't in ["Single-line mode"](https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options#single-line-mode) and you are using `\n` instead of `\\n`. But this is a total guess.

Comment (gregg-miskelly) [reply test/DebuggerTesting/OpenDebug/Events/OutputEvent.cs:34]: I tried running the snippet on Linux, and I got the result we expect, so no idea. You might just have to debug it.

Comment (WardenGnaw) [reply test/DebuggerTesting/OpenDebug/Events/OutputEvent.cs:34]: Did you try this in WSL or a Linux VM? I was able to have it succeeded in WSL locally, but the test machine runs fail.

My next steps is running in a VM.

Comment (gregg-miskelly) [reply test/DebuggerTesting/OpenDebug/Events/OutputEvent.cs:34]: I was also using WSL. Did you try `"^Log Message\\b"`?

Comment (WardenGnaw) [reply test/DebuggerTesting/OpenDebug/Events/OutputEvent.cs:34]: Both `"Log Message(\r\n|\r|\n)"` and `"^Log Message\\b"` succeeded in WSL.

Comment (WardenGnaw) [reply test/DebuggerTesting/OpenDebug/Events/OutputEvent.cs:34]: Yep. Reproduces in a Linux VM.

File: test/DebugAdapterRunner/DebugAdapterCommand.cs [hunk 4]
```diff
@@ -140,10 +140,33 @@ public override void Run(DebugAdapterRunner runner)
...
             // Loop until we have received as many expected responses as expected
             while (currentExpectedResponseIndex < this.ExpectedResponses.Count)
             {
+                // Check if previous messages contained the expected response
+                if (previousExpectedResponseIndex != currentExpectedResponseIndex)
+                {
+                    DebugAdapterResponse expected = this.ExpectedResponses[currentExpectedResponseIndex];
+                    foreach (var response in responseList)
+                    {
+                        if (Utils.CompareObjects(expected.Response, response, expected.IgnoreOrder))
```
WardenGnaw, line 154: This will have an issue of accidently matching a single response to multiple expected responses.

test/DebugAdapterRunner/DebugAdapterCommand.cs
```diff
@@ -140,10 +140,33 @@ public override void Run(DebugAdapterRunner runner)
...
+                        {
+                            expected.Match = response;
+                            break;
+                        }
+                    }
+
+                    if (expected.Match != null)
+                    {
+                        currentExpectedResponseIndex++;
+                        break;
```
WardenGnaw, line 164: Oops, accidentally ignoring the rest of the responses expected. 

```suggestion
                        continue;
```

test/DebugAdapterRunner/DebugAdapterCommand.cs
```diff
@@ -138,12 +151,38 @@ public override void Run(DebugAdapterRunner runner)
...
+                // Check if previous messages contained the expected response
+                if (previousExpectedResponseIndex != currentExpectedResponseIndex)
+                {
+                    DebugAdapterResponse expected = this.ExpectedResponses[currentExpectedResponseIndex];
+                    foreach (ResponsePair responsePair in responseList)
+                    {
+                        // Make sure we have not seen this response and check to see if it the response we are expecting.
+                        if (!responsePair.Seen && Utils.CompareObjects(expected.Response, responsePair.Response, expected.IgnoreOrder))
+                        {
+                            expected.Match = responsePair.Response;
```
WardenGnaw, line 170: need to mark `responsePair.seen = true`

test/DebugAdapterRunner/DebugAdapterCommand.cs
```diff
@@ -138,12 +151,38 @@ public override void Run(DebugAdapterRunner runner)
...
             while (currentExpectedResponseIndex < this.ExpectedResponses.Count)
             {
+                // Check if previous messages contained the expected response
+                if (previousExpectedResponseIndex != currentExpectedResponseIndex)
+                {
+                    DebugAdapterResponse expected = this.ExpectedResponses[currentExpectedResponseIndex];
+                    foreach (ResponsePair responsePair in responseList)
+                    {
+                        // Make sure we have not seen this response and check to see if it the response we are expecting.
+                        if (!responsePair.Seen && Utils.CompareObjects(expected.Response, responsePair.Response, expected.IgnoreOrder))
```
WardenGnaw, line 168: Do we want to add another boolean flag to `DebugAdapterResponse` to be `IngoreResponseOrder` to only allow this for specifc cases / only for `event` types?

gregg-miskelly: I think so. Seems like there are cases when we would want to catch cases where we unintentionally re-oredered events.

gregg-miskelly:
>                     string expectedResponseText = JsonConvert.SerializeObject(this.ExpectedResponses[currentExpectedResponseIndex].Response);

To make these issues easier to debug in the future, I would suggest we change the expected format to be more like the actual.

Something like:

```
for (int i = 0; i < ExpectedResponses.Lenght; i++)
{
    string status;
    if (i < currentExpectedResponseIndex)
        status = "Found" ;
    else if (i == currentExpectedResponseIndex)
        status = "Not Found";
    else
        status = "Not searched yet";
    expectedResponseText += string.Format(CultureInfo.CurrentCulture, "{0}. ({1}) {1}\n", (i + 1), status, JsonConvert.SerializeObject(ExpectedResponses[i].Response));
}
```

---
Refers to: test/DebugAdapterRunner/DebugAdapterCommand.cs:242 in 7249fbd. [](commit_id = 7249fbd9e305e64f2e5714228b1a2ab5ae84b624, deletion_comment = False)

WardenGnaw:
> > ```
> >                 string expectedResponseText = JsonConvert.SerializeObject(this.ExpectedResponses[currentExpectedResponseIndex].Response);
> > ```
> 
> To make these issues easier to debug in the future, I would suggest we change the expected format to be more like the actual.
> 
> Something like:
> 
> ```
> for (int i = 0; i < ExpectedResponses.Lenght; i++)
> {
>     string status;
>     if (i < currentExpectedResponseIndex)
>         status = "Found" ;
>     else if (i == currentExpectedResponseIndex)
>         status = "Not Found";
>     else
>         status = "Not searched yet";
>     expectedResponseText += string.Format(CultureInfo.CurrentCulture, "{0}. {1}: {2}\n", (i + 1), status, JsonConvert.SerializeObject(ExpectedResponses[i].Response));
> }
> ```
> 
> Refers to: test/DebugAdapterRunner/DebugAdapterCommand.cs:242 in [7249fbd](https://github.com/microsoft/MIEngine/commit/7249fbd9e305e64f2e5714228b1a2ab5ae84b624). [](commit_id = [7249fbd](https://github.com/microsoft/MIEngine/commit/7249fbd9e305e64f2e5714228b1a2ab5ae84b624), deletion_comment = False)

Example Test failure log:
```
 DARException: Expected response not found before timeout.
 Expected = 1. Found: {"success":true,"command":"continue"}
 2. Not Found: {"body":{"output":"^Log Message\\b","category":"console"},"event":"output"}
 3. Not searched yet: {"body":{},"event":"exited"}
 4. Not searched yet: {"event":"terminated"}

 Actual Responses =
 1. {"Seen":false,"Response":{"event":"breakpoint","body":{"reason":"changed","breakpoint":{"id":2,"verified":true,"line":52,"BoundBreakpoints":[]}},"seq":40,"type":"event"}}
 2. {"Seen":false,"Response":{"event":"output","body":{"category":"stdout","output":"Calling recursive function.\n"},"seq":42,"type":"event"}}
 3. {"Seen":false,"Response":{"event":"output","body":{"category":"stdout","output":"\n"},"seq":43,"type":"event"}}
 4. {"Seen":false,"Response":{"event":"output","body":{"category":"stdout","output":"Breakpoint 3, Calling::CoreRun (this=0x7ffffffedd90) at /mnt/d/dd/MIEngine/bin/DebugAdapterProtocolTests/Debug/CppTests/debuggees/instance/kitchensink-3/src/calling.cpp:52\n"},"seq":44,"type":"event"}}
 5. {"Seen":false,"Response":{"event":"output","body":{"category":"stdout","output":"52\t    this->Log(\"Calling variable arg function.\");\n"},"seq":45,"type":"event"}}
 6. {"Seen":true,"Response":{"success":true,"message":null,"request_seq":6,"command":"continue","body":{},"running":false,"refs":null,"seq":41,"type":"response"}}
 7. {"Seen":false,"Response":{"event":"stopped","body":{"reason":"breakpoint","threadId":6195,"allThreadsStopped":true,"source":{"name":"calling.cpp","path":"/mnt/d/dd/MIEngine/bin/DebugAdapterProtocolTests/Debug/CppTests/debuggees/instance/kitchensink-3/src/calling.cpp","sources":[],"checksums":[]},"line":52,"column":1},"seq":46,"type":"event"}}

```

gregg-miskelly (approved):

PR #1209, opened 2021-08-30
Fix PlatformUtilities methods

WardenGnaw:
Use 'RuntimeInformation.IsOSPlatform' instead of
'Environment.OSVersion.Platform'

Environment.OSVersion.Platform on .NET 5 is returning 'Unix' for macOS
which is causing a bunch of our cases to not work.

Tested RuntimeInformation.IsOSPlatform(OSPlatform.Windows) on net472 on
Windows and it returns true.

### DIFF
--- a/src/MICore/PlatformUtilities.cs
+++ b/src/MICore/PlatformUtilities.cs
@@ -32,17 +32,19 @@ namespace MICore
 
         private static RuntimePlatform CalculateRuntimePlatform()
         {
-            switch (Environment.OSVersion.Platform)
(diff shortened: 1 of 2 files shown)


gregg-miskelly (approved):

WardenGnaw:
Investigating failing macOS tests. May be an indicator of other macOS issues.

WardenGnaw:
Disabling Tests since they seem to be a test machine only issue / lldb-mi external issue.

Choosing to complete this PR instead since without these changes, it causes the IsOSX checks to return false on macOS.

PR #1225, opened 2021-09-30
AutoComplete: Address non ` or -exec command

WardenGnaw:
This PR addresses completion commands so it works only for <code>\`</code> or `-exec`.  

This also moves the logic to be in MIDebugEngine instead of duplicating the command code also in OpenDebugAD7.


gregg-miskelly (approved):

src/MIDebugEngine/Engine.Impl/EngineUtils.cs
```diff
@@ -109,6 +109,33 @@ internal static bool ProcIdEquals(AD_PROCESS_ID pid1, AD_PROCESS_ID pid2)
...
+            string execCommandString = "-exec ";
+            if (command.StartsWith(execCommandString, StringComparison.Ordinal))
+            {
+                prefix = execCommandString;
+                strippedCommand = command.Substring(execCommandString.Length);
+                return true;
+            }
+            else if (command[0] == '`')
+            {
+                prefix = "`";
```
gregg-miskelly, line 132: >" [](http://example.com/codeflow?start=27&length=1)

Do we need to include the whitespace here also?

gregg-miskelly: In other words, instead of using TrimStart, do we need to check for whitespace, and make the cut off the first non-whitespace character?

src/MIDebugEngine/Engine.Impl/EngineUtils.cs
```diff
@@ -109,6 +109,33 @@ internal static bool ProcIdEquals(AD_PROCESS_ID pid1, AD_PROCESS_ID pid2)
...
+                strippedCommand = command.Substring(1).TrimStart(); // remove spaces if any
```
gregg-miskelly, line 133: >strippedCommand [](http://example.com/codeflow?start=16&length=15)

Should this check that there is something left after the whitespace?

src/MIDebugEngine/Engine.Impl/EngineUtils.cs
WardenGnaw: There is no whitespace needed here. The prefixes are `-exec <COMMAND>` and <code>`\<COMMAND\></code>

src/MIDebugEngine/Engine.Impl/EngineUtils.cs
```diff
@@ -109,6 +109,33 @@ internal static bool ProcIdEquals(AD_PROCESS_ID pid1, AD_PROCESS_ID pid2)
...
+            prefix = string.Empty;
+            strippedCommand = string.Empty;
+            string execCommandString = "-exec ";
+            if (command.StartsWith(execCommandString, StringComparison.Ordinal))
+            {
+                prefix = execCommandString;
+                strippedCommand = command.Substring(execCommandString.Length);
+                return true;
+            }
+            else if (command[0] == '`')
```
gregg-miskelly, line 130: >command [](http://example.com/codeflow?start=21&length=7)

Do we need to check that `command.Length > 0`?

src/MIDebugEngine/Engine.Impl/EngineUtils.cs
gregg-miskelly: Or does GDB handle that okay?

src/MIDebugEngine/Engine.Impl/EngineUtils.cs
WardenGnaw: Are you saying in the case of `        -exec <COMMAND>`, we should still autocomplete and return with the spaces prepended? 
```
        -exec ls
        -exec list
```

gregg-miskelly: Just to make sure my question is understood. I ask because the prefix is being returned to the client as part of the match. So you are saying that VS Code is happy if the input text is `` ` pw`` (note extra spaces) but the output is <code>`pwd</code> (note the spaces went away).

src/MIDebugEngine/Engine.Impl/EngineUtils.cs
WardenGnaw: GDB seems to be ok with extranous whitespace:
image

gregg-miskelly: Sorry, I mean is GDB happy if you ask for complete after `-exec` (no command typed yet)?

WardenGnaw: Ah, VS Code does not seem to send an autocomplete on a string ending in whitespace or `. We kind of lucked out there. 

But to be extra careful, do we want to do `TrimEnd`?

gregg-miskelly: What about `-exec`?

WardenGnaw: `-exec` does not match `-exec ` (with a space) 😅

gregg-miskelly: Ahh... Cool.

src/MIDebugEngine/Engine.Impl/EngineUtils.cs
WardenGnaw: Heres the results with TrimStart() removed. 

image

gregg-miskelly (approved):

PR #1231, opened 2021-10-18
Change whitespace characters to space

paulmaybee:
Auto's processing in recent main builds is passing the expression "()\r\n{\n" into the MIEngine at debugger launch. This was causing the MIEngine to pop-up a message from gdb: 'Undefined command: "" Try "help"'. Not sure if this is by-design, or a debugger bug. But now the MIEngine preprocesses expression strings and replaces white space characters with a simple space.

Changing all whatespace was too agressive, backed off to only disallow \r and \n.


src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -609,6 +610,18 @@ internal string EscapeQuotes(string str)
             return str.Replace("\"", "\\\"");
         }
 
+        internal string HandleInvalidChars(string str)
+        {
+            char[] invalidChars = { '\r', '\n' };
```
gregg-miskelly, line 615: Do we also need to remove `\0`? Or is that an error in your description?

paulmaybee: Not following you. Where did I mention \0? Or are you misreading () as 0?

PR #1234, opened 2021-10-20
make radix callback on calling thread

paulmaybee:
Not sure this is the right fix since this doesn't seem to be causing a problem in any other scenario. Perhaps ImageWatch is not calling into the engine correctly?

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -1255,7 +1255,7 @@ namespace Microsoft.MIDebugEngine
                 if (!string.IsNullOrEmpty(resultVar))
                 {
                     ReturnValue = new VariableInformation("$ReturnValue", resultVar, cxt, Engine, (AD7Thread)thread.Client, isParameter: false);
-                    await ReturnValue.Eval(Engine.CurrentRadix());
+                    await ReturnValue.Eval(radix: 0);
                 }
                 _callback.OnStepComplete(thread);
             }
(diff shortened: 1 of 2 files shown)


src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -493,11 +494,11 @@ public string EvalDependentExpression(string expr)
             return val;
         }
 
-        internal async Task Eval(enum_EVALFLAGS dwFlags = 0, DAPEvalFlags dwDAPFlags = 0)
+        internal async Task Eval(enum_EVALFLAGS dwFlags = 0, DAPEvalFlags dwDAPFlags = 0, uint radix = 0)
```
gregg-miskelly, line 497: A few suggestion:
1. We probably should make radix required
2. We should pass it from AsyncEval as well
3. If we pass zero, I am thinking instead of querying the UI we should just assume that it is already correct.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1255,7 +1255,7 @@ private async Task HandleBreakModeEvent(ResultEventArgs results, BreakRequest br
                 if (!string.IsNullOrEmpty(resultVar))
                 {
                     ReturnValue = new VariableInformation("$ReturnValue", resultVar, cxt, Engine, (AD7Thread)thread.Client, isParameter: false);
-                    await ReturnValue.Eval();
+                    await ReturnValue.Eval(Engine.CurrentRadix());
```
gregg-miskelly, line 1258: Probably should just pass zero here

gregg-miskelly (approved):
Otherwise LGTM

gregg-miskelly (approved):

PR #1235, opened 2021-10-21
Bug 1358033

gc46:


src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -556,7 +562,7 @@ public virtual IEnumerable<Guid> GetSupportedExceptionCategories()
         /// exceptions in the category. Note that this clear all previous exception breakpoints set in this category.</param>
         /// <param name="exceptionBreakpointState">Indicates when the exception breakpoint should fire</param>
         /// <returns>Task containing the exception breakpoint id's for the various set exceptions</returns>
-        public virtual Task<IEnumerable<ulong>> SetExceptionBreakpoints(Guid exceptionCategory, /*OPTIONAL*/ IEnumerable<string> exceptionNames, ExceptionBreakpointStates exceptionBreakpointState)
+        public virtual async Task<IEnumerable<ulong>> SetExceptionBreakpoints(Guid exceptionCategory, /*OPTIONAL*/ IEnumerable<string> exceptionNames, ExceptionBreakpointStates exceptionBreakpointState)
```
WardenGnaw, line 565: Just a heads up. `lldb-mi` does not support catchpoints. You will want to keep this method throwing Not Implemented exception and override this method in [gdb.cs](https://github.com/microsoft/MIEngine/blob/main/src/MICore/CommandFactories/gdb.cs)

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -569,12 +575,53 @@ public virtual Task<IEnumerable<ulong>> SetExceptionBreakpoints(Guid exceptionCa
...
+            {
+
+            }
+            */
+
+            return breakpointNumbers;
         }
 
-        public virtual Task RemoveExceptionBreakpoint(Guid exceptionCategory, IEnumerable<ulong> exceptionBreakpoints)
+        public virtual async Task RemoveExceptionBreakpoint(Guid exceptionCategory, IEnumerable<ulong> exceptionBreakpoints)
```
WardenGnaw, line 618: Same issue for `lldb-mi` as `SetExceptionBreakpoints`

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -543,7 +546,10 @@ public virtual async Task BreakCondition(string bkptno, string expr)
 
         public virtual IEnumerable<Guid> GetSupportedExceptionCategories()
         {
-            return new Guid[0];
+            // return new Guid[0];
+            List<Guid> categories = new List<Guid>();
+            categories.Add(new Guid("{3A12D0B7-C26C-11D0-B442-00A0244A1DD2}")); // maybe there is a better way than this...
```
gregg-miskelly, line 551: > // maybe there is a better way than this... [](http://example.com/codeflow?start=79&length=44)

```
const string CppExceptionCategoryString = "{3A12D0B7-C26C-11D0-B442-00A0244A1DD2}";

return new Guid[] { new Guid(CppExceptionCategoryString) };
```

gregg-miskelly:
>             // For C++, we have a bunch of C++ projection exceptions that we will not want to send down to GDB. Ignore these.

Should we reject all characters that have special meaning in RegEx:

`"\\^$.|?+()[]{}"`

---
Refers to: src/MIDebugEngine/Engine.Impl/ExceptionManager.cs:547 in 24188cb. [](commit_id = 24188cbf9362e461e41cb5509f3f711d81ba3ec2, deletion_comment = False)

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -569,12 +575,53 @@ public virtual Task<IEnumerable<ulong>> SetExceptionBreakpoints(Guid exceptionCa
...
+                Results result = await _debugger.CmdAsync(command, ResultClass.done);
+                var breakpointNumber = result.Find("bkpt").FindUint("number");
+                breakpointNumbers.Add((ulong)breakpointNumber);
+
+            }
+            else
+            {
+                foreach (string exceptionName in exceptionNames)
+                {
+                    string command = string.Format(CultureInfo.InvariantCulture, "-catch-throw {0}", exceptionName);
```
gregg-miskelly, line 596: >exceptionName [](http://example.com/codeflow?start=101&length=13)

Do we need to escape `*` in case the exception raised is a pointer?

src/MICore/CommandFactories/MICommandFactory.cs
WardenGnaw: @gregg-miskelly Do we need to add `ExceptionSettings` to [cppdbg.ad7Engine.json](https://github.com/microsoft/MIEngine/blob/main/src/OpenDebugAD7/cppdbg.ad7Engine.json)?

E.g.
```
"exceptionSettings": {
   "categories": [
      {
         "name": "CppExceptionCategory",
         "id": "3A12D0B7-C26C-11D0-B442-00A0244A1DD2"
      }
   ]
}
```

or should we change the code to ignore it for VS Code for now since we wont have the DAP side of it implemented for this feature?
https://github.com/microsoft/MIEngine/blob/8456c49b4731971b885c2909f9feaac4bdcd494d/src/DebugEngineHost.VSCode/HostConfigurationStore.cs#L55-L65

gregg-miskelly: Got it. Yes, it sounds like we need to add the place holder value that you have above.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -425,7 +425,7 @@ private async Task FlushSettingsUpdates()
             }
         }
 
-        private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings categorySettings, SettingsUpdates updates)
+        private async Task UpdateCategory(Guid categoryId, ExceptionCategorySettings categorySettings, SettingsUpdates updates)
         {
```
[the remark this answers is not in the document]
gc46: Resolving this conversation as gdb does not work with negative lookahead. Thanks!

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -544,7 +544,8 @@ public virtual async Task BreakCondition(string bkptno, string expr)
 
         public virtual IEnumerable<Guid> GetSupportedExceptionCategories()
         {
-            return new Guid[0];
+            const string CppExceptionCategoryString = "{3A12D0B7-C26C-11D0-B442-00A0244A1DD2}";
+            return new Guid[] {new Guid(CppExceptionCategoryString)};
```
gregg-miskelly, line 548: This should be in gdb.cs

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -434,11 +434,26 @@ private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings cat
...
+                    {
+                        await _commandFactory.RemoveExceptionBreakpoint(categoryId, new ulong[] { breakpointId });
+                    }
+                    if (categorySettings.CurrentRules.TryGetValue("winrt::hresult_error", out breakpointId))
+                    {
+                        await _commandFactory.RemoveExceptionBreakpoint(categoryId, new ulong[] { breakpointId });
+                    }
+                }
+
                 categorySettings.CurrentRules.Clear();
```
gregg-miskelly, line 451: >categorySettings.CurrentRules.Clear(); [](http://example.com/codeflow?start=16&length=38)

We don't want to do this until we deleted all the old rules.

gregg-miskelly: Oh, I guess that is what you were doing in the if block above.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -434,11 +434,26 @@ private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings cat
             {
                 ExceptionBreakpointStates newCategoryState = updates.NewCategoryState.Value;
                 categorySettings.CategoryState = newCategoryState;
+
+                if (newCategoryState != ExceptionBreakpointStates.BreakThrown)
+                {
```
gregg-miskelly, line 439: I think you want to do this regardless of what the new state is

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -434,11 +434,26 @@ private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings cat
...
+                    ulong breakpointId;
+                    if (categorySettings.CurrentRules.TryGetValue("*", out breakpointId))
+                    {
+                        await _commandFactory.RemoveExceptionBreakpoint(categoryId, new ulong[] { breakpointId });
+                    }
+                    if (categorySettings.CurrentRules.TryGetValue("winrt::hresult_error", out breakpointId))
+                    {
+                        await _commandFactory.RemoveExceptionBreakpoint(categoryId, new ulong[] { breakpointId });
+                    }
```
gregg-miskelly, line 448: Sorry if this is just prototype code and it is obvious, but I believe you want: `await _commandFactory.RemoveExceptionBreakpoint(categoryId, categorySettings.CurrentRules.Values);`

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -480,17 +495,32 @@ private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings cat
                     }
                 }
 
-                IEnumerable<ulong> breakpointIds = await _commandFactory.SetExceptionBreakpoints(categoryId, exceptionNames, grouping.Key);
-
-                int count = exceptionNames.Zip(breakpointIds, (exceptionName, breakpointId) =>
+                if (grouping.Key.ToString().Contains(ExceptionBreakpointStates.BreakThrown.ToString())) // only set catchpoints on BreakThrown - problematic when grouping.Key is BreakThrown | something
```
gregg-miskelly, line 498: >grouping.Key.ToString().Contains(ExceptionBreakpointStates.BreakThrown.ToString()) [](http://example.com/codeflow?start=20&length=82)

You don't want to turn things into a string to see if it contains a flag.

`grouping.Key.HasFlag(ExceptionBreakpointStates.BreakThrown)`

Though really we can get rid of this whole `.GroupBy` code now since we only care about rules with `grouping.Key.HasFlag(ExceptionBreakpointStates.BreakThrown)` except to remember that we are in an unsupported state so we need to warn when we stop.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -480,17 +495,32 @@ private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings cat
...
+                    ulong breakpointId;
+                    if (categorySettings.CurrentRules.TryGetValue("*", out breakpointId))
+                    {
+                        await _commandFactory.RemoveExceptionBreakpoint(categoryId, new ulong[] { breakpointId });
+                    }
+                    if (categorySettings.CurrentRules.TryGetValue("winrt::hresult_error", out breakpointId))
+                    {
+                        await _commandFactory.RemoveExceptionBreakpoint(categoryId, new ulong[] { breakpointId });
+                    }
+                }
```
gregg-miskelly, line 523: Remove

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gc46: Yes, this is perfect. Thank you for bringing this up :)

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1162,8 +1162,16 @@ private async Task HandleBreakModeEvent(ResultEventArgs results, BreakRequest br
                 string bkptno = results.Results.FindString("bkptno");
                 ulong addr = cxt.pc ?? 0;
 
+                EXCEPTION_INFO exceptionInfo;
                 bool fContinue;
                 TupleValue frame = results.Results.TryFind<TupleValue>("frame");
+
+                // test -- need to delete
+                if (ExceptionManager.TryGetExceptionBreakpoint(bkptno, out exceptionInfo) == Constants.S_OK)
```
WardenGnaw, line 1170: You can move this block to line 1200.

The flow would be that if we fail to find it as a regular breakpoint, it would check to see if it is an exception breakpoint (catchpoint).

As of now you are going to fire two events.

You should also be able to declare the type within the out parameter.

```suggestion
                if (ExceptionManager.TryGetExceptionBreakpoint(bkptno, out EXCEPTION_INFO exceptionInfo) == Constants.S_OK)
```

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -278,6 +278,27 @@ public void SetAllExceptions(enum_EXCEPTION_STATE dwState)
             }
         }
 
+        public int TryGetExceptionBreakpoint(string bkptno, out EXCEPTION_INFO exceptionInfo)
```
WardenGnaw, line 281: You can make this function simpler by saving a `Dictionary<ulong, EXCEPTION_INFO>` and have this method `TryGetValue`.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -434,13 +455,21 @@ private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings cat
...
 
-                IEnumerable<ulong> breakpointIds = await _commandFactory.SetExceptionBreakpoints(categoryId, null, newCategoryState);
-                if (newCategoryState != ExceptionBreakpointStates.None)
+                // IEnumerable<ulong> breakpointIds = await _commandFactory.SetExceptionBreakpoints(categoryId, null, newCategoryState);
+                if (newCategoryState == ExceptionBreakpointStates.BreakThrown) // only set catchpoint on BreakThrown
                 {
-                    ulong breakpointId = breakpointIds.Single();
-                    categorySettings.CurrentRules.Add("*", breakpointId);
+                    IEnumerable<ulong> breakpointIds = await _commandFactory.SetExceptionBreakpoints(categoryId, null, newCategoryState);
+                    ulong singleBreakpointId = breakpointIds.Single();
```
WardenGnaw, line 471: Here is where you have the breakpoint id (long), category guid, name, and state.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1162,8 +1162,16 @@ private async Task HandleBreakModeEvent(ResultEventArgs results, BreakRequest br
...
+                {
+                    _callback.OnException(thread, exceptionInfo.bstrExceptionName, "", exceptionInfo.dwCode,exceptionInfo.guidType);
```
WardenGnaw, line 1172: The reason why everything is seen as an uncaught exception is because `state` was not set here.

Everything is defaulting to `ExceptionBreakpointStates.None` which leads to `EXCEPTION_STOP_SECOND_CHANCE`

```suggestion
                    _callback.OnException(thread, exceptionInfo.bstrExceptionName, "", 0, exceptionInfo.guidType, state);
```

WardenGnaw (commented):
@gc46 Instead of using `EXCEPTION_INFO`, I think all you need is to keep track of the `guid` and `state`.

E.g.
`Dictionary<ulong, Tuple<Guid, ExceptionBreakpointStates >>`

PR #1237, opened 2021-11-09
Add support for Disassembly Source for GDB

WardenGnaw:
This PR adds Location and Line to Disassembly Response.


src/MIDebugEngine/Engine.Impl/Disassembly.cs
```diff
@@ -305,17 +305,153 @@ private void DeleteRangeFromCache(DisassemblyBlock block)
...
+                EndAddress = instruction.Addr;
+                AddressToInstruction = new Dictionary<ulong, DisasmInstruction>()
+                {
+                    {instruction.Addr, instruction}
+                };
+            }
+
+            public void UpdateEndAddress(DisasmInstruction instruction)
+            {
+                EndAddress = instruction.Addr;
```
WardenGnaw, line 326: Should `EndAddress` be `instruction.Addr + 1`?

src/MIDebugEngine/Engine.Impl/Disassembly.cs
```diff
@@ -305,17 +305,153 @@ private void DeleteRangeFromCache(DisassemblyBlock block)
...
+                         * asm_insns=[
+                         *  src_and_asm_line={
+                         *      line="15",
+                         *      file="main.cpp",
+                         *      fullname="/home/cpp/main.cpp",
+                         *      line_asm_insn=[ ... ]
+                         *  }
+                         * ]
+                         */
+                        ResultListValue asm_insns = results.Find<ResultListValue>("asm_insns");
```
WardenGnaw, line 436: Not sure if throwing exception here when `Find` fails, slows down this process a lot.

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -89,6 +89,20 @@ private DisassemblyData FetchBadInstruction(enum_DISASSEMBLY_STREAM_FIELDS dwFie
                 dis.dwFields |= enum_DISASSEMBLY_STREAM_FIELDS.DSF_OPCODE;
                 dis.bstrOpcode = "??";
             }
+
+            if ((dwFields & enum_DISASSEMBLY_STREAM_FIELDS.DSF_DOCUMENTURL) != 0)
+            {
+                dis.dwFields |= enum_DISASSEMBLY_STREAM_FIELDS.DSF_DOCUMENTURL;
+                dis.bstrDocumentUrl = string.Empty;
+            }
```
gregg-miskelly, line 97: I don't think you need this. It should be fine to leave the documentUrl empty + no DSF_DOCUMENTURL for instructions without a source file.

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -89,6 +89,20 @@ private DisassemblyData FetchBadInstruction(enum_DISASSEMBLY_STREAM_FIELDS dwFie
...
+
+            if ((dwFields & enum_DISASSEMBLY_STREAM_FIELDS.DSF_POSITION) != 0)
+            {
+                dis.dwFields |= enum_DISASSEMBLY_STREAM_FIELDS.DSF_POSITION;
+                dis.posBeg = new TEXT_POSITION();
+                dis.posEnd = new TEXT_POSITION();
+            }
```
gregg-miskelly, line 104: Same

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -162,6 +176,28 @@ public int Read(uint dwInstructions, enum_DISASSEMBLY_STREAM_FIELDS dwFields, ou
                     }
                 }
 
+                if (!string.IsNullOrWhiteSpace(instruction.File))
```
gregg-miskelly, line 179: >if [](http://example.com/codeflow?start=16&length=2)

We need to compute `DF_DOCUMENTCHANGE`. Note that that flag isn't trivial as it requires either going back an extra instruction to compute, or remembering where we left off in the case of multiple calls to `Next` without a call to Seek.

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -162,6 +176,28 @@ public int Read(uint dwInstructions, enum_DISASSEMBLY_STREAM_FIELDS dwFields, ou
...
+                {
+                    if ((dwFields & enum_DISASSEMBLY_STREAM_FIELDS.DSF_DOCUMENTURL) != 0)
+                    {
+                        prgDisassembly[iOp].dwFields |= enum_DISASSEMBLY_STREAM_FIELDS.DSF_DOCUMENTURL;
+                        prgDisassembly[iOp].bstrDocumentUrl = instruction.File;
+                    }
```
gregg-miskelly, line 185: Technically, we should also be returning this if it is different from the last instruction.

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -162,6 +176,28 @@ public int Read(uint dwInstructions, enum_DISASSEMBLY_STREAM_FIELDS dwFields, ou
...
+
+                    if ((dwFields & enum_DISASSEMBLY_STREAM_FIELDS.DSF_POSITION) != 0)
+                    {
+                        prgDisassembly[iOp].dwFields |= enum_DISASSEMBLY_STREAM_FIELDS.DSF_POSITION;
+                        prgDisassembly[iOp].posBeg = new TEXT_POSITION()
+                        {
+                            dwLine = instruction.Line
```
gregg-miskelly, line 192: >dwLine = instruction.Line [](http://example.com/codeflow?start=28&length=25)

We might want to expand this range to cover more than just the exsact line that GDB gives us. Consider something like:

```
   printf("%d, %d, %d, %d\n",
     Func1(),
     Func2(),
     Func3(),
     Func4()
     );
```

If you only give back one line, then that is all that will show up in disassembly. So you would see:

```
   printf("%d, %d, %d, %d\n", // or worse mabe ');'
   mov ...
```

We probably instead want to remember the line of the previous instruction, and expand the range to fill in the space (with some limits to avoid dumping huge ranges).

src/MIDebugEngine/Engine.Impl/Disassembly.cs
```diff
@@ -305,17 +305,153 @@ private void DeleteRangeFromCache(DisassemblyBlock block)
...
                 return null;
             }
 
-            return DecodeDisassemblyInstructions(results.Find<ValueListValue>("asm_insns").AsArray<TupleValue>());
+            DisasmInstruction[] instructions =  DecodeDisassemblyInstructions(results.Find<ValueListValue>("asm_insns").AsArray<TupleValue>());
+
+            if (instructions != null && instructions.Length != 0)
+            {
+                // Map 'Symbol' (Function Name) to Address Range
+                Dictionary<string, DisasmAddressRange> funcToAddressRange = new Dictionary<string, DisasmAddressRange>();
```
gregg-miskelly, line 391: >Dictionary<string, DisasmAddressRange> [](http://example.com/codeflow?start=16&length=38)

I don't think you want a map here because if you see something like:

MethodA
MethodB
MethodA

You don't want to combine the two MethodA blocks together. So you just need to remember the last (range/function name) pair.

src/MIDebugEngine/Engine.Impl/Disassembly.cs
gregg-miskelly: Seems like it is worth fixing to me.

gregg-miskelly:
>                     }

Remove and start returning this flag correctly

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:2072 in 31f3dac. [](commit_id = 31f3dac6a8e6a10833591ae8cb6f729b319e08a7, deletion_comment = False)

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2076,6 +2076,16 @@ protected override void HandleDisassembleRequestAsync(IRequestResponder<Disassem
...
                         Symbol = data.bstrSymbol
                     };
+
+                    if (!string.IsNullOrWhiteSpace(data.bstrDocumentUrl))
+                    {
+                        instruction.Location = new Source()
+                        {
+                            Path = data.bstrDocumentUrl
+                        };
+                        instruction.Line = (int)data.posBeg.dwLine;
```
gregg-miskelly, line 2086: >instruction [](http://example.com/codeflow?start=24&length=11)

Return endLine also

gregg-miskelly: Actually, no reason not to return everything.

gregg-miskelly, line 2086: >(int)data.posBeg.dwLine [](http://example.com/codeflow?start=43&length=23)

m_pathConverter.ConvertDebuggerLineToClient

gregg-miskelly: Which might mean that you are missing a `- 1` in the line calculations in AD7Disassembly.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2076,6 +2076,16 @@ protected override void HandleDisassembleRequestAsync(IRequestResponder<Disassem
                         Instruction = data.bstrOpcode,
                         Symbol = data.bstrSymbol
                     };
+
+                    if (!string.IsNullOrWhiteSpace(data.bstrDocumentUrl))
+                    {
```
gregg-miskelly, line 2081: This should be checking for the hasSource flag, and it should either return `data.bstrDocumentUrl` (if set) or the last `bstrDocumentUrl` if the current one is null.

If you want to reduce the data payload a bit, you could also omit this if the line and document didn't change.

gregg-miskelly: Wait, that wasn't quite right. We can set `instruction.Location` whenever `data.bstrDocumentUrl` is not null or empty. It is just the line/endline etc that should be based on `DF_HASSOURCE`.

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
WardenGnaw: Still need to address expanding the range over multiple lines.

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -118,8 +125,12 @@ public int Read(uint dwInstructions, enum_DISASSEMBLY_STREAM_FIELDS dwFields, ou
                 prgDisassembly[iOp] = FetchBadInstruction(dwFields);
             }
 
-            foreach (DisasmInstruction instruction in instructions)
+            IEnumerator<DisasmInstruction> instructionEnumerator = instructions.GetEnumerator();
```
gregg-miskelly, line 128: >IEnumerator [](http://example.com/codeflow?start=12&length=11)

nit: you need a `using()`

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -162,6 +173,64 @@ public int Read(uint dwInstructions, enum_DISASSEMBLY_STREAM_FIELDS dwFields, ou
...
 
+                if (idx == 0 && m_lastSourceInfoStale)
+                {
+                    m_lastSourceInfoStale = false;
+
+                    ulong previousInstructionAddress = instruction.Addr - 1;
+
+                    if (previousInstructionAddress >= 0 && previousInstructionAddress < (ulong)prgDisassembly.Length)
+                    {
+                        DisassemblyData previousInstruction = prgDisassembly[previousInstructionAddress];
```
gregg-miskelly, line 184: >previousInstructionAddress [](http://example.com/codeflow?start=77&length=26)

I am not following this logic. `prgDisassembly` is an array. So you can't used an instruction address here.

Instead, what I think you want is:
* We need to SeekBack an extra instruction, so:
   * Add a new bool field to know if we have already seaked back an extra instruction (something like `m_skipNextInstruction`)
   * In SeekBack, if the last instruction data is stale, SeekBack an extra instruction and set the m_skipNextInstruction to true
   * In the start of this method, if the previous instruction data is stale and m_skipNextInstruction is false, call `SeekBack(1)`.
   * In this method, if `m_skipNextInstruction` is true then we need to increment the requested number of instructions.
   * In this method, if `m_skipNextInstruction` is true then we need to:
      * Do a `.FirstOrDefault()` to process the first instruction and fill in the previous instruction data
      * Do a `.Skip(1)` on the instruction enumerator

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -162,6 +173,64 @@ public int Read(uint dwInstructions, enum_DISASSEMBLY_STREAM_FIELDS dwFields, ou
...
+                        m_dwLastSourceLine = (int)previousInstruction.posBeg.dwLine - 1;
+                        _pLastDocument = previousInstruction.bstrDocumentUrl;
+                    }
+                }
+
+                if ((dwFields & enum_DISASSEMBLY_STREAM_FIELDS.DSF_POSITION) != 0)
```
gregg-miskelly, line 190: >) [](http://example.com/codeflow?start=81&length=1)

I am assuming `instruction.Line` is 0 if it isn't set? If so, you need `&& instruction.Line != 0`

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -162,6 +173,64 @@ public int Read(uint dwInstructions, enum_DISASSEMBLY_STREAM_FIELDS dwFields, ou
...
+                    prgDisassembly[iOp].posEnd = new TEXT_POSITION()
+                    {
+                        dwLine = instruction.Line,
+                        dwColumn = 0
+                    };
+
+                    prgDisassembly[iOp].dwFields |= enum_DISASSEMBLY_STREAM_FIELDS.DSF_POSITION;
+
+                    if (m_dwLastSourceLine != instruction.Line)
+                    {
```
gregg-miskelly, line 208: I don't think you want this if

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -14,13 +14,20 @@ internal class AD7DisassemblyStream : IDebugDisassemblyStream2
         private AD7Engine _engine;
         private ulong _addr;
         private enum_DISASSEMBLY_STREAM_SCOPE _scope;
+        private string _pLastDocument;
+        private int m_dwLastSourceLine;
```
gregg-miskelly, line 18: >dw [](http://example.com/codeflow?start=22&length=2)

nit: remove hungarian prefix

gregg-miskelly:
>             if (instructions == null || (instructions.First().Addr - _addr > dwInstructions))

Do you understand this condition? I would have thought it would be `instructions.First().Addr != _addr`

---
Refers to: src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs:122 in d63c422. [](commit_id = d63c4222fc85209c794106d2b3a848434cf33673, deletion_comment = False)

gregg-miskelly:
>                 for (iOp = 0; iOp < dwInstructions; _addr++, ++iOp)

Need to handle skipping the first instruction in this path.

---
Refers to: src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs:125 in d63c422. [](commit_id = d63c4222fc85209c794106d2b3a848434cf33673, deletion_comment = False)

gregg-miskelly:
>                 return Constants.S_OK;

```
m_lastSourceInfoStale = false;
m_dwLastSourceLine = 0;
_pLastDocument = null;
``




---
Refers to: src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs:130 in d63c422. [](commit_id = d63c4222fc85209c794106d2b3a848434cf33673, deletion_comment = False)

**Comment by gregg-miskelly** [comment]
>             if (instructions == null || (instructions.First().Addr - _addr > dwInstructions))

Ah. Figured it out.

---
In reply to: [970990930](https://github.com/microsoft/MIEngine/pull/1237#issuecomment-970990930) [](http://example.com/codeflow?ancestors=970990930)

---
Refers to: src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs:122 in d63c422. [](commit_id = d63c4222fc85209c794106d2b3a848434cf33673, deletion_comment = False)

File: src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs [hunk 16]
```diff
@@ -118,52 +136,114 @@ public int Read(uint dwInstructions, enum_DISASSEMBLY_STREAM_FIELDS dwFields, ou
                 prgDisassembly[iOp] = FetchBadInstruction(dwFields);
             }
```
gregg-miskelly, line 137: You need a similar change here

gregg-miskelly: It seems like we could probably combine the two code paths together.

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -118,52 +136,114 @@ public int Read(uint dwInstructions, enum_DISASSEMBLY_STREAM_FIELDS dwFields, ou
...
 
-            foreach (DisasmInstruction instruction in instructions)
+            using (IEnumerator<DisasmInstruction> instructionEnumerator = instructions.GetEnumerator())
             {
-                if (iOp >= dwInstructions)
+                int idx = 0;
+                if (m_skipNextInstruction)
                 {
```
gregg-miskelly, line 143: >{ [](http://example.com/codeflow?start=16&length=1)

`m_skipNextInstruction = false;`

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -118,52 +136,114 @@ public int Read(uint dwInstructions, enum_DISASSEMBLY_STREAM_FIELDS dwFields, ou
...
                 }
 
-                if ((dwFields & enum_DISASSEMBLY_STREAM_FIELDS.DSF_SYMBOL) != 0)
+                for (; idx < dwInstructions; idx++)
                 {
-                    if (instruction.Offset == 0)
+                    instructionEnumerator.MoveNext();
+                    DisasmInstruction instruction = instructionEnumerator.Current;
+
+                    if (iOp >= dwInstructions)
```
gregg-miskelly, line 158: >dwInstructions [](http://example.com/codeflow?start=31&length=14)

prgDisassembly.Length

gregg-miskelly: Might make sense to put this in the loop condition of the for loop

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -118,52 +136,114 @@ public int Read(uint dwInstructions, enum_DISASSEMBLY_STREAM_FIELDS dwFields, ou
...
+                        };
+                        prgDisassembly[iOp].posEnd = new TEXT_POSITION()
+                        {
+                            dwLine = instruction.Line,
+                            dwColumn = 0
+                        };
+
+                        prgDisassembly[iOp].dwFields |= enum_DISASSEMBLY_STREAM_FIELDS.DSF_POSITION;
+
+                        if (m_dwLastSourceLine != instruction.Line && instruction.Line != 0)
```
gregg-miskelly, line 217: >m_dwLastSourceLine != instruction.Line && [](http://example.com/codeflow?start=28&length=41)

I think you want to get rid of this part of the if. It should just be `instruction.Line != 0`.

gregg-miskelly: Oh wait, the outer if aready has that check, so we should also set `DF_HASSOURCE`

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
WardenGnaw: This is what we get with the VC++ Debug Engine:
image

What I have currently:
image

src/MIDebugEngine/AD7.Impl/AD7Disassembly.cs
```diff
@@ -213,11 +213,15 @@ public int Read(uint dwInstructions, enum_DISASSEMBLY_STREAM_FIELDS dwFields, ou
 
                     if ((dwFields & enum_DISASSEMBLY_STREAM_FIELDS.DSF_POSITION) != 0 && currentLine != 0)
                     {
+                        // If we have a new line and the current line is greater than the previously seen source line.
+                        // Try to grab the last seen source line + 1 and show a group of source code.
+                        // Else, just show the single line.
+                        uint startLine = isNewLine && currentLine > _dwLastSourceLine ? _dwLastSourceLine + 1 : currentLine;
```
gregg-miskelly, line 219:
```suggestion
                        uint startLine = isNewLine && !isNewDocument && currentLine > _dwLastSourceLine ? _dwLastSourceLine + 1 : currentLine;
```

WardenGnaw:
Based on godbolt, it looks like this is intended.
Should revert https://github.com/microsoft/MIEngine/pull/1237/commits/b7870f554426b4b15d4806d3c0767d1d20375dc8

haneefdm:
If it is alright, I can do a Pull Request for interleaving source in assembly. I had to do it for my own debugger and made similar changes to AD7. There are some unexpected gotchas in the parsing of the gdb MI results for mixed source/asm.

My changes are on https://github.com/haneefdm/MIEngine
disassembly

Will work with gdb version 8+ (Jun 4 2017) where they revamped (and deprecated) some of the ways disassembly is produced.

haneefdm:
I can't do a PR for MIEngine with this PR being outstanding. I looked at the files in this PR. There are a couple of things jump out at me. My changes are much simpler and not sure we have to worry about source lines being in order, etc. One thing missing (and should be fixed) in my changes however are calls to ConvertDebuggerLineToClient and ConvertDebuggerColumnToClient

My own debugger is here

https://github.com/Marus/cortex-debug/blob/master/src/backend/disasm.ts

I have two disassemblers in there and hoping to get rid of the old one and replace it with the new DAP protocol version so it can work nicely with the VSCode debug infrastructure.

WardenGnaw:
@haneefdm One issue that came up with this feature was when requesting address ranges outside of the user's space or a huge size request, GDB would return with no source information, which required us to find each range per symbol and re-request disassembly from GDB.

It seemed like GDB would only return source information if it actually had matching symbol information, but any anypoint if it was missing, it would remove it from the whole data-disassemble response.

src/MIDebugEngine/Engine.Impl/Disassembly.cs
```diff
@@ -305,17 +305,145 @@ private void DeleteRangeFromCache(DisassemblyBlock block)
...
+                 *              opcodes="75 07",
+                 *              inst="jne    0x8001495 <main(int, char**)+332>"
+                 *          }
+                 *      ]
+                 *   }
+                 *  ]
+                 */
+                foreach (TupleValue src_and_asm_line in src_and_asm_lines)
+                {
+                    uint line = src_and_asm_line.FindUint("line");
```
haneefdm, line 361: Watch out, this can fail (Exception) because "line" can be missing. When "line" is missing so will the "file" or "filename". I did it backwards. First I check for the file and if it exists and the line exists, then I call `FindUnit("line")`. It happens when a few instructions have source info but a few don't in the same addr-rane. If all don't have source-info, then there will be no  `src_and_asm_line` either. [My code looks like](https://github.com/haneefdm/MIEngine/blob/507a8140aa0de4f02e20fd10387da084d5fddc9e/src/MIDebugEngine/Engine.Impl/Disassembly.cs#L343).

```C#
                    string file = process.GetMappedFileFromTuple(item);
                    uint line = !String.IsNullOrEmpty(file) && item.Contains("line") ? item.FindUint("line") : 0;
```

haneefdm: Oh, I see you are disassembling twice. I did not find that necessary, but you may have seen things I haven't. You can ignore my previous comment but I would still guard against not having file/source info in the middle of normal looking stuff.

haneefdm:
From what I saw, the request for source info (mode 4 or 5) degenerates into a request of request for mode 2. At least, the results will look like that. This just baffled me at first. This happens especially with bad/unknown memory regions and happened even within a proper memory region when source information was missing.

Yup, address ranges are a big problem. I am still experimenting with this, btw. I am reading gdb source code to confirm what I guessed -- lot left to read though. Also want to see what Ecliipse does. Here is what I plan to do in the short term for my debugger
* Have a limit on a number of instructions (let us say 10K or something reasonable). The client can always chunk it up if they are serious. Anyone making huge requests, chances are it was a bug and also interferes with responsiveness -- what was their purpose? I know the DAP does not place limits, but I intend to. Perhaps as part of capabilities we can advertise the max request size.
* Let gdb fail on getting the instructions that are outside of the memory range. This is another reason why we should be getting instructions in small chunks. Maybe later I can add code to find the true boundary.
* My cases are embedded cases where I do have some information in the executable to know if the requests are venturing out of known boundaries.
* In your case, you have native and remote cases but more than likely also using virtual memory.
* For native programs, if you so wish, you can find out the extent(s) of the process allocated pages. For instance on Linux you can look at /proc/pid/maps. Not worth it I think
* For remote programs (those involving a gdb-server) the options are even more limited.

The following does not work for native programs because technically the entire memory is the same and super large. So, no regions are defined.
```
(gdb) info mem
Using memory regions provided by the target.
Num Enb Low Addr   High Addr  Attrs 
0   y  	0x00000000 0x10000000 rw nocache 
1   y  	0x10000000 0x10100000 flash blocksize 0x200 nocache 
2   y  	0x10100000 0x14000000 rw nocache 
3   y  	0x14000000 0x14008000 flash blocksize 0x200 nocache 
4   y  	0x14008000 0x16000000 rw nocache 
5   y  	0x16000000 0x16008000 flash blocksize 0x200 nocache 
6   y  	0x16008000 0x90700000 rw nocache 
7   y  	0x90700000 0x90700400 flash blocksize 0x400 nocache 
8   y  	0x90700400 0x100000000 rw nocache 
(gdb)
```
The above is what is possible even if very little is used (actual boundaries). For the same executable, using objdump, I get the following which is what this executable is using. In your case, you won't know which memory region got mapped to which virtual one.
```
% arm-none-eabi-objdump -h build/CY8CKIT-062-WIFI-BT/Debug/mtb-example-psoc6-dual-cpu-ipc-pipes.elf

build/CY8CKIT-062-WIFI-BT/Debug/mtb-example-psoc6-dual-cpu-ipc-pipes.elf:     file format elf32-littlearm

Sections:
Idx Name          Size      VMA       LMA       File off  Algn
  0 .cy_m0p_image 000025d4  10000000  10000000  00010000  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  1 .text         000093f8  10008000  10008000  00018000  2**4
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
  2 .ARM.exidx    00000008  100113f8  100113f8  000213f8  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  3 .copy.table   00000018  10011400  10011400  00021400  2**0
                  CONTENTS, ALLOC, LOAD, DATA
  4 .zero.table   00000008  10011418  10011418  00021418  2**0
                  CONTENTS, ALLOC, LOAD, DATA
  5 .ramVectors   0000028c  08008000  08008000  000288a0  2**3
                  CONTENTS, READONLY
  6 .data         00000604  0800828c  10011420  0002828c  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
  7 .cy_sharedmem 0000000c  08008890  10011a24  00028890  2**2
                  CONTENTS, ALLOC, LOAD, DATA
  8 .noinit       00000094  080088a0  10011a38  000288a0  2**3
                  ALLOC
  9 .bss          00000434  08008934  10011acc  000288a0  2**2
                  ALLOC
 10 .heap         0003da98  08008d68  10011f00  000288a0  2**3
                  ALLOC, READONLY
 11 .stack_dummy  00001000  08046800  08046800  00028b30  2**3
                  CONTENTS, READONLY
 12 .ARM.attributes 0000002e  00000000  00000000  00029b30  2**0
                  CONTENTS, READONLY
 13 .comment      00000049  00000000  00000000  00029b5e  2**0
                  CONTENTS, READONLY
 14 .debug_line   0001b471  00000000  00000000  00029ba7  2**0
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 15 .debug_info   000611ec  00000000  00000000  00045018  2**0
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 16 .debug_abbrev 00005ef3  00000000  00000000  000a6204  2**0
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 17 .debug_aranges 00001740  00000000  00000000  000ac0f8  2**3
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 18 .debug_str    000206c8  00000000  00000000  000ad838  2**0
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 19 .debug_ranges 00002290  00000000  00000000  000cdf00  2**0
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 20 .debug_frame  00004e74  00000000  00000000  000d0190  2**2
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 21 .debug_loc    00017b87  00000000  00000000  000d5004  2**0
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 22 .stab         0000009c  00000000  00000000  000ecb8c  2**2
                  CONTENTS, READONLY, DEBUGGING
 23 .stabstr      0000014f  00000000  00000000  000ecc28  2**0
                  CONTENTS, READONLY, DEBUGGING
% 
```

PR #1239, opened 2021-11-18
Allow passwords in ssh connection strings

paulmaybee:
Supports an Office scenario for attaching to a process running somewhere in a pool of servers. Such connections are temporary; they do not get added to the Connection Manager.

### DIFF
--- a/src/SSHDebugPS/ConnectionManager.cs
+++ b/src/SSHDebugPS/ConnectionManager.cs
@@ -5,8 +5,10 @@ using System;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
+using System.Security;
 using System.Text.RegularExpressions;
 using System.Threading;
+using System.Windows;
 using System.Windows.Interop;
 using EnvDTE;
 using liblinux;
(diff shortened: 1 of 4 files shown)


src/SSHDebugPS/ConnectionManager.cs
```diff
@@ -82,19 +84,48 @@ public static SSHConnection GetSSHConnection(string name)
...
+                            // Attempt to open the temp connection so that the host key can be verified.
+                            try
+                            {
+                                using (var system = new UnixSystem())
+                                {
+                                    system.Connect(connectionInfo);
+                                }
+                            }
+                            catch (RemoteConnectivityException e) when (!String.IsNullOrEmpty(e.Fingerprint))
+                            {
```
gregg-miskelly, line 113: >{ [](http://example.com/codeflow?start=28&length=1)

I think we need to throw if we aren't on the UI thread.

paulmaybee: First line in this method does that.

gregg-miskelly: Thanks. Somehow missed that.

src/SSHDebugTests/SSHConnectionStringTests.cs
```diff
@@ -45,9 +46,9 @@ public void IPv6ConnectionStrings()
             ipv6TestStrings.Add(
                 new ConnectionStringTestItem()
                 {
-                    // valid with custom username
+                    // valid with username:password
```
gregg-miskelly, line 49: >:password [](http://example.com/codeflow?start=42&length=9)

I think you want to restore this comment

src/SSHDebugTests/SSHConnectionStringTests.cs
```diff
@@ -143,6 +144,15 @@ public void Ipv4ConnectionStrings()
...
                     expectedPort = ConnectionManager.DefaultSSHPort
                 });
+            ipv4TestStrings.Add(
+                 new ConnectionStringTestItem()
+                 {
+                     // valid username no port with password
+                     rawConnectionString = "user:pass@10.10.10.10",
+                     expectedUsername = "user",
+                     expectedHostname = "10.10.10.10",
+                     expectedPort = ConnectionManager.DefaultSSHPort
```
gregg-miskelly, line 154: >expectedPort [](http://example.com/codeflow?start=21&length=12)

Should we add expected password testing also?

gregg-miskelly (approved):
Otherwise looks good to me

src/SSHDebugTests/SSHConnectionStringTests.cs
paulmaybee: I started too, but given the secure string I can't check the value. I can check that the expected password is present?

src/SSHDebugTests/SSHConnectionStringTests.cs
paulmaybee: Isn't the new comment more explicit about what is being tested with this case?

gregg-miskelly: Right, but didn't you change that test case so it does NOT have a password?

gregg-miskelly: Oh, I see. That test case is weird. I wonder if we should get rid of it in favor of your new one?

gregg-miskelly: Never mind. I see that it is good as an IPv6 test. I wonder what in the world it was trying to test before...

src/SSHDebugTests/SSHConnectionStringTests.cs
gregg-miskelly: You can convert a SecureString to a string with this:

```
            IntPtr bstr = System.Runtime.InteropServices.Marshal.SecureStringToBSTR(s);
            string value = System.Runtime.InteropServices.Marshal.PtrToStringBSTR(bstr);
            System.Runtime.InteropServices.Marshal.FreeBSTR(bstr);
```

paulmaybee: Excellent thanks. Added the password checking.

WardenGnaw (approved):

gregg-miskelly (approved):

PR #1245, opened 2021-11-22
Linux Debugging: Debugger does not stop for handled C++ exceptions

gc46:
Spent a Sunday debugging my cross-platform C++ exceptions implementation. Functionally okay after some testing (this was confirmed with Command Window using `info breakpoints`), but may need to do a bit of cleanup on the UI side. Pretty happy right now and am ready for PR comments. Never give up :)


src/MICore/CommandFactories/gdb.cs
```diff
@@ -295,5 +295,46 @@ public override async Task<string[]> AutoComplete(string command, int threadId,
...
+
+        public override async Task<IEnumerable<ulong>> SetExceptionBreakpoints(Guid exceptionCategory, IEnumerable<string> exceptionNames, ExceptionBreakpointStates exceptionBreakpointStates)
+        {
+            string command;
+            Results result;
+            List<ulong> breakpointNumbers = new List<ulong>();
+
+            if (exceptionNames == null) // set breakpoint for all exceptions in exceptionCategory
+            {
+                command = string.Format(CultureInfo.InvariantCulture, "-catch-throw");
```
gregg-miskelly, line 313: >string.Format(CultureInfo.InvariantCulture,  [](http://example.com/codeflow?start=26&length=44)

nit: no need to use `String.Format` here -- you aren't passing an args

src/MICore/CommandFactories/gdb.cs
```diff
@@ -295,5 +295,46 @@ public override async Task<string[]> AutoComplete(string command, int threadId,
...
+                result = await _debugger.CmdAsync(command, ResultClass.done);
+                var breakpointNumber = result.Find("bkpt").FindUint("number");
+                breakpointNumbers.Add(breakpointNumber);
+            }
+            else // set breakpoint for each exceptionName in exceptionNames
+            {
+                command = string.Format(CultureInfo.InvariantCulture, "-catch-throw -r ");
+                foreach (string exceptionName in exceptionNames)
+                {
+                    result = await _debugger.CmdAsync(command + exceptionName, ResultClass.done);
```
gregg-miskelly, line 323: >exceptionName [](http://example.com/codeflow?start=64&length=13)

Do we need to do anything to prevent partial string matches here? For example, suppose the debugger is configured to break on `Foo` but the exception raised is a `NFoo` or `FooN`. Do we need to surrond `exceptionName` with `\b` or something like that?

src/MICore/CommandFactories/gdb.cs
```diff
@@ -295,5 +295,46 @@ public override async Task<string[]> AutoComplete(string command, int threadId,
...
+            if (exceptionNames == null) // set breakpoint for all exceptions in exceptionCategory
+            {
+                command = string.Format(CultureInfo.InvariantCulture, "-catch-throw");
+                result = await _debugger.CmdAsync(command, ResultClass.done);
+                var breakpointNumber = result.Find("bkpt").FindUint("number");
+                breakpointNumbers.Add(breakpointNumber);
+            }
+            else // set breakpoint for each exceptionName in exceptionNames
+            {
+                command = string.Format(CultureInfo.InvariantCulture, "-catch-throw -r ");
```
gregg-miskelly, line 320: >string.Format(CultureInfo.InvariantCulture, [](http://example.com/codeflow?start=26&length=43)

nit: Same here - remove `string.Format`

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1189,6 +1189,11 @@ private async Task HandleBreakModeEvent(ResultEventArgs results, BreakRequest br
                     bplist.AddRange(bkpt);
                     _callback.OnBreakpoint(thread, bplist.AsReadOnly());
                 }
+                else if (ExceptionManager.TryGetExceptionBreakpoint(bkptno, out EXCEPTION_INFO exceptionInfo) == Constants.S_OK) // exception breakpoint hit
```
gregg-miskelly, line 1192: >== Constants.S_OK [](http://example.com/codeflow?start=110&length=17)

nit: I would suggest returning a `bool` instead of an HRESULT. Both for methods that our "bool sounding" like "Try..." and also just for internal methods in general when in C# code.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -278,6 +278,28 @@ public void SetAllExceptions(enum_EXCEPTION_STATE dwState)
...
+            exceptionInfo = new EXCEPTION_INFO();
+            ExceptionCategorySettings categorySettings;
+            if (_categoryMap.TryGetValue(new Guid(CppExceptionCategoryString), out categorySettings))
+            {
+                ulong breakpointNumber = Convert.ToUInt32(bkptno, CultureInfo.InvariantCulture);
+                if (categorySettings.CurrentRules.ContainsValue(breakpointNumber))
+                {
+                    // test -- need to delete; these are placeholder values
+                    exceptionInfo.bstrExceptionName = categorySettings.CategoryName;
+                    exceptionInfo.dwCode = (uint)breakpointNumber;
```
gregg-miskelly, line 293: >exceptionInfo.dwCode = (uint)breakpointNumber; [](http://example.com/codeflow?start=20&length=46)

Remove

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -278,6 +278,28 @@ public void SetAllExceptions(enum_EXCEPTION_STATE dwState)
...
+                    bool isBreakThrown = categorySettings.CategoryState == ExceptionBreakpointStates.BreakThrown; // this is true, but why is dwState still NONE?
+                    exceptionInfo.dwState = isBreakThrown ? (enum_EXCEPTION_STATE.EXCEPTION_STOP_FIRST_CHANCE | enum_EXCEPTION_STATE.EXCEPTION_STOP_USER_FIRST_CHANCE) : enum_EXCEPTION_STATE.EXCEPTION_NONE;
```
gregg-miskelly, line 295: >enum_EXCEPTION_STATE.EXCEPTION_STOP_FIRST_CHANCE | enum_EXCEPTION_STATE.EXCEPTION_STOP_USER_FIRST_CHANCE [](http://example.com/codeflow?start=61&length=104)

You can just always use this value (isBreakThrown will always be true)

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -278,6 +278,28 @@ public void SetAllExceptions(enum_EXCEPTION_STATE dwState)
             }
         }
 
+        public int TryGetExceptionBreakpoint(string bkptno, out EXCEPTION_INFO exceptionInfo)
+        {
+            const string CppExceptionCategoryString = "{3A12D0B7-C26C-11D0-B442-00A0244A1DD2}";
+            exceptionInfo = new EXCEPTION_INFO();
+            ExceptionCategorySettings categorySettings;
+            if (_categoryMap.TryGetValue(new Guid(CppExceptionCategoryString), out categorySettings))
```
gregg-miskelly, line 286: >new Guid(CppExceptionCategoryString) [](http://example.com/codeflow?start=41&length=36)

I would suggest making a static readonly property for this.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -278,6 +278,28 @@ public void SetAllExceptions(enum_EXCEPTION_STATE dwState)
...
+            {
+                ulong breakpointNumber = Convert.ToUInt32(bkptno, CultureInfo.InvariantCulture);
+                if (categorySettings.CurrentRules.ContainsValue(breakpointNumber))
```
gregg-miskelly, line 289: >CurrentRules [](http://example.com/codeflow?start=37&length=12)

This violates the threading rule on line 97. We need to either start locking the dictionary.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -278,6 +278,28 @@ public void SetAllExceptions(enum_EXCEPTION_STATE dwState)
...
+                {
+                    // test -- need to delete; these are placeholder values
+                    exceptionInfo.bstrExceptionName = categorySettings.CategoryName;
```
gregg-miskelly, line 292: >categorySettings.CategoryName [](http://example.com/codeflow?start=54&length=29)

FYI: This should be whatever name is associated with the breakpoint number. So the category name for the category rule, and otherwise the name of the exception type.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1189,6 +1189,11 @@ private async Task HandleBreakModeEvent(ResultEventArgs results, BreakRequest br
...
+                {
+                    ExceptionBreakpointStates exceptionBreakpointStates = ExceptionManager.ToExceptionBreakpointState(exceptionInfo.dwState);
+                    _callback.OnException(thread, exceptionInfo.bstrExceptionName, "", 0, exceptionInfo.guidType, exceptionBreakpointStates);
```
gregg-miskelly, line 1195: >OnException [](http://example.com/codeflow?start=30&length=11)

nit: We should probably get more agreement between `TryGetExceptionBreakpoint` and `OnException` on the types used, at the least they should both probably agree on enum_EXCEPTION_STATE vs ExceptionBreakpointStates.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -434,11 +473,17 @@ private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings cat
...
+                // remove exception breakpoints before categorySettings.CurrentRules is cleared
+                await _commandFactory.RemoveExceptionBreakpoint(categoryId, categorySettings.CurrentRules.Values);
+
+                // question: why does categorySettings.CurrentRules need to be cleared?
                 categorySettings.CurrentRules.Clear();
 
-                IEnumerable<ulong> breakpointIds = await _commandFactory.SetExceptionBreakpoints(categoryId, null, newCategoryState);
-                if (newCategoryState != ExceptionBreakpointStates.None)
+                // only do a generic catch throw if C++ exceptions category is checked
+                if (IsCppExceptionsCategoryChecked(updates))
```
gregg-miskelly, line 484: >if (IsCppExceptionsCategoryChecked(updates)) [](http://example.com/codeflow?start=16&length=44)

This should give you the same result as `updates.NewCategoryState.GetValueOrDefault(ExceptionBreakpointStates.None).HasFlag(ExceptionBreakpointStates.BreakThrown)`

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -480,17 +525,30 @@ private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings cat
                     }
                 }
 
-                IEnumerable<ulong> breakpointIds = await _commandFactory.SetExceptionBreakpoints(categoryId, exceptionNames, grouping.Key);
-
-                int count = exceptionNames.Zip(breakpointIds, (exceptionName, breakpointId) =>
+                bool isBreakThrown = grouping.Key.HasFlag(ExceptionBreakpointStates.BreakThrown);
+                if (!IsCppExceptionsCategoryChecked(updates) && isBreakThrown)
```
gregg-miskelly, line 529: >IsCppExceptionsCategoryChecked [](http://example.com/codeflow?start=21&length=30)

Same as comment as above

gregg-miskelly:
>                     settingsUpdateHolder.Value.RulesToAdd.Remove(exceptionName);

This should be removed

---
Refers to: src/MIDebugEngine/Engine.Impl/ExceptionManager.cs:232 in ef3b6e2. [](commit_id = ef3b6e28f827f5950307346ac8a2d70561d455d4, deletion_comment = False)

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -480,17 +525,30 @@ private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings cat
...
+                {
+                    // IEnumerable<ulong> breakpointIds = await _commandFactory.SetExceptionBreakpoints(categoryId, exceptionNames, grouping.Key);
+                    // this approach is very messy, but we need to update categorySettings.CurrentRules
+                    foreach (string exceptionName in exceptionNames)
+                    {
+                        IEnumerable<ulong> breakpointIds = await _commandFactory.SetExceptionBreakpoints(categoryId, new string[] { exceptionName }, grouping.Key);
+                        categorySettings.CurrentRules.Add(exceptionName, breakpointIds.First());
+                    }
```
gregg-miskelly, line 537: Rather than sending the exceptions one at a time, we should really use the old code here.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -480,17 +525,30 @@ private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings cat
...
+                }
```
gregg-miskelly, line 538: >} [](http://example.com/codeflow?start=16&length=1)

We should send a warning to the output window when the category is checked but there is an unchecked exception.

gregg-miskelly:
>                     _callback.OnError(string.Format(CultureInfo.CurrentCulture, ResourceStrings.ExceptionSettingsError, e.Message));

This will pop up a MessageBox on failure in the VS case. I am not sure how long GDB has supported `-catch-throw` command. Before you checkin we probably need to test Android and some embedded scenarios to make sure we don't need to change this to an output window warning instead.

---
Refers to: src/MIDebugEngine/Engine.Impl/ExceptionManager.cs:428 in ef3b6e2. [](commit_id = ef3b6e28f827f5950307346ac8a2d70561d455d4, deletion_comment = False)

src/MICore/CommandFactories/gdb.cs
gc46: Thank you for bringing this up. Fixed using `^` to match the start of the string and `$` to match the end of the string. Source: https://stackoverflow.com/questions/4581412/how-to-make-regexs-only-match-on-exact-matches/4581431

gregg-miskelly: You might want `\b` (or similar if `\b` isn't supported) instead of `^` and `$`, but you will need to test to find out: what happens if you raise an exception of type `const Foo` or `Foo*`?

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gc46: I understand this. That said, I did a `foreach` here because `CurrentRules` keeps track of the `exceptionName` and `breakpointId` pairs. Unless there is a better way of tracking this (perhaps with a lambda expression?)...

Notice that {exceptionName, breakpointId} is added to `CurrentRules` after `SetExceptionsBreakpoints()` is called. And notice {exceptionName, breakpointId} is removed from `CurrentRules` after `RemoveExceptionBreakpoint()` is called.

gregg-miskelly: That is why the old code was using `.Zip` to zip together the two enumerables:

```
                IEnumerable<ulong> breakpointIds = await _commandFactory.SetExceptionBreakpoints(categoryId, exceptionNames, grouping.Key);

                int count = exceptionNames.Zip(breakpointIds, (exceptionName, breakpointId) =>
                {
                    categorySettings.CurrentRules[exceptionName] = breakpointId;
                    return 1;
                }).Sum();

#if DEBUG
                Debug.Assert(count == exceptionNames.Count());
#endif
```

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gc46: Okay. Will lock the dictionary.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gc46: > isBreakThrown will always be true

Makes sense when reason == "breakpoint-hit". That said, I can just remove the `exceptionInfo.dwState` assignment then.

Instead of the following in DebuggedProcess.cs, Line 1194
```
ExceptionBreakpointStates exceptionBreakpointStates = ExceptionManager.ToExceptionBreakpointState(exceptionInfo.dwState);
_callback.OnException(thread, exceptionInfo.bstrExceptionName, "", 0, exceptionInfo.guidType, exceptionBreakpointStates);
```

I can simply use `_callback.OnException(thread, exceptionInfo.bstrExceptionName, "", 0, exceptionInfo.guidType, ExceptionBreakpointStates.BreakThrown);`

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gc46: Using `updates.NewCategoryState.HasValue && updates.NewCategoryState.Value != ExceptionBreakpointStates.None` will NOT isolate the C++ exceptions category is checked case.

More specifically, when C++ exceptions category is checked, `updates.NewCategoryState` is `ExceptionBreakpointStates.BreakThrown` AND every key in `RulesToAdd` has value `BreakUserHandled | BreakThrown`.

`updates.NewCategoryState` is ALSO `ExceptionBreakpointStates.BreakThrown` when "<All C++ Exceptions not in this list>" is checked. BUT every key in `RulesToAdd` only has value `BreakUserHandled`.

In short, the only way to distinguish between checked "C++ Exceptions" (category) and checked "<All C++ Exceptions not in this list>" is to check `RulesToAdd`.

src/MICore/CommandFactories/gdb.cs
gc46: Using `^(exceptionName)$` with

`const Foo`
image

but the catchpoint does not hit when an exception of type `Foo*` is raised.

gc46: With `\b(exceptionName)\b`, however, the catchpoint hits in both test scenarios.

`const Foo`
image

`Foo*`
image

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
gc46: > at the least they should both probably agree on enum_EXCEPTION_STATE vs ExceptionBreakpointStates.

Since `categorySettings.CategoryState` is always `BreakThrown` as you mentioned when `TryGetExceptionBreakpoint()` is called in the "breakpoint-hit" condition, we can remove the `exceptionInfo.dwState` assignment completely and replace

```
ExceptionBreakpointStates exceptionBreakpointStates = ExceptionManager.ToExceptionBreakpointState(exceptionInfo.dwState);
_callback.OnException(thread, exceptionInfo.bstrExceptionName, "", 0, exceptionInfo.guidType, exceptionBreakpointStates);
```

with `_callback.OnException(thread, exceptionInfo.bstrExceptionName, "", 0, exceptionInfo.guidType, ExceptionBreakpointStates.BreakThrown);`

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gregg-miskelly: Yup. Though in that case I would suggest not using `EXCEPTION_INFO`, and instead just return the guid+name from this method.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gregg-miskelly: You can get rid of `BreakUserUnhandled` if you want -- it isn't useful anymore.

> `updates.NewCategoryState` is ALSO ExceptionBreakpointStates.BreakThrown when "<All C++ Exceptions not in this list>" is checked. BUT every key in RulesToAdd only has value BreakUserHandled.

Any time "<All C++ Exceptions not in this list>" is checked -- you want to break on all C++ exceptions. Even the ones in the list. We can't support 'all but not these'.

src/MICore/CommandFactories/gdb.cs
gregg-miskelly: Excellent. So you want `\b` then.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gc46: Okay. Seems like I should remove `IsCppExceptionsCategoryChecked()` then.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gc46: I am tempted to keep `IsCppExceptionsCategoryChecked()` as we can use `if (!updates.RulesToAdd[rule].HasFlag(ExceptionBreakpointStates.BreakThrown))` to check for unchecked exceptions when category is checked and send the warning before returning false. Thoughts on this? 

And by sending a warning to the output window, we want to write to console, correct? Would a MessageBox be better as it is more visible? But then again, sending the warning for each unchecked exception in RulesToAdd would be bothersome from a customer's perspective.

```
private bool IsCppExceptionsCategoryChecked(SettingsUpdates updates)
{
    // C++ exceptions category is checked when NewCategoryState is BreakThrown and all RulesToAdd contain BreakThrown
    if (updates.NewCategoryState == ExceptionBreakpointStates.BreakThrown)
    {
        foreach (string rule in updates.RulesToAdd.Keys)
        {
            if (!updates.RulesToAdd[rule].HasFlag(ExceptionBreakpointStates.BreakThrown))
            {
                // WE CAN SEND WARNING HERE
                return false;
            }
        }
        return true;
    }
    return false;
}
```

gregg-miskelly: > I am tempted to keep `IsCppExceptionsCategoryChecked()`

You shouldn't need it. You can find if you have exceptions that are unchecked while the category is checked with:

```
else if (grouping.Key != categorySettings.CategoryState && !grouping.Key.HasFlag(ExceptionBreakpointStates.BreakThrown))
{
   // Send warning
}
```

gregg-miskelly: > And by sending a warning to the output window, we want to write to console, correct?  Would a MessageBox be better as it is more visible? But then again, sending the warning for each unchecked exception in RulesToAdd would be bothersome from a customer's perspective.

By the console do you mean the program's console? Or the output window? My suggestion was the debug pain of the output window. We don't want to write to the target program's console.

A message box would definitely be more visible, but it would also be more annoying if you have the same solution that you are going back and forth between Windows and Linux and you don't actually even care about stopping on C++ exceptions in Linux and none are raised, so it is a bit of a trade-off.

If there is a way to get extra text to show up when we stop on an exception indicating that we don't support unclicking specific exceptions then that would probably be the best option, and then I would definitely go with just an output window warning for exception configuration time.

If we can't make that happen then maybe we are both right: a message box the first time we see exceptions configured this way in a debug session, and an output window warning and any subsequent times.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -94,7 +96,7 @@ private class ExceptionCategorySettings
             public readonly ExceptionBreakpointStates DefaultCategoryState;
             public readonly ReadOnlyDictionary<string, ExceptionBreakpointStates> DefaultRules;
 
-            // Threading note: these are only read or updated by the FlushSettingsUpdates thread (in UpdateCatagory), and we
+            // Threading note: these are only read or updated by the FlushSettingsUpdates thread (in UpdateCategory), and we
             // guarantee that there will only be one active FlushSettingsUpdates task at a time
```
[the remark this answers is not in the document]
gc46: I can simply change this to 

`// Threading note: these are only read or updated by the FlushSettingsUpdates thread (in UpdateCategory) and TryGetExceptionBreakpoint, and we`

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gc46: Strange. I tried using `Debug.WriteLine("insert-string-here");`, but the warning did not show in the Output window...

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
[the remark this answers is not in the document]
gc46: Used `HostOutputWindow.WriteLaunchError("There are unchecked exceptions in a checked exceptions category.");` to resolve this issue. Thanks @caslan.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gc46: Used `HostOutputWindow.WriteLaunchError("There are unchecked exceptions in a checked exceptions category.");` to resolve this issue. Thanks @caslan.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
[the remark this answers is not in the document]
gregg-miskelly: I think I would instead suggest something like:

```
            // Threading: this is only read and updated by FlushSettingsUpdates thread, and there will only be one thread at a time
            public ExceptionBreakpointStates CategoryState;
            // Threading: this is locked by itself
            public readonly Dictionary<string, ulong> CurrentRules = new Dictionary<string, ulong>();
```

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -278,6 +279,31 @@ public void SetAllExceptions(enum_EXCEPTION_STATE dwState)
...
+            exceptionCategoryGuid = Guid.Empty;
+            ExceptionCategorySettings categorySettings;
+            if (_categoryMap.TryGetValue(CppExceptionCategoryGuid, out categorySettings))
+            {
+                ulong breakpointNumber = Convert.ToUInt32(bkptno, CultureInfo.InvariantCulture);
+                lock (categorySettings.CurrentRules)
+                {
+                    if (categorySettings.CurrentRules.ContainsValue(breakpointNumber))
+                    {
+                        exceptionName = categorySettings.CurrentRules.First(pair => pair.Value == breakpointNumber).Key;
```
gregg-miskelly, line 294: nit: instead of traversing the collection twice, use:

```
string exceptionName = categorySettings.CurrentRules.FirstOrDefault(pair => pair.Value == breakpointNumber)?.Key;
if (exceptionName != null)
{
...
}
```

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -473,24 +510,61 @@ private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings cat
...
 #if DEBUG
-                Debug.Assert(count == exceptionNames.Count());
+                        Debug.Assert(count == exceptionNames.Count());
 #endif
+                    }
+                }
+                else if (grouping.Key != categorySettings.CategoryState && !isBreakThrown)
+                {
+                    // Send warning when there are unchecked exceptions in a checked exceptions category
+                    HostOutputWindow.WriteLaunchError("There are unchecked exceptions in a checked exceptions category.");
```
gregg-miskelly, line 550: >HostOutputWindow.WriteLaunchError("There are unchecked exceptions in a checked exceptions category."); [](http://example.com/codeflow?start=20&length=102)

A few changes:

1. `HostOutputWindow.WriteLaunchError` is supposed to just be for launch errors, so you don't want that. Instead you want a new method on EngineCallback to send a message event with `enum_MESSAGETYPE.MT_OUTPUTSTRING`. Something like `_callback.OnOutputWindowWarning(string message)`
2. You want this string to be in a .resx so it can be localized.
3. I think I would suggest the text be tweaked to give the user more context. Something like:
`Warning: one or more C++ exceptions are unchecked while '<All C++ Exceptions not in this list>' is enabled. This is unsupported for GDB and the debugger will break on all C++ exceptions.\n'

gregg-miskelly:
Otherwise LGTM

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gc46: If I may ask, why is a new method on EngineCallback better than
```
_callback.OnOutputMessage(new OutputMessage("Warning: one or more C++ exceptions are unchecked while \'<All C++ Exceptions not in this list>\' is enabled. This is unsupported for GDB and the debugger will break on all C++ exceptions.\n", enum_MESSAGETYPE.MT_OUTPUTSTRING, OutputMessage.Severity.Warning));
```

gregg-miskelly: Oh, I missed that. That works.

gc46: Okay. Will add the string to the .resx and then will commit my changes. Thanks, @gregg-miskelly!

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -289,15 +289,16 @@ public bool TryGetExceptionBreakpoint(string bkptno, out string exceptionName, o
                 ulong breakpointNumber = Convert.ToUInt32(bkptno, CultureInfo.InvariantCulture);
                 lock (categorySettings.CurrentRules)
                 {
-                    if (categorySettings.CurrentRules.ContainsValue(breakpointNumber))
+                    exceptionName = categorySettings.CurrentRules.FirstOrDefault(pair => pair.Value == breakpointNumber).Key;
```
gregg-miskelly, line 292: You need `?.Key`

src/MIDebugEngine/ResourceStrings.resx
```diff
@@ -253,4 +253,7 @@ See https://aka.ms/miengine-gdb-troubleshooting for details.</value>
   <data name="DebuggingWillAbort" xml:space="preserve">
     <value>Debugging will now abort.</value>
   </data>
+  <data name="Warning_UncheckedExceptionsInCheckedCategory" xml:space="preserve">
+    <value>Warning: One or more C++ exceptions are unchecked while '&lt;All C++ Exceptions not in this list&gt;' is enabled. This is unsupported for GDB and the debugger will break on all C++ exceptions.</value>
```
gregg-miskelly, line 257: I am fairly certain you need a new line at the end of this (or to add it when creating the `OutputMessage` object).

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gc46: I tried this before removing the `?`, but received the below error:
image

gregg-miskelly: Oh, I am sorry.

src/MIDebugEngine/ResourceStrings.resx
gc46: You are correct. I do need a new line. However, I tried adding `\n` to the end of the string in the .resx file, and it shows up in the Output window as well. 

```
Warning: One or more C++ exceptions are unchecked while '<All C++ Exceptions not in this list>;' is enabled. This is unsupported for GDB and the debugger will break on all C++ exceptions.\n
```

This seems to be a common problem when adding the new line character in resx files...

gregg-miskelly: Just open up the resource editor in Visual Studio and add a real new line (hit enter)

WardenGnaw: In `.resx`, new lines are just `enter` instead of escaped new line character. 

E.g.
```xml
  <data name="Error_Message" xml:space="preserve">
    <value>First Line
Second Line
Third Line

Fifth Line</value>
```

gregg-miskelly (approved):

gregg-miskelly (approved):
Changes look good to me. Before we merge this, someone needs to do a quick test for:
* Android (someone on the core debugger team)
* Common embedded scenarios (someone on the C++ embedded team)

To make sure we don't need to better handle the case that GDB doesn't support `-catch-throw`

WardenGnaw:
RE: 
> Before we merge this, someone needs to do a quick test for: Android (someone on the core debugger team)

image

then it shows 
image

gregg-miskelly:
> ```
> 3: (216721) <-1024-catch-throw -r \bwinrt::hresult_error\b
> 3: (216727) ->1024^error,msg="Undefined MI command: catch-throw",code="undefined-command"
> ```

Sorry if this is obvious, but I think what we probably want to do is: 
* In gdb.cs: if we get a failure back with `code="undefined-command"`, convert that to `throw new NotSupportedException();`
* In ExceptionManager.cs: catch NotSupportedException, and cause that to:
    * Output a warning to the output window
    * Disable any further exception processing

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -425,22 +452,30 @@ private async Task FlushSettingsUpdates()
...
+
+                    // remove exception breakpoints before categorySettings.CurrentRules is cleared
+                    _commandFactory.RemoveExceptionBreakpoint(categoryId, categorySettings.CurrentRules.Values);
+
+                    categorySettings.CurrentRules.Clear();
+
+                    // only do a generic catch throw if C++ exceptions category is checked
+                    if (newCategoryState != ExceptionBreakpointStates.None)
+                    {
+                        IEnumerable<ulong> breakpointIds = _commandFactory.SetExceptionBreakpoints(categoryId, null, newCategoryState).Result;
```
WardenGnaw, line 475: It is bad practice to use `.Result` which causes it to be synchronous. 

```suggestion
                        await IEnumerable<ulong> breakpointIds = _commandFactory.SetExceptionBreakpoints(categoryId, null, newCategoryState);
```

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -425,22 +452,30 @@ private async Task FlushSettingsUpdates()
...
+                    updates.NewCategoryState.Value != ExceptionBreakpointStates.None || // send down a rule if the category isn't in the default state
+                    categorySettings.CurrentRules.Count != 0)) // Or if we have other rules for the category that we need to blow away
                 {
-                    ulong breakpointId = breakpointIds.Single();
-                    categorySettings.CurrentRules.Add("*", breakpointId);
+                    ExceptionBreakpointStates newCategoryState = updates.NewCategoryState.Value;
+                    categorySettings.CategoryState = newCategoryState;
+
+                    // remove exception breakpoints before categorySettings.CurrentRules is cleared
+                    _commandFactory.RemoveExceptionBreakpoint(categoryId, categorySettings.CurrentRules.Values);
```
WardenGnaw, line 468:
```suggestion
                    await _commandFactory.RemoveExceptionBreakpoint(categoryId, categorySettings.CurrentRules.Values);
```

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -425,22 +452,30 @@ private async Task FlushSettingsUpdates()
...
         }
 
-        private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings categorySettings, SettingsUpdates updates)
+        private async Task UpdateCategory(Guid categoryId, ExceptionCategorySettings categorySettings, SettingsUpdates updates)
         {
             // Update the category
-            if (updates.NewCategoryState.HasValue && (
-                updates.NewCategoryState.Value != ExceptionBreakpointStates.None || // send down a rule if the category isn't in the default state
-                categorySettings.CurrentRules.Count != 0)) // Or if we have other rules for the category that we need to blow away
+            lock (categorySettings.CurrentRules)
```
WardenGnaw, line 458: Move this lock to be just around 
`categorySettings.CurrentRules.Clear();` and `categorySettings.CurrentRules.Add("*", breakpointId);` so you can use await on `_commandFactory`.

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -94,7 +96,7 @@ private class ExceptionCategorySettings
             public readonly ExceptionBreakpointStates DefaultCategoryState;
             public readonly ReadOnlyDictionary<string, ExceptionBreakpointStates> DefaultRules;
 
-            // Threading note: these are only read or updated by the FlushSettingsUpdates thread (in UpdateCatagory), and we
+            // Threading note: these are only read or updated by the FlushSettingsUpdates thread (in UpdateCategory) and TryGetExceptionBreakpoint, and we
```
WardenGnaw, line 99: Update this comment to say: `// Since this is only written on the FlushSettingsUpdate thread, we can freely read this without a lock. However, all updates must take the lock.`

WardenGnaw (changes requested):
I was able to fix the hang with the following changes:

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
gregg-miskelly: I would suggest:

```
// Threading note: these are only read or updated by the FlushSettingsUpdates thread (in UpdateCategory) and TryGetExceptionBreakpoint. The following rules apply:
// 1. Writes and reads when *NOT* on the FlushSettingsUpdates thread - collection should be locked on itself
// 2. Reads on the FlushSettingsUpdates thread - no locking is needed
```

gc46:
> > ```
> > 3: (216721) <-1024-catch-throw -r \bwinrt::hresult_error\b
> > 3: (216727) ->1024^error,msg="Undefined MI command: catch-throw",code="undefined-command"
> > ```
> 
> Sorry if this is obvious, but I think what we probably want to do is:
> 
> * In gdb.cs: if we get a failure back with `code="undefined-command"`, convert that to `throw new NotSupportedException();`
> * In ExceptionManager.cs: catch NotSupportedException, and cause that to:
>   
>   * Output a warning to the output window
>   * Disable any further exception processing

As the current implementation for `SetExceptionBreakpoints()` calls the following:

```
result = await _debugger.CmdAsync(command, ResultClass.done);
```

is there a better way to detect when ResultClass is ResultClass.error? Thanks!

gregg-miskelly:
> is there a better way to detect/integrate when ResultClass is ResultClass.error? Thanks!

You want something like:

```
   result = await _debugger.CmdAsync(command, ResultClass.None);
   switch (results.ResultClass)
   {
     case ResultClass.done:
      ...
     case ResultClass.errror:
      ...
     default:
     ...
   }
```

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -473,24 +524,69 @@ private async Task UpdateCatagory(Guid categoryId, ExceptionCategorySettings cat
...
+                                return 1;
+                            }).Sum();
 
 #if DEBUG
-                Debug.Assert(count == exceptionNames.Count());
+                            Debug.Assert(count == exceptionNames.Count());
 #endif
+                        }
+                    }
+                    catch (NotSupportedException ex)
```
WardenGnaw, line 563:
```suggestion
                    catch (NotSupportedException)
```

WardenGnaw (changes requested):
Sorry if you are still working on this:
> In ExceptionManager.cs: catch NotSupportedException, and cause that to:
> - Disable any further exception processing

You will want to add a flag in the `ExceptionManager` class to represent that it can process exceptions and default it to `true`. You will set it to false when you see the `NotImplementedException`.

You will want to check this flag in `SetException`, `RemoveSetException`,  `RemoveAllSetExceptions`, and `SetAllExceptions`, and do nothing..

WardenGnaw (approved):

PR #1251, opened 2021-12-13
Add 'All Exceptions' for ExceptionFilter in OpenDebugAD7

WardenGnaw:
This PR adds an exception filter in the VS Code window "All Exceptions".

When checked, any exception thrown will be stopped at as a breakpoint.
image

This PR also updates the exception description for the AD7Exception
event. The description will indicate the address and function that it
was thrown.

**Image of `All C++ Exceptions` check**
image

**Image of `All C++ Exceptions` with a condition**
image

**Image of a specific exception in Visual Studio 2022**
image

TODO: 
- [ ] Update https://github.com/microsoft/vscode-cpptools/blob/main/Documentation/Debugger/ExceptionSettings.md

### DIFF
--- a/src/DebugEngineHost.VSCode/VSCode/ExceptionBreakpointFilter.cs
+++ b/src/DebugEngineHost.VSCode/VSCode/ExceptionBreakpointFilter.cs
@@ -45,12 +45,13 @@ namespace Microsoft.DebugEngineHost.VSCode
         [JsonRequired]
         public string conditionDescription { get; set; }
 
+        [JsonRequired]
+        public Guid categoryId { get; set; }
+
         /// <summary>
         /// The default state for the button
         /// </summary>
         public bool @default { get; set; }
-
-        public Guid categoryId { get; set; }
 
         [JsonIgnore]
         public enum_EXCEPTION_STATE State { get; set; } = enum_EXCEPTION_STATE.EXCEPTION_STOP_SECOND_CHANCE | enum_EXCEPTION_STATE.EXCEPTION_STOP_FIRST_CHANCE | enum_EXCEPTION_STATE.EXCEPTION_STOP_USER_FIRST_CHANCE;
--- a/src/DebugEngineHost.VSCode/VSCode/ExceptionBreakpointFilter.cs
+++ b/src/DebugEngineHost.VSCode/VSCode/ExceptionBreakpointFilter.cs
@@ -46,6 +46,12 @@ namespace Microsoft.DebugEngineHost.VSCode
             }
         }
 
+        [JsonRequired]
+        public bool supportsCondition { get; set; }
+
+        [JsonRequired]
+        public string conditionDescription { get; set; }
+
         /// <summary>
         /// The default state for the button
         /// </summary>
--- a/src/DebugEngineHost.VSCode/VSCode/ExceptionBreakpointFilter.cs
+++ b/src/DebugEngineHost.VSCode/VSCode/ExceptionBreakpointFilter.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.VisualStudio.Debugger.Interop;
 using Newtonsoft.Json;
 using System;
 using System.Diagnostics;
@@ -13,9 +14,6 @@ namespace Microsoft.DebugEngineHost.VSCode
     /// </summary>
     sealed public class ExceptionBreakpointFilter
     {
-        public const string Filter_All = "all";
-        public const string Filter_UserUnhandled = "user-unhandled";
-
         private string _filter;
 
         /// <summary>
@@ -25,7 +23,7 @@ namespace Microsoft.DebugEngineHost.VSCode
         public string label { get; set; }
 
         /// <summary>
-        /// The identifier for this filter. Currently this should be 'all' or 'user-unhandled'
+        /// The identifier for this filter.
         /// </summary>
         [JsonRequired]
         public string filter
@@ -37,11 +35,6 @@ namespace Microsoft.DebugEngineHost.VSCode
 
             set
             {
-                if (value != Filter_All && value != Filter_UserUnhandled)
-                {
-                    Debug.Fail("Invalid ExceptionBreakpointFilter");
-                    throw new ArgumentOutOfRangeException("filter");
-                }
                 _filter = value;
             }
         }
@@ -56,5 +49,11 @@ namespace Microsoft.DebugEngineHost.VSCode
         /// The default state for the button
         /// </summary>
         public bool @default { get; set; }
+
+        public Guid categoryId { get; set; }
+
+        [JsonIgnore]
+        public enum_EXCEPTION_STATE State { get; set; } = enum_EXCEPTION_STATE.EXCEPTION_STOP_SECOND_CHANCE | enum_EXCEPTION_STATE.EXCEPTION_STOP_FIRST_CHANCE | enum_EXCEPTION_STATE.EXCEPTION_STOP_USER_FIRST_CHANCE;
+
     }
 }
--- a/src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
+++ b/src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
@@ -308,19 +308,27 @@ namespace Microsoft.MIDebugEngine
                     exceptionName = categorySettings.CurrentRules.FirstOrDefault(pair => pair.Value == breakpointNumber).Key;
                     if (exceptionName != null)
                     {
+                        // The string to use when displaying which exception caused the breakpoint to hit.
+                        // It is empty if it uses the category name
+                        string displayException = string.Empty;
+
                         if (exceptionName.Length < 1 || exceptionName == "*") // if exceptionName is "*", the exceptions category is selected
                         {
                             exceptionName = categorySettings.CategoryName;
+                        }
+                        else
+                        {
+                            displayException = string.Format(CultureInfo.InvariantCulture, " '{0}'", exceptionName);
                         }
 
                         string functionName = frame?.TryFindString("func");
                         if (string.IsNullOrWhiteSpace(functionName))
                         {
-                            exceptionDescription = string.Format(CultureInfo.CurrentCulture, ResourceStrings.Exception_Thrown, address);
+                            exceptionDescription = string.Format(CultureInfo.CurrentCulture, ResourceStrings.Exception_Thrown, displayException, address);
                         }
                         else
                         {
-                            exceptionDescription = string.Format(CultureInfo.CurrentCulture, ResourceStrings.Exception_Thrown_with_Source, address, functionName);
+                            exceptionDescription = string.Format(CultureInfo.CurrentCulture, ResourceStrings.Exception_Thrown_with_Source, displayException, address, functionName);
                         }
 
                         exceptionCategoryGuid = CppExceptionCategoryGuid;
--- a/src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
+++ b/src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
@@ -294,9 +294,10 @@ namespace Microsoft.MIDebugEngine
             }
         }
 
-        public bool TryGetExceptionBreakpoint(string bkptno, out string exceptionName, out Guid exceptionCategoryGuid)
+        public bool TryGetExceptionBreakpoint(string bkptno, ulong address, TupleValue frame, out string exceptionName, out string exceptionDescription, out Guid exceptionCategoryGuid)
         {
-            exceptionName = null;
+            exceptionName = string.Empty;
+            exceptionDescription = string.Empty;
             exceptionCategoryGuid = Guid.Empty;
             ExceptionCategorySettings categorySettings;
             if (_categoryMap.TryGetValue(CppExceptionCategoryGuid, out categorySettings))
@@ -311,6 +312,17 @@ namespace Microsoft.MIDebugEngine
                         {
                             exceptionName = categorySettings.CategoryName;
                         }
+
+                        string functionName = frame?.TryFindString("func");
+                        if (string.IsNullOrWhiteSpace(functionName))
+                        {
+                            exceptionDescription = string.Format(CultureInfo.CurrentCulture, ResourceStrings.Exception_Thrown, address);
+                        }
+                        else
+                        {
+                            exceptionDescription = string.Format(CultureInfo.CurrentCulture, ResourceStrings.Exception_Thrown_with_Source, address, functionName);
+                        }
+
                         exceptionCategoryGuid = CppExceptionCategoryGuid;
                         return true;
 
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -671,15 +671,22 @@ namespace OpenDebugAD7
             return flags;
         }
 
-        private void SetAllExceptions(enum_EXCEPTION_STATE state)
+        private void SetAllExceptions(Guid categoryId, enum_EXCEPTION_STATE state)
         {
-            foreach (ExceptionSettings.CategoryConfiguration category in m_engineConfiguration.ExceptionSettings.Categories)
+            ExceptionSettings.CategoryConfiguration category = m_engineConfiguration.ExceptionSettings.Categories.FirstOrDefault(x => x.Id == categoryId);
+
+            if (category != null)
             {
                 var exceptionInfo = new EXCEPTION_INFO[1];
                 exceptionInfo[0].dwState = state;
-                exceptionInfo[0].guidType = category.Id;
+                exceptionInfo[0].guidType = categoryId;
                 exceptionInfo[0].bstrExceptionName = category.Name;
+
                 m_engine.SetException(exceptionInfo);
+            }
+            else
+            {
+                Debug.Fail(categoryId + " is a referencing a non-existant category. This should have been caught in ExceptionSettings.ValidateExceptionFilters.");
             }
         }
 
@@ -873,6 +880,13 @@ namespace OpenDebugAD7
                 });
             }
 
+            // -catch-throw is not supported in lldb-mi
+            List<ExceptionBreakpointsFilter> filters = new List<ExceptionBreakpointsFilter>();
+            if (!Utilities.IsOSX())
+            {
+                filters = m_engineConfiguration.ExceptionSettings.ExceptionBreakpointFilters.Select(item => new ExceptionBreakpointsFilter() { Default = item.@default, Filter = item.filter, Label = item.label, SupportsCondition = item.supportsCondition, ConditionDescription = item.conditionDescription }).ToList();
+            }
+
             InitializeResponse initializeResponse = new InitializeResponse()
             {
                 SupportsConfigurationDoneRequest = true,
@@ -881,9 +895,8 @@ namespace OpenDebugAD7
                 SupportsSetVariable = true,
                 SupportsFunctionBreakpoints = m_engineConfiguration.FunctionBP,
                 SupportsConditionalBreakpoints = m_engineConfiguration.ConditionalBP,
-                ExceptionBreakpointFilters = m_engineConfiguration.ExceptionSettings.ExceptionBreakpointFilters.Select(item => new ExceptionBreakpointsFilter() { Default = item.@default, Filter = item.filter, Label = item.label, SupportsCondition = item.supportsCondition, ConditionDescription = item.conditionDescription }).ToList(),
-                SupportsExceptionConditions = true,
-                SupportsExceptionFilterOptions = true,
+                ExceptionBreakpointFilters = filters,
+                SupportsExceptionFilterOptions = filters.Any(),
                 SupportsClipboardContext = m_engineConfiguration.ClipboardContext,
                 SupportsLogPoints = true,
                 SupportsReadMemoryRequest = m_engine is IDebugMemoryBytesDAP, // TODO: Read from configuration or query engine for capabilities.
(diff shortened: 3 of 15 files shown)


gregg-miskelly:
I am assuming you know this, but just in case: if we want to, we could add support for exception conditions to allow customizing the types that we stop on.

src/OpenDebugAD7/cppdbg.ad7Engine.json
```diff
@@ -7,9 +7,15 @@
...
     "categories": [
       {
-        "name": "CppExceptionCategory",
+        "name": "C++ Exceptions",
         "id": "3A12D0B7-C26C-11D0-B442-00A0244A1DD2"
       }
+    ],
+    "exceptionBreakpointFilters": [
+      {
+        "label": "All Exceptions",
```
gregg-miskelly, line 16: >Exceptions [](http://example.com/codeflow?start=22&length=10)

I wonder if we should name this `All C++ Exceptions` since someday we might support signals?

src/MIDebugEngine/Engine.Impl/ExceptionManager.cs
```diff
@@ -311,6 +312,17 @@ public bool TryGetExceptionBreakpoint(string bkptno, out string exceptionName, o
                         {
                             exceptionName = categorySettings.CategoryName;
                         }
+
+                        string functionName = frame?.TryFindString("func");
+                        if (string.IsNullOrWhiteSpace(functionName))
+                        {
+                            exceptionDescription = string.Format(CultureInfo.CurrentCulture, ResourceStrings.Exception_Thrown, address);
```
gregg-miskelly, line 319: >Exception_Thrown [](http://example.com/codeflow?start=109&length=16)

Does this string show up in VS? If so, should we include the name of the exception that we are stopping on when this isn't a category exception?

WardenGnaw:
> I am assuming you know this, but just in case: if we want to, we could add support for exception conditions to allow customizing the types that we stop on.

How do we want to handle the condition? Do we have special keywords like `name` so the condition can be `name == \"std::out_of_range\"`

Should I change up [HandleSetExceptionBreakpointsRequestAsync](https://github.com/microsoft/MIEngine/blob/876f81514cf2e62e566a452a1ccc647f7a5f9ec2/src/OpenDebugAD7/AD7DebugSession.cs#L2303-L2338) so it only supports the all filters case? I'm assuming most of the code here was MI for clrdbg.

gregg-miskelly:
> How do we want to handle the condition? Do we have special keywords like `name` so the condition can be `name == \"std::out_of_range\"`

Here is what the C# extension did: [docs](https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger.md#exception-settings)
It seems like doing the same thing here would make sense.

> Should I change up [HandleSetExceptionBreakpointsRequestAsync](https://github.com/microsoft/MIEngine/blob/876f81514cf2e62e566a452a1ccc647f7a5f9ec2/src/OpenDebugAD7/AD7DebugSession.cs#L2303-L2338) so it only supports the all filters case?

It wouldn't break anything to remove all the user-unhandled support, but it also wouldn't be a problem to leave it. Up to you. Yup, all that was for clrdbg.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2302,34 +2307,91 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
-                }
-                else if (filter.Contains(ExceptionBreakpointFilter.Filter_All))
-                {
-                    enum_EXCEPTION_STATE state = enum_EXCEPTION_STATE.EXCEPTION_STOP_SECOND_CHANCE | enum_EXCEPTION_STATE.EXCEPTION_STOP_FIRST_CHANCE | enum_EXCEPTION_STATE.EXCEPTION_STOP_USER_FIRST_CHANCE;
+                    if (filterOption.FilterId == ExceptionBreakpointFilter.Filter_All)
+                    {
+                        if (!string.IsNullOrWhiteSpace(filterOption.Condition))
+                        {
+                            string[] conditions = filterOption.Condition.Split(',');
+                            ExceptionSettings.CategoryConfiguration category = m_engineConfiguration.ExceptionSettings.Categories.First(); // How to map filterOption.FilterId to Category Name / Guid.
```
WardenGnaw, line 2320: I was not sure how protocol handled the exception category for our engine's `EXCEPTION_INFO` structure.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -879,7 +881,10 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
                 SupportsSetVariable = true,
                 SupportsFunctionBreakpoints = m_engineConfiguration.FunctionBP,
                 SupportsConditionalBreakpoints = m_engineConfiguration.ConditionalBP,
-                ExceptionBreakpointFilters = m_engineConfiguration.ExceptionSettings.ExceptionBreakpointFilters.Select(item => new ExceptionBreakpointsFilter() { Default = item.@default, Filter = item.filter, Label = item.label }).ToList(),
+                ExceptionBreakpointFilters = m_engineConfiguration.ExceptionSettings.ExceptionBreakpointFilters.Select(item => new ExceptionBreakpointsFilter() { Default = item.@default, Filter = item.filter, Label = item.label, SupportsCondition = item.supportsCondition, ConditionDescription = item.conditionDescription }).ToList(),
+                SupportsExceptionConditions = true,
+                SupportsExceptionOptions = true,
```
WardenGnaw, line 886: I think I turned on too many capabilities.
```suggestion
```

gregg-miskelly: Yup

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: If we wanted to support this, `ExceptionBreakpointFilter` should have a category GUID that we used internally.

WardenGnaw: Do I just map the "all" filterId to the GUID?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2302,34 +2307,91 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                                m_engine.SetException(exceptionInfo);
+
+                                m_exceptionBreakpoints.Add(condition);
+                            }
+
+                            IEnumerable<string> toRemove = m_exceptionBreakpoints.Except(conditions).ToList();
+                            foreach (string condition in toRemove)
+                            {
+                                if (m_exceptionBreakpoints.Contains(condition))
+                                {
```
gregg-miskelly, line 2340: How could this fail?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2302,34 +2307,91 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                            IEnumerable<string> toAdd = conditions.Except(m_exceptionBreakpoints);
+                            foreach (string condition in toAdd)
+                            {
+                                var exceptionInfo = new EXCEPTION_INFO[1];
+                                exceptionInfo[0].dwState = state;
+                                exceptionInfo[0].guidType = category.Id;
+                                exceptionInfo[0].bstrExceptionName = condition;
+                                m_engine.SetException(exceptionInfo);
+
+                                m_exceptionBreakpoints.Add(condition);
```
gregg-miskelly, line 2333: >m_exceptionBreakpoints [](http://example.com/codeflow?start=32&length=22)

Does `Except` get angry that the collections are changing while enumerating?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2302,34 +2307,91 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
-                    SetAllExceptions(enum_EXCEPTION_STATE.EXCEPTION_STOP_SECOND_CHANCE);
-                }
-                else if (filter.Contains(ExceptionBreakpointFilter.Filter_All))
-                {
-                    enum_EXCEPTION_STATE state = enum_EXCEPTION_STATE.EXCEPTION_STOP_SECOND_CHANCE | enum_EXCEPTION_STATE.EXCEPTION_STOP_FIRST_CHANCE | enum_EXCEPTION_STATE.EXCEPTION_STOP_USER_FIRST_CHANCE;
+                    if (filterOption.FilterId == ExceptionBreakpointFilter.Filter_All)
+                    {
+                        if (!string.IsNullOrWhiteSpace(filterOption.Condition))
+                        {
+                            string[] conditions = filterOption.Condition.Split(',');
```
gregg-miskelly, line 2319: >conditions [](http://example.com/codeflow?start=37&length=10)

Do we want to validate the name contains valid C++ identifier customers (in case people try to write something like `== Foo` instead of `Foo`)?

gregg-miskelly: VS Code supports returning failures back for these

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: Yep.

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: If you wanted to fix this for real I would suggest:
* Get rid of the special understanding of 'all', and make any exception filter act like 'all' in that it sets the state to: `enum_EXCEPTION_STATE.EXCEPTION_STOP_SECOND_CHANCE | enum_EXCEPTION_STATE.EXCEPTION_STOP_FIRST_CHANCE | enum_EXCEPTION_STATE.EXCEPTION_STOP_USER_FIRST_CHANCE`
* Have each filter associated with a category

This would be enough to support signals in the future.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: Ah, residual from when I was originally looping through all of the conditions in m_exceptionBreakpoints. 

Will remove

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: Looks like VSCodeDebugAdapterHost has not been updated to return the BreskpointList.

I'll mark this in an issue for TODO.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: I hope the latest commit is how you expected this fix to be.

src/DebugEngineHost.VSCode/VSCode/ExceptionBreakpointFilter.cs
```diff
@@ -37,18 +35,25 @@ public string filter
...
+
+        [JsonRequired]
+        public string conditionDescription { get; set; }
+
         /// <summary>
         /// The default state for the button
         /// </summary>
         public bool @default { get; set; }
+
+        public Guid categoryId { get; set; }
```
gregg-miskelly, line 53: >categoryId [](http://example.com/codeflow?start=20&length=10)

Should this be `[JsonRequired]`?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -879,7 +881,9 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
                 SupportsSetVariable = true,
                 SupportsFunctionBreakpoints = m_engineConfiguration.FunctionBP,
                 SupportsConditionalBreakpoints = m_engineConfiguration.ConditionalBP,
-                ExceptionBreakpointFilters = m_engineConfiguration.ExceptionSettings.ExceptionBreakpointFilters.Select(item => new ExceptionBreakpointsFilter() { Default = item.@default, Filter = item.filter, Label = item.label }).ToList(),
+                ExceptionBreakpointFilters = m_engineConfiguration.ExceptionSettings.ExceptionBreakpointFilters.Select(item => new ExceptionBreakpointsFilter() { Default = item.@default, Filter = item.filter, Label = item.label, SupportsCondition = item.supportsCondition, ConditionDescription = item.conditionDescription }).ToList(),
+                SupportsExceptionConditions = true,
```
gregg-miskelly, line 885: >SupportsExceptionConditions = true, [](http://example.com/codeflow?start=16&length=35)

You don't want `SupportsExceptionConditions`. That is a VS-only extension which is part of exception options.

```
        "supportsExceptionConditions": {
          "type": "boolean",
          "description": "The debug adapter supports using exception conditions in exception options."
        }
```

src/OpenDebugAD7/LanguageUtilities.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <returns>'true' if this is a valid identifier name. 'false' if otherwise.</returns>
+        public static bool IsValidIdentifier(string identifier)
+        {
+            // Check to see if identifier is null or blank.
+            if (string.IsNullOrWhiteSpace(identifier))
+            {
+                return false;
+            }
+
+            string[] parts = identifier.Split();
```
gregg-miskelly, line 24: >string[] parts = identifier.Split(); [](http://example.com/codeflow?start=12&length=36)

Is there a reason you are splitting this?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2302,34 +2306,88 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                    if (filter != null)
+                    { 
+                        if (!string.IsNullOrWhiteSpace(filterOption.Condition))
+                        {
+                            string[] conditions = filterOption.Condition.Split(',');
+
+                            // Validate condition strings
+                            List<string> validConditions = new List<string>();
+                            foreach (string condition in conditions)
+                            {
```
gregg-miskelly, line 2325: >{ [](http://example.com/codeflow?start=28&length=1)

I think you want to do a `.Trim()` on the condition.

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: So don't you need a `.ToList()` (and maybe a .Distinct()?) on line 2336?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2302,34 +2306,88 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                                {
+                                    validConditions.Add(condition);
+                                }
+                                else
+                                {
+                                    m_logger.WriteLine(LoggingCategory.StdErr, string.Format(CultureInfo.CurrentCulture, AD7Resources.Error_Invalid_Exception_Condition, condition));
+                                }
+                            }
+
+                            IEnumerable<string> toAdd = conditions.Except(m_exceptionBreakpoints);
```
gregg-miskelly, line 2336: >conditions [](http://example.com/codeflow?start=56&length=10)

You want `validConditions`

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2302,34 +2306,88 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                                var exceptionInfo = new EXCEPTION_INFO[1];
+                                exceptionInfo[0].dwState = filter.State;
+                                exceptionInfo[0].guidType = filter.categoryId;
+                                exceptionInfo[0].bstrExceptionName = condition;
+                                m_engine.SetException(exceptionInfo);
+
+                                m_exceptionBreakpoints.Add(condition);
+                            }
+
+                            IEnumerable<string> toRemove = m_exceptionBreakpoints.Except(conditions).ToList();
```
gregg-miskelly, line 2348: >conditions [](http://example.com/codeflow?start=89&length=10)

Same

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2302,34 +2306,88 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                                exceptionInfo[0].guidType = filter.categoryId;
+                                exceptionInfo[0].bstrExceptionName = condition;
+
+                                m_engine.RemoveSetException(exceptionInfo);
+                                m_exceptionBreakpoints.Remove(condition);
+                            }
+                        }
+                        else
+                        {
+                            SetAllExceptions(filter.State);
```
gregg-miskelly, line 2362: >SetAllExceptions [](http://example.com/codeflow?start=28&length=16)

This method should really go away and we should use `filter.categoryId`

src/OpenDebugAD7/AD7Resources.resx
```diff
@@ -304,4 +304,12 @@
   <data name="Error_UnableToSetInstructionBreakpoint" xml:space="preserve">
     <value>Error setting instruction breakpoint: {0}</value>
   </data>
+  <data name="Error_FilterOption_Not_Supported" xml:space="preserve">
+    <value>'{0}' is not a supported exception filter.</value>
+    <comment>{0} is the name of the exception filter.</comment>
+  </data>
+  <data name="Error_Invalid_Exception_Condition" xml:space="preserve">
+    <value>ERROR: '{0}' is an invalid exception condition</value>
```
gregg-miskelly, line 312: >condition [](http://example.com/codeflow?start=48&length=9)

Include aka.ms link to documentation?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2302,34 +2306,88 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
-                    }
-                    else
+                    foreach (string filter in filters)
                     {
-                        // TODO: once VS Code has UI to break on more than just 'uncaught' and 'all' we will need to enhance this with more features
-                        Debug.Fail("Unexpected exception filter string");
+                        ExceptionBreakpointFilter breakpointFilter = m_engineConfiguration.ExceptionSettings.ExceptionBreakpointFilters.FirstOrDefault(ebf => ebf.filter == filter);
+                        if (breakpointFilter != null)
+                        {
+                            SetAllExceptions(breakpointFilter.State);
```
gregg-miskelly, line 2385: >SetAllExceptions [](http://example.com/codeflow?start=28&length=16)

This should use `filter.categoryId` for `guidType`.

test/CppTests/Tests/ExceptionTests.cs
```diff
@@ -281,6 +281,94 @@ public void RaisedReThrowException(ITestSettings settings)
             }
         }
 
+        [Theory]
```
gregg-miskelly, line 284: >Theory [](http://example.com/codeflow?start=9&length=6)

Disable new tests for LLDB?

gregg-miskelly (changes requested):
Otherwise LGTM

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2302,34 +2306,88 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
-
-                    SetAllExceptions(state);
+                }
+            }
+            else
+            {
+                List<string> filters = responder.Arguments.Filters;
+                if (filters == null || filters.Count == 0)
+                {
+                    SetAllExceptions(enum_EXCEPTION_STATE.EXCEPTION_STOP_SECOND_CHANCE);
```
gregg-miskelly, line 2376: I don't know if you want to fix this now, but we should probably remember which filters in `m_engineConfiguration.ExceptionSettings.ExceptionBreakpointFilters` were NOT referenced, and set those categories to second chance.

test/CppTests/Tests/ExceptionTests.cs
WardenGnaw: Oops, forgot if this was supported on macOS. 

I'll have to update the capabilities too

src/OpenDebugAD7/AD7Resources.resx
WardenGnaw: Also need to update https://github.com/microsoft/vscode-cpptools/blob/main/Documentation/Debugger/ExceptionSettings.md for the cppdbg case.

src/OpenDebugAD7/LanguageUtilities.cs
WardenGnaw: I was going to do something more specific but I think I'm good with just validating first character and equal comparisons.

test/CppTests/Tests/ExceptionTests.cs
WardenGnaw: Are these tests valid for VsDbg?

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: This path also gets hit when we disable the filter, so wouldn't it also need to include the referenced filter?

gregg-miskelly: Right, so what we should do is something create a dictionary from filterName->ExceptionBreakpointFilter. Whenever we find that filter in either `responder.Arguments.FilterOptions` or `responder.Arguments.Filters` we remove it from the dictionary. Any element left in the dictionary is the disabled filters, and they should be set to `enum_EXCEPTION_STATE.EXCEPTION_STOP_SECOND_CHANCE`.

The code works for now, because there is only one filter. But it wouldn't work correctly with multiple.

test/CppTests/Tests/ExceptionTests.cs
gregg-miskelly: VsDbg also calls it `all` and has the same syntax (+ some more stuff), so it seems like it should work (famous last words).

WardenGnaw: Also there seems to be a bug with exceptions with the MinGW GDB version on the CI Windows machine.

> &(did not find exception probe (does libstdcxx have SDT probes?)

gregg-miskelly: It might not be in GDB but rather the version of libstdcxx that MinGW uses.

Do we output a reasonable warning when this happens?

WardenGnaw: It's seems to be an output from GDB when it tries to evaluate the exception breakpoint and we forward it on to the UI as a log message.

gregg-miskelly (approved):

PR #1257, opened 2022-01-07
Data breakpoints in OpenDebugAD7

calgagi:
This PR implements `DataBreakpointInfo` and `SetDataBreakpoints` for OpenDebugAD7's debug session as a part of the [Debug Adapter Protocol (DAP)](https://microsoft.github.io/debug-adapter-protocol/specification).

Notes:
1. Refactors most of `HandleEvaluateRequestAsync` into a `TryEvaluate` function that `DataBreakpointInfo` can use for evaluating a variable to find it's address/size (gotten via `IDebugProperty160.GetDataBreakpointInfo160`).
2. Makes `AD7Property.EnumChildren` support filtering with the child's name (specified [here with pszNameFilter](https://docs.microsoft.com/en-us/visualstudio/extensibility/debugger/reference/idebugproperty2-enumchildren?view=vs-2022))
3. Supports `BPLT_DATA_STRING` in `AD7BreakpointRequest` (returns the proper address/size).
4. The DataId passed to/from VS Code as part of both DAP functions is "{address},{size}"

Potential problems/things to improve:
1. If users have too many data breakpoints, then we receive an exception that GDB can't continue the debuggee process and we display a very mysterious error message: "Debugger was unable to continue the process."
2. If a variable size is too big to set a data breakpoint on, then we simply don't display the "Break When Value Changes" context menu item. It might be worthwhile in the future to say why it doesn't display.
3. Add test :)

### DIFF
--- a/src/DebugEngineHost.VSCode/HostMarshal.cs
+++ b/src/DebugEngineHost.VSCode/HostMarshal.cs
@@ -4,6 +4,7 @@
 using Microsoft.DebugEngineHost.VSCode;
 using Microsoft.VisualStudio.Debugger.Interop;
 using System;
+using System.Runtime.InteropServices;
 
 namespace Microsoft.DebugEngineHost
 {
@@ -110,7 +111,7 @@ namespace Microsoft.DebugEngineHost
         /// <returns></returns>
         public static string GetDataBreakpointStringForIntPtr(IntPtr stringId)
         {
-            throw new NotImplementedException();
+            return Marshal.PtrToStringBSTR(stringId);
         }
 
         /// <summary>
@@ -120,7 +121,7 @@ namespace Microsoft.DebugEngineHost
         /// <returns>IntPtr to a BSTR which can be returned to VS.</returns>
         public static IntPtr GetIntPtrForDataBreakpointAddress(string address)
         {
-            throw new NotImplementedException();
+            return Marshal.StringToBSTR(address);
         }
 
         /// <summary>
--- a/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
@@ -292,9 +292,6 @@ namespace Microsoft.MIDebugEngine
             {
                 if (e.GetBaseException() is InvalidCoreDumpOperationException)
                     return AD7_HRESULT.E_CRASHDUMP_UNSUPPORTED;
-                else if (e.GetBaseException() is ArgumentException && (enum_BP_LOCATION_TYPE)_bpRequestInfo.bpLocation.bpLocationType == enum_BP_LOCATION_TYPE.BPLT_DATA_STRING)
-                    // handle invalid data bp address/size
-                    return Constants.E_FAIL;
                 else
                     return EngineUtils.UnexpectedException(e);
             }
@@ -323,6 +320,11 @@ namespace Microsoft.MIDebugEngine
                 //replaced by the real mi debugger error text
                 this.SetError(new AD7ErrorBreakpoint(this, ResourceStrings.LongBind, enum_BP_ERROR_TYPE.BPET_SEV_LOW | enum_BP_ERROR_TYPE.BPET_TYPE_WARNING), true);
                 return Constants.S_FALSE;
+            }
+            else if (this._BPError != null)
+            {
+                // Ran into some sort of error
+                return Constants.E_FAIL;
             }
             else
             {
@@ -367,7 +369,16 @@ namespace Microsoft.MIDebugEngine
                 }
                 else
                 {
-                    bindResult = await PendingBreakpoint.Bind(_address, _size, _engine.DebuggedProcess, _condition, this);
+                    try
+                    {
+                        bindResult = await PendingBreakpoint.Bind(_address, _size, _engine.DebuggedProcess, _condition, this);
+                    }
+                    catch (ArgumentException ex)
+                    {
+                        // There was an error binding at the address with the specified size. This could happen if the size is greater
+                        // than the max size a data bp can watch.
+                        bindResult = new PendingBreakpoint.BindResult(ex.Message);
+                    }
                 }
 
                 lock (_boundBreakpoints)
--- a/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
@@ -3,7 +3,6 @@
 
 using MICore;
 using Microsoft.DebugEngineHost;
-using Microsoft.MIDebugEngine.Natvis;
 using Microsoft.VisualStudio.Debugger.Interop;
 using System;
 using System.Collections.Generic;
--- a/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
@@ -3,6 +3,7 @@
 
 using MICore;
 using Microsoft.DebugEngineHost;
+using Microsoft.MIDebugEngine.Natvis;
 using Microsoft.VisualStudio.Debugger.Interop;
 using System;
 using System.Collections.Generic;
@@ -292,6 +293,9 @@ namespace Microsoft.MIDebugEngine
             {
                 if (e.GetBaseException() is InvalidCoreDumpOperationException)
                     return AD7_HRESULT.E_CRASHDUMP_UNSUPPORTED;
+                else if (e.GetBaseException() is ArgumentException && (enum_BP_LOCATION_TYPE)_bpRequestInfo.bpLocation.bpLocationType == enum_BP_LOCATION_TYPE.BPLT_DATA_STRING)
+                    // handle invalid data bp address/size
+                    return Constants.E_FAIL;
                 else
                     return EngineUtils.UnexpectedException(e);
             }
--- a/src/MIDebugEngine/AD7.Impl/AD7Property.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Property.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using EnvDTE;
 using Microsoft.VisualStudio.Debugger.Interop;
 using System;
 using System.Diagnostics;
@@ -141,10 +142,31 @@ namespace Microsoft.MIDebugEngine
                 {
                     _engine.DebuggedProcess.Natvis.WaitDialog.ShowWaitDialog(_variableInformation.Name);
                     var children = _engine.DebuggedProcess.Natvis.Expand(_variableInformation);
-                    DEBUG_PROPERTY_INFO[] properties = new DEBUG_PROPERTY_INFO[children.Length];
+
+                    // Count number of children that fit filter (results saved in "fitsFilter")
+                    int propertyCount = children.Length;
+                    bool[] fitsFilter = null;
+                    if (!string.IsNullOrEmpty(pszNameFilter))
+                    {
+                        fitsFilter = new bool[children.Length];
+                        for (int i = 0; i < children.Length; i++)
+                        {
+                            fitsFilter[i] = string.Compare(children[i].Name, pszNameFilter, StringComparison.Ordinal) == 0;
+                            if (!fitsFilter[i])
+                            {
+                                propertyCount--;
+                            }
+                        }
+                    }
+
+                    // Create property array
+                    DEBUG_PROPERTY_INFO[] properties = new DEBUG_PROPERTY_INFO[propertyCount];
                     for (int i = 0; i < children.Length; i++)
                     {
-                        properties[i] = (new AD7Property(_engine, children[i])).ConstructDebugPropertyInfo(dwFields);
+                        if (fitsFilter == null || fitsFilter[i])
+                        {
+                            properties[i] = (new AD7Property(_engine, children[i])).ConstructDebugPropertyInfo(dwFields);
+                        }
                     }
                     ppEnum = new AD7PropertyEnum(properties);
                     return Constants.S_OK;
--- a/src/MIDebugEngine/AD7.Impl/AD7Property.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Property.cs
@@ -161,11 +161,12 @@ namespace Microsoft.MIDebugEngine
 
                     // Create property array
                     DEBUG_PROPERTY_INFO[] properties = new DEBUG_PROPERTY_INFO[propertyCount];
-                    for (int i = 0; i < children.Length; i++)
+                    for (int i = 0, j = 0; i < children.Length; i++)
                     {
                         if (fitsFilter == null || fitsFilter[i])
                         {
-                            properties[i] = (new AD7Property(_engine, children[i])).ConstructDebugPropertyInfo(dwFields);
+                            properties[j] = (new AD7Property(_engine, children[i])).ConstructDebugPropertyInfo(dwFields);
+                            ++j; // increment j if we fit filter, this allows us to traverse "properties" array properly.
                         }
                     }
                     ppEnum = new AD7PropertyEnum(properties);
@@ -466,7 +467,7 @@ namespace Microsoft.MIDebugEngine
         {
             try
             {
-                pbstrAddress = _variableInformation.Address() + "," + _variableInformation.FullName();
+                pbstrAddress = _variableInformation.Address();
                 pSize = _variableInformation.Size();
                 pbstrDisplayName = _variableInformation.FullName();
                 pbstrError = "";
--- a/src/MIDebugEngine/AD7.Impl/AD7Property.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Property.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using EnvDTE;
 using Microsoft.VisualStudio.Debugger.Interop;
 using System;
 using System.Diagnostics;
@@ -151,7 +150,7 @@ namespace Microsoft.MIDebugEngine
                         fitsFilter = new bool[children.Length];
                         for (int i = 0; i < children.Length; i++)
                         {
-                            fitsFilter[i] = string.Compare(children[i].Name, pszNameFilter, StringComparison.Ordinal) == 0;
+                            fitsFilter[i] = string.Equals(children[i].Name, pszNameFilter, StringComparison.Ordinal);
                             if (!fitsFilter[i])
                             {
                                 propertyCount--;
@@ -467,7 +466,7 @@ namespace Microsoft.MIDebugEngine
         {
             try
             {
-                pbstrAddress = _variableInformation.Address();
+                pbstrAddress = _variableInformation.Address() + "," + _variableInformation.FullName();
                 pSize = _variableInformation.Size();
                 pbstrDisplayName = _variableInformation.FullName();
                 pbstrError = "";
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -1,4 +1,5 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+﻿
+// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
@@ -51,6 +52,8 @@ namespace OpenDebugAD7
 
         private Dictionary<string, IDebugPendingBreakpoint2> m_functionBreakpoints;
         private Dictionary<ulong, IDebugPendingBreakpoint2> m_instructionBreakpoints;
+        private Dictionary<string, IDebugPendingBreakpoint2> m_dataBreakpoints;
+
         private List<string> m_exceptionBreakpoints;
         private readonly HandleCollection<IDebugStackFrame2> m_frameHandles;
 
@@ -122,6 +125,7 @@ namespace OpenDebugAD7
             m_breakpoints = new Dictionary<string, Dictionary<int, IDebugPendingBreakpoint2>>();
             m_functionBreakpoints = new Dictionary<string, IDebugPendingBreakpoint2>();
             m_instructionBreakpoints = new Dictionary<ulong, IDebugPendingBreakpoint2>();
+            m_dataBreakpoints = new Dictionary<string, IDebugPendingBreakpoint2>();
             m_exceptionBreakpoints = new List<string>();
             m_variableManager = new VariableManager();
         }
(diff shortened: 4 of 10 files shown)


src/DebugEngineHost.VSCode/HostMarshal.cs
```diff
@@ -110,7 +111,7 @@ public static IDebugFunctionPosition2 GetDebugFunctionPositionForIntPtr(IntPtr f
         /// <returns></returns>
         public static string GetDataBreakpointStringForIntPtr(IntPtr stringId)
         {
-            throw new NotImplementedException();
+            return Marshal.PtrToStringBSTR(stringId);
```
gregg-miskelly, line 114: >PtrToStringBSTR [](http://example.com/codeflow?start=27&length=15)

I am not sure this API will actually work on Linux/Mac.

src/DebugEngineHost.VSCode/HostMarshal.cs
```diff
@@ -120,7 +121,7 @@ public static string GetDataBreakpointStringForIntPtr(IntPtr stringId)
         /// <returns>IntPtr to a BSTR which can be returned to VS.</returns>
         public static IntPtr GetIntPtrForDataBreakpointAddress(string address)
         {
-            throw new NotImplementedException();
+            return Marshal.StringToBSTR(address);
```
gregg-miskelly, line 124: >return [](http://example.com/codeflow?start=12&length=6)

You probably should have a matching free function. I am assuming right now we just leak the string.

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -292,6 +292,9 @@ int IDebugPendingBreakpoint2.Bind()
             {
                 if (e.GetBaseException() is InvalidCoreDumpOperationException)
                     return AD7_HRESULT.E_CRASHDUMP_UNSUPPORTED;
+                else if (e.GetBaseException() is ArgumentException && (enum_BP_LOCATION_TYPE)_bpRequestInfo.bpLocation.bpLocationType == enum_BP_LOCATION_TYPE.BPLT_DATA_STRING)
+                    // handle invalid data bp address/size
+                    return Constants.E_FAIL;
```
gregg-miskelly, line 297: I don't think this is where you want this - if binding takes a while we will give up synchronisouly waiting.

gregg-miskelly: You probably want BindAsync to return a Task<int> instead. Or maybe you don't want to return E_FAIL at all but rather raise a breakpoint errer event.

src/MIDebugEngine/AD7.Impl/AD7Property.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using EnvDTE;
```
gregg-miskelly, line 4: What is using this? I don't think you want this. If so, we probably need to ship another dll.

src/MIDebugEngine/AD7.Impl/AD7Property.cs
```diff
@@ -141,10 +142,32 @@ public int EnumChildren(enum_DEBUGPROP_INFO_FLAGS dwFields, uint dwRadix, ref Gu
...
                     _engine.DebuggedProcess.Natvis.WaitDialog.ShowWaitDialog(_variableInformation.Name);
                     var children = _engine.DebuggedProcess.Natvis.Expand(_variableInformation);
-                    DEBUG_PROPERTY_INFO[] properties = new DEBUG_PROPERTY_INFO[children.Length];
-                    for (int i = 0; i < children.Length; i++)
+
+                    // Count number of children that fit filter (results saved in "fitsFilter")
+                    int propertyCount = children.Length;
+                    bool[] fitsFilter = null;
+                    if (!string.IsNullOrEmpty(pszNameFilter))
+                    {
```
gregg-miskelly, line 150: Rather than doing it this way, I wonder if we want OpenDebugAD7 keep a collection of children by name. This is what vsdbg does (see m_variablesChildren in the Concord repo).

src/MIDebugEngine/AD7.Impl/AD7Property.cs
```diff
@@ -141,10 +142,32 @@ public int EnumChildren(enum_DEBUGPROP_INFO_FLAGS dwFields, uint dwRadix, ref Gu
...
+                        fitsFilter = new bool[children.Length];
+                        for (int i = 0; i < children.Length; i++)
+                        {
+                            fitsFilter[i] = string.Compare(children[i].Name, pszNameFilter, StringComparison.Ordinal) == 0;
```
gregg-miskelly, line 154: >string.Compare [](http://example.com/codeflow?start=44&length=14)

nit: `string.Equals`

src/MIDebugEngine/AD7.Impl/AD7Property.cs
```diff
@@ -444,7 +467,7 @@ public int GetDataBreakpointInfo160(out string pbstrAddress, out uint pSize, out
         {
             try
             {
-                pbstrAddress = _variableInformation.Address() + "," + _variableInformation.FullName();
```
gregg-miskelly, line 447: > + "," + _variableInformation.FullName(); [](http://example.com/codeflow?start=61&length=41)

I don't *think* you want this. I think [this code](https://github.com/microsoft/MIEngine/blob/00a3aee314701adb598846acd23808f1248a70ba/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs#L248) depends on the address containing this.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -905,6 +908,7 @@ protected override void HandleInitializeRequestAsync(IRequestResponder<Initializ
                 SupportsSetVariable = true,
                 SupportsFunctionBreakpoints = m_engineConfiguration.FunctionBP,
                 SupportsConditionalBreakpoints = m_engineConfiguration.ConditionalBP,
+                SupportsDataBreakpoints = true,
```
gregg-miskelly, line 911: >true [](http://example.com/codeflow?start=42&length=4)

You want to define an engine configuration option for this.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2331,229 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+            DataBreakpointInfoResponse response = new DataBreakpointInfoResponse();
+            string name = responder.Arguments.Name;
+            IDebugStackFrame2 frame = null;
+            IDebugProperty2 property = null;
+            string errorMessage = null;
+
+            // Did our request come with a parent object?
+            if (responder.Arguments.VariablesReference.HasValue)
+            {
+                m_variableManager.TryGet(responder.Arguments.VariablesReference.Value, out object variableObj);
```
gregg-miskelly, line 2351: >m_variableManager [](http://example.com/codeflow?start=16&length=17)

Fail if this returns false?

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2331,229 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                        {
+                            DEBUG_PROPERTY_INFO[] propertyInfo = new DEBUG_PROPERTY_INFO[1];
+                            if (children.Next(childrenCount, propertyInfo, out _) == HRConstants.S_OK)
+                            {
+                                property = propertyInfo[0].pProperty;
+                            }
+                        }
+                        else
+                        {
+                            errorMessage = $"Unable to get property from parent: there are {childrenCount} children properties with name {name}";
```
gregg-miskelly, line 2375: >$"Unable to get property from parent: there are [](http://example.com/codeflow?start=43&length=47)

Localize string

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2331,229 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+            }
+
+            DataBreakpointInfoResponse response = new DataBreakpointInfoResponse();
+            string name = responder.Arguments.Name;
+            IDebugStackFrame2 frame = null;
+            IDebugProperty2 property = null;
+            string errorMessage = null;
+
+            // Did our request come with a parent object?
+            if (responder.Arguments.VariablesReference.HasValue)
```
gregg-miskelly, line 2349: >if [](http://example.com/codeflow?start=12&length=2)

I will repeat the suggestion here too -- rather than re-evaluating, I wonder if we want to keep a collection of <parent,name>->evaluation result to avoid needing to reenumerate children to return this info (and other requests that work the same way).

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2331,229 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                return;
+            }
+
+            // If we don't have a property for what we're trying to set the data bp on yet, attempt to get it by evaluating the variable name.
+            if (property == null)
+            {
+                // If we don't have a frame, just use top frame.
+                if (frame == null)
+                {
+                    m_frameHandles.TryGetFirst(out frame);
```
gregg-miskelly, line 2398: >m_frameHandles [](http://example.com/codeflow?start=20&length=14)

Throw if this is null.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2331,229 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                            }
+                        }
+                        else
+                        {
+                            errorMessage = $"Unable to get property from parent: there are {childrenCount} children properties with name {name}";
+                        }
+                    }
+                    else
+                    {
+                        errorMessage = "ParentProperty.EnumChildren failed";
```
gregg-miskelly, line 2380: >errorMessage = "ParentProperty.EnumChildren failed"; [](http://example.com/codeflow?start=24&length=52)

Use ErrorBuilder.CheckHR

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2331,229 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
 
+        protected override void HandleDataBreakpointInfoRequestAsync(IRequestResponder<DataBreakpointInfoArguments, DataBreakpointInfoResponse> responder)
+        {
+            if (responder.Arguments.Name == null)
+            {
+                responder.SetError(new ProtocolException("SetDataBreakpointRequest failed: Missing 'breakpoints'."));
+                return;
+            }
+
+            DataBreakpointInfoResponse response = new DataBreakpointInfoResponse();
```
gregg-miskelly, line 2342: >DataBreakpointInfoResponse [](http://example.com/codeflow?start=12&length=26)

Wrap body in try/catch (all of these methods should, but I wouldn't advise fixing that in this PR)

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2331,229 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                if (property160.GetDataBreakpointInfo160(out string address, out uint size, out string displayName, out errorMessage) == HRConstants.S_OK)
+                {
+                    // Test to see if we can set a data bp
+                    AD7BreakPointRequest pBPRequest = new AD7BreakPointRequest(address, size);
+                    int hr = m_engine.CreatePendingBreakpoint(pBPRequest, out IDebugPendingBreakpoint2 pendingBp);
+                    if (hr == HRConstants.S_OK && pendingBp != null)
+                    {
+                        hr = pendingBp.Bind();
+                    }
+                    pendingBp?.Delete();
```
gregg-miskelly, line 2422: Remove. You don't want to bind breakpoints as port of DataBreakpointInfoRequest. If we need any more error checking it should be done in GetDataBreakpointInfo160.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2331,229 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+
+                    if (hr == HRConstants.S_OK)
+                    {
+                        // If we succeeded, return response that we can set a data bp.
+                        string dataId = $"{address},{size.ToString(CultureInfo.InvariantCulture)}";
+                        response.DataId = dataId;
```
gregg-miskelly, line 2428: >dataId [](http://example.com/codeflow?start=42&length=6)

This should be `address`.

gregg-miskelly: Nevermind.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2331,229 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                        response.Description = $"{address}, {displayName}";
```
gregg-miskelly, line 2429: >Description [](http://example.com/codeflow?start=33&length=11)

This is supposed to be human readable, so it should just be `displayName`

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2331,229 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                        }
+                    }
+                }
+
+                // Bind the new data bp
+                if (!m_dataBreakpoints.ContainsKey(b.DataId))
+                {
+                    int hr = HRConstants.S_OK;
+
+                    string[] splits = b.DataId.Split(',');
```
gregg-miskelly, line 2518: >string[] splits = b.DataId.Split(','); [](http://example.com/codeflow?start=20&length=38)

You don't want to split it, because the engine provided engine id can contain a comma. Instead, find the last index of a comma and split it into two.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2331,229 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+            else
+            {
+                response.Description = "Property could not be found or does not implement IDebugProperty160.";
+            }
+
+            responder.SetResponse(response);
+        }
+
+        protected override void HandleSetDataBreakpointsRequestAsync(IRequestResponder<SetDataBreakpointsArguments, SetDataBreakpointsResponse> responder)
+        {
```
gregg-miskelly, line 2451: >{ [](http://example.com/codeflow?start=8&length=1)

Same as the other, this method you have a try/catch around it.

src/OpenDebugAD7/AD7DebugSession.cs
gregg-miskelly: Also you want to throw if there is no comma.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2331,229 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                        hr = pendingBp.Bind();
+                    }
+
+                    if (hr == HRConstants.S_OK)
+                    {
+                        newBreakpoints[b.DataId] = pendingBp;
+                        response.Breakpoints.Add(new Breakpoint()
+                        {
+                            Id = (int)pBPRequest.Id,
+                            Verified = true,
```
gregg-miskelly, line 2536: >Verified = true, [](http://example.com/codeflow?start=28&length=16)

<Aside>I am fairly certain this wrong, but looks like we have the same bug throughout OpenDebugAD7 -- a breakpoint isn't verified until we get a bound breakpoint event from the engine.</Aside>

src/MIDebugEngine/AD7.Impl/AD7Property.cs
calgagi: Ahhhh got it, I missed the `this.AddressId` taking the FullName as well

src/DebugEngineHost.VSCode/HostMarshal.cs
gregg-miskelly: Actually, I guess this is a general bug we have in MIEngine.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: To expand on this:
1. Add `dataBP` to [cppdbg.ad7Engine.json](https://github.com/microsoft/MIEngine/blob/main/src/OpenDebugAD7/cppdbg.ad7Engine.json)
2. Match it with [EngineConfiguration.cs](https://github.com/microsoft/MIEngine/blob/cb8e84f8f350da178b75f43c5582dae71e60c166/src/DebugEngineHost.VSCode/VSCode/EngineConfiguration.cs)

src/DebugEngineHost.VSCode/HostMarshal.cs
calgagi: I tested it on my laptop which already has Debian 11 installed, seems to work fine.

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
calgagi: Yup, worked it out in order to raise a breakpoint error event (but also still return E_FAIL so that the caller won't return an invalid address)

src/OpenDebugAD7/AD7DebugSession.cs
calgagi: We could have data breakpoints with the same name then, is that okay? For ex, if we have a recursive function with a local variable in it and a data bp on the same variable in 2 different frames.

gregg-miskelly: VS + native engine provides the address as the name in the breakpoints window. Here is what I see as an example:

image

Does VS + MIEngine do something similar? 
If not: Sounds like a deficiency in the implementation of GetDataBreakpointInfo160 in MIEngine.
If so: I am not sure how to fix this well, as `address` is supposed to be something internal

calgagi: Looks like the format for VS + MIEngine is:

image

I'll emulate this format

src/MIDebugEngine/AD7.Impl/AD7Property.cs
calgagi: Should we remove these changes now? I think it's fine to add, but also this is no longer used.

src/OpenDebugAD7/VariableManager.cs
```diff
@@ -56,6 +77,24 @@ internal int Create(VariableScope scope)
             return m_variableHandles.Create(scope);
         }
 
+        public void AddChildVariable(int parentHandle, IDebugProperty2 prop)
```
gregg-miskelly, line 80: >IDebugProperty2 prop [](http://example.com/codeflow?start=55&length=20)

You might want to pass in a property info instead of a property, as the caller would already have the name

src/OpenDebugAD7/VariableManager.cs
```diff
@@ -56,6 +77,24 @@ internal int Create(VariableScope scope)
...
+        {
+            var propInfo = new DEBUG_PROPERTY_INFO[1];
+            prop.GetPropertyInfo(enum_DEBUGPROP_INFO_FLAGS.DEBUGPROP_INFO_NAME, Constants.EvaluationRadix, Constants.EvaluationTimeout, null, 0, propInfo);
+            string name = propInfo[0].bstrName;
+
+            m_variablesChildren.Add(Tuple.Create(parentHandle, name), prop);
+        }
+
+        public void AddFrameVariable(IDebugStackFrame2 frame, IDebugProperty2 prop)
```
gregg-miskelly, line 89: >IDebugProperty2 prop [](http://example.com/codeflow?start=62&length=20)

Same

src/OpenDebugAD7/AD7Impl/AD7BreakPointRequest.cs
```diff
@@ -25,13 +25,16 @@ static public uint GetNextBreakpointId()
...
+        public uint DataSize { get; private set; }
+
         // Used for Releasing the MemoryContext.
         // Caller of AD7BreakPointRequest(MemoryContext) is required to
         // release it with HostMarshal.ReleaseCodeContextId
         public IntPtr MemoryContextIntPtr { get; private set;  }
 
         // Unique identifier for breakpoint when communicating with VSCode
-        public uint Id { get; private set; }
+        public uint Id { get; private set; } = GetNextBreakpointId();
```
gregg-miskelly, line 37: >private set; [](http://example.com/codeflow?start=30&length=12)

You can remove `private set;` now

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2333,198 @@
...
+                    }
+                }
+                else if (response.Description == null)
+                {
+                    response.Description = string.Format(CultureInfo.CurrentCulture, AD7Resources.Error_DataBreakpointInfoFail, AD7Resources.Error_ChildPropertyNotFound);
+                }
+            }
+            catch (Exception ex)
+            {
+                response.Description = string.Format(CultureInfo.CurrentCulture, AD7Resources.Error_DataBreakpointInfoFail, "");
```
gregg-miskelly, line 2415: >response.Description [](http://example.com/codeflow?start=16&length=20)

Should this be `responder.SetError(...)` instead and move `responder.SetResponse` into the try block?

gregg-miskelly: Oh, I see.

gregg-miskelly, line 2415: >string.Format(CultureInfo.CurrentCulture, AD7Resources.Error_DataBreakpointInfoFail, ""); [](http://example.com/codeflow?start=39&length=89)

If this is an AD7Exception, I believe you want to use the Message property.

src/MIDebugEngine/AD7.Impl/AD7Property.cs
gregg-miskelly: Up to you.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2333,198 @@
...
+                        {
+                            newBreakpoints[b.DataId] = pendingBp;
+                        }
+                        response.Breakpoints.Add(pBPRequest.BindResult);
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+            }
```
gregg-miskelly, line 2519: >} [](http://example.com/codeflow?start=12&length=1)

I think you want something like `responder.SetError(ex.Message)` and move the `SetResponse` to the end of the try.

src/OpenDebugAD7/AD7DebugSession.cs
calgagi: Got it, I was wondering if I should've included `ex.Message` into the string format. So if it's an AD7Exception, we should be good

gregg-miskelly:
Otherwise LGTM

src/MIDebugEngine/AD7.Impl/AD7Property.cs
calgagi: I'll leave them

WardenGnaw (approved):

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2327,6 +2333,201 @@ protected override void HandleSetBreakpointsRequestAsync(IRequestResponder<SetBr
...
+                }
+                else if (response.Description == null)
+                {
+                    response.Description = string.Format(CultureInfo.CurrentCulture, AD7Resources.Error_DataBreakpointInfoFail, AD7Resources.Error_ChildPropertyNotFound);
+                }
+            }
+            catch (Exception ex)
+            {
+                if (ex is AD7Exception ad7ex)
+                    response.Description = string.Format(CultureInfo.CurrentCulture, AD7Resources.Error_DataBreakpointInfoFail, ad7ex.Message);
```
gregg-miskelly, line 2416: >string.Format(CultureInfo.CurrentCulture, AD7Resources.Error_DataBreakpointInfoFail, ad7ex.Message); [](http://example.com/codeflow?start=43&length=100)

I think you just want `ad7ex.Message`. It should already have the `Error_DataBreakpointInfoFail` prefix.

calgagi: Wow good catch, because of the ErrorBuilder!

gregg-miskelly (approved):

PR #1276, opened 2022-02-22
Remove MI Debugger from Help -> About

WardenGnaw:
MI Debugger does not need to show up in Help -> About, this is only for
top level "brands".
 
@davkean


davkean:
I think you also want to delete: https://github.com/microsoft/MIEngine/blob/main/src/MIDebugPackage/Resources/Package.ico.

chuckries (approved):

PR #1281, opened 2022-03-01
Visual Studio: add support for custom UI visualizers

paulmaybee:
Add support for natvis UIVisualizers and VSIX defined natvis. 
1. At startup (when VS hosted debugging) the host queries VS for VSIX defined natvis files and adds those to the set searched for visualizers.
2. UIVisualizer natvis elements are parsed and returned with property values when they are found.

### DIFF
--- a/src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
+++ b/src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
@@ -395,6 +395,16 @@ public sealed class HostLogger
         }
 
         /// <summary>
+        /// Searches the extension manager for natvis files, invoking the loader on any which are found.
+        /// </summary>
+        /// <param name="loader">Natvis loader method to invoke</param>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Natvis")]
+        public static void FindNatvisInVSIX(NatvisLoader loader)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
         /// Return the solution's root directory, null if no solution
         /// </summary>
         public static string FindSolutionRoot()
--- a/src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
+++ b/src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
@@ -385,21 +385,11 @@ public sealed class HostLogger
         public delegate void NatvisLoader(string path);
 
         /// <summary>
-        /// Searches the solution for natvis files, invoking the loader on any which are found.
+        /// Searches the solution and VSIXs for natvis files, invoking the loader on any which are found.
         /// </summary>
         /// <param name="loader">Natvis loader method to invoke</param>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Natvis")]
-        public static void FindNatvisInSolution(NatvisLoader loader)
-        {
-            throw new NotImplementedException();
-        }
-
-        /// <summary>
-        /// Searches the extension manager for natvis files, invoking the loader on any which are found.
-        /// </summary>
-        /// <param name="loader">Natvis loader method to invoke</param>
-        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Natvis")]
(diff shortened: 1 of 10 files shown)


src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
```diff
@@ -394,6 +394,16 @@ public static void FindNatvisInSolution(NatvisLoader loader)
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Searches the extension manager for natvis files, invoking the loader on any which are found.
+        /// </summary>
+        /// <param name="loader">Natvis loader method to invoke</param>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Natvis")]
+        public static void FindNatvisInVSIX(NatvisLoader loader)
```
gregg-miskelly, line 402: >FindNatvisInVSIX [](http://example.com/codeflow?start=27&length=16)

nit: How about `FindNatvisExtensions`?

src/DebugEngineHost/HostNatvisProject.cs
```diff
@@ -207,6 +237,26 @@ public static string FindSolutionRootImpl()
...
+                    var rgbstrContentLocations = new string[contentLocations];
+                    var rgbstrUniqueStrings = new string[contentLocations];
+
+                    var hr = pem.GetEnabledExtensionContentLocations(name, contentLocations, rgbstrContentLocations, rgbstrUniqueStrings, out var actualContentLocations);
+                    if (hr == VSConstants.S_OK && actualContentLocations > 0)
+                    {
+                        for (var i = 0; i < actualContentLocations; ++i)
+                        {
+                            paths.Add(rgbstrContentLocations[i]);
+                        }
```
gregg-miskelly, line 255: nit: any reason to not just return this array instead of adding it to a list?

src/MIDebugEngine/AD7.Impl/AD7Property.cs
```diff
@@ -111,6 +115,15 @@ public DEBUG_PROPERTY_INFO ConstructDebugPropertyInfo(enum_DEBUGPROP_INFO_FLAGS
                     propertyInfo.dwAttrib |= enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_VALUE_RAW_STRING;
                 }
                 propertyInfo.dwAttrib |= variable.Access;
+
+                if (_uiVisualizers != null && _uiVisualizers.Any())
+                {
+                    propertyInfo.dwAttrib |= enum_DBG_ATTRIB_FLAGS.DBG_ATTRIB_VALUE_CUSTOM_VIEWER;
+                    if (_uiVisualizers.Count() > 1)
```
gregg-miskelly, line 122: >.Count() [](http://example.com/codeflow?start=38&length=8)

nit: this will wind up doing too passes over the enumerable. Should we just turn the enumerable into an array in the constructor?

src/DebugEngineHost.Stub/DebugEngineHost.ref.cs
gregg-miskelly: Actually, can we just change `FindNatvisInSolution` to `FindNatvis` and have it do both? `FindNatvis` MIEngine currently violates VS Threading rules and does all this stuff on a background thread which turns around and calls shell services. It would be good to just have one spot to fix instead of more than one.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -133,6 +135,15 @@ internal class VisualizerInfo
             public VisualizerType Visualizer { get; private set; }
             public Dictionary<string, string> ScopedNames { get; private set; }
 
+            public IEnumerable<(string,int)> GetUIVisualizers()
```
gregg-miskelly, line 138: >(string,int) [](http://example.com/codeflow?start=31&length=12)

nit: we use this in a bunch of spots. Probably better to create a struct for it.

gregg-miskelly (approved):
Otherwise LGTM

gregg-miskelly:
nit: Please add a real description when you merge.

src/DebugEngineHost/HostNatvisProject.cs
paulmaybee: works cleaner with caller as is, especially now that caller is the new "FindNatvis".

gregg-miskelly: In that case: switch to `paths.AddRange(rgbstrContentLocations);`

gregg-miskelly (approved):

PR #1284, opened 2022-03-03
Fix linux CI Build

WardenGnaw:
This pr adds `sudo apt update` to make the `sudo apt-get install` succeed.


WardenGnaw:
I'll be investigating macOS in a different PR since that one is really flaky but not completely dead.

gregg-miskelly (approved):

PR #1295, opened 2022-03-25
Add threadId to thread name

WardenGnaw:
This PR adds a threadId to the Threads response to help users see what
the thread is in the CallStacks view.

**Old:**
image

**New:**
image

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -9,6 +9,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.DebugEngineHost;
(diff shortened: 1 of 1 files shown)


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2043,11 +2044,24 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
...
             // iterate over the collection asking the engine for the name
-            foreach (var pair in threads)
+            foreach ((int id, IDebugThread2 thread) in threads)
             {
-                string name;
-                pair.Value.GetName(out name);
-                response.Threads.Add(new OpenDebugThread(pair.Key, name));
+                string name = null;
+
+                if (thread != null && thread.GetName(out name) == HRConstants.S_OK)
```
gregg-miskelly, line 2051: >thread != nul [](http://example.com/codeflow?start=20&length=13)

I don't know how thread could ever be null (probably couldn't be), but if we want to handle that case, we should just ignore it rather than adding it to the collection.

gregg-miskelly, line 2051: >thread.GetName(out name) == HRConstants.S_OK [](http://example.com/codeflow?start=38&length=44)

Seems like we should either:
1. Treat GetName failures as fatal for that thread, and ignore the thread.
-or-
2. Treat GetName failures the same as if string.Empty was returned, and use a name of `[<thread-id-here>]`

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -2043,11 +2044,24 @@ protected override void HandleThreadsRequestAsync(IRequestResponder<ThreadsArgum
...
+                {
+                    // Append the thread id as a suffix unless the engine is already including it
+                    if (thread.GetThreadId(out uint threadId) == HRConstants.S_OK)
```
gregg-miskelly, line 2054: >thread.GetThreadId(out uint threadId) == HRConstants.S_OK [](http://example.com/codeflow?start=24&length=57)

If thread.GetThreadId fails I think we should just ignore the thread.

gregg-miskelly (approved):
Aside from those nits, LGTM

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: This will be handled as #2 in the OpenDebugThread class.

See https://github.com/microsoft/MIEngine/blob/496f56bbed40ab22f3cbaea58cb7e6ebf18f310c/src/OpenDebugAD7/OpenDebug/CustomProtocolObjects.cs#L75-L78

gregg-miskelly (approved):

PR #1300, opened 2022-03-30
Bug 1447876: Local window variable value can not be synchronized with…

gc46:
… Immediate window value change

We want `_hasGottenLocalsAndParams` to be false when the user modifies the value of a Locals variable in the Immediate window. By resetting `_hasGottenLocalsAndParams` to its default value before the call to `EnsureLocalsAndParameters()`, the Locals variable values are updated correctly.

**Note: there is a very noticeable lag to the change in the Locals window for the current implementation.**

**Before**
image

**After**
image

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
@@ -22,7 +22,6 @@ namespace Microsoft.MIDebugEngine
 
         private string _functionName;
         private MITextPosition _textPosition;
-        private bool _hasGottenLocalsAndParams = false;
         private uint _radix;
         private AD7MemoryAddress _codeCxt;
         private AD7DocumentContext _documentCxt;
@@ -249,7 +248,7 @@ namespace Microsoft.MIDebugEngine
                     await Engine.UpdateRadixAsync(radix);
                 });
             }
-            if (_textPosition != null && (!_hasGottenLocalsAndParams || radix != _radix))
+            if (_textPosition != null && radix != _radix)
             {
                 _radix = radix;
                 List<VariableInformation> localsAndParameters = null;
@@ -271,8 +270,6 @@ namespace Microsoft.MIDebugEngine
                         _locals.Add(vi);
                     }
                 }
-
-                _hasGottenLocalsAndParams = true;
             }
         }
 
@@ -399,15 +396,6 @@ namespace Microsoft.MIDebugEngine
                     guidFilter == AD7Guids.guidFilterLocals ||
                     guidFilter == AD7Guids.guidFilterLocalsPlusArgs)
                 {
-                    /*
-                    Question: do we want to filter on guidFilter?
-
-                    if (guidFilter == AD7Guids.guidFilterLocalsPlusArgs)
-                    {
-                        _hasGottenLocalsAndParams = false;
-                    }
-                    */
-                    _hasGottenLocalsAndParams = false;
                     EnsureLocalsAndParameters();
                 }
 
--- a/src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
@@ -26,13 +26,6 @@ namespace Microsoft.MIDebugEngine
         private AD7MemoryAddress _codeCxt;
         private AD7DocumentContext _documentCxt;
 
-        // An array of this frame's parameters
-        private readonly List<VariableInformation> _parameters = new List<VariableInformation>();
-
-        // An array of this frame's locals
-        private readonly List<VariableInformation> _locals = new List<VariableInformation>();
-
-
         public AD7StackFrame(AD7Engine engine, AD7Thread thread, ThreadContext threadContext)
         {
             Debug.Assert(threadContext != null, "ThreadContext is null");
@@ -238,8 +231,11 @@ namespace Microsoft.MIDebugEngine
             }
         }
 
-        private void EnsureLocalsAndParameters()
+        private void GetLocalsAndParameters(out List<VariableInformation> locals, out List<VariableInformation> parameters)
         {
+            parameters = new List<VariableInformation>();
+            locals = new List<VariableInformation>();
+
             uint radix = Engine.CurrentRadix();
             if (radix != Engine.DebuggedProcess.MICommandFactory.Radix)
             {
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
```diff
@@ -399,6 +399,15 @@ int IDebugStackFrame2.EnumProperties(enum_DEBUGPROP_INFO_FLAGS dwFields, uint nR
...
                 {
+                    /*
+                    Question: do we want to filter on guidFilter?
+
+                    if (guidFilter == AD7Guids.guidFilterLocalsPlusArgs)
+                    {
+                        _hasGottenLocalsAndParams = false;
+                    }
+                    */
+                    _hasGottenLocalsAndParams = false;
```
gregg-miskelly, line 410: >_hasGottenLocalsAndParams = false; [](http://example.com/codeflow?start=20&length=34)

This change makes `_hasGottenLocalsAndParams`/`_parameters`/`_locals` useless as a cache -- we always recalcuale them. So, assuming this is the right way to go (which I think is true), we should get rid of those member variables.

I wouldn't expect that the UI would repeatedly call EnumProperties. So I think ditching the cache is probably the way to go. Is that what you saw?

gregg-miskelly: As long as you aren't seeing the locals repeatedly queried in cases where we should be able to cache things -- yes, I think it makes sense to just delete the cache. The other option would be to change your fix to invalidate the cache when particular things happened, like maybe when we evaluate from the immediate window that should invalidate. But I think I am of the opinion that we don't really know any better than our caller when things have changed, so we shouldn't be caching.

gc46: I can't find a viable way to remove `_parameters` and `_locals`. Do you have any suggestions for how to do so? If I remove both arrays and their associated functions completely, the Locals window will be completely blank.

gregg-miskelly: You want to change the code to:
1. Delete this `if` block that you are changing
2. Rename `EnsureLocalsAndParameters()` to `GetLocalsAndParameters(out List<VariableInformation> parameters, out List<VariableInformation> locals)`
3. Change all the `CreateXYZProperties` functions to call `GetLocalsAndParameters` instead of using `_parameters` and `_locals`
4. Remove all the member variables.

gc46: I have been trying to make the `_parameters` and `_locals` changes without the Locals window going blank to no avail:
image

Will investigate in the meantime.

gc46: Update: I fixed the blank Locals window issue. However, I had to remove this check in `GetLocalsAndParameters` to do so: `if (_textPosition != null && radix != _radix).
`

src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
```diff
@@ -249,31 +244,29 @@ private void EnsureLocalsAndParameters()
                     await Engine.UpdateRadixAsync(radix);
                 });
             }
-            if (_textPosition != null && (!_hasGottenLocalsAndParams || radix != _radix))
+            //if (_textPosition != null && radix != _radix)
```
gregg-miskelly, line 247: >if [](http://example.com/codeflow?start=14&length=2)

Eliminating the `radix` check is definitely what you want since `_textPosition != null && (!_hasGottenLocalsAndParams || radix != _radix)` would evaluate to `_textPosition != null && (!false || radix != _radix))` which just turns into `_textPosition != null`.

I am assuming the `_textPosition != null` check is there to avoid trying to fetch locals from a frame without symbols. I am not 100% sure if that check is correct, but I think I would put that part of the check back since I can see why that might be correct.

src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
```diff
@@ -249,31 +244,29 @@ private void EnsureLocalsAndParameters()
...
-                    localsAndParameters = await Engine.DebuggedProcess.GetLocalsAndParameters(Thread, ThreadContext);
-                });
+                localsAndParameters = await Engine.DebuggedProcess.GetLocalsAndParameters(Thread, ThreadContext);
+            });
 
-                _parameters.Clear();
-                _locals.Clear();
-                foreach (VariableInformation vi in localsAndParameters)
+            parameters.Clear();
+            locals.Clear();
```
gregg-miskelly, line 257: nit: remove

src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
```diff
@@ -319,12 +314,14 @@ private void CreateLocalsPlusArgsProperties(enum_DEBUGPROP_INFO_FLAGS dwFields,
         // Construct an instance of IEnumDebugPropertyInfo2 for the locals collection only.
         private void CreateLocalProperties(enum_DEBUGPROP_INFO_FLAGS dwFields, out uint elementsReturned, out IEnumDebugPropertyInfo2 enumObject)
         {
-            elementsReturned = (uint)_locals.Count;
-            DEBUG_PROPERTY_INFO[] propInfo = new DEBUG_PROPERTY_INFO[_locals.Count];
+            GetLocalsAndParameters(out List<VariableInformation> locals, out List<VariableInformation> parameters);
```
gregg-miskelly, line 317: >List<VariableInformation> parameters [](http://example.com/codeflow?start=77&length=36)

nit: Use `_` here since we don't care about parameters.

src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
```diff
@@ -334,12 +331,14 @@ private void CreateLocalProperties(enum_DEBUGPROP_INFO_FLAGS dwFields, out uint
         // Construct an instance of IEnumDebugPropertyInfo2 for the parameters collection only.
         private void CreateParameterProperties(enum_DEBUGPROP_INFO_FLAGS dwFields, out uint elementsReturned, out IEnumDebugPropertyInfo2 enumObject)
         {
-            elementsReturned = (uint)_parameters.Count;
-            DEBUG_PROPERTY_INFO[] propInfo = new DEBUG_PROPERTY_INFO[_parameters.Count];
+            GetLocalsAndParameters(out List<VariableInformation> locals, out List<VariableInformation> parameters);
```
gregg-miskelly, line 334: >List<VariableInformation> locals [](http://example.com/codeflow?start=39&length=32)

nit: Use `_` since we don't care about locals.

src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
```diff
@@ -334,12 +331,14 @@ private void CreateLocalProperties(enum_DEBUGPROP_INFO_FLAGS dwFields, out uint
...
+
+            elementsReturned = (uint)parameters .Count;
```
gregg-miskelly, line 336: >  [](http://example.com/codeflow?start=47&length=1)

nit: remove the extra space

src/MIDebugEngine/AD7.Impl/AD7StackFrame.cs
```diff
@@ -399,7 +398,7 @@ int IDebugStackFrame2.EnumProperties(enum_DEBUGPROP_INFO_FLAGS dwFields, uint nR
                     guidFilter == AD7Guids.guidFilterLocals ||
                     guidFilter == AD7Guids.guidFilterLocalsPlusArgs)
                 {
```
gregg-miskelly, line 400: Delete this whole if block -- you don't need to pre-call `GetLocalsAndParameters` any more since each of the `CreateXyzProperties` does this.

gregg-miskelly (approved):

PR #1301, opened 2022-03-30
Fix GitHub Action Test runs

WardenGnaw:
This PR revert "Remove downgrade to GNU toolchain (#1298)" since
machines seem to be flaky on which versions of GDB we get.

This reverts commit 8d70a37d2243862eccba70a6226bf684708871c7.

We also introduce retry logic for dotnet test on flaky tests.

Also does a null check on writer.close since that seems to be the cause
of most tests failing when disconnecting.

### DIFF
--- a/.github/workflows/Build-And-Test.yml
+++ b/.github/workflows/Build-And-Test.yml
@@ -195,7 +195,7 @@ jobs:
       if: ${{ always() }}
       with:
         name: linux_x64_results
-        path: ${{ github.workspace }}/bin/DebugAdapterProtocolTests/Debug/CppTests/results*.trx
+        path: ${{ github.workspace }}/bin/DebugAdapterProtocolTests/Debug/CppTests/results.trx
 
   osx_build:
     runs-on: macos-latest
@@ -210,11 +210,6 @@ jobs:
       with:
         dotnet-version: 6.0.x
 
-    # BSD grep will fail on -o and -P. Install GNU Grep
-    - run: |
-        brew install grep
-        echo 'export PATH="/usr/local/opt/grep/libexec/gnubin:$PATH"' >> .bash_profile
-
     - name: Build
       run: dotnet build ${{ github.workspace }}/src/MIDebugEngine-Unix.sln
 
@@ -222,7 +217,6 @@ jobs:
         ${{ github.workspace }}/eng/Scripts/CI-Build.sh
 
     - run: |
-        [ -f .bash_profile ] && source .bash_profile
         ${{ github.workspace }}/eng/Scripts/CI-Test.sh
 
     - name: 'Upload Test Results'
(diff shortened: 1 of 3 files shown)


eng/Scripts/CI-Test.sh
```diff
@@ -19,4 +19,30 @@ if [ ! -f "$RootDir/bin/DebugAdapterProtocolTests/Debug/CppTests/config.xml" ];
...
+    # capture dotnet test exit status, different from tee
+    exitcode=${PIPESTATUS[0]}
+    if [ "$exitcode" == 0 ]
+    then
+        exit 0
+    fi
+    # cat output, get failed test names, join as DisplayName=TEST with |, remove trailing |.
+    filter=$(grep -o -P '(?<=\sFailed\s)\w*' < ./output.log | awk 'BEGIN { ORS="|" } { print("DisplayName=" $0) }' | grep -o -P '.*(?=\|$)')
+    if [ -z "$filter" ] && [ "$(uname)" = "Darwin" ]; then
+        echo -e "Failed to set a filter. Make sure you are using GNU grep from homebrew."
```
WardenGnaw, line 44: nit: exit

eng/Scripts/CI-Test.sh
```diff
@@ -19,4 +19,30 @@ if [ ! -f "$RootDir/bin/DebugAdapterProtocolTests/Debug/CppTests/config.xml" ];
     fi
 fi
 
-dotnet test "$RootDir"/bin/DebugAdapterProtocolTests/Debug/CppTests/CppTests.dll --logger "trx;LogFileName=$RootDir/bin/DebugAdapterProtocolTests/Debug/CppTests/results.trx"
\ No newline at end of file
+# Adapted from https://www.cazzulino.com/dotnet-test-retry.html
+counter=0
+exitcode=0
+while [ $counter -lt 6 ]
```
gregg-miskelly, line 25: >6 [](http://example.com/codeflow?start=21&length=1)

I don't think we should treat a test that passes 1 out of 6 runs as passed.

Would it be possible to instead use the logic from the C# extension? That will rerun failed tests, but:
* There is a maximum number of tests we will retry
-and-
* If a test fails initially, it needs to repeatedly pass on rerun (4 times is what the C# tests use)

See RetrySettings in /src/vsdbg/VsDbg-UITests/TestRunner/CustomTestCase.cs.

.github/workflows/Build-And-Test.yml
```diff
@@ -85,6 +85,57 @@ jobs:
         install: >-
           mingw-w64-x86_64-toolchain
 
+    # As of Nov 8, 2021, mingw-w64-x86_64-toolchain-11.1.x
```
gregg-miskelly, line 88: ># [](http://example.com/codeflow?start=4&length=1)

Is there a different GDB that we should use that is consistent but also supports the exception stuff that you wanted?

WardenGnaw: 10.2 should support `-catch-throw`. The version I see that dosent support it is `8.1`.

gregg-miskelly (approved):

Issue #1129, opened 2021-04-14
DataBreakpointVariables collection is not updated if binding is slow


gregg-miskelly:
While reviewing code for another PR, I noticed that the [code to update DataBreakpointVariables](https://github.com/microsoft/MIEngine/blob/8456c49b4731971b885c2909f9feaac4bdcd494d/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs#L284) is in the wrong spot -- this should be done as part of BindAsync so that if it takes a long time to do the bind we still update the collection.

Repro steps:
* Modify BindAsync to add an `await Task.Delay(3000);` to simulate it taking a long time to finish binding
* Add something to the watch window
* Invoke "Break when value changes"
* You should see a breakpoint bind warning about binding taking a while, and then soon after the breakpoint should turn solid
* Delete and re-add the variable to the watch window

Expected:
* Watch window item should show it has a breakpoint

Actual:
It doesn't

PR #1302, opened 2022-03-31
Fix deleting breakpoints created as pending

WardenGnaw:
This PR fixes https://github.com/microsoft/vscode-cpptools/issues/9095.
The issue is that breakpoints that are created on debug session start
have not had their module loaded yet and BoundBreakpoints will be null
and with the current check we will return E_FAIL and never cache the
breakpoint.

The original usage of this was to capture errors for data breakpoints,
but they will come through as BreakpointEvents instead of the response
of SetBreakpoints / SetDataBreakpoints.

Example of error caught by BreakpointEvent
image

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
@@ -321,11 +321,6 @@ namespace Microsoft.MIDebugEngine
                 this.SetError(new AD7ErrorBreakpoint(this, ResourceStrings.LongBind, enum_BP_ERROR_TYPE.BPET_SEV_LOW | enum_BP_ERROR_TYPE.BPET_TYPE_WARNING), true);
                 return Constants.S_FALSE;
             }
-            else if (this._BPError != null)
-            {
-                // Ran into some sort of error
-                return Constants.E_FAIL;
-            }
             else
             {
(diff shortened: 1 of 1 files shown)


src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -321,11 +321,6 @@ private int BindWithTimeout()
                 this.SetError(new AD7ErrorBreakpoint(this, ResourceStrings.LongBind, enum_BP_ERROR_TYPE.BPET_SEV_LOW | enum_BP_ERROR_TYPE.BPET_TYPE_WARNING), true);
                 return Constants.S_FALSE;
             }
-            else if (this._BPError != null)
-            {
-                // Ran into some sort of error
-                return Constants.E_FAIL;
```
gregg-miskelly, line 327: Should the block still exist but return S_FALSE? At least to avoid going into the block starting on line 331?

calgagi (approved):

calgagi:
Approved w/ Gregg's comment. The data breakpoint code checks for S_OK so S_FALSE should be fine.

src/MIDebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
WardenGnaw: Keeping the deletion but moved away block of 331 to address  #1129

gregg-miskelly (approved):

PR #1309, opened 2022-05-31
Fixed deadlock in HandleStackTraceRequestAsync where lock was hold too long

### LINKED ISSUE #1308: No response for loading stack trace
I'm using VS Code on Windows with a QNX adjusted GDB client. I noticed that sometimes when I reach a breakpoint and the stacktrace command is send, loading the stack trace hangs forever.
image
This seems to only happen I have two breakpoints in different threads and one of the thread waits for the other. These are the last logs I see when putting on trace logging:
<--   C (stackTrace-27): {"command":"stackTrace","arguments":{"threadId":-21,"startFrame":0,"levels":20},"type":"request","seq":27}
<--   C (stackTrace-28): {"command":"stackTrace","arguments":{"threadId":-2,"startFrame":0,"levels":20},"type":"request","seq":28}
--> R (stackTrace-27): {"type":"response","request_seq":27,"success":true,"command":"stackTrace","body":{"stackFrames": ....

It seems that there is no problem loading the stack trace of the running thread. But it does not seem to be able to load the stack of the thread waiting for the other thread.
If I quit the debug session, I see the stack trace opening for a few milliseconds. In that situation there is nothing I can do except quiting the debug session.

JoergMeier106:
Resolves #1308

The lock in HandleStackTraceRequestAsync was hold too long. This caused the behavior described in the the linked issue.

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -246,7 +246,7 @@ namespace OpenDebugAD7
         {
             string miMode = args.GetValueAsString("MIMode");
 
-            // If MIMode is not provided, set default to GDB. 
+            // If MIMode is not provided, set default to GDB.
             if (string.IsNullOrEmpty(miMode))
             {
                 args["MIMode"] = "gdb";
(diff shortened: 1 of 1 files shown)


gregg-miskelly (approved):
Thank you so much for taking the time to find and fix this!

WardenGnaw:
osx_build failure is just a flaky test fail that is unrelated to this change.

Thank you for the PR. We will get this in the next version of VS Code C/C++ Extension.

PR #1310, opened 2022-06-05
Add Tests for NatVis

WardenGnaw:
This PR adds support for testing MIEngine Natvis.

- sending `visualizerFile` in the launch configuration for CppTests. 
- Added tests for:
  - DisplayView
  - ArrayItems
  - LinkedListItems
  - TreeItems

### DIFF
--- a/src/MIDebugEngine/Natvis.Impl/Natvis.cs
+++ b/src/MIDebugEngine/Natvis.Impl/Natvis.cs
@@ -516,237 +516,25 @@ namespace Microsoft.MIDebugEngine.Natvis
                     IVariableInformation expr = GetExpression(item.Value, variable, visualizer.ScopedNames, item.Name);
                     children.Add(expr);
                 }
-                else if (i is ArrayItemsType)
+                else if (i is ArrayItemsType arrayItemsType)
                 {
-                    ArrayItemsType item = (ArrayItemsType)i;
-                    if (!EvalCondition(item.Condition, variable, visualizer.ScopedNames))
-                    {
-                        continue;
-                    }
-                    uint size = 0;
-                    string val = GetExpressionValue(item.Size, variable, visualizer.ScopedNames);
-                    size = MICore.Debugger.ParseUint(val, throwOnError: true);
-                    size = size > MAX_EXPAND ? MAX_EXPAND : size;   // limit expansion
-                    ValuePointerType[] vptrs = item.ValuePointer;
-                    foreach (var vp in vptrs)
-                    {
-                        if (EvalCondition(vp.Condition, variable, visualizer.ScopedNames))
-                        {
-                            IVariableInformation ptrExpr = GetExpression("*(" + vp.Value + ")", variable, visualizer.ScopedNames);
-                            string typename = ptrExpr.TypeName;
-                            if (String.IsNullOrWhiteSpace(typename))
-                            {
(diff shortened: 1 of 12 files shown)


WardenGnaw:
Investigating why the test machines say `map[0]` is `-72` instead of `-100`.
Locally it produces:
image

WardenGnaw:
**DAP Data:**

Local:
```
<--   C (variables-23): {"command":"variables","arguments":{"variablesReference":1017},"type":"request","seq":23}
--> R (variables-23): {"type":"response","request_seq":23,"success":true,"command":"variables","body":{"variables":[{"name":"Count","value":"6","type":"int","evaluateName":"((map).numElements)","variablesReference":0,"memoryReference":"0x0000000000000006"},{"name":"[0]","value":"-100","type":"int","evaluateName":"[0]","variablesReference":0},{"name":"[1]","value":"-72","type":"int","evaluateName":"[1]","variablesReference":0},{"name":"[2]","value":"-35","type":"int","evaluateName":"[2]","variablesReference":0},{"name":"[3]","value":"0","type":"int","evaluateName":"[3]","variablesReference":0,"memoryReference":"0x0000000000000000"},{"name":"[4]","value":"4","type":"int","evaluateName":"[4]","variablesReference":0,"memoryReference":"0x0000000000000004"},{"name":"[5]","value":"15","type":"int","evaluateName":"[5]","variablesReference":0,"memoryReference":"0x000000000000000f"},{"name":"[Raw View]","value":"{...}","evaluateName":"[Raw View]","variablesReference":1020}]},"seq":119}
```

Test Machine:
```
 <--   C (variables-8): {"command":"variables","arguments":{"variablesReference":1004},"seq":8,"type":"request"}
 --> R (variables-8): {"type":"response","request_seq":8,"success":true,"command":"variables","body":{"variables":[{"name":"Count","value":"6","type":"int","evaluateName":"((map).numElements)","variablesReference":0,"memoryReference":"0x0000000000000006"},{"name":"[0]","value":"-72","type":"int","evaluateName":"[0]","variablesReference":0},{"name":"[Raw View]","value":"{...}","evaluateName":"[Raw View]","variablesReference":1006}]},"seq":32}
```

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -516,246 +516,266 @@ private IVariableInformation[] ExpandVisualized(IVariableInformation variable)
                     IVariableInformation expr = GetExpression(item.Value, variable, visualizer.ScopedNames, item.Name);
```
WardenGnaw, line 516: This file was not meant to be checked in here. Tests will validate that these changes are valid 😅

calgagi (approved):

gregg-miskelly:
For the 'map' issue: sorry if this is obvious, but the map traversal code is also deciding that the map only has one element.

WardenGnaw:
> For the 'map' issue: sorry if this is obvious, but the map traversal code is also deciding that the map only has one element.

 Yep. I realized that my `refactor` broke MapItems. My goal was to add tests before I actually refactors just so I would catch bugs like this. My local state does not have the natvis.cs changes but it does remotely. I have some weird `git` issue.

WardenGnaw:
osx_build is running into a flaky test failure.

PR #1312, opened 2022-06-06
Add IndexListItems Natvis test and enable on macOS

WardenGnaw:
This PR adds a test for IndexListItems and enables all but TestArrayItems for macOS

### DIFF
--- a/test/CppTests/debuggees/natvis/src/SimpleArray.h
+++ b/test/CppTests/debuggees/natvis/src/SimpleArray.h
@@ -1,0 +1,21 @@
+#include <stdlib.h>
+
+class SimpleArray
+{
+public:
+  SimpleArray(int size)
+  {
+    _array = (int *)calloc(size, sizeof(int));
+    _size = size;
+  }
+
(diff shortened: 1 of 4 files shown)


test/CppTests/debuggees/natvis/src/SimpleArray.h
```diff
@@ -0,0 +1,21 @@
...
+{
+public:
+  SimpleArray(int size)
+  {
+    _array = (int *)calloc(size, sizeof(int));
+    _size = size;
+  }
+
+  int operator [] (int i) const {
+    return _array[i];
```
gregg-miskelly, line 13: >i [](http://example.com/codeflow?start=18&length=1)

I wonder if you want to do something slightly weird with the index to better test `IndexListItems` (since this class should really be used with `ArrayItems`). My suggestion: store the items in reverse order instead:

```
int operator [] (int i) const {
    return _reverseArray[_size - i - 1];
  }
```

gregg-miskelly (approved):
Otherwise LGTM

PR #1313, opened 2022-06-07
Adding UnitTests for Natvis Parsing (MIDebugEngineUnitTests)

WardenGnaw:
Adding UnitTests for `TypeNames.Parse(...)`

Required adding an IVT for MIDebugEngine and updating the build to run the tests.

Caught that both: `long long int` and `long double` did not succeeded in parsing and fixed it in NatvisName.cs regex.

### DIFF
--- a/src/MICore/Logger.cs
+++ b/src/MICore/Logger.cs
@@ -18,6 +18,8 @@ namespace MICore
     public interface ILogger
     {
         void WriteLine(string line);
+
+        void WriteLine(string format, params object[] args);
     }
 
     /// <summary>
--- a/src/MICore/Logger.cs
+++ b/src/MICore/Logger.cs
@@ -15,10 +15,15 @@ using Microsoft.DebugEngineHost;
 
 namespace MICore
 {
+    public interface ILogger
+    {
+        void WriteLine(string line);
+    }
+
     /// <summary>
     /// Class which implements logging. The logging is control by a registry key. If enabled, logging goes to %TMP%\Microsoft.MIDebug.log
     /// </summary>
-    public class Logger
+    public class Logger : ILogger
     {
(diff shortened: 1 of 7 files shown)


src/MIDebugEngine/Properties/AssemblyInfo.cs
```diff
@@ -4,6 +4,7 @@
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Security.Cryptography;
```
WardenGnaw, line 7: nit: remove

src/MICore/Logger.cs
```diff
@@ -15,10 +15,15 @@
 
 namespace MICore
 {
+    public interface ILogger
+    {
+        void WriteLine(string line);
+    }
```
gregg-miskelly, line 21: >} [](http://example.com/codeflow?start=4&length=1)

Also add `public void WriteLine(string format, params object[] args)` since, in general, we should probably perfer code to call that overload?

gregg-miskelly:
>                 logger.WriteLine("Natvis failed to parse typename: " + fullyQualifiedName);

Like this should really be:
``
logger.WriteLine("Natvis failed to parse typename: {0}", fullyQualifiedName);
```
So we aren't creating strings when logging is disabled...

---
Refers to: src/MIDebugEngine/Natvis.Impl/NatvisNames.cs:149 in 793863e. [](commit_id = 793863e0eafa2ba11d51ccf287e8be29a8876f1f, deletion_comment = False)

gregg-miskelly (approved):
Otherwise LGTM

WardenGnaw:
Flaky windows VSCode test failure.

PR #1319, opened 2022-06-22
Fix Windows VS Code Test Runs and upgrade GDB to 12.1

WardenGnaw:
Windows VS Code tests were failing when falling back to 10.x.

When upgrading to 12.x, GDB was failing because a shared library could not be found and it would error with:

> ERROR: Unable to start debugging. Unexpected GDB output from command \"-exec-run\". During startup program exited with code 0xc0000139

In this PR, the changes made are:
1. Delete downgrade of GDB.
2. Added `inherit` for environment variables so we can use `dotnet` in `MSYS` shell.
3. Converted windows paths needed for `dotnet test` with `cygpath -u` to convert to unix styled paths.
4. Upgrades the version of GNU Compiler and GDB to `12.1`

### DIFF
--- a/.github/workflows/Build-And-Test.yml
+++ b/.github/workflows/Build-And-Test.yml
@@ -80,61 +80,11 @@ jobs:
     - name: Setup MSYS2
       uses: msys2/setup-msys2@v2
       with:
-        msystem: mingw64
+        msystem: MINGW64
+        path-type: inherit
         update: true
         install: >-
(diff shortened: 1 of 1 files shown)


gregg-miskelly (approved):

gregg-miskelly:
Minor note: `inherit` was misspelled in your original PR description, so be sure to take the updated text in your commit message.

PR #1320, opened 2022-06-22
Upgrade Newtonsoft.Json from 12.0.2 to 13.0.1

WardenGnaw:
Addressing https://github.com/microsoft/MIEngine/security/dependabot/1


gregg-miskelly:
Looks like current VS main uses 13.0.1 already, so this looks save.

gregg-miskelly (approved):

PR #1323, opened 2022-07-01
Override Newtonsoft.JSON for CG

WardenGnaw:
This PR adds an explcit reference to Newtonsoft.Json for packages the
reference MS.NET.Test.SDK and MS.VS.Shell until they have properly
published packages using Newtonsoft.JSON v13.0.1

### DIFF
--- a/src/OpenDebugAD7/OpenDebugAD7.csproj
+++ b/src/OpenDebugAD7/OpenDebugAD7.csproj
@@ -49,6 +49,16 @@
 
   <Import Project="..\..\build\Debugger.PIAs.Portable.Packages.settings.targets" />
 
+  <!-- 
+    This ItemGroup is to override the packages that being depended on for the PackageReferences above. 
+    The issue is that Microsoft.VisualStudio.Shared.VSCodeDebugProtocol ->
+                      Newtonsoft.Json v12.0.2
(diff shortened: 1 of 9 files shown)


src/OpenDebugAD7/OpenDebugAD7.csproj
```diff
@@ -49,6 +49,16 @@
 
   <Import Project="..\..\build\Debugger.PIAs.Portable.Packages.settings.targets" />
 
+  <!-- 
+    This ItemGroup is to override the packages that being depended on for the PackageReferences above. 
+    The issue is that Microsoft.VisualStudio.Shared.VSCodeDebugProtocol ->
```
gregg-miskelly, line 54: >Microsoft.VisualStudio.Shared.VSCodeDebugProtocol [](http://example.com/codeflow?start=22&length=49)

Since we own this one, should we just fix it?

gregg-miskelly (approved):

gregg-miskelly:
Looks like we should switch the implementation of `HandleSetExceptionBreakpointsRequestAsync` to the new overload that has a `SetExceptionBreakpointsResponse` arg.

```
void HandleSetExceptionBreakpointsRequestAsync(IRequestResponder<SetExceptionBreakpointsArguments, SetExceptionBreakpointsResponse> responder)
```

PR #1333, opened 2022-08-09
[MIEngine Natvis] Add support for visualizing past 50 elements

gc46:
ArrayItemsType
image

TreeItemsType
image

LinkedListItemsType
image

IndexListItemsType
image

### DIFF
--- a/src/MIDebugEngine/Natvis.Impl/Natvis.cs
+++ b/src/MIDebugEngine/Natvis.Impl/Natvis.cs
@@ -110,7 +110,7 @@ namespace Microsoft.MIDebugEngine.Natvis
     internal sealed class LinkedListContinueWrapper : PaginatedVisualizerWrapper
     {
         public readonly IVariableInformation ContinueNode;
-        public LinkedListContinueWrapper(string name, AD7Engine engine, IVariableInformation underlyingVariable, Natvis.VisualizerInfo viz, bool isVisualizerView, IVariableInformation continueNode, uint startIndex = 0)
+        public LinkedListContinueWrapper(string name, AD7Engine engine, IVariableInformation underlyingVariable, Natvis.VisualizerInfo viz, bool isVisualizerView, IVariableInformation continueNode, uint startIndex)
             : base(name, engine, underlyingVariable, viz, isVisualizerView, startIndex)
         {
             ContinueNode = continueNode;
--- a/src/MIDebugEngine/Natvis.Impl/Natvis.cs
+++ b/src/MIDebugEngine/Natvis.Impl/Natvis.cs
@@ -910,12 +910,12 @@ namespace Microsoft.MIDebugEngine.Natvis
         /// Traverse tree based on specified startIndex.
         /// Then add wrappers for Natvis tree visualizations.
         /// </summary>
-        /// <param name="root"></param>
-        /// <param name="goLeft"></param>
-        /// <param name="goRight"></param>
-        /// <param name="getValue"></param>
-        /// <param name="content"></param>
-        /// <param name="size"></param>
+        /// <param name="root">Root of the tree</param>
+        /// <param name="goLeft">Left child of root</param>
+        /// <param name="goRight">Right child of root</param>
+        /// <param name="getValue">Root value</param>
+        /// <param name="content">List of IVariableInformation children</param>
+        /// <param name="size">Number of nodes in tree</param>
         /// <param name="variable">Tree to traverse if size <= 50. Otherwise, expandable continue wrapper.</param>
         /// <param name="parent">The tree to traverse</param>
         /// <param name="startIndex">Index to start traversing from</param>
--- a/src/MIDebugEngine/Natvis.Impl/Natvis.cs
+++ b/src/MIDebugEngine/Natvis.Impl/Natvis.cs
@@ -964,7 +964,7 @@ namespace Microsoft.MIDebugEngine.Natvis
             }
         }
 
-        private void TraverseList(IVariableInformation root, Traverse goNext, Traverse getValue, List<IVariableInformation> content, uint size, bool noValueInRoot, IVariableInformation parent, uint startIndex=0)
+        private void TraverseList(IVariableInformation root, Traverse goNext, Traverse getValue, List<IVariableInformation> content, uint size, bool noValueInRoot, IVariableInformation parent, uint startIndex)
         {
             uint i = startIndex;
             IVariableInformation node = root;
--- a/src/MIDebugEngine/Natvis.Impl/Natvis.cs
+++ b/src/MIDebugEngine/Natvis.Impl/Natvis.cs
@@ -694,7 +694,6 @@ namespace Microsoft.MIDebugEngine.Natvis
                         {
                             string val = GetExpressionValue(s.Value, variable, visualizer.ScopedNames);
                             size = MICore.Debugger.ParseUint(val);
-                            // size = size > MAX_EXPAND ? MAX_EXPAND : size;   // limit expansion
                             break;
                         }
                     }
@@ -711,10 +710,9 @@ namespace Microsoft.MIDebugEngine.Natvis
                         {
                             string processedExpr = ReplaceNamesInExpression(v.Value, variable, visualizer.ScopedNames);
                             Dictionary<string, string> indexDic = new Dictionary<string, string>();
-                            // for (uint index = 0; index < size; ++index)
-                            // for (uint index = 0; index < MAX_EXPAND; ++index) // limit expansion to first 50 elements
                             uint currentIndex = (variable as VisualizerWrapper).StartIndex;
-                            for (uint index = currentIndex; index < currentIndex + 3; ++index) // limit expansion to first 3 elements
+                            uint maxIndex = currentIndex + MAX_EXPAND > size ? size : currentIndex + MAX_EXPAND;
+                            for (uint index = currentIndex; index < maxIndex; ++index) // limit expansion to first 50 elements
                             {
                                 indexDic["$i"] = index.ToString(CultureInfo.InvariantCulture);
                                 string finalExpr = ReplaceNamesInExpression(processedExpr, null, indexDic);
@@ -723,10 +721,11 @@ namespace Microsoft.MIDebugEngine.Natvis
                                 children.Add(expressionVariable);
                             }
 
+                            currentIndex += MAX_EXPAND;
                             if (size > currentIndex)
                             {
-                                IVariableInformation testVariable = new VisualizerWrapper("[More...]", _process.Engine, variable, visualizer, isVisualizerView: true, currentIndex+3);
-                                children.Add(testVariable);
+                                IVariableInformation moreVariable = new VisualizerWrapper(ResourceStrings.MoreView, _process.Engine, variable, visualizer, isVisualizerView: true, currentIndex);
+                                children.Add(moreVariable);
                             }
 
                             break;
--- a/src/MIDebugEngine/Natvis.Impl/Natvis.cs
+++ b/src/MIDebugEngine/Natvis.Impl/Natvis.cs
@@ -599,7 +599,12 @@ namespace Microsoft.MIDebugEngine.Natvis
                             arrayExpr.EnsureChildren();
                             if (arrayExpr.CountChildren != 0)
                             {
-                                uint currentIndex = (variable as PaginatedVisualizerWrapper).StartIndex;
+                                uint currentIndex = 0;
+                                var pvwVariable = variable as PaginatedVisualizerWrapper;
+                                if (pvwVariable != null)
+                                {
+                                    currentIndex = (variable as PaginatedVisualizerWrapper).StartIndex;
+                                }
                                 uint maxIndex = currentIndex + MAX_EXPAND > size ? size : currentIndex + MAX_EXPAND;
                                 for (uint index = currentIndex; index < maxIndex; ++index)
                                 {
@@ -675,7 +680,10 @@ namespace Microsoft.MIDebugEngine.Natvis
                             startIndex = visualizerWrapper.StartIndex;
                             TraverseTree(headVal, goLeft, goRight, getValue, children, size, visualizerWrapper.Parent, startIndex);
                         }
-                        TraverseTree(headVal, goLeft, goRight, getValue, children, size, (variable as SimpleWrapper).Parent, startIndex);
+                        else
+                        {
+                            TraverseTree(headVal, goLeft, goRight, getValue, children, size, (variable as SimpleWrapper).Parent, startIndex);
+                        }
                     }
                 }
                 else if (i is LinkedListItemsType)
@@ -750,7 +758,10 @@ namespace Microsoft.MIDebugEngine.Natvis
                             startIndex = visualizerWrapper.StartIndex;
                             TraverseList(headVal, goNext, getValue, children, size, item.NoValueHeadPointer, visualizerWrapper.Parent, startIndex);
                         }
-                        TraverseList(headVal, goNext, getValue, children, size, item.NoValueHeadPointer, (variable as SimpleWrapper).Parent, startIndex);
+                        else
+                        {
+                            TraverseList(headVal, goNext, getValue, children, size, item.NoValueHeadPointer, (variable as SimpleWrapper).Parent, startIndex);
+                        }
                     }
                 }
                 else if (i is IndexListItemsType)
@@ -795,7 +806,12 @@ namespace Microsoft.MIDebugEngine.Natvis
                         {
                             string processedExpr = ReplaceNamesInExpression(v.Value, variable, visualizer.ScopedNames);
                             Dictionary<string, string> indexDic = new Dictionary<string, string>();
-                            uint currentIndex = (variable as PaginatedVisualizerWrapper).StartIndex;
+                            uint currentIndex = 0;
+                            var pvwVariable = variable as PaginatedVisualizerWrapper;
+                            if (pvwVariable != null)
+                            {
+                                currentIndex = (variable as PaginatedVisualizerWrapper).StartIndex;
+                            }
                             uint maxIndex = currentIndex + MAX_EXPAND > size ? size : currentIndex + MAX_EXPAND;
                             for (uint index = currentIndex; index < maxIndex; ++index) // limit expansion to first 50 elements
                             {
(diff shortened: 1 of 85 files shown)


WardenGnaw (changes requested):
One thing to test is to enable `showDisplayString` in your launch.json. 

https://code.visualstudio.com/docs/cpp/launch-json-reference#_showdisplaystring

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -72,25 +72,65 @@ public uint Size()
...
 
-    internal sealed class VisualizerWrapper : SimpleWrapper
+    internal class VisualizerWrapper : SimpleWrapper
     {
         public readonly Natvis.VisualizerInfo Visualizer;
         private readonly bool _isVisualizerView;
+        private readonly uint _startIndex;
 
-        public VisualizerWrapper(string name, AD7Engine engine, IVariableInformation underlyingVariable, Natvis.VisualizerInfo viz, bool isVisualizerView)
+        public VisualizerWrapper(string name, AD7Engine engine, IVariableInformation underlyingVariable, Natvis.VisualizerInfo viz, bool isVisualizerView, uint startIndex = 0)
```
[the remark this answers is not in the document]
gc46: `startIndex` is used in `TraverseTree/TreeItemsType` and `TraverseList/LinkedListItemsType` as well...

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -788,27 +864,14 @@ private Traverse GetTraverse(string direction, IVariableInformation node)
...
-                Content = v;
-                State = ScanState.left;
-            }
-        }
-
-        private void TraverseTree(IVariableInformation root, Traverse goLeft, Traverse goRight, Traverse getValue, List<IVariableInformation> content, uint size)
-        {
-            uint i = 0;
+            uint i = startIndex;
             var nodes = new Stack<Node>();
```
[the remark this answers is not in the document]
gc46: Looks like I need to pass in `variable` to `TraverseTree()`. Otherwise, there is no way to access `nodes`.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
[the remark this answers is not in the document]
WardenGnaw: Yep. You can create a `PaganatedVisualizerWrapper` that adds in the StartIndex and then have the other classes extend off of that.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -708,14 +786,24 @@ private IVariableInformation[] ExpandVisualized(IVariableInformation variable)
                         {
                             string processedExpr = ReplaceNamesInExpression(v.Value, variable, visualizer.ScopedNames);
                             Dictionary<string, string> indexDic = new Dictionary<string, string>();
-                            for (uint index = 0; index < size; ++index)
+                            uint currentIndex = (variable as VisualizerWrapper).StartIndex;
```
WardenGnaw, line 789: The assumption that variable will always be a `VisualizerWrapper` is causing the test to fail because that is not true with the `showDisplayString` case. You will need to handle where variable is only a `VariableInformation` from the engine.

image

src/MIDebugEngine/Natvis.Impl/Natvis.cs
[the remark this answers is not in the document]
gc46: The functionality breaks when I add `PaginatedVisualizerWrapper`. May you take a look at [this commit](https://github.com/microsoft/MIEngine/pull/1333/commits/0fdf5bbbea0b9ce95c66d04d33ee4d756cbf218b) and perhaps catch what I may have missed? Thanks!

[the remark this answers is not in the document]
gc46: Fixed.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -91,6 +91,55 @@ public override string FullName()
...
+        {
+            _startIndex = startIndex;
+        }
+        public uint StartIndex { get { return _startIndex; } }
+    }
+
+    internal sealed class LinkedListContinueWrapper : PaginatedVisualizerWrapper
+    {
+        public readonly IVariableInformation ContinueNode;
+        public LinkedListContinueWrapper(string name, AD7Engine engine, IVariableInformation underlyingVariable, Natvis.VisualizerInfo viz, bool isVisualizerView, IVariableInformation continueNode, uint startIndex = 0)
```
WardenGnaw, line 109: Make `startIndex` required since this is built on top of `PaginatedVisualizerWrapper`
```suggestion
        public LinkedListContinueWrapper(string name, AD7Engine engine, IVariableInformation underlyingVariable, Natvis.VisualizerInfo viz, bool isVisualizerView, IVariableInformation continueNode, uint startIndex)
```

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -91,6 +91,55 @@ public override string FullName()
...
+            Content = v;
+            State = ScanState.left;
+        }
+    }
+
+    internal sealed class TreeContinueWrapper : PaginatedVisualizerWrapper
+    {
+        public readonly Node ContinueNode;
+        public readonly Stack<Node> Nodes;
+        public TreeContinueWrapper(string name, AD7Engine engine, IVariableInformation underlyingVariable, Natvis.VisualizerInfo viz, bool isVisualizerView, Node continueNode, Stack<Node> nodes, uint startIndex = 0)
```
WardenGnaw, line 135: Make `startIndex` required since this is built on top of `PaginatedVisualizerWrapper`
```suggestion
        public TreeContinueWrapper(string name, AD7Engine engine, IVariableInformation underlyingVariable, Natvis.VisualizerInfo viz, bool isVisualizerView, Node continueNode, Stack<Node> nodes, uint startIndex)
```

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -506,7 +555,7 @@ private IVariableInformation[] ExpandVisualized(IVariableInformation variable)
             }
             foreach (var i in expandType.Items)
             {
-                if (i is ItemType)
+                if (i is ItemType && variable.Name != ResourceStrings.MoreView) // we do not want to display Size/Count when expanding the "[More...]" node
```
WardenGnaw, line 558: Someone could named their visualized variable to be `[More...]`.

It would be better to check to see if it is a `PaginatedVisualizerWrapper` type.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -551,7 +599,24 @@ private IVariableInformation[] ExpandVisualized(IVariableInformation variable)
                             arrayExpr.EnsureChildren();
                             if (arrayExpr.CountChildren != 0)
                             {
-                                children.AddRange(arrayExpr.Children);
+                                uint currentIndex = 0;
+                                var pvwVariable = variable as PaginatedVisualizerWrapper;
```
WardenGnaw, line 603: You can use the `is` syntax here

```suggestion
                                if (variable is PaginatedVisualizerWrapper pvwVariable )
                                {
                                   currentIndex =pvwVariable.StartIndex;
                                }
```

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -551,7 +599,24 @@ private IVariableInformation[] ExpandVisualized(IVariableInformation variable)
...
+                                if (pvwVariable != null)
+                                {
+                                    currentIndex = (variable as PaginatedVisualizerWrapper).StartIndex;
+                                }
+                                uint maxIndex = currentIndex + MAX_EXPAND > size ? size : currentIndex + MAX_EXPAND;
+                                for (uint index = currentIndex; index < maxIndex; ++index)
+                                {
+                                    children.Add(arrayExpr.Children[index]);
```
WardenGnaw, line 611: I would be careful here.

Can you try a natvis where the actual number of children in the array is like 5 but the user hard-codes it to be 500?
We could be getting an indexOutOfBoundsException

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -601,7 +673,20 @@ private IVariableInformation[] ExpandVisualized(IVariableInformation variable)
                         {
                             continue;
                         }
-                        TraverseTree(headVal, goLeft, goRight, getValue, children, size);
+                        uint startIndex = 0;
+                        IVariableInformation parent = variable;
+                        var visualizerWrapper = variable as PaginatedVisualizerWrapper;
```
WardenGnaw, line 678: Use the `is` syntax here.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -662,7 +754,20 @@ private IVariableInformation[] ExpandVisualized(IVariableInformation variable)
                         {
                             continue;
                         }
-                        TraverseList(headVal, goNext, getValue, children, size, item.NoValueHeadPointer);
+                        uint startIndex = 0;
+                        IVariableInformation parent = variable;
+                        var visualizerWrapper = variable as PaginatedVisualizerWrapper;
+                        var simpleWrapper = variable as SimpleWrapper;
+                        if (visualizerWrapper != null)
```
WardenGnaw, line 761: Use the `is` syntax here.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -708,14 +812,29 @@ private IVariableInformation[] ExpandVisualized(IVariableInformation variable)
                         {
                             string processedExpr = ReplaceNamesInExpression(v.Value, variable, visualizer.ScopedNames);
                             Dictionary<string, string> indexDic = new Dictionary<string, string>();
-                            for (uint index = 0; index < size; ++index)
+                            uint currentIndex = 0;
+                            var pvwVariable = variable as PaginatedVisualizerWrapper;
+                            if (pvwVariable != null)
```
WardenGnaw, line 817: Use the `is` syntax here.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -788,27 +907,22 @@ private Traverse GetTraverse(string direction, IVariableInformation node)
             return go;
         }
 
-        private class Node
+        private void TraverseTree(IVariableInformation root, Traverse goLeft, Traverse goRight, Traverse getValue, List<IVariableInformation> content, uint size, IVariableInformation variable, IVariableInformation parent, uint startIndex = 0)
```
WardenGnaw, line 910: Don't make `startIndex` default

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -788,27 +907,22 @@ private Traverse GetTraverse(string direction, IVariableInformation node)
...
         {
-            public enum ScanState
+            uint i = startIndex;
+            var nodes = new Stack<Node>();
+            var tcwVariable = variable as TreeContinueWrapper;
```
WardenGnaw, line 914: Use the `is` syntax here.

WardenGnaw (changes requested):
Please add testing for this new scenario.

See the `// TODO` in https://github.com/microsoft/MIEngine/blob/main/test/CppTests/Tests/NatvisTests.cs

src/MIDebugEngine/Natvis.Impl/Natvis.cs
gc46: May you elaborate on this? If the user hardcodes the array to be of size 500, how would the number of children in the array be 5? Unless you mean we have an array `int arr [5]` and the user tries to access `arr[500]`?

WardenGnaw: Change the SimpleVector definition in Natvis to hardcode a size outside of the actual code size.
```
  <Type Name="SimpleVector">
   <DisplayString>{{ size={_size} }}</DisplayString>
   <Expand>
    <Item Name="Size">_size</Item>
    <ArrayItems>
      <Size>500</Size>
      <ValuePointer>_start</ValuePointer>
    </ArrayItems>
   </Expand>
  </Type>
```

We should only be displaying values from valid indexes.

WardenGnaw: Actually this seems to be ok. The internal IVariableInformation allocates enough space for Children that we will not be indexing out of bounds.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -708,14 +810,28 @@ private IVariableInformation[] ExpandVisualized(IVariableInformation variable)
                         {
                             string processedExpr = ReplaceNamesInExpression(v.Value, variable, visualizer.ScopedNames);
                             Dictionary<string, string> indexDic = new Dictionary<string, string>();
-                            for (uint index = 0; index < size; ++index)
+                            uint currentIndex = 0;
+                            if (variable is PaginatedVisualizerWrapper pvwVariable)
+                            {
+                                currentIndex = (variable as PaginatedVisualizerWrapper).StartIndex;
```
WardenGnaw, line 816:
```suggestion
                                currentIndex = pvwVariable.StartIndex;
```

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -847,15 +957,22 @@ private void TraverseTree(IVariableInformation root, Traverse goLeft, Traverse g
...
             }
+            if (size > i)
+            {
+                IVariableInformation tcw = new TreeContinueWrapper(ResourceStrings.MoreView, _process.Engine, parent, FindType(parent), isVisualizerView: true, nodes.Peek(), nodes, i);
+                content.Add(tcw);
+            }
         }
 
-        private void TraverseList(IVariableInformation root, Traverse goNext, Traverse getValue, List<IVariableInformation> content, uint size, bool noValueInRoot)
+        private void TraverseList(IVariableInformation root, Traverse goNext, Traverse getValue, List<IVariableInformation> content, uint size, bool noValueInRoot, IVariableInformation parent, uint startIndex=0)
```
WardenGnaw, line 967:
```suggestion
        private void TraverseList(IVariableInformation root, Traverse goNext, Traverse getValue, List<IVariableInformation> content, uint size, bool noValueInRoot, IVariableInformation parent, uint startIndex)
```

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -506,7 +558,7 @@ private IVariableInformation[] ExpandVisualized(IVariableInformation variable)
             }
             foreach (var i in expandType.Items)
             {
-                if (i is ItemType)
+                if (i is ItemType && !(variable is PaginatedVisualizerWrapper)) // we do not want to display Size/Count when expanding the "[More...]" node
```
WardenGnaw, line 561: nit: This comment should be:
```
// we do not want to repeatedly display other ItemTypes when expanding the "[More...]" node
```

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -90,6 +90,58 @@ public override string FullName()
...
+        }
+        public ScanState State { get; set; }
+        public IVariableInformation Content { get; private set; }
+        public Node(IVariableInformation v)
+        {
+            Content = v;
+            State = ScanState.left;
+        }
+    }
+
```
WardenGnaw, line 133: Add comment for the class.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -788,27 +904,21 @@ private Traverse GetTraverse(string direction, IVariableInformation node)
             return go;
         }
 
-        private class Node
+        private void TraverseTree(IVariableInformation root, Traverse goLeft, Traverse goRight, Traverse getValue, List<IVariableInformation> content, uint size, IVariableInformation variable, IVariableInformation parent, uint startIndex)
```
WardenGnaw, line 907: This method now has many parameters and is getting complex. Can you document each argument in the function?

WardenGnaw (changes requested):

src/MIDebugEngine/Natvis.Impl/Natvis.cs
gc46: Made changes. Added comments for `variable`, `parent`, and `startIndex` as they are the newly added params. Do we want to elaborate on the preexisting params? Thanks!

WardenGnaw: Yes. Please document all the parameters so it is easier for the next person to understand what it does when they have to modify this method.

gc46: Done. Let me know if I need to fix the wording of any of them. Thanks!

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -788,27 +906,35 @@ private Traverse GetTraverse(string direction, IVariableInformation node)
             return go;
         }
 
-        private class Node
+        /// <summary>
+        /// Traverse tree based on specified startIndex.
+        /// Then add wrappers for Natvis tree visualizations.
+        /// </summary>
+        /// <param name="root">Root of the tree</param>
+        /// <param name="goLeft">Left child of root</param>
```
WardenGnaw, line 914: `goLeft` and `goRight` are not directly the child nodes. They are `Traverse` callbacks to handle grabbing the left child of the node passed into it. 

Same with `getValue`, it is a callback to obtain the value of the given node

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -788,27 +906,35 @@ private Traverse GetTraverse(string direction, IVariableInformation node)
...
+        /// <param name="goRight">Right child of root</param>
+        /// <param name="getValue">Root value</param>
+        /// <param name="content">List of IVariableInformation children</param>
```
WardenGnaw, line 917: `content` is more like the list of variables to be displayed given the current variable.

WardenGnaw (approved):

src/MIDebugEngine/Natvis.Impl/Natvis.cs
gc46: Okay will fix.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
gc46: Will fix. Thanks.

Issue #1236, opened 2021-10-25
The result of GDB -exec evaluate request in all contexts is printed in debug console.


awulkiew:
When `vscode.DebugSession.customRequest('evaluate', ...)` is called for expression starting with `-exec` the result is printed in debug console no matter what context was specified. It is not returned to the caller in `DebugProtocol.EvaluateResponse.result`. On the other hand if it doesn't start with `-exec` it is returned to the caller even if the context is `'repl'`.

**How to reproduce:**

In vscode extension, `registerDebugAdapterTrackerFactory`, define `onDidSendMessage` and catch `'stopped'` `'event'`. Then get `threadId` and execute `'stackTrace'` request to get `frameId`. Then call:
```
let exprArgs : DebugProtocol.EvaluateArguments = { expression: '-exec whatis name', frameId: frameId, context: 'watch' };
let test = await session.customRequest('evaluate', exprArgs);
```
with GDB and `name` being some identifier in C++ program.

awulkiew:
FYI, I'm willing to work on a PR addressing this issue.

tingleby:
We also have seen this behavior when working on an extension to enhance the debug session.

WardenGnaw:
We will be shipping this in the next VS Code Release 1.12.1

PR #1334, opened 2022-08-10
Only print evaluate requests in REPL context

WardenGnaw:
This PR fixes the EvaluateRequest with an `-exec` command and assign the return protocol with the correct `Value`.

This causes the DebugConsole in VS Code to show the output as a REPL output instead of an engine output.

Old: 
ret-exec-val-old

New:
ret-exec-val

Fixes: https://github.com/microsoft/MIEngine/issues/1236


andrewcrawley (approved):
Commit message should be updated, but otherwise LGTM!

PR #1344, opened 2022-08-29
Address visualizing past 1000 ArrayItems

WardenGnaw:
We have a limit of 1000 child items that can be returned when we call `EnsureChildren`.
https://github.com/microsoft/MIEngine/blob/d5a05958d3cca45995b335645005d6917b515839/src/MICore/CommandFactories/MICommandFactory.cs#L353

Since we only ever need at most 50 at a time, this PR modified the request to be `(T[<At most 50>])*(<ValuePointer> +
StartOffset)` to handle child values past 1000.

Also added a test case to see if we can see past 1000 by testing with an
ArrayItem visualized type with 2000 elements.

### DIFF
--- a/src/MIDebugEngine/Natvis.Impl/Natvis.cs
+++ b/src/MIDebugEngine/Natvis.Impl/Natvis.cs
@@ -605,7 +605,7 @@ namespace Microsoft.MIDebugEngine.Natvis
                             //   If requestedSize > 1000, the evaluation will only grab the first 1000 elements.
 
                             // We want to limit it to at most 50.
-                            uint requestedSize = totalSize - startIndex > MAX_EXPAND ? MAX_EXPAND : totalSize - startIndex;
+                            uint requestedSize = Math.Min(MAX_EXPAND, totalSize - startIndex);
 
                             StringBuilder arrayBuilder = new StringBuilder();
                             arrayBuilder.Append('(');
--- a/src/MIDebugEngine/Natvis.Impl/Natvis.cs
+++ b/src/MIDebugEngine/Natvis.Impl/Natvis.cs
@@ -577,9 +577,16 @@ namespace Microsoft.MIDebugEngine.Natvis
                     {
                         continue;
                     }
-                    uint size = 0;
+                    uint totalSize = 0;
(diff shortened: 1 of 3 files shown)


test/CppTests/Tests/NatvisTests.cs
```diff
@@ -167,14 +167,69 @@ public void TestArrayItems(ITestSettings settings)
...
+
+                    this.Comment("Verifying ArrayItems natvis");
+                    var ll = currentFrame.GetVariable("vec");
+                    Assert.Equal("{ size=2000 }", ll.Value);
+
+                    // Custom Item in natvis
+                    Assert.Equal("2000", ll.GetVariable("Size").Value);
+
+                    var more = ll.GetVariable("[More...]");
+                    for (int i = 1; i < 20; i++)
```
calgagi, line 226: Aren't we doing 2000 elements here? 20*50 = 1000, so shouldn't we have 40 iterations instead of 20?

calgagi (approved):

WardenGnaw:
Looks like my debuggee changes did not make it in.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -591,34 +598,41 @@ private IVariableInformation[] ExpandVisualized(IVariableInformation variable)
...
                                 continue;
                             }
+                            // Creates an expression: (T[50])*(<ValuePointer> + 50)
+                            // This evaluates for 50 elements of type T, starting at <ValuePointer> with an offet of 50 elements.
+                            // E.g. This will grab elements 50 - 99 from <ValuePointer>.
+                            // Note:
+                            //   If requestedSize > 1000, the evaluation will only grab the first 1000 elements.
+
+                            // We want to limit it to at most 50.
+                            uint requestedSize = totalSize - startIndex > MAX_EXPAND ? MAX_EXPAND : totalSize - startIndex;
```
gregg-miskelly, line 608: >totalSize - startIndex [](http://example.com/codeflow?start=49&length=22)

nit: `Math.Min(MAX_EXPAND, totalSize - startIndex);`

gregg-miskelly (approved):

gregg-miskelly (approved):

PR #1345, opened 2022-08-30
Add support for comma format specifier for lldb

WardenGnaw:
This PR enables the comma format specifier for lldb scenarios.

Also fixes the issue that `,[#]` was not being handled correctly.
Adds a test for the `,[#]` and updates the test name.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -341,11 +341,12 @@ namespace Microsoft.MIDebugEngine
         /// This callback is used when we need to call into the engine for additional information for
         /// the format specifier.
         ///
-        /// <param name="int">threadId</param>
-        /// <param name="uint">frameLevel</param>
+        /// <param name="threadId">The threadId to use when calling the engine.</param>
+        /// <param name="frameLevel">The frameLevel to use when calling the engine.</param>
         /// <returns>The expression to send to the engine</returns>
         /// </summary>
-        private Func<int, uint, Task<string>> _deferedFormatExpression;
+        private delegate Task<string> DeferedFormatExpression(int threadId, uint frameLevel);
+        private DeferedFormatExpression _deferedFormatExpression;
         private IVariableInformation _parent;
         private string _format;
         private string _strippedName;  // "Name" stripped of format specifiers
@@ -452,7 +453,7 @@ namespace Microsoft.MIDebugEngine
                 {
                     _deferedFormatExpression = async (int threadId, uint frameLevel) =>
                     {
-                        string derefType = await GetDereferencedTypeString(expr, threadId, frameLevel);
+                        string derefType = await GetDereferencedTypeStringAsync(expr, threadId, frameLevel);
 
                         if (!string.IsNullOrEmpty(derefType))
                         {
@@ -474,7 +475,7 @@ namespace Microsoft.MIDebugEngine
             return exp;
         }
 
-        private async Task<string> GetDereferencedTypeString(string expr, int threadId, uint frameLevel)
+        private async Task<string> GetDereferencedTypeStringAsync(string expr, int threadId, uint frameLevel)
         {
             // TODO: Should we error if the current type is not a pointer type?
 
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -213,7 +213,7 @@ namespace Microsoft.MIDebugEngine
             : this(ctx, engine, thread)
         {
             // strip off formatting string
-            _strippedName = StripFormatSpecifier(expr, out _format, thread.GetDebuggedThread().Id, ctx.Level);
+            _strippedName = ProcessFormatSpecifiers(expr, out _format);
             Name = displayName;
             IsParameter = isParameter;
             _parent = null;
@@ -225,7 +225,7 @@ namespace Microsoft.MIDebugEngine
             : this(parent.ThreadContext, engine, parent.Client)
         {
             // strip off formatting string
-            _strippedName = StripFormatSpecifier(expr, out _format, parent.Client.GetDebuggedThread().Id, parent.ThreadContext.Level);
+            _strippedName = ProcessFormatSpecifiers(expr, out _format);
             Name = displayName ?? expr;
             _parent = parent;
             VariableNodeType = NodeType.Synthetic;
@@ -236,7 +236,7 @@ namespace Microsoft.MIDebugEngine
             : this(parent._ctx, parent._engine, parent.Client)
         {
             // strip off formatting string
-            _strippedName = StripFormatSpecifier(expr, out _format, parent.Client.GetDebuggedThread().Id, parent._ctx.Level);
+            _strippedName = ProcessFormatSpecifiers(expr, out _format);
             Name = expr;
             VariableNodeType = NodeType.Root;
         }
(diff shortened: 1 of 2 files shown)


src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -428,24 +428,48 @@ private string StripFormatSpecifier(string exp, out string formatSpecifier)
...
+                    // return *<expression>@<count> which is only supported in GDB
+                    return $"*{expr}@{count}";
+                }
+                else
+                {
+                    string derefType = GetDereferencedTypeString(expr, threadId, frameLevel);
 
-                return exp.Substring(0, lastComma);
+                    // return '*(T(*)[n])(exp)'
+                    return $"*({derefType}(*)[{count}])({expr})";
```
gregg-miskelly, line 447: >(*) [](http://example.com/codeflow?start=42&length=3)

Do you understand why this is `(*)`?

gregg-miskelly, line 447: >$ [](http://example.com/codeflow?start=27&length=1)

Use Invarint locale

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -365,7 +365,7 @@ public enum NodeType
 
         private static Regex s_isFunction = new Regex(@".+\(.*\).*");
 
-        private string StripFormatSpecifier(string exp, out string formatSpecifier)
+        private string StripFormatSpecifier(string exp, out string formatSpecifier, int threadId, uint frameLevel)
```
gregg-miskelly, line 368: >StripFormatSpecifier [](http://example.com/codeflow?start=23&length=20)

nit: Maybe rename to `ProcessFormatSpecifiers` since this now does a lot more than 'stripping'

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -428,24 +428,48 @@ private string StripFormatSpecifier(string exp, out string formatSpecifier)
...
 
             return exp;
         }
 
+        private string GetDereferencedTypeString(string expr, int threadId, uint frameLevel)
+        {
+            // TODO: Should we error if the current type is not a pointer type?
+
+            Task<string> eval = Task.Run(async () =>
+            {
```
gregg-miskelly, line 463: Yuck. Can you use WorkerThread.RunOperation?

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -428,24 +428,48 @@ private string StripFormatSpecifier(string exp, out string formatSpecifier)
...
+                // Evaluates: *expr
+                Results results = await _engine.DebuggedProcess.MICommandFactory.VarCreate($"*({expr})", threadId, frameLevel, 0, ResultClass.None);
```
gregg-miskelly, line 465: >VarCreate [](http://example.com/codeflow?start=81&length=9)

Also need:

```
                                    string varName = results.TryFindString("name");
                                    if (!String.IsNullOrWhiteSpace(varName))
                                    {
                                        // Remove the variable we created as we don't track it.
                                        await MICommandFactory.VarDelete(varName);
                                    }
```

gregg-miskelly, line 465: >results [](http://example.com/codeflow?start=24&length=7)

Need to check that this succeeded.

src/MIDebugEngine/Engine.Impl/Variables.cs
gregg-miskelly, line 447: >return [](http://example.com/codeflow?start=20&length=6)

Need to only do this if we got the type.

src/MIDebugEngine/Engine.Impl/Variables.cs
gregg-miskelly: Maybe a better option: delay this work until 'Eval'. Then you would have a reasonable path for handling errors as well.

gregg-miskelly (changes requested):
See comments

src/MIDebugEngine/Engine.Impl/Variables.cs
WardenGnaw: Not really sure, but my guess is casting it to an array of T breaks with `C-style cast from 'int' to 'int [10]' is not allowed`, but if we cast the `exp` to a function that returns an `array with size 'n' of T elements` it works.

WardenGnaw: This is what I was using for reference to clarify the `(*)`.
https://stackoverflow.com/a/48133012

gregg-miskelly: For clarity, this isn't a pointer to a function, but rather a pointer to an array. Weird syntax, but I guess it makes sense.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -337,9 +337,18 @@ public VariableInformation FindChildByName(string name)
...
         private bool _isReadonly;
+        /// <summary>
+        /// This callback is used when we need to call into the engine for additional information for
+        /// the format specifier.
+        ///
+        /// <param name="int">threadId</param>
+        /// <param name="uint">frameLevel</param>
+        /// <returns>The expression to send to the engine</returns>
+        /// </summary>
+        private Func<int, uint, Task<string>> _deferedFormatExpression;
```
WardenGnaw, line 348: Would this be better?
```csharp
        /// <summary>
        /// This callback is used when we need to call into the engine for additional information for
        /// the format specifier.
        ///
        /// <param name="threadId">The threadId to use when calling the engine.</param>
        /// <param name="frameLevel">The frameLevel to use when calling the engine.</param>
        /// <returns>The expression to send to the engine</returns>
        /// </summary>
        private delegate Task<string> DeferedFormatExpression(int threadId, uint frameLevel);
        private DeferedFormatExpression _deferedFormatExpression;
```

gregg-miskelly: Sounds good to me.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -446,6 +474,27 @@ private string StripFormatSpecifier(string exp, out string formatSpecifier)
             return exp;
         }
 
+        private async Task<string> GetDereferencedTypeString(string expr, int threadId, uint frameLevel)
```
gregg-miskelly, line 477: >GetDereferencedTypeString [](http://example.com/codeflow?start=35&length=25)

nit: `GetDereferencedTypeStringAsync`

gregg-miskelly, line 477: >string [](http://example.com/codeflow?start=27&length=6)

If this fails, do we want to either throw or return a `Task<(string /*expression*/, string /*error*/)>` instead?

gregg-miskelly: I guess it doesn't really matter, since that expression should fail anyway.

gregg-miskelly (approved):

src/MIDebugEngine/Engine.Impl/Variables.cs
WardenGnaw: I'm hoping to get the NatvisDiagnostic work in to help with logging errors.

gregg-miskelly (approved):

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -428,16 +438,35 @@ private string StripFormatSpecifier(string exp, out string formatSpecifier)
...
             if (m.Success)
             {
+                string count = m.Groups[1].Value; // (\d+) capture group
+                string expr = exp.Substring(0, lastComma);
+
                 if (_engine.DebuggedProcess.MICommandFactory.Mode == MIMode.Gdb)
-                    return "*" + exp.Substring(0, lastComma) + "@" + trimmed;  // this does not work for lldb
+                {
+                    // return *<expression>@<count> which is only supported in GDB
+                    return FormattableString.Invariant($"*{expr}@{count}");
```
Trass3r, line 450: Does this work with more complex expressions?
Parens around expr might be needed.

Trass3r: Also I think this syntax works for dynamic sizes too, see L472.

PR #1350, opened 2022-09-06
[#1349] Update the default stack trace format to include values.

### LINKED ISSUE #1349: Unnecessary `-var-create` and `-var-delete` commands during a stack trace can cause noticeable pause each time the debuggee stops
### Summary

Each time the debuggee stops, Visual Studio Code requests a stack trace of the threads that are expanded in the CALL STACK window. Turning on `engineLogging` shows that MIEngine implements a stack trace for a thread by issuing a `-stack-list-arguments` for the thread, and then issuing a `-var-create` and `-var-delete` command for every argument in every frame in the stack.

When there are many stack frames with many arguments, or if the debugger is behind a sluggish network connection, or if the user installs a hook that runs on every GDB prompt, the time taken to execute these commands can amount to a significant delay.

These `-var-create` and `-var-delete` commands ought to be unnecessary as explained below.

### Reproduction

1. Put this program in a suitable file, say recurse.c:
    ```
    static void
    r(int a, int b, int c, int d, int e, int f, int g, int h, int i, int j)
    {
        if (a)
        {
            r(b, c, d, e, f, g, h, i, j, 0);
        }
    }

    int
    main(void)
    {
        r(1, 2, 3, 4, 5, 6, 7, 8, 9, 10);
        return 0;
    }
    ```

2. Compile it with debugging information, for example `gcc -g -o recurse recurse.c`.

3. Open Visual Studio Code and create a new launch configuration using the "(gdb) launch" template, supplying `recurse` as the value for the `"program"` key, and turning on engineLogging:

    ```
    "logging": {
        "engineLogging": true
    }
    ```

4. Select View ⟶ Run to switch to the run panel, select the new launch configuration in the RUN AND DEBUG dropdown, and press F5 to start debugging.

5. Click "Step Into" a few times until there are multiple frames on the stack.

6. Look at the DEBUG CONSOLE to see the GDB/MI commands sent by MIEngine. I took a copy of the commands issued by a single Step Over operation with ten frames on the stack and attached them here: [stack-list-arguments.gz](https://github.com/microsoft/MIEngine/files/9497978/stack-list-arguments.gz). Here's a summary of the commands issued by MIEngine to GDB:

   | Command | Count |
   | ------- | ----- |
   | `-stack-list-arguments` | 2 |
   | `-stack-list-frames` | 1 |
   | `-stack-list-variables` | 1 |
   | `-stack-select-frame` | 20 |
   | `-var-create` | 210 |
   | `-var-delete` | 210 |
   
   If each of the `-var-create` and `-var-delete` commands takes only a millisecond due to network lag or a GDB before-prompt hook, then there's a 0.4-second pause each time the debuggee stops.

### Analysis

You can see from the log output that each `-var-create` is followed immediately by a `-var-delete`. This means that MIEngine is only issuing these commands to collect information about the variables (not to get a variable reference that it can later use). Why is this? The culprit is `DebuggedProcess.GetParameterInfoOnly()` which has a comment explaining what's happening:

https://github.com/microsoft/MIEngine/blob/c594534c5af5b9cc3ad81b10588f00649db1ce6f/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs#L1975-L1976

The problem is that `GetParameterInfoOnly()` was called with `values = false` and so it passed `PrintValue.NoValues` to the `-stack-list-arguments` call and then followed up with `-var-create` and `-var-delete` commands to get the types of the arguments.

The reason why `values = false` here is that `AD7DebugSession.HandleStackTraceRequestAsync()` did not set the `FIF_FUNCNAME_ARGS_VALUES` flag when calling `AD7Thread.EnumFrameInfo()`. And the reason why this flag is unset is that VSCode did not pass a `format` argument to the `StackTrace` request, and so we are in the "default format" case:

https://github.com/microsoft/MIEngine/blob/c594534c5af5b9cc3ad81b10588f00649db1ce6f/src/OpenDebugAD7/AD7DebugSession.cs#L1706-L1710

### Solution ideas

1. The default format could include `FIF_FUNCNAME_ARGS_VALUES`: then `GetParameterInfoOnly()` would pass `PrintValue.SimpleValues` to the the `-stack-list-arguments` call, which would return the types immediately and there would be no need for the subsequent `-var-create` and `-var-delete` calls.

2. `GetParameterInfoOnly()` could pass `PrintValue.SimpleValues` to the the `-stack-list-arguments` call if _either_ types or values were required, discarding the values information if it is not needed.

### Software versions

Cpptools extension: 1.12.4
VSCode: 1.71.0
Commit: 784b0177c56c607789f9638da7b6bf3230d47a8c
Date: 2022-09-01T07:25:10.472Z
Electron: 19.0.12
Chromium: 102.0.5005.167
Node.js: 16.14.2
V8: 10.2.154.15-electron.0
OS: Linux x64 5.4.0-124-generic
Sandboxed: No

gareth-rees:
### Summary

This reduces the number of round-trips to the debugger, because when values are requested, the `-stack-list-arguments` command also returns the type of each argument immediately, avoiding the need for subsequent `-var-create` and `-var-delete` commands to get these types.

Fixes #1349 (Unnecessary -var-create and -var-delete commands during a stack trace can cause noticeable pause each time the debuggee stops)

### Details

I repeated the reproduction procedure from #1349, took a copy of the commands issued by a single Step Over operation with ten frames on the stack and attached them here: [stack-list-arguments-after.gz](https://github.com/microsoft/MIEngine/files/9498177/stack-list-arguments-after.gz). Here's a comparison of the before and after counts of commands:

| Command | Counts (main) | Counts (this branch) |
| ------- | ----- | ----- |
| `-stack-list-arguments` | 2 | 1 |
| `-stack-list-frames` | 1 | 1 |
| `-stack-list-variables` | 1 | 1 |
| `-stack-select-frame` | 20 | 0 |
| `-var-create` | 210 | 10 |
| `-var-delete` | 210 | 10 |

### Notes

A side-effect of this change is that we get values in the CALL STACK window in Visual Studio Code (see screenshots below). I think this is an improvement, but since it is a big UX change you need to take it into consideration.

| main | this branch |
| ---- | ----------- |
|Screenshot from 2022-09-06 16-02-07|Screenshot from 2022-09-06 16-03-40|

If this is not acceptable, let me know, and I can implement solution 2 from #1349 instead.

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -1703,10 +1703,11 @@ namespace OpenDebugAD7
                 }
                 else
                 {
-                    // No formatting flags provided in the request - use the default format, which includes the module name and argument names / types
+                    // No formatting flags provided in the request - use the default format, which includes the module name and argument names / types / values
                     flags |= enum_FRAMEINFO_FLAGS.FIF_FUNCNAME_MODULE |
                                 enum_FRAMEINFO_FLAGS.FIF_FUNCNAME_ARGS |
                                 enum_FRAMEINFO_FLAGS.FIF_FUNCNAME_ARGS_TYPES |
+                                enum_FRAMEINFO_FLAGS.FIF_FUNCNAME_ARGS_VALUES |
                                 enum_FRAMEINFO_FLAGS.FIF_FUNCNAME_ARGS_NAMES;
                 }
 


gregg-miskelly:
I believe this is going to regress the scenario fixed in https://github.com/Microsoft/MIEngine/pull/673 where the problem was that GDB could take a long time to obtain values.

I wonder if Gareth's scenario could be fixed by kicking off all the `-var-create`'s in parallel, and not waiting on the results of the `-var-deletes`.

gregg-miskelly:
I haven't had time to test it, but I created a branch to play with that kind of fix to see if it helps your scenario: https://github.com/gregg-miskelly/MIEngine/pull/new/ParallelGetTypeOfArgAsync

gareth-rees:
Thanks for the link to #673, this explains why the default format excludes values. But this is surprising to me, since the GDB documentation says that

> if _print-values_ is 2 or --simple-values, print the name, type and value for simple data types, and the name and type for arrays, structures and unions

so I didn't expect that values of large objects would be printed by `--simple-values`. The relevant GDB source code is here:

https://github.com/bminor/binutils-gdb/blob/3c4e2282566d54b974f6929ee5c1270fb3cea4e5/gdb/mi/mi-cmd-stack.c#L652-L657

which means that C++ reference types (`TYPE_CODE_REF` and `TYPE_CODE_RVALUE_REF`) get treated as "simple" values. Possibly this is something that can be fixed in GDB, which would allow MIEngine to add values to the default format (guarded by a [suitable GDB feature](https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI-Support-Commands.html#index-supported-GDB_002fMI-features_002c-list)). I will investigate whether this is possible.

(I doubt that parallelizing the `-var-create` commands will help my case, because I'm in the "GDB before-prompt hook" case and so all the hooks have to be run before any new operations can take place. I can pursue other options here, for example disabling the hook during the stack trace. So don't spend too much time on this.)

gareth-rees:
I think it makes sense to close this PR: I will open another one if I come up with a better approach.

PR #1351, opened 2022-09-08
Add configuration for handling of unknown breakpoints

benmcmorran:
Today, breakpoints set outside of a DAP or AD7 command are treated as application exceptions. This can be unexpected, particularly in cases where the breakpoint was also set in the launch configuration. For example, `postRemoteConnectCommands` might include `tbreak main`. This PR adds a new `unknownBreakpointHandling` configuration option to for these unknown breakpoints.

- `throw` current behavior
- `stop` pause the debug session but do not simulate an application exception
- `ignore` immedately continue

### DIFF
--- a/src/MIDebugPackage/OpenFolderSchema.json
+++ b/src/MIDebugPackage/OpenFolderSchema.json
@@ -185,6 +185,15 @@
               },
               "hardwareBreakpoints": {
                 "$ref": "#/definitions/cpp_schema/definitions/hardwareBreakpointsOptions"
+              },
+              "unknownBreakpointHandling": {
+                "type": "string",
+                "enum": [
+                  "throw",
+                  "stop",
+                  "ignore"
+                ],
+                "description": "Controls how breakpoints set externally (usually via raw GDB commands) are handled when hit.\nAllowed values are \"throw\", which acts as if an exception was thrown by the application, \"stop\", which only pauses the debug session, and \"ignore\", which continues execution without pausing. The default value is \"throw\"."
               }
             },
             "definitions": {
--- a/src/MIDebugPackage/OpenFolderSchema.json
+++ b/src/MIDebugPackage/OpenFolderSchema.json
@@ -190,10 +190,9 @@
                 "type": "string",
                 "enum": [
                   "throw",
-                  "stop",
-                  "ignore"
+                  "stop"
                 ],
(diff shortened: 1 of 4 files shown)


WardenGnaw (approved):
:shipit:

gregg-miskelly:
@benmcmorran was there a scenario you had in mind when `ignore` would be useful? Feels kind of dangerous to me -- if we are getting a bunch of stopping events it is going to make execution super slow, and also break stepping if you are stepping over a function triggering these. Though maybe it is okay to be complete.

src/MIDebugPackage/OpenFolderSchema.json
```diff
@@ -185,6 +185,15 @@
...
                 "$ref": "#/definitions/cpp_schema/definitions/hardwareBreakpointsOptions"
+              },
+              "unknownBreakpointHandling": {
+                "type": "string",
+                "enum": [
+                  "throw",
+                  "stop",
+                  "ignore"
+                ],
+                "description": "Controls how breakpoints set externally (usually via raw GDB commands) are handled when hit.\nAllowed values are \"throw\", which acts as if an exception was thrown by the application, \"stop\", which only pauses the debug session, and \"ignore\", which continues execution without pausing. The default value is \"throw\"."
```
gregg-miskelly, line 196: I would suggest putting in a warning for `ignore`. Maybe something like:

> Note that `"ignore"` can cause side effects such as causing step over to become a go, and poor performance. So, it should generally not be used.

gregg-miskelly (commented):
Otherwise, LGTM

benmcmorran:
@gregg-miskelly I don't have a specific use-case for `ignore`. It was just easy to add for completeness and I was already editing that part of the codebase. Given that it sounds like it could be a dangerous option, I'm happy to remove it, although I'd like to keep the configuration option as an enum to allow for additional handling options in the future.

gregg-miskelly:
Makes sense. I am fine with either leaving it in, but adding a disclaimer, or taking it out.

src/MIDebugPackage/OpenFolderSchema.json
benmcmorran: `ignore` is removed in latest update.

gregg-miskelly (approved):

PR #1357, opened 2022-10-06
Enable NatvisDiagnostics for VS Debug Option

WardenGnaw:
This PR enables Natvis Diagnostics for VS.
If the NatvisDiagnostics in the VS Debug Options is enabled, it will output to the debug pane.

### DIFF
--- a/src/DebugEngineHost/HostConfigurationSection.cs
+++ b/src/DebugEngineHost/HostConfigurationSection.cs
@@ -42,5 +42,7 @@ namespace Microsoft.DebugEngineHost
         {
             return _key.GetValueNames();
         }
+
+        public IntPtr YOLOHandle => _key.Handle.DangerousGetHandle();
     }
 }
--- a/src/DebugEngineHost/HostNatvisProject.cs
+++ b/src/DebugEngineHost/HostNatvisProject.cs
@@ -18,9 +18,26 @@ using Microsoft.VisualStudio.Threading;
 using Microsoft.VisualStudio.Workspace;
 using Microsoft.VisualStudio.Workspace.Indexing;
 using Microsoft.VisualStudio.Workspace.VSIntegration.Contracts;
+using Microsoft.Win32;
 
 namespace Microsoft.DebugEngineHost
 {
+    internal class RegisterMonitorWrapper : IDisposable
+    {
+        public RegistryMonitor CurrentMonitor { get; set; }
+
+        internal RegisterMonitorWrapper(RegistryMonitor currentMonitor)
+        {
+            CurrentMonitor = currentMonitor;
+        }
+
+        public void Dispose()
+        {
+            CurrentMonitor.Dispose();
+            CurrentMonitor = null;
+        }
+    }
+
     /// <summary>
     /// Provides interactions with the host's source workspace to locate and load any natvis files
     /// in the project.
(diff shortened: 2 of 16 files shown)


src/DebugEngineHost/HostConfigurationSection.cs
```diff
@@ -42,5 +42,7 @@ public IEnumerable<string> GetValueNames()
         {
             return _key.GetValueNames();
         }
+
+        public IntPtr YOLOHandle => _key.Handle.DangerousGetHandle();
```
WardenGnaw, line 46: nit: Rename

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1365,5 +1370,10 @@ private string GetDisplayNameFromArrayIndex(uint arrayIndex, int rank, uint[] di
             return displayName.ToString();
         }
 
+        public void Dispose()
```
WardenGnaw, line 1373: Address Warning

src/DebugEngineHost/HostConfigurationSection.cs
gregg-miskelly, line 46: >Handle [](http://example.com/codeflow?start=41&length=6)

I am fairly certain that you can pass SafeHandles in P/Invoke signatures. You might want to do that instead of using `DangerousGetHandle`.

src/DebugEngineHost/RegistryMonitor.cs
```diff
@@ -0,0 +1,148 @@
...
+        {
+            _section = section;
+            _watchSubtree = watchSubtree;
+        }
+
+        public void Start()
+        {
+            Thread registryMonitor = new Thread(Monitor);
+            registryMonitor.IsBackground = true;
+            registryMonitor.Name = "RegistryMonitor";
```
gregg-miskelly, line 86: >RegistryMonitor [](http://example.com/codeflow?start=36&length=15)

"Microsoft.DebugEngineHost.RegistryMonitor" ?

src/DebugEngineHost/RegistryMonitor.cs
```diff
@@ -0,0 +1,148 @@
...
+
+            m_stoppedEvent = new AutoResetEvent(false);
+            m_changeEvent = new AutoResetEvent(false);
+
+            IntPtr handle = m_changeEvent.SafeWaitHandle.DangerousGetHandle();
+
+            int errorCode = RegNotifyChangeKeyValue(_section.YOLOHandle, _watchSubtree, RegChangeNotifyFilter.Key | RegChangeNotifyFilter.Attribute | RegChangeNotifyFilter.Value | RegChangeNotifyFilter.Security, handle, true);
+            if (errorCode != 0) // 0 is ERROR_SUCCESS
+            {
+                throw new Win32Exception(errorCode);
```
gregg-miskelly, line 112: >throw new Win32Exception(errorCode); [](http://example.com/codeflow?start=16&length=36)

This will crash VS. I think you want to log this and go away.

src/DebugEngineHost/RegistryMonitor.cs
```diff
@@ -0,0 +1,148 @@
...
+        #region Native Methods
+
+        /// <summary>
+        /// Filter for notifications reported by <see cref="RegistryMonitor"/>.
+        /// </summary>
+        [Flags]
+        public enum RegChangeNotifyFilter
+        {
+            /// <summary>Notify the caller if a subkey is added or deleted.</summary>
+            Key = 1,
```
gregg-miskelly, line 30: >Key [](http://example.com/codeflow?start=12&length=3)

nit: It is slightly weird that these names don't match up with MSDN. Might be worth at least including the Win32 constant names (ex: `REG_NOTIFY_CHANGE_NAME`): [https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regnotifychangekeyvalue](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regnotifychangekeyvalue)

src/DebugEngineHost/RegistryMonitor.cs
```diff
@@ -0,0 +1,148 @@
...
+                        stopped = true;
+                    }
+                    else
+                    {
+                        errorCode = RegNotifyChangeKeyValue(_section.YOLOHandle, _watchSubtree, RegChangeNotifyFilter.Key | RegChangeNotifyFilter.Attribute | RegChangeNotifyFilter.Value | RegChangeNotifyFilter.Security, handle, true);
+                        if (errorCode != 0) // 0 is ERROR_SUCCESS
+                        {
+                            throw new Win32Exception(errorCode);
+                        }
+                        RegChanged.Invoke(this, null);
```
gregg-miskelly, line 131: >. [](http://example.com/codeflow?start=34&length=1)

Should this be: `?.`

src/DebugEngineHost/RegistryMonitor.cs
```diff
@@ -0,0 +1,148 @@
...
+        private void Monitor()
+        {
+            bool stopped = false;
+
+            m_stoppedEvent = new AutoResetEvent(false);
+            m_changeEvent = new AutoResetEvent(false);
+
+            IntPtr handle = m_changeEvent.SafeWaitHandle.DangerousGetHandle();
+
+            int errorCode = RegNotifyChangeKeyValue(_section.YOLOHandle, _watchSubtree, RegChangeNotifyFilter.Key | RegChangeNotifyFilter.Attribute | RegChangeNotifyFilter.Value | RegChangeNotifyFilter.Security, handle, true);
```
gregg-miskelly, line 109: >RegChangeNotifyFilter.Key | RegChangeNotifyFilter.Attribute | RegChangeNotifyFilter.Value | RegChangeNotifyFilter.Security [](http://example.com/codeflow?start=88&length=122)

Do you care about `Attributes` or `Security`?

src/DebugEngineHost/HostNatvisProject.cs
```diff
@@ -18,9 +18,26 @@
 using Microsoft.VisualStudio.Workspace;
 using Microsoft.VisualStudio.Workspace.Indexing;
 using Microsoft.VisualStudio.Workspace.VSIntegration.Contracts;
+using Microsoft.Win32;
 
 namespace Microsoft.DebugEngineHost
 {
+    internal class RegisterMonitorWrapper : IDisposable
```
gregg-miskelly, line 25: >RegisterMonitorWrapper [](http://example.com/codeflow?start=19&length=22)

Maybe I will eventually figure this out, but why do we need a wrapper instead of just directly holding the registry monitor?

src/DebugEngineHost/HostNatvisProject.cs
```diff
@@ -49,6 +66,111 @@ public static void FindNatvis(NatvisLoader loader)
...
+                    break;
+                case "error":
+                    logLevel = LogLevel.Error;
+                    break;
+                case "warning":
+                    logLevel = LogLevel.Warning;
+                    break;
+                case "verbose":
+                default: // Unknown, default to verbose
+                    logLevel = LogLevel.Verbose;
```
gregg-miskelly, line 156: >logLevel = LogLevel.Verbose; [](http://example.com/codeflow?start=20&length=28)

If the key is missing, is Verbose the correct default?

src/DebugEngineHost/RegistryMonitor.cs
```diff
@@ -0,0 +1,148 @@
...
+            /// include adding or deleting a value, or changing an existing value.</summary>
+            Value = 4,
+            /// <summary>Notify the caller of changes to the security descriptor
+            /// of the key.</summary>
+            Security = 8,
+        }
+
+        [DllImport("advapi32.dll", SetLastError = true)]
+        private static extern int RegOpenKeyEx(IntPtr hKey, string subKey, uint options, int samDesired,
+                                               out IntPtr phkResult);
```
gregg-miskelly, line 44: This wasn't used

gregg-miskelly (approved):
Otherwise LGTM

src/DebugEngineHost/HostNatvisProject.cs
WardenGnaw: I think this was before I discovered the MIDebugEngine did not really need to know about the RegistorMonitor class and we could just cover cleanup with IDisposable.

WardenGnaw: Nope just remembered why myself now 😅

It was because of `WatchNatvisOptionSetting` and we need to switch out the registry we are watching since we are watching two different registry keys.

src/DebugEngineHost/RegistryMonitor.cs
WardenGnaw: `Microsoft.Win32.RegistryKey.OpenSubKey` opens with `0x2001F` which covered `KEY_NOTIFY (0x0010)`

src/DebugEngineHost/HostNatvisProject.cs
WardenGnaw: I think setting it in the middle would be better. Setting it to be Warning

PR #1360, opened 2022-10-12
Catch ProtocolException in HandleEvaluateAsync

WardenGnaw:
Catch ProtocolException from GetDebugPropertyFromExpression. If frames change and then we get a request,
GetDebugPropertyFromException would throw and crash the debug adapter.

This PR catches the exception and sets it to the responder of the Caller in HandleEvaluateAsync

Based off of @jfblazquez changes at https://github.com/microsoft/MIEngine/compare/main...jfblazquez:MIEngine:EvaluateRequestAsyncProtocolException

Related: https://github.com/microsoft/vscode-cpptools/issues/9438


jaredpar:
@WardenGnaw think you @'d the wrong person here. I didn't make those changes.

WardenGnaw:
> @WardenGnaw think you @'d the wrong person here. I didn't make those changes.

Whoops, autocompleted to wrong person. Updated to jfblazquez

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -3034,7 +3034,16 @@ protected override void HandleEvaluateRequestAsync(IRequestResponder<EvaluateArg
                 propertyInfoFlags |= (enum_DEBUGPROP_INFO_FLAGS)enum_DEBUGPROP_INFO_FLAGS110.DEBUGPROP110_INFO_NOSIDEEFFECTS;
             }
 
-            GetDebugPropertyFromExpression(eb, expression, frameId, isExecInConsole, flags, dapEvalFlags, out IDebugProperty2 property);
+            IDebugProperty2 property;
+            try
+            {
+                GetDebugPropertyFromExpression(eb, expression, frameId, isExecInConsole, flags, dapEvalFlags, out property);
+            }
+            catch (ProtocolException pe)
```
gregg-miskelly, line 3042: >catch (ProtocolException pe) [](http://example.com/codeflow?start=12&length=28)

While we are at it, should we add a `try`/`catch` around the whole method?

gregg-miskelly (approved):

PR #1362, opened 2022-10-13
Fix LogPoint stepping

WardenGnaw:
As mentioned in: https://github.com/microsoft/vscode-cpptools/issues/9995

Due to https://github.com/microsoft/MIEngine/commit/f91dec6efdf5bd70b06ded5d82301a0968e02b05 we lost the state that we were stepping. This PR re-adds the flag that we are in the middle of the step so the log point will stop hit as a bp. 

@gareth-rees as an FYI since I think you really wanted to just clear the following states in that commit.
```csharp
            m_isStopped = false;
            m_variableManager.Reset();
            m_frameHandles.Reset();
            m_gotoCodeContexts.Clear();
```


gregg-miskelly:
To be clear: the design here is that if you hit a trace point while we are stepping, we will just stop, and leave it up to the user to get back to the function they were stepping, correct? I think this is the best that is possible with GDB, but just wanted to be clear that log points + stepping can't really work well together.

gregg-miskelly:
>             m_isStepping = true;

Should we remove this?

---
Refers to: src/OpenDebugAD7/AD7DebugSession.cs:863 in 5f35bf3. [](commit_id = 5f35bf3eef1a71ae31ce3534ea34e639aba3016c, deletion_comment = False)

WardenGnaw:
> To be clear: the design here is that if you hit a trace point while we are stepping, we will just stop, and leave it up to the user to get back to the function they were stepping, correct? I think this is the best that is possible with GDB, but just wanted to be clear that log points + stepping can't really work well together.

Yep. Any stepping that causes a LogPoint to fire results in the LogPoint becoming a breakpoint since we are unable to continue with the original step command. 

The user also expects the same:
> Expected: the logpoint is treated as a breakpoint so the program stops at the return 7 line.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -887,6 +887,7 @@ private void StepInternal(int threadId, enum_STEPKIND stepKind, SteppingGranular
             // The program should now be stepping, so it is safe to discard the
             // cached program state.
             BeforeContinue();
+            m_isStepping = true;
```
gregg-miskelly, line 890: >m_isStepping [](http://example.com/codeflow?start=12&length=12)

I am not sure if you want to deal with it, but [https://github.com/microsoft/MIEngine/commit/f91dec6efdf5bd70b06ded5d82301a0968e02b05](https://github.com/microsoft/MIEngine/commit/f91dec6efdf5bd70b06ded5d82301a0968e02b05) also introduces a race here where if we get a stopping event extremely quickly, we could wind up blowing away the stopped state. We really need some sort of lock around the call to step.

gregg-miskelly (approved):

nickatundo:
Hi, I reported the original issue: https://github.com/microsoft/vscode-cpptools/issues/9995. Thank you for jumping on this and coming up with a fix so quickly.

Is there anything I can do to help move this along toward merge? Having briefly looked at the test failures it would take me a while to understand what's going on there, but I can try to get into it if nobody else has time.

src/OpenDebugAD7/AD7DebugSession.cs
WardenGnaw: Tracking at https://github.com/microsoft/MIEngine/issues/1372.

PR #1367, opened 2022-10-31
Add "sendInvalidate" request

sbobko:
This PR adds support for OpenDebugAD7 to handle `sendInvalidate` requests which will respond if it successfully fired an InvalidatedEvent to the UI.

Here is the schema:
```json
"SendInvalidateRequest": {
    "allOf": [
        {
            "$ref": "#/definitions/Request"
        },
        {
            "type": "object",
            "properties": {
               "areas": {
                  "type": "array",
                  "description": "Set of logical areas that got invalidated. This property has a hint characteristic: a client can only be expected to make a 'best effort' in honoring the areas but there are no guarantees. If this property is missing, empty, or if values are not understood, the client should assume a single value `all`.",
                  "items": {
                     "$ref": "#/definitions/InvalidatedAreas"
                  }
               },
                "threadId": {
                    "type": "int",
                },
                "stackFrameId": {
                    "type": "int",
                },
            },
            "required": [
                "areas"
            ]
        }
    ]
},
```

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -128,8 +128,54 @@ namespace OpenDebugAD7
             m_dataBreakpoints = new Dictionary<string, IDebugPendingBreakpoint2>();
             m_exceptionBreakpoints = new List<string>();
             m_variableManager = new VariableManager();
+
+            //Register sendInvalidate request
+            Protocol.RegisterRequestType<SendInvalidateRequest, SendInvalidateArguments, SendInvalidateResponse>(r => this.HandleSendInvalidateRequestAsync(r));
+    
         }
 
+        private void HandleSendInvalidateRequestAsync(IRequestResponder<SendInvalidateArguments> responder)
+        {
+            InvalidatedEvent invalidated = new InvalidatedEvent();
+            SendInvalidateResponse response = new SendInvalidateResponse();
+            
+            // Setting the area and adding it to the result
+            switch (responder.Arguments.Areas.ToString())
+            {
+                case "Threads":
+                    invalidated.Areas.Add(InvalidatedAreas.Threads);
+                    break;
+                case "Stacks":
+                    invalidated.Areas.Add(InvalidatedAreas.Stacks);
+                    break;
+                case "Variables":
+                    invalidated.Areas.Add(InvalidatedAreas.Variables);
+                    break;
+                case "Unknown":
+                    invalidated.Areas.Add(InvalidatedAreas.Unknown);
+                    break;
+                default:
+                    invalidated.Areas.Add(InvalidatedAreas.All);
+                    break;
+            }
+
+            // Setting the StackFrameId if passed (and the 'threadId' is ignored).
+            if (null != responder.Arguments.StackFrameId)
+            {
+                invalidated.StackFrameId = responder.Arguments.StackFrameId;
+            }
+
+            // Setting the ThreadId if passed
+            else if (null != responder.Arguments.ThreadId)
+            {
+                invalidated.ThreadId = responder.Arguments.ThreadId;
+            }
+
+
+            Protocol.SendEvent(invalidated);
+            response.body.result = JsonConvert.SerializeObject(responder);
+            responder.SetResponse(response);
+        }
         #endregion
 
         #region Utility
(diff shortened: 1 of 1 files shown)


iAbadia:
I'm interested in having this PR merged

sbobko:
> 
Hello @iAbadia , Unfortunately, I'm busy right now and I can't spare the time. You can use the code for inspiration and make your own contribution.

iAbadia:
Thanks @sbobko , I'll do that!

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -3874,4 +3930,35 @@ int IDebugSettingsCallback110.ShouldSuppressImplicitToStringCalls(out int pfSupp
...
+
+    internal class SendInvalidateRequest : DebugRequestWithResponse<SendInvalidateArguments, SendInvalidateResponse>
+    {
+ 
+        public SendInvalidateRequest(): base("sendInvalidate")
+        {
+        }
+    }
+
+    internal class SendInvalidateResponse:ResponseBody
```
WardenGnaw, line 3942:
```suggestion
    internal class SendInvalidateResponse : ResponseBody
```

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -3874,4 +3930,35 @@ int IDebugSettingsCallback110.ShouldSuppressImplicitToStringCalls(out int pfSupp
...
+    {
+        public sealed class Body
+        {
+            [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
```
WardenGnaw, line 3946: Can we define all the fields here instead of creating a json string? Let the JSON librarly handle the serialize of the class.

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -128,8 +128,64 @@ public AD7DebugSession(Stream debugAdapterStdIn, Stream debugAdapterStdOut, List
...
+    
         }
 
+        private void HandleSendInvalidateRequestAsync(IRequestResponder<SendInvalidateArguments> responder)
+        {
+            InvalidatedEvent invalidated = new InvalidatedEvent();
+            SendInvalidateResponse response = new SendInvalidateResponse();
+            
+            // Setting the area and adding it to the result
+            response.body.result += "sendInvalidate: { Area: ";
```
WardenGnaw, line 143: I'm not a fan of creating json strings by hand, can you define this as a class that can be serialized by the JSON library?

WardenGnaw (changes requested):

src/OpenDebugAD7/AD7DebugSession.cs
sbobko: Removed all JsonProperty

src/OpenDebugAD7/AD7DebugSession.cs
sbobko: I changed the output format a bit, now it's a serializable class, so there shouldn't be any problems

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -3874,4 +3920,31 @@ int IDebugSettingsCallback110.ShouldSuppressImplicitToStringCalls(out int pfSupp
...
+            public string result;
+        }
+
+        public Body body = new Body();
+    }
+
+    internal class SendInvalidateArguments : DebugRequestArguments
+    {
+
+        public string Areas { get; set; }
```
WardenGnaw, line 3945: I missed this in an earlier review, are you able to use `InvalidatedAreas`?

This would make lines 143 - 160 be:
```csharp
invalidated.Areas.Add(responder.Arguments.Areas)
```

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -3874,4 +3920,31 @@ int IDebugSettingsCallback110.ShouldSuppressImplicitToStringCalls(out int pfSupp
             }
         }
     }
+
+    internal class SendInvalidateRequest : DebugRequestWithResponse<SendInvalidateArguments, SendInvalidateResponse>
```
WardenGnaw, line 3924: In your extension, are you checking the contents of this response?

If you are, could you modify it to be similar to https://github.com/microsoft/VSDebugAdapterHost/blob/3bcc0b01df7e604227e1a0334a4160777901c2ea/src/sample/SampleDebugAdapter.VSIX/CustomProtocolExtension/SampleExtension.cs#L84-L97

E.g.
```csharp
internal class SendInvalidateResponse : ResponseBody
{
        public InvalidatedAreas Areas { get; set; }
        public int? ThreadId { get; set; }
        public int? StackFrameId { get; set; }
}
```

If not, then we could remove `SendInvalidateResponse` and just use `DebugRequest`
```suggestion
    internal class SendInvalidateRequest : DebugRequest<SendInvalidateArguments>
```

src/OpenDebugAD7/AD7DebugSession.cs
sbobko: done

src/OpenDebugAD7/AD7DebugSession.cs
sbobko: In my case, response does not play any role, I have removed all mentions

WardenGnaw (approved):
LGTM Thanks for the contribution. Once the PR checks pass I'll check it in and it should show in the cpptools changelog when it is in.

PR #1377, opened 2022-11-29
Fix race on stopping monitor

WardenGnaw:
This PR fixes the race condition when disposing of the RegistryMonitor and when detecting the stop.

This PR modifies Dispose to call Stop() which will cause the wait handler to fire and stop the Monitor thread which will properly dispose of the object.

### DIFF
--- a/src/DebugEngineHost/RegistryMonitor.cs
+++ b/src/DebugEngineHost/RegistryMonitor.cs
@@ -47,14 +47,11 @@ namespace Microsoft.DebugEngineHost
 
         #endregion
 
-        private HostConfigurationSection _section;
+        private readonly HostConfigurationSection _section;
         private readonly bool _watchSubtree;
 
-        // Set when registry value is changed
-        private AutoResetEvent m_changeEvent;
-
         // Set when monitoring is stopped
-        private AutoResetEvent m_stoppedEvent;
+        private AutoResetEvent _stoppedEvent;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/DebugEngineHost/RegistryMonitor.cs
```diff
@@ -80,10 +77,7 @@ public void Start()
 
         public void Stop()
         {
-            if (m_stoppedEvent != null)
-            {
-                m_stoppedEvent.Set();
-            }
+            _stoppedEvent?.Set();
```
gregg-miskelly, line 80: >_stoppedEvent?.Set(); [](http://example.com/codeflow?start=12&length=21)

Is it possible to call `Stop` more than once? If so, this could throw `ObjectDisposedException`.

gregg-miskelly (approved):

src/DebugEngineHost/RegistryMonitor.cs
WardenGnaw: I think we should be able to call stop multiple times safely. 

However, I think we should only allow Dispose to be called once?

gregg-miskelly: If you want to allow calling `Stop` more than once, you need something like an `isStopped` field and take a lock or do an interlocked operation in order to make sure that `_stoppedEvent.Set()` is only called once. Calling it a second time may throw since the other thread disposes it.

gregg-miskelly (approved):

PR #1380, opened 2023-01-12
Update Newtonsoft.JSON to 13.0.2

WardenGnaw:


gregg-miskelly:
Visual Studio main is still using 13.0.1. Are you sure you want this? Are the two versions compatible?

PR #1381, opened 2023-01-24
User Story 1565289: [MIEngine Natvis] Support multiple natvis files in visualizerFile launch configuration

gc46:


test/CppTests/OpenDebug/CrossPlatCpp/LaunchCommand.cs
```diff
@@ -34,12 +37,24 @@ public sealed class CppLaunchCommandArgs : LaunchCommandArgs
         public string MIMode;
 
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
-        public string VisualizerFile;
+        public List<string> VisualizerFile { get; set; }
```
WardenGnaw, line 40: You can remove this since if we want to use a visualizer file use one of the other classes instead.

test/CppTests/OpenDebug/CrossPlatCpp/LaunchCommand.cs
```diff
@@ -4,15 +4,18 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Text.Json;
```
WardenGnaw, line 7: Remove `using System.Text*`

test/CppTests/CppTests.csproj
```diff
@@ -39,16 +39,18 @@
     <ContentFiles Include="TestConfigurations\**" />
   </ItemGroup>
 
+  <ItemGroup>
```
WardenGnaw, line 42: Remove, already captured on line 35.

WardenGnaw (changes requested):
You will need to fix https://github.com/microsoft/MIEngine/pull/1381/files#file-src-androiddebuglauncher-launcher-cs-L462

test/CppTests/Tests/NatvisTests.cs
```diff
@@ -480,6 +480,56 @@ public void TestCommaFormatWithSquareBrackets(ITestSettings settings)
...
+                runner.SetBreakpoints(writerBreakpoints);
+
+                runner.Expects.StoppedEvent(StoppedReason.Breakpoint).AfterConfigurationDone();
+
+                using (IThreadInspector threadInspector = runner.GetThreadInspector())
+                {
+                    IFrameInspector currentFrame = threadInspector.Stack.First();
+
+                    this.Comment("Verifying TreeItems natvis");
+                    var map = currentFrame.GetVariable("map");
```
WardenGnaw, line 514: This only tests to see if `Simple2.natvis` got parsed correctly. Can you also test for `Simple1,natvis`?

What happens if we have collisions in the natvis files.

test/CppTests/OpenDebug/CrossPlatCpp/LaunchCommand.cs
WardenGnaw: yes you will need to fix those based on removing this.

test/CppTests/Tests/NatvisTests.cs
WardenGnaw: You are only testing map. you are not testing any variable that is being visualized by Simple1.natvis.

src/MICore/LaunchOptions.cs
```diff
@@ -960,11 +960,11 @@ public string AdditionalSOLibSearchPath
             }
         }
 
-        private string _visualizerFile;
+        private List<string> _visualizerFile;
         /// <summary>
         /// [Optional] Natvis file name - from install location
         /// </summary>
-        public string VisualizerFile
+        public List<string> VisualizerFile
```
gregg-miskelly, line 967: >VisualizerFile [](http://example.com/codeflow?start=28&length=14)

nit: rename to `VisualizerFiles`

gregg-miskelly:
>             }

I think you want to remove the `set` and backing field and just have:

```
        /// <summary>
        /// Collection of natvis files to use when evaluating
        /// </summary>
        public List<string> VisualizerFile { get; } = new List<string>();
```

---
Refers to: src/MICore/LaunchOptions.cs:974 in 50400ad. [](commit_id = 50400ad92d55785042d3851a4d472b8fbf985902, deletion_comment = False)

src/MICore/LaunchOptions.xsd.types.designer.cs
```diff
@@ -387,7 +388,7 @@ public partial class BaseLaunchOptions {
         
         /// <remarks/>
         [System.Xml.Serialization.XmlAttributeAttribute()]
-        public string VisualizerFile;
+        public List<string> VisualizerFile;
```
gregg-miskelly, line 391: >public List<string> VisualizerFile; [](http://example.com/codeflow?start=8&length=35)

This is a generated file, and you didn't change the schema, so it shouldn't have changed

src/MICore/JsonLaunchOptions.cs
```diff
@@ -42,7 +42,8 @@ public abstract partial class BaseOptions
         /// .natvis file to be used when debugging this process. This option is not compatible with GDB pretty printing. Please also see "showDisplayString" if using this setting.
```
gregg-miskelly, line 42: >file [](http://example.com/codeflow?start=20&length=4)

nit: 'files'

src/MICore/JsonLaunchOptions.cs
```diff
@@ -123,6 +124,33 @@ public abstract partial class BaseOptions
...
+    internal class VisualizerFileConverter : JsonConverter
+    {
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            List<string> visualizerFile = new List<string>();
+            if (reader.TokenType == JsonToken.StartArray)
+            {
+                visualizerFile = serializer.Deserialize<List<string>>(reader);
+            }
+            else
```
gregg-miskelly, line 136: >else [](http://example.com/codeflow?start=12&length=4)

I wonder if we should check if the TokenType is a string so that if the user does something like `"visualizerFile": 12` they get a failure about bogus json.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -321,6 +321,20 @@ public void Initialize(string fileName)
...
+        /*
+         * Handle multiple Natvis files
+         */
+        public void Initialize(List <string> fileNames)
+        {
+            if (fileNames.Count > 0)
+            {
+                foreach (var file in fileNames)
+                {
+                    Initialize(file);
```
gregg-miskelly, line 333: >Initialize(file); [](http://example.com/codeflow?start=20&length=17)

Not all of Initialize should be repeated, and parts of Initialize look like they should be run even if there are zero files.

test/CppTests/OpenDebug/CrossPlatCpp/LaunchCommand.cs
```diff
@@ -34,12 +35,24 @@
         public string MIMode;
 
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
-        public string VisualizerFile;
+        public List<string> VisualizerFile { get; set; }
```
gregg-miskelly, line 38: >List<string> [](http://example.com/codeflow?start=15&length=12)

You might want to make this an `object` instead so that you can test both the old and new format for this.

gregg-miskelly: Oh, I see. Nevermind.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
WardenGnaw: > Not all of Initialize should be repeated,

Yep. There does not need to be multiple registrations of the natvis registration key. 

>and parts of Initialize look like they should be run even if there are zero files.

Everything else seems to be under if (!string.IsNullOrEmpty(fileName))

src/MICore/LaunchOptions.xsd.types.designer.cs
WardenGnaw: I don't think we need to support this for XML. This is mostly used for VS scenarios and there is already a project system natvis file detector. We mostly need JSON support.

gregg-miskelly: Makes sense to me. We just need to backout the changes to this file.

gregg-miskelly:
> Why do we not want the `set()`?

Generally, properties with a collection type shouldn't have a `set` method. Instead, they get assigned in the constructor and items are added to them.

test/CppTests/Tests/NatvisTests.cs
```diff
@@ -480,6 +480,69 @@ public void TestCommaFormatWithSquareBrackets(ITestSettings settings)
             }
         }
 
+        [Theory]
+        [DependsOnTest(nameof(CompileNatvisDebuggee))]
+        [RequiresTestSettings]
+        public void TestMultipleNatvisFiles(ITestSettings settings)
+        {
+            this.TestPurpose("This test checks if TreeItems are visualized.");
```
WardenGnaw, line 488: nit: Fix Test Purpose

src/MIDebugEngine/Natvis.Impl/Natvis.cs
WardenGnaw: I believe you will want to move the following to its own method and call it if `List<string> filenames` is not empty,
```
void InitializeNatvisServices() {
            try
            {
                _natvisSettingWatcher = HostNatvisProject.WatchNatvisOptionSetting(_configStore, _process.Logger.NatvisLogger);
                HostNatvisProject.FindNatvis((s) => LoadFile(s));
            }
            catch (FileNotFoundException)
            {
                // failed to find the VS Service
            }
}
```

Call this method in the list<string> initialize function.

WardenGnaw (approved):

src/MICore/LaunchOptions.xsd.types.designer.cs
```diff
@@ -13,8 +13,9 @@
 // 
 namespace MICore.Xml.LaunchOptions {
```
gregg-miskelly, line 14: >namespace [](http://example.com/codeflow?start=0&length=9)

You either need to reset this file back to where it was (recommended approach) or make the changes to the LaunchOptions XSD and regenerate it.

gregg-miskelly:
>             }

This is still active.

If it makes your life easier, you can allow a set method, but in this case I would make these other changes:
* Still initialize the backing field to an empty collection so you don't need those null checks.
* Only allow `set` if the backing field is still an empty collection


---
In reply to: [1404353242](https://github.com/microsoft/MIEngine/pull/1381#issuecomment-1404353242) [](http://example.com/codeflow?ancestors=1404353242,1404353242)

---
Refers to: src/MICore/LaunchOptions.cs:974 in 50400ad. [](commit_id = 50400ad92d55785042d3851a4d472b8fbf985902, deletion_comment = False)

gregg-miskelly:
>         /// Don't change this test without checking with C++ team.

We are changing this test. 

---
Refers to: src/MICoreUnitTests/BasicLaunchOptionsTests.cs:255 in 3cb7931. [](commit_id = 3cb7931ff5764cf126edadcdc14d80876a287471, deletion_comment = False)

src/MICoreUnitTests/BasicLaunchOptionsTests.cs
```diff
@@ -263,7 +263,8 @@ internal void VerifyCoreApisPresent()
             launchOptions.WorkingDirectory = "/home/user";
             launchOptions.DebuggerMIMode = MIMode.Gdb;
             launchOptions.WaitDynamicLibLoad = false;
-            launchOptions.VisualizerFile = @"c:\myproject\file.natvis";
+            launchOptions.VisualizerFiles = new List<string>();
```
gregg-miskelly, line 266: @paulmaybee Do you know if the C++ team still owns something that relies on this property?

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -321,6 +324,21 @@ public void Initialize(string fileName)
...
         }
 
+        /*
+         * Handle multiple Natvis files
+         */
+        public void Initialize(List <string> fileNames)
+        {
+            if (fileNames != null && fileNames.Count > 0)
+            {
+                InitializeNatvisServices();
```
gregg-miskelly, line 334: >InitializeNatvisServices(); [](http://example.com/codeflow?start=16&length=27)

This should be moved outside the `if`

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -260,7 +260,7 @@ internal Natvis(DebuggedProcess process, bool showDisplayString, HostConfigurati
             _configStore = configStore;
         }
 
-        public void Initialize(string fileName)
+        public void InitializeNatvisServices()
```
gregg-miskelly, line 263: >public [](http://example.com/codeflow?start=8&length=6)

`private`?

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -271,7 +271,10 @@ public void Initialize(string fileName)
             {
                 // failed to find the VS Service
             }
+        }
 
+        public void Initialize(string fileName)
+        {
```
gregg-miskelly, line 277: >{ [](http://example.com/codeflow?start=8&length=1)

Call `InitializeNatvisServices`?

test/CppTests/OpenDebug/CrossPlatCpp/LaunchCommand.cs
```diff
@@ -50,10 +52,16 @@ public class LaunchCommand : LaunchCommand<CppLaunchCommandArgs>
...
+        public LaunchCommand(IDebuggerSettings settings, string program, object visualizerFile = null, bool isAttach = false, params string[] args)
         {
+            if (!(visualizerFile == null || visualizerFile is string || visualizerFile is List<string>))
+            {
+                throw new ArgumentOutOfRangeException(nameof(visualizerFile));
+            }
+
             this.Timeout = TimeSpan.FromSeconds(15);
 
+            this.Args = new CppLaunchCommandArgs();
```
gregg-miskelly, line 64: >this.Args = new CppLaunchCommandArgs(); [](http://example.com/codeflow?start=12&length=39)

You shouldn't need this change

src/MICore/LaunchOptions.xsd.types.designer.cs
WardenGnaw: yep you can run
`git checkout origin/main src/MICore/LaunchOptions.xsd.types.designer.cs` to reset it back to the state in `main`

src/MICoreUnitTests/BasicLaunchOptionsTests.cs
WardenGnaw: One note is that we still support sending over this field as a string in JSON, just in the codebase everything is represented as a list so this had to change.

WardenGnaw:
> > > ```
> > >         }
> > > ```
> > 
> > 
> > This is still active.
> > If it makes your life easier, you can allow a set method, but in this case I would make these other changes:
> > 
> > * Still initialize the backing field to an empty collection so you don't need those null checks.
> > * Only allow `set` if the backing field is still an empty collection
> > 
> > In reply to: [1404353242](https://github.com/microsoft/MIEngine/pull/1381#issuecomment-1404353242)
> > Refers to: src/MICore/LaunchOptions.cs:974 in [50400ad](https://github.com/microsoft/MIEngine/commit/50400ad92d55785042d3851a4d472b8fbf985902). [](commit_id = [50400ad](https://github.com/microsoft/MIEngine/commit/50400ad92d55785042d3851a4d472b8fbf985902), deletion_comment = False)
> 
> The reason I allowed for `set` is due to the `VisualizerFiles` assignment in other places throughout the codebase such as the one shown below: image
> 
> We can disallow the ability to set `VisualizerFiles` though if you deem it fine.

You can let `LaunchOptions.VisualizerFiles` be created during the constructor and it will be empty.
When we need to assign/pass the Users strings you can use `this.VisualizerFiles.AddRange(option.VisualizerFiles)`.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
WardenGnaw: InitializeNatvisServices should only be called once. I think these we should add this per natvis file.

src/MICore/LaunchOptions.cs
```diff
@@ -960,19 +960,11 @@
             }
         }
 
-        private string _visualizerFile;
+        private List<string> _visualizerFile;
```
gregg-miskelly, line 963: >private [](http://example.com/codeflow?start=8&length=7)

Remove the field

src/MICore/LaunchOptions.cs
```diff
@@ -1579,9 +1571,9 @@
             {
                 DebugChildProcesses = suppOptions.DebugChildProcesses;
             }
-            if (string.IsNullOrWhiteSpace(VisualizerFile))
+            if (VisualizerFiles == null || VisualizerFiles.Count == 0)
```
gregg-miskelly, line 1574: >VisualizerFiles == null ||  [](http://example.com/codeflow?start=16&length=27)

nit: no need to check for null anymore

src/MICore/LaunchOptions.cs
```diff
@@ -1841,8 +1836,8 @@
             if (string.IsNullOrEmpty(this.WorkingDirectory))
                 this.WorkingDirectory = source.WorkingDirectory;
 
-            if (string.IsNullOrEmpty(this.VisualizerFile))
-                this.VisualizerFile = source.VisualizerFile;
+            if (this.VisualizerFiles == null || this.VisualizerFiles.Count == 0)
```
gregg-miskelly, line 1839: >if (this.VisualizerFiles == null || this.VisualizerFiles.Count == 0) [](http://example.com/codeflow?start=12&length=68)

This check doesn't make sense. You want: `if (!string.IsNullOrEmpty(source.VisualizerFile))`

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -271,7 +271,11 @@ public void Initialize(string fileName)
...
```
gregg-miskelly, line 277: I failed to think about this earlier, but you got rid of the only call site to this method. You can just change this method to take the file list, you don't need two different Initialize methods with different signatures.

src/MICore/LaunchOptions.cs
WardenGnaw: This specific instance is for the XML merge case where its using the XML.VisualizerFile not the JSON one. 

I would remove the outer loop gregg suggested and just do the inner check with `file` as suppOptions.VisualizerFile

src/MICore/JsonLaunchOptions.cs
```diff
@@ -137,6 +137,10 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
             {
                 visualizerFile.Add(reader.Value.ToString());
             }
+            else
+            {
+                throw new Exception("Input is neither a List nor a String");
```
WardenGnaw, line 142: You will want to Localize this.

Add an entry in https://github.com/microsoft/MIEngine/blob/5b50551e6827381fcd238c3669142fab0bcc8097/src/MICore/MICoreResources.resx

src/MICore/JsonLaunchOptions.cs
```diff
@@ -123,6 +124,37 @@ public abstract partial class BaseOptions
...
+            {
+                visualizerFile = serializer.Deserialize<List<string>>(reader);
+            }
+            else if (reader.TokenType == JsonToken.String)
+            {
+                visualizerFile.Add(reader.Value.ToString());
+            }
+            else
+            {
+                throw new Exception("Input is neither a List nor a String");
```
gregg-miskelly, line 142: You probably want to use a JSON serialization exception rather than `throw new Exception`. Looks like the exception would be thrown if you didn't use a JsonConverter would be `Newtonsoft.Json.JsonReaderException`, so I think I would suggest that.

src/MICore/MICoreResources.resx
```diff
@@ -326,4 +326,7 @@ Error: {1}</value>
   <data name="Error_SourceFileMapInvalidEditorPath" xml:space="preserve">
     <value>'editorPath' can not be null or empty</value>
   </data>
+  <data name="Error_IncorrectVisualizerFileFormat" xml:space="preserve">
+    <value>Input is neither a List nor a String.</value>
```
gregg-miskelly, line 330: How about: "'visualizerFile' must be a string or array of strings"

gregg-miskelly (approved):

PR #1390, opened 2023-03-07
Add SDL Tasks

WardenGnaw:
FYI @isadorasophia


eng/pipelines/MIDebugEngine-CI.yml
```diff
@@ -1,5 +1,15 @@
 ---
 name: $(Date:yyyMMdd).$(Rev:r)
+
+schedules:
+# Run on the 15th and 30th of every month
```
gregg-miskelly, line 5: >30th [](http://example.com/codeflow?start=22&length=4)

Is it an issue that there will not be a run between February 15th and March 15th?

eng/pipelines/steps/APIScan.yml
```diff
@@ -8,14 +8,17 @@ steps:
     SourceFolder: ${{ parameters.FolderToScan }}
     Contents: |
       **\*Microsoft@(*.dll|*.pdb|*.exe)
-      **\*Newtonsoft@(*.dll|*.pdb|*.exe)
       **\*OpenDebugAD7@(*.dll|*.pdb|*.exe)
       **\*WindowsDebugLauncher@(*.dll|*.pdb|*.exe)
       !**\*.resources.dll
+      !**\Microsoft.VisualStudio.Debugger.Interop.15.0.dll
```
gregg-miskelly, line 14: >!**\Microsoft.VisualStudio.Debugger.Interop.15.0.dll [](http://example.com/codeflow?start=6&length=52)

Is this because the scan tool needs to find all the dependencies?

gregg-miskelly (approved):
Otherwise LGTM

eng/pipelines/steps/APIScan.yml
WardenGnaw: This is to ignore dependent binaries since we should only be scanning what was created.

eng/pipelines/MIDebugEngine-CI.yml
WardenGnaw: I thought about this last night. I'll switch it to 1st and 15th to be even across the year
Although the max length of time we have is 30 days.

isadorasophia (approved):

eng/pipelines/MIDebugEngine-CI.yml
WardenGnaw: and I did change that in the cron, just not the description :D

eng/pipelines/steps/APIScan.yml
gregg-miskelly: Ah. Should we just ignore all interop dlls?

WardenGnaw: All except for `Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier.DesignTime.dll` 
15 is the only one dropped for now but I can update the expression to include UnixPortSupplier and no other Microsoft.VisualStudio.Debugger.Interop* like we do for the ones under `vscode` folder.

https://github.com/microsoft/MIEngine/tree/main/src/Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier

gregg-miskelly: If it is easy to do, I think that would be a good idea since we will probably add additional interop dlls in the future and forget to update this list.

gregg-miskelly (approved):

gregg-miskelly (approved):

PR #1398, opened 2023-05-03
Added parsing support for the QNX platform Coredump's thread ids format and fixed VSCODE .devcontainer compilation failure

khvysofq:
This PR fixes two issues.

1. It fixes the compilation error issue in vscode/.devcontainer. The .Net version in the Docker image in the current version is different from the .Net version used in the actual main branch, and it is not possible to generate the latest available compilation image correctly through the Dockerfile. The fix uses a fixed .Net official image, the Linux platform version, and does not require the installation of gcc-related tools, so it can be compiled directly. https://github.com/microsoft/vscode-dev-containers/tree/main/containers/dotnet
2. It adds support for the parsing format of Thread ID in QNX Coredump. The format of threads in Coredump in the QNX platform is "pid \<pid\> tid \<pid\>", e.g., "pid 4194373 tid 308", and this format is currently not supported.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
@@ -44,7 +44,7 @@ namespace Microsoft.MIDebugEngine
         private List<DebuggedThread> _deadThreads;
         private List<DebuggedThread> _newThreads;
         private Dictionary<string, List<int>> _threadGroups;
-        private static uint s_targetId = int.MaxValue;
+        private static uint s_targetId = 0; // Thread ids should be start at 0 in the default scenario
         private const string c_defaultGroupId = "i1";  // gdb's default group id, also used for any process without group ids
 
         private List<DebuggedThread> DeadThreads


src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -44,7 +44,7 @@ internal class ThreadCache
         private List<DebuggedThread> _deadThreads;
         private List<DebuggedThread> _newThreads;
         private Dictionary<string, List<int>> _threadGroups;
-        private static uint s_targetId = uint.MaxValue;
+        private static uint s_targetId = 0; // Thread ids should be start at 0 in the default scenario
```
gregg-miskelly, line 47: Can you talk about why you changed this part? Unfortunately there are no comments to say for sure, but I **think** the idea of starting with `uint.MaxValue` was to select something that was highly unlikely to clash with actual thread ids in the case that the thread id couldn't be parsed.

This code was added in commit [12632f3280da94787c009c8a97accbe797ac6994](https://github.com/microsoft/MIEngine/commit/12632f3280da94787c009c8a97accbe797ac6994), but the author didn't add any more details, and I am sure the author no longer remembers.

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -368,9 +368,19 @@ private bool TryGetTidFromTargetId(string targetId, out uint tid)
                 // In gdb coredumps the thread name is in the form:" LWP <thread-id>"
                 return true;
             }
+            else if (targetId.StartsWith("pid ", StringComparison.OrdinalIgnoreCase))
+            {
+                // In QNX gdb coredumps the thread name is in the form:"pid <pid> tid <pid>", eg "pid 4194373 tid 308"
+                int tid_pos = targetId.IndexOf("tid ", StringComparison.Ordinal);
+                int len = targetId.Length - (tid_pos + 4);
```
gregg-miskelly, line 375: Need to put the rest of this block in `if (tid_pos >= 0)`

src/MIDebugEngine/Engine.Impl/DebuggedThread.cs
```diff
@@ -368,9 +368,19 @@ private bool TryGetTidFromTargetId(string targetId, out uint tid)
...
+                if (len > 0 && System.UInt32.TryParse(targetId.Substring(tid_pos + 4, len), out tid) && tid != 0)
```
gregg-miskelly, line 376: nit: Use a constant for `tid ` and use '<const_name>.Length` instead of '4' so it is obvious what the '4' means.

Example:
```
const string tidPrefix = "tid ";
int tid_pos = targetId.IndexOf(tidPrefix, ...)
int len = targetId.Lenth - (tid_pos + tidPrefix.Length);
if (... && UInt32.TryParse(targetId.Substring(tid_pos + tidPrefix.Length...)
```

gregg-miskelly (changes requested):

.devcontainer/devcontainer.json
```diff
@@ -21,10 +21,12 @@
...
 		"jmrog.vscode-nuget-package-manager"
 	],
-	"postCreateCommand": "",
-  	"build": {
-    		"dockerfile": "Dockerfile"
-  	}
+	"image":"mcr.microsoft.com/vscode/devcontainers/dotnet:6.0-focal",
+	"remoteUser": "root"
+	// "postCreateCommand": "",
+  // 	"build": {
```
gregg-miskelly, line 27: Please just delete the commented out code

Issue #1382, opened 2023-02-01
vscode debugger pauses on SIGINT after adding breakpoint (remote gdb)


FabioGNR:
When you add a breakpoint in vscode using ms-vscode.cpptools v1.13.9 the debugger pauses on SIGINT "Interrupted".
I believe this was introduced with #1366.
src/MICore/Debugger.cs contains `var res = CmdAsync("-exec-interrupt", ResultClass.done);` which to me would appear to always generate SIGINT, but I could be wrong here.

GDB version 10.2, target is ARM.
Doing remote debugging on target from a Ubuntu 22 system.

I will try a smaller repro scenario which I can share.

FabioGNR:
Small gist to repro: https://gist.github.com/FabioGNR/42a2f1c18e4c5820022f5f7242e978d7
It only happens with remote debug (i.e. gdbserver)

aallrd:
Reproduce the issue with GDB 13.1 on RHEL 8.

PR #1399, opened 2023-05-05
Fix Remote AsyncBreak

WardenGnaw:
https://github.com/microsoft/MIEngine/commit/5f6213cf60fdd271cd7bf6b03b4225a914a10da5 accidently broke async-break for gdb remote scenarios as we send "-exec-interrupt" for async-break instead of sending a 'kill SIGTRAP'.

This PR modifies the `IsAsyncBreakSignal` to return a newly created enum `AsyncBreakSignal`, and renames it to `GetAsyncBreakSignal`. `GetAsyncBreakSignal` will now return if it saw a `SIGINT`, `SIGTRAP`, or `None` if its an unknown signal related to async break or no signal was retrieved at all.

This will be used along with a new flag called `IsUsingExecInterrupt`, which will be set before `-exec-interrupt` is sent to the debugger.

`IsUsingExecInterrupt` will be reset when the engine resolves the `async-break` from the user (see DebuggedProcess.cs changes) or when we are resolving an internal async break (See Debugger.cs DoInternalBreakActions).

Resolves: #1382

### DIFF
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -675,7 +675,6 @@ namespace MICore
                 {
                     return MICore.AsyncBreakSignal.SIGINT;
                 }
-                return MICore.AsyncBreakSignal.Unknown;
             }
 
             return MICore.AsyncBreakSignal.None;
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -660,7 +660,7 @@ namespace MICore
             {
                 if (results.TryFindString("signal-name") == "SIGTRAP" || 
                     // If we are remote, we use -exec-interrupt to async-break. GDB will send a SIGINT and we will need to handle the async break
-                    (_debugger.IsRemoteGdbTarget() && results.TryFindString("signal-name") == "SIGINT"))
+                    (_debugger.IsRemoteGdbTarget() && (_debugger.IsRequestingInternalAsyncBreak || _debugger.IsRequestingRealAsyncBreak) && results.TryFindString("signal-name") == "SIGINT"))
                 {
                     isAsyncBreak = true;
                 }
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -667,11 +667,12 @@ namespace MICore
         {
             if (results.TryFindString("reason") == "signal-received")
             {
-                if (results.TryFindString("signal-name") == "SIGTRAP")
+                string signalName = results.TryFindString("signal-name");
+                if (signalName == "SIGTRAP")
                 {
                     return MICore.AsyncBreakSignal.SIGTRAP;
                 }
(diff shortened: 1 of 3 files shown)


src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -658,7 +658,9 @@ public virtual bool IsAsyncBreakSignal(Results results)
 
             if (results.TryFindString("reason") == "signal-received")
             {
-                if (results.TryFindString("signal-name") == "SIGTRAP")
+                if (results.TryFindString("signal-name") == "SIGTRAP" || 
+                    // If we are remote, we use -exec-interrupt to async-break. GDB will send a SIGINT and we will need to handle the async break
+                    (_debugger.IsRemoteGdbTarget() && (_debugger.IsRequestingInternalAsyncBreak || _debugger.IsRequestingRealAsyncBreak) && results.TryFindString("signal-name") == "SIGINT"))
```
gregg-miskelly, line 663: >_debugger.IsRemoteGdbTarget() && (_debugger.IsRequestingInternalAsyncBreak || _debugger.IsRequestingRealAsyncBreak) && [](http://example.com/codeflow?start=21&length=118)

Should we instead set before we do `CmdAsync("-exec-interrupt", ResultClass.done)` since the logic for when we use `-exec-interrupt` is pretty complicated (ex: looks like it is used for LLDB, and only only sometimes for remote GDB).

WardenGnaw: Would we clear that bool flag here?

gregg-miskelly: I think it probably makes more sense to do in the caller. Maybe be returning an enum instead of a bool?

gregg-miskelly: Yeah, I think we would need to since it looks like `IsAsyncBreakSignal` could be called twice for the same stop.

WardenGnaw: I created the `AsyncBreakSignal` enum to handle this.

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -652,19 +663,21 @@ public virtual bool SupportsFrameFormatting
...
         }
 
-        public virtual bool IsAsyncBreakSignal(Results results)
+        public virtual AsyncBreakSignal GetAsyncBreakSignal(Results results)
         {
-            bool isAsyncBreak = false;
-
             if (results.TryFindString("reason") == "signal-received")
             {
                 if (results.TryFindString("signal-name") == "SIGTRAP")
```
gregg-miskelly, line 670: >results.TryFindString("signal-name") [](http://example.com/codeflow?start=20&length=36)

nit: save in a local so this doesn't need to be called twice?

gregg-miskelly (approved):
Otherwise LGTM

paulmaybee (approved):

mpeyrotc (approved):

gregg-miskelly (approved):

PR #1400, opened 2023-05-09
[#1349] Always use `--simple-values` in newer versions of GDB.

### LINKED ISSUE #1349: Unnecessary `-var-create` and `-var-delete` commands during a stack trace can cause noticeable pause each time the debuggee stops

gareth-rees:
## Summary

In newer versions of GDB, the `--simple-values` option to the `-stack-list-arguments`, `-stack-list-locals` and
`-stack-list-variables` commands no longer prints the value for references to compound types. This improves the performance of these commands when the stack has references to large data structures.

When these versions of GDB are available, take advantage by using `--simple-values` in `DebuggedProcess.GetParameterInfoOnly` to fetch names and types in a single `-stack-list-arguments` command. This is faster than the old approach of using `--no-values` followed with `-var-create` and `-var-delete` for each parameter to get the type.

The new method `MICommandFactory.SupportsSimpleValuesExcludesRefTypes` determines if the debugger supports the improved behaviour of the `--simple-values` option, by executing the `-list-features` command and checking for the `"simple-values-ref-types"` feature in the output. We cache the result on the `DebuggedProcess` object as the set of supported features does not change during the lifetime of the debug session.

Fixes #1349 (Unnecessary `-var-create` and `-var-delete` commands during a stack trace can cause noticeable pause each time the debuggee stops)

### Details

The behaviour of GDB was changed in [commit 51f8dafba8](https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;h=51f8dafba8175a8d59716f220a46de4e626e5073) and this change will be included in the next release of GDB. I understand that Microsoft developers are not allowed to look at GDB's source code (see [comment from Gregg Miskelly here](https://github.com/microsoft/MIEngine/issues/1349#issuecomment-1239671665)), so I have summarized the effect of the change above. Let me know if there is anything else you'd like to know about this change.

I repeated the reproduction procedure from #1349, took a copy of the commands issued by a single Step Over operation with ten frames on the stack and attached them here: [stack-list-arguments-after.gz](https://github.com/microsoft/MIEngine/files/11429955/stack-list-arguments-after.gz). The table below shows the counts of GDB/MI commands issued for the Step Over operation by the MIEngine from `main` against GDB 13.1 and the MIEngine from this branch against the updated GDB.

| Command | Counts (main) | Counts (this branch) |
| ------- | ----- | ----- |
| `-stack-list-arguments` | 2 | 2 |
| `-stack-list-frames` | 1 | 1 |
| `-stack-list-variables` | 1 | 1 |
| `-stack-select-frame` | 20 | 0 |
| `-var-create` | 210 | 10 |
| `-var-delete` | 210 | 10 |

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -52,7 +52,7 @@ namespace Microsoft.MIDebugEngine
         private IProcessSequence _childProcessHandler;
         private bool _deleteEntryPointBreakpoint;
         private string _entryPointBreakpoint = string.Empty;
-        private bool _simpleValuesExcludesRefTypes = false;
+        private bool? _simpleValuesExcludesRefTypes = null;
 
         public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngineCallback callback, WorkerThread worker, BreakpointManager bpman, AD7Engine engine, HostConfigurationStore configStore, HostWaitLoop waitLoop = null) : base(launchOptions, engine.Logger)
         {
@@ -551,8 +551,6 @@ namespace Microsoft.MIDebugEngine
 
             try
             {
-                _simpleValuesExcludesRefTypes = await this.MICommandFactory.SupportsSimpleValuesExcludesRefTypes();
-
                 await this.MICommandFactory.EnableTargetAsyncOption();
 
                 await this.CheckCygwin(_launchOptions as LocalLaunchOptions);
@@ -1988,13 +1986,25 @@ namespace Microsoft.MIDebugEngine
         //NOTE: eval is not called
         public async Task<List<ArgumentList>> GetParameterInfoOnly(AD7Thread thread, bool values, bool types, uint low, uint high)
         {
-            // If SimpleValues excludes printing values for references to compound types,
-            // use SimpleValues (even if values are not requested) as it gets types too.
-            // Otherwise, if values are requested, use SimpleValues, but if not, the
-            // potential performance penalty of fetching values for references to compound
-            // types is too high, so use NoValues and follow up with -var-create to get
-            // the types.
-            PrintValue printValue = (_simpleValuesExcludesRefTypes || values) ? PrintValue.SimpleValues : PrintValue.NoValues;
+            PrintValue printValue = values ? PrintValue.SimpleValues : PrintValue.NoValues;
+            if (types && !values)
+            {
+                // We want types but not values. There is no PrintValue option for this, but if
+                // SimpleValues excludes printing values for references to compound types, then
+                // the fastest approach is to use SimpleValues (and ignore the values).
+                // Otherwise, the potential performance penalty of fetching values for
+                // references to compound types is too high, so use NoValues and follow up with
+                // -var-create to get the types.
+                if (!_simpleValuesExcludesRefTypes.HasValue)
+                {
+                    _simpleValuesExcludesRefTypes  = await this.MICommandFactory.SupportsSimpleValuesExcludesRefTypes();
+                }
+                if (_simpleValuesExcludesRefTypes.Value)
+                {
+                    printValue = PrintValue.SimpleValues;
+                }
+            }
+
             var frames = await MICommandFactory.StackListArguments(printValue, thread.Id, low, high);
             List<ArgumentList> parameters = new List<ArgumentList>();
 
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -52,6 +52,7 @@ namespace Microsoft.MIDebugEngine
         private IProcessSequence _childProcessHandler;
         private bool _deleteEntryPointBreakpoint;
         private string _entryPointBreakpoint = string.Empty;
+        private bool _simpleValuesExcludesRefTypes = false;
 
         public DebuggedProcess(bool bLaunched, LaunchOptions launchOptions, ISampleEngineCallback callback, WorkerThread worker, BreakpointManager bpman, AD7Engine engine, HostConfigurationStore configStore, HostWaitLoop waitLoop = null) : base(launchOptions, engine.Logger)
         {
@@ -550,6 +551,8 @@ namespace Microsoft.MIDebugEngine
 
             try
             {
+                _simpleValuesExcludesRefTypes = await this.MICommandFactory.SupportsSimpleValuesExcludesRefTypes();
+
                 await this.MICommandFactory.EnableTargetAsyncOption();
 
                 await this.CheckCygwin(_launchOptions as LocalLaunchOptions);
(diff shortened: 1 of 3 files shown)


gareth-rees:
@gregg-miskelly You looked at my previous attempt to fix this issue, pull request #1350, so it would valuable if you could take a look at this attempt too.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -52,6 +52,7 @@ internal class DebuggedProcess : MICore.Debugger
         private IProcessSequence _childProcessHandler;
         private bool _deleteEntryPointBreakpoint;
         private string _entryPointBreakpoint = string.Empty;
+        private bool _simpleValuesExcludesRefTypes = false;
```
gregg-miskelly, line 55: I would suggest making this a `bool?` and initializing it on first use rather than initializing it immediately on start debugging -- I am not sure if GDB will always be able to give you an accurate answer immediately on startup even before connecting.

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -1981,8 +1984,14 @@ public async Task<List<SimpleVariableInformation>> GetParameterInfoOnly(AD7Threa
...
         {
-            // If values are requested, request simple values, otherwise we'll use -var-create to get the type of argument it is.
-            var frames = await MICommandFactory.StackListArguments(values ? PrintValue.SimpleValues : PrintValue.NoValues, thread.Id, low, high);
+            // If SimpleValues excludes printing values for references to compound types,
+            // use SimpleValues (even if values are not requested) as it gets types too.
+            // Otherwise, if values are requested, use SimpleValues, but if not, the
+            // potential performance penalty of fetching values for references to compound
+            // types is too high, so use NoValues and follow up with -var-create to get
+            // the types.
+            PrintValue printValue = (_simpleValuesExcludesRefTypes || values) ? PrintValue.SimpleValues : PrintValue.NoValues;
```
gregg-miskelly, line 1993: We only need `SimpleValues` if the caller is requested `values` or `types` so I think you want something like:

```
PrintValue printValue = values ? PrintValue.SimpleValues : PrintValue.NoValues;
if (!values && types)
{
    if (!_simpleValuesExcludesRefTypes.HasValue)
    {
        _simpleValuesExcludesRefTypes  = await this.MICommandFactory.SupportsSimpleValuesExcludesRefTypes();
    }
    if (_simpleValuesExcludesRefTypes.Value)
    {
        printValue = PrintValue.SimpleValues;
    }
}

gregg-miskelly (changes requested):
Otherwise sounds good to me

src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
gareth-rees: I pushed a fixup commit with your suggested changes.

gregg-miskelly (approved):

gregg-miskelly:
@WardenGnaw do you want to take a look as well? These look good to me.

WardenGnaw:
LGTM. Linux tests are just being flaky and not failing due to the changes.

Thanks again for contributing @gareth-rees!

PR #1401, opened 2023-05-16
MIEngine: Introduce --thread and --frame options

intel-rganesh:
Remove -thread-select and -stack-select-frame and introduce --thread and --frame options instead.

It is general recommendation to not use -thread-select and -stack-select-frame and use --thread and --frame instead. More info can be found here:
https://sourceware.org/gdb/onlinedocs/gdb/Context-management.html

And here are the deprecation notes:
-thread-select
-thread-select: https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI-Thread-Commands.html#GDB_002fMI-Thread-Commands

-stack-select-frame
-stack-select-frame: https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI-Stack-Manipulation.html#GDB_002fMI-Stack-Manipulation

### DIFF
--- a/src/MICore/CommandFactories/MICommandFactory.cs
+++ b/src/MICore/CommandFactories/MICommandFactory.cs
@@ -54,6 +54,8 @@ namespace MICore
 
         public abstract string Name { get; }
 
+        internal int MajorVersion { get; set; }
+
         public static MICommandFactory GetInstance(MIMode mode, Debugger debugger)
         {
             MICommandFactory commandFactory;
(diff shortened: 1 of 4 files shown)


intel-rganesh:
@WardenGnaw Hi Andrew, since -thread-select and -stack-select-frame commands are deprecated in favor of --thread and --frame options respectively, we (at Intel) modified the fork of MI Engine we use for our gdb plugin, which we use with Intel oneAPI. We also wanted to contribute these changes to the upstream MI Engine repository. Could you review the PR and let me know if these changes look good? Please let me know if I am required to report this as an issue and track this in the PR additionally.

WardenGnaw:
@intel-rganesh Thanks for contributing this PR!

Do you know if there was a specific version of gdb that `--thread` and `--frame` was included? Would there be issues of a very old version of GDB not being able to support those flags?

intel-rganesh:
@WardenGnaw Source code of gdb 7.0.0 (Release date: Oct 06, 2009) includes this, and 6.8 does not. Please find the commit details here:
https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=1e92afda99a58f4427293937e49ef1ebe6bb3968

If we intend to support really old versions, I can try to fetch the version value from the log and use deprecated thread-select for older versions.

WardenGnaw:
@intel-rganesh I'm not able to see source code of GDB based on its licensing agreements.

However, version checks are useful since we do have some embedded engineers who may be using a really old version.
We have a version parser already at https://github.com/microsoft/MIEngine/blob/f530168188f23c4cf2c5799ce309ade66f854b4e/src/MICore/Debugger.cs#LL1033C1-L1048C10

src/MICore/CommandFactories/MICommandFactory.cs
```diff
@@ -54,6 +54,8 @@ public abstract class MICommandFactory
 
         public abstract string Name { get; }
 
+        internal string Version { get; set; }
```
WardenGnaw, line 57: Can this be an int so we dont have to constantly run `int.TryParse(Version.Split('.').FirstOrDefault(), out int gdbMainVersion);`?

If its `null`, we could use majorVersion as -1 for invalid.

```suggestion
        internal int majorVersion { get; set; }
```

WardenGnaw (changes requested):
Sorry for the delay. Just one more comment.

src/MICore/CommandFactories/MICommandFactory.cs
intel-rganesh: I have changed the property to int.
Since the return of the method GdbVersionFromLog is a string, I had to parse it one time though.
Please let me know if this is alright!

WardenGnaw (approved):

WardenGnaw:
@intel-rganesh Thank you for your contribution! We will try to get these changes in the next release as soon as we can.

PR #1402, opened 2023-05-21
Align use of sendInvalidate request arguments

iAbadia:
Using the `sendInvalidate` request I realised that I was forced to provide a single `Area` as an argument. This PR aligns the behaviour of `sendInvalidate` request with the `invalidate` event.

Thanks to @sbobko for his original contribution, I only hope this patch doesn't contradict the original intent of the behaviour of `sendInvalidate`

### DIFF
--- a/src/OpenDebugAD7/AD7DebugSession.cs
+++ b/src/OpenDebugAD7/AD7DebugSession.cs
@@ -137,22 +137,25 @@ namespace OpenDebugAD7
         private void HandleSendInvalidateRequestAsync(IRequestResponder<SendInvalidateArguments> responder)
         {
             InvalidatedEvent invalidated = new InvalidatedEvent();
-            
-            // Setting the area and adding it to the result
-            invalidated.Areas.Add(responder.Arguments.Areas);
+            // Set the Arguments only if passed
+            if (null != responder.Arguments) {
+                // Setting the Areas if passed
+                if (null != responder.Arguments.Areas || responder.Arguments.Areas.Length != 0) {
+                    invalidated.Areas = responder.Arguments.Areas.ToList();
+                }
 
-            // Setting the StackFrameId if passed (and the 'threadId' is ignored).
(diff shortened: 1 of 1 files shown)


src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -137,23 +137,26 @@ public AD7DebugSession(Stream debugAdapterStdIn, Stream debugAdapterStdOut, List
         private void HandleSendInvalidateRequestAsync(IRequestResponder<SendInvalidateArguments> responder)
         {
             InvalidatedEvent invalidated = new InvalidatedEvent();
-            
-            // Setting the area and adding it to the result
-            invalidated.Areas.Add(responder.Arguments.Areas);
+            // Set the Arguments only if passed
+            if (null != responder.Arguments) {
+                // Setting the Areas if passed
+                if (null != responder.Arguments.Areas || responder.Arguments.Areas.Length != 0) {
```
WardenGnaw, line 143:
```suggestion
                if (null != responder.Arguments.Areas) {
```

It is ok to send an empty array with invalidateEvent.

>  If this property is missing, empty, or if values are not understood, the client should assume a single value `all`.",

src/OpenDebugAD7/AD7DebugSession.cs
```diff
@@ -3913,7 +3916,7 @@ public SendInvalidateRequest(): base("sendInvalidate")
     internal class SendInvalidateArguments : DebugRequestArguments
     {
 
-        public InvalidatedAreas Areas { get; set; }
+        public InvalidatedAreas[]? Areas { get; set; }
```
WardenGnaw, line 3919:
```suggestion
        public List<InvalidatedAreas> Areas { get; set; }
```

Working with a collection is easier than a raw array.

WardenGnaw (approved):

WardenGnaw:
@iAbadia Thank you for contributing this PR.

Issue #1403, opened 2023-06-01
Wrong numeration after first 50 items with multidimensional arrays in natvis (Linux)


NValerij:
Hi.
I’m using natvis to view my rather complex custom data type in VS Code (1.78.2, C/C++ v1.15.4, Ubuntu 20.04).
There is a fragment of matrix visualizer (unfortunately, `CustomListItems` does not work in Linux, #929):
```xml
            <ArrayItems Condition="Memory != 0">
                <Direction>Forward</Direction>
                <Rank>2</Rank>
                <Size>$i == 0 ? Rows : Cols</Size>
                <ValuePointer>(($T1*)(Memory->Segments[Segment].Data + Offset))</ValuePointer>
            </ArrayItems>
```
Imagine, I have matrix with `Rows = 1` and `Cols = 256`. In Variables window it looks like this:
![image](https://github.com/microsoft/MIEngine/assets/11822289/a9d666eb-59f4-4e1a-9ed6-85bec2ee015e)

I expect that after `[0, 49]` I’ll see `[0, 50]` and so on until `[0, 255]`, but couters are reset after `[0, 49]`. Values to the right are correct (I’ve checked them on another example).

PR #1406, opened 2023-07-07
Fix Rank > 1 Display Strings for Natvis

WardenGnaw:
This PR fixes the display string for ArrayItems with Rank > 1 when the "[More...]" is expanded.
Before "[More...]" will expand starting at [0,0] instead of [0,50] for the first expansion since the offsets are at 50.

This change updated the call to GetDisplayNameFromArrayIndex from just `index` which would always start at 0, to `startIndex + index` to include the offet.
Added a test to cover this case and refactored the Matrix Natvis syntax to use rol/col instead of random behavior. 

Fixes: #1403

### DIFF
--- a/test/CppTests/debuggees/natvis/src/SimpleMatrix.h
+++ b/test/CppTests/debuggees/natvis/src/SimpleMatrix.h
@@ -1,27 +1,20 @@
 class SimpleMatrix
 {
 public:
-    int m_size1;
-    int m_size2;
-    bool m_fUseSize1;
+    int m_rows;
+    int m_cols;
     int* m_pData;
 
-    SimpleMatrix(int size1, int size2, bool fUseSize1)
+    SimpleMatrix(int row, int col, bool fUseSize1)
     {
-        m_size1 = size1;
-        m_size2 = size2;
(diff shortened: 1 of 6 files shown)


gregg-miskelly (approved):

test/CppTests/debuggees/natvis/src/SimpleMatrix.h
```diff
@@ -1,27 +1,20 @@
...
 public:
-    int m_size1;
-    int m_size2;
-    bool m_fUseSize1;
+    int m_rows;
+    int m_cols;
     int* m_pData;
 
-    SimpleMatrix(int size1, int size2, bool fUseSize1)
+    SimpleMatrix(int row, int col, bool fUseSize1)
```
mpeyrotc, line 8: Remove

mpeyrotc, line 8:
```suggestion
    SimpleMatrix(int row, int col)
```

mpeyrotc (approved):

gregg-miskelly (approved):

WardenGnaw:
Flaky test failing.

PR #1412, opened 2023-08-05
Only use ps 'flags' for macOS

WardenGnaw:
This PR adds in an operating system check via uname to determine if we should run the PS command with the 'flags' flag. This feature was originally used to determine the architecture of a process for macOS M1 machines that can run as x64 or as arm64.

AD7Process only uses the flags field in macOS so this PR updates the command sent to only use it on macOS machines.

Other changes:
- Updated the setup.csx script to update VS with VS.list instead of VS.Codespaces.list
- Updated tests to use the new PSOutputParser
- Added a macOS test

Testing:
- [X] macOS SSH
- [X] WSL Ubuntu

### DIFF
--- a/src/SSHDebugPS/IConnection.cs
+++ b/src/SSHDebugPS/IConnection.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Threading;
 using Microsoft.DebugEngineHost;
 using Microsoft.SSHDebugPS.Utilities;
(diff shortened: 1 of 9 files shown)


tools/Setup.csx
```diff
@@ -196,15 +196,21 @@ class Setup {
...
+                        {
+                            TargetPath = extension;
+                            break;
+                        }
                     }
                 }
+                else 
+                {
+                    throw new InvalidOperationException("Unable to find an installation of VS Code C++ Extension.");
+                }
```
gregg-miskelly, line 213: Isn't this telling you if you failed to find any extension and you need to move line 203 into a `.Where` clause on line 198?

src/SSHDebugPS/IConnection.cs
```diff
@@ -121,15 +123,37 @@ public Process(uint id, string arch, uint flags, string userName, string command
         }
     }
 
+    internal static class OperatingSystemStringConverter
+    {
+        internal static PlatformID ConvertToPlatformID(this string value)
```
gregg-miskelly, line 128: >this [](http://example.com/codeflow?start=55&length=4)

Nit: why make it an extension method on `string` instead of just a normal static method?

gregg-miskelly (approved):
Otherwise LGTM

CarlosGuerraVz (approved):

src/SSHDebugPS/IConnection.cs
WardenGnaw: I was being too fancy, but after thinking about it since its not going to be a common API for string, its just make it a non-extension method.

gregg-miskelly (approved):

PR #1415, opened 2023-08-15
Fixing File writing for EngineLogger

WardenGnaw:
With the Natvis Diagnostics Logger work, the file logger would assume fully qualified paths and would try to create it in the root drive, this would fail since it would not have access. The original change also did not close the ChannelLoggers so this PR fixes that.

This PR also runs `File.CreateText` in MIDebugCommandDispatcher so errors can be caught earlier and not set an invalid file path.

### DIFF
--- a/src/DebugEngineHost/HostLogger.cs
+++ b/src/DebugEngineHost/HostLogger.cs
@@ -54,7 +54,9 @@ namespace Microsoft.DebugEngineHost
 
         public static void Reset()
         {
+            s_natvisLogChannel?.Close();
             s_natvisLogChannel = null;
+            s_engineLogChannel?.Close();
             s_engineLogChannel = null;
         }
     }
--- a/src/MIDebugEngine/MIDebugCommandDispatcher.cs
+++ b/src/MIDebugEngine/MIDebugCommandDispatcher.cs
@@ -4,6 +4,7 @@
 using MICore;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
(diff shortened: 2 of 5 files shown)


paulmaybee (approved):

gregg-miskelly (approved):
I am not sure if this is worth the effort since I don't know how often this is used, but ideally we would probably stop holding the log file path as a string, and instead hold it as a `StreamWriter` since otherwise we could still hit failures when we try to re-open the stream.

Otherwise, LGTM

WardenGnaw:
> I am not sure if this is worth the effort since I don't know how often this is used, but ideally we would probably stop holding the log file path as a string, and instead hold it as a StreamWriter since otherwise we could still hit failures when we try to re-open the stream.

I'll take a stab at moving to StreamWriter. I also noticed there will be a bug if you try to run the command while debugging. Do we want to throw a warning if you try to change the output file mid-debug session?

gregg-miskelly:
> I'll take a stab at moving to StreamWriter. I also noticed there will be a bug if you try to run the command while debugging. Do we want to throw a warning if you try to change the output file mid-debug session?

Makes sense, or fix it if that is just as easy.

WardenGnaw:
> > I'll take a stab at moving to StreamWriter. I also noticed there will be a bug if you try to run the command while debugging. Do we want to throw a warning if you try to change the output file mid-debug session?
> 
> Makes sense, or fix it if that is just as easy.

Making it a Stream was a bit difficult, I just added a check to see if a debug session is running and we will throw an error.

gregg-miskelly (approved):

PR #1430, opened 2023-12-19
Theme Docker Container Selector Dialog

WardenGnaw:
This PR updates the Docker Container Selector Dialog in Visual Studio to be themed. The following changes were done:
- Updates VS packages since some of the newer theming was in newer packages.
- Newer packages had new attributes that needed to be handled with MakePIAPortableTool.
- Converted ListBox to ListView for ContainerPickerDialog and modified associated automation peers

Old:
image

New (Dark Theme):
image

New (Dark Theme) with Error:
image

New (Blue Contrast):
image

### DIFF
--- a/src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
+++ b/src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
@@ -447,7 +447,7 @@
     </platformui:DialogWindow.Resources>
 
     <!-- Main layout grid -->
-    <Grid Margin="12,0,12,12" Focusable="False">
+    <Grid Focusable="False">
         <Grid.RowDefinitions>
             <RowDefinition Height="32" />
             <RowDefinition Height="*" />
@@ -462,7 +462,7 @@
 
             <TextBlock Grid.Column="0"
                     FontSize="14"
-                    Margin="0,4,0,0"
+                    Margin="12,2,0,0"
                     VerticalAlignment="Center"
                     Text="{x:Static local:UIResources.DialogTitle}">
             </TextBlock>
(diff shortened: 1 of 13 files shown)


src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
```diff
@@ -296,6 +296,21 @@ private bool IsLibLinuxAvailable()
             }
         }
 
+        // Default is to Support SSH Connections
```
WardenGnaw, line 299: Remove

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml
```diff
@@ -448,32 +446,77 @@
...
+                    Text="{x:Static local:UIResources.DialogTitle}">
+            </TextBlock>
+            <StackPanel Grid.Column="1"
+                    Orientation="Horizontal"
+                    HorizontalAlignment="Right"
+                    WindowChrome.IsHitTestVisibleInChrome="true">
+                <Button Style="{DynamicResource TitleBarButtonStyle}" Click="MaximizeOrRestoreButton_Click">
+                    <StackPanel>
+                        <!-- Maximize -->
+                        <Path Visibility="{Binding IsMaximized, Converter={StaticResource BoolToVisibilityConverter}, ConverterParameter=True, Mode=OneWay}"
```
asundheim, line 475: Suggestion: VisibleIfFalseConverter (defined with parameter), VisibleIfTrueConverter rather than BoolToVisibilityConverter with anonymous parameter

asundheim (approved):

CarlosGuerraVz:
I'm not sure this dialog needs a Maximize button, but if you do add it check how it looks when maximized, as other times I've used a WindowChrome for adding a tittle bar it causes the dialog to get slightly bigger than the screen when maximized, looking something like this:
image
A quick solution is to conditionally add a padding to the whole window whenever maximized.

CarlosGuerraVz (approved):

gregg-miskelly (approved):
:shipit:

PR #1431, opened 2023-12-21
Fix missing s32 format specifier support

nikitalita:
https://learn.microsoft.com/en-us/visualstudio/debugger/format-specifiers-in-cpp?view=vs-2022
the `s32` and `s32b` format specifiers for unicode 32-bit strings were not being parsed, I have added them.


chuckries (approved):

WardenGnaw (approved):

WardenGnaw:
@nikitalita Thank you for your contribution!

We will try to target this for the next VS Code C++ Extension and VS release.

PR #1437, opened 2024-01-24
MIEngine: New methods and framework update

intel-rganesh:
This PR covers the following commits:

- Add null condition check in DebuggedThread.GetThreadContext​
- Add a method to get the process state of the last process​ and another method that sets the selected thread
- Updating .net target framework to net472 for WindowsDebugLauncher​
- Correct VSIX file so that plugin integration is performed for ALL users​
- Option to execute MI Commands using CmdAsync alone without result-class

Signed-off-by: intel-rganesh [rakesh.ganesh@intel.com](mailto:rakesh.ganesh@intel.com)

### DIFF
--- a/src/MIDebugEngine/MIDebugCommandDispatcher.cs
+++ b/src/MIDebugEngine/MIDebugCommandDispatcher.cs
@@ -41,7 +41,7 @@ namespace Microsoft.MIDebugEngine
             return GetLastProcess().ProcessState;
         }
 
-        public static async Task<Results> SetSelectedThread(int threadId)
+        public static async Task<Results> ViewSelectedThreadInfo(int threadId)
         {
             GetLastProcess().MICommandFactory.DefineCurrentThread(threadId);
             return await Task.FromResult<Results>(new Results(ResultClass.done));
@@ -60,7 +60,7 @@ namespace Microsoft.MIDebugEngine
             }
             else
             {
-                return null;
+                throw new ArgumentOutOfRangeException(nameof(command));
             }
         }
 
--- a/src/MIDebugEngine/MIDebugCommandDispatcher.cs
+++ b/src/MIDebugEngine/MIDebugCommandDispatcher.cs
@@ -15,7 +15,7 @@ namespace Microsoft.MIDebugEngine
     {
         private readonly static List<DebuggedProcess> s_processes = new List<DebuggedProcess>();
 
-        public static async Task<Results> ExecuteMICommandWithResultsObject(string command)
+        private static DebuggedProcess GetLastProcess()
         {
             DebuggedProcess lastProcess;
             lock (s_processes)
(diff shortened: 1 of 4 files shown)


WardenGnaw (approved):

src/MIDebugEngine/MIDebugCommandDispatcher.cs
```diff
@@ -27,7 +27,46 @@ public static Task<string> ExecuteCommand(string command)
...
+
+            return lastProcess;
+        }
+
+        public static MICore.ProcessState GetProcessState()
+        {
+            return GetLastProcess().ProcessState;
+        }
+
+        public static async Task<Results> SetSelectedThread(int threadId)
```
gregg-miskelly, line 44: Can you tell us more about the goal of this method?

src/MIDebugEngine/MIDebugCommandDispatcher.cs
```diff
@@ -27,7 +27,46 @@ public static Task<string> ExecuteCommand(string command)
...
+        }
+
+        public static async Task<Results> ExecuteMICommandWithResultsObject(string command)
+        {
+            if (string.IsNullOrWhiteSpace(command))
+                throw new ArgumentNullException(nameof(command));
+
+            command = command.Trim();
+
+            if (command[0] == '-')
```
gregg-miskelly, line 57: Should this throw `ArgumentOutOfRange` instead?

src/MIDebugEngine/MIDebugCommandDispatcher.cs
intel-rganesh: When we want to look into the values of variables in Locals and Watch windows etc for another hardware thread (and eventually lanes, when this is merged upstream into gdb) we need to be able to switch to another thread. So, this  method shall enable us to do so.

https://www.intel.com/content/www/us/en/docs/distribution-for-gdb/tutorial-debugging-dpcpp-windows/2024-0/debug-an-application-on-a-gpu.html#:~:text=the%20SIMD%20lane.-,Switch,-to%20another%20active

src/MIDebugEngine/MIDebugCommandDispatcher.cs
intel-rganesh: This method enables a user to obtain the Results object without stringification that is done inside the method ExecuteMiCommand.
Please see:
https://github.com/microsoft/MIEngine/blob/782c97ded5747639a947e094f8e9d7d06c153cf4/src/MIDebugEngine/MIDebugCommandDispatcher.cs#L84-L87

https://github.com/microsoft/MIEngine/blob/782c97ded5747639a947e094f8e9d7d06c153cf4/src/MICore/MIResults.cs#L582

gregg-miskelly: Sorry, what I mean is, instead of:

```
   if (command[0] != '-') return null;
```

I think this should be:
```
if (command[0] != '-') throw new ArgumentOutOfRangeException(nameof(command));
```

src/MIDebugEngine/MIDebugCommandDispatcher.cs
gregg-miskelly: In your scenario, can you just use `--thread {threadId} --frame {frameLevel}` as part of the command string? Or do you need to use versions of GDB and/or commands that don't support that.

There are two problems with this approach:
1. Anything else can come along and change the thread
2. `DefineCurrentThread` isn't really meant to switch the thread, but rather it is meant to update the command factory to know what the current thread is after we enter stop state

intel-rganesh: Setting the thread Id (i.e., defining the selected thread in MICommnadFactory) in a stopped state calls -var-create, -stack-list-arguments, -stack-list-frames, -data-evaluate-expression etc with the intended --thread {threadId} --frame {frameLevel} command without having to manually call each one of them. It is intended that the normal process flow resets the thread the when the program is continued.
Maybe the method here needs to be renamed to ViewSelectedThreadInfo or DisplaySelectedThreadInfo.
Do you have a suggestion here?

gregg-miskelly: My problem isn't with the name of the method, but rather that it promises something that it will not deliver on, and make break the UI -- anything else in the UI may change the current thread at any moment. So you are temporarily changing the current thread, but there are no guarantees that the other queries you want to execute against a certain thread will work that way, and doing so without taking the `CommandLock` make break the UI. 

So I would get rid of `ViewSelectedThreadInfo`, and do one of:

1. `ExecuteMICommandWithResultsObject` should take a (optional?) thread id (and maybe frame id while you are at it?) and `MICommandFactory.ThreadCmdAsync` (or `ThreadFrameCmdAsync` if there is a `frameLevel` also) should be used if a thread/frame is provided.
-or-
2. If you only need to deal with versions of GDB that support `--thread {threadId} --frame {frameLevel}`, then you don't need to update MIEngine to solve this part of the problem.

intel-rganesh: Removing ViewSelectedThreadInfo as part of the discussion. ExecuteMICommandWithResultsObject takes an argument which can contain threadId, frameLevel (and laneId when needed, for the future).

gregg-miskelly (approved):

PR #1441, opened 2024-02-28
MIEngine: Set MajorVersion from the correct place

intel-rganesh:
We were previously setting the MajorGdbVersion when the process exited and this was too late.
We need to set this OnStateChanged so we can use this to opt for the latest gdb MI commands.
Additionally include Intel GNU gdb example when fetching the GDB version.

### DIFF
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -161,6 +161,8 @@ namespace MICore
         // The key is the thread group, the value is the pid
         private Dictionary<string, int> _debuggeePids;
 
+        private string _gdbVersion = string.Empty;
+
         public Debugger(LaunchOptions launchOptions, Logger logger)
         {
             _launchOptions = launchOptions;
@@ -316,8 +318,29 @@ namespace MICore
             this.OnStateChanged(mode, _miResults.ParseResultList(strresult));
         }
 
+        private void SetGdbVersion()
+        {
+            if (string.IsNullOrWhiteSpace(_gdbVersion))
+            {
+                _gdbVersion = GdbVersionFromLog();
+            }
+        }
+
+        private void SetMajorGdbVersion()
+        {
+            int majorVersion = -1;
+            if (!string.IsNullOrWhiteSpace(_gdbVersion))
+            {
+                int.TryParse(_gdbVersion.Split('.').FirstOrDefault(), out majorVersion);
+            }
+            MICommandFactory.MajorVersion = majorVersion;
+        }
+
         protected void OnStateChanged(string mode, Results results)
         {
+            SetGdbVersion();
+            SetMajorGdbVersion();
+
             if (mode == "stopped")
             {
                 OnStopped(results);
@@ -492,8 +515,6 @@ namespace MICore
             {
                 _consoleDebuggerInitializeCompletionSource = null;
 
-                // We no longer care about keeping these, so empty them out
-                _initializationLog = null;
                 _initialErrors = null;
             }
         }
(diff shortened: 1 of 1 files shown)


src/MICore/Debugger.cs
```diff
@@ -316,8 +318,29 @@ protected virtual void OnStateChanged(string mode, string strresult)
...
+                _gdbVersion = GdbVersionFromLog();
+            }
+        }
+
+        private void SetMajorGdbVersion()
+        {
+            int majorVersion = -1;
+            if (!string.IsNullOrWhiteSpace(_gdbVersion))
+            {
+                int.TryParse(_gdbVersion.Split('.').FirstOrDefault(), out majorVersion);
```
gregg-miskelly, line 334: >int [](http://example.com/codeflow?start=16&length=3)

Handle failures here?

src/MICore/Debugger.cs
```diff
@@ -978,24 +999,16 @@ public void OnDebuggerProcessExit(/*OPTIONAL*/ string exitCode)
...
-                            }
-                            MICommandFactory.MajorVersion = majorVersion;
-
+                            
                             // We can't use IsMinGW or IsCygwin because we never connected to the debugger
                             bool isMinGWOrCygwin = _launchOptions is LocalLaunchOptions &&
                                     PlatformUtilities.IsWindows() &&
                                     this.MICommandFactory.Mode == MIMode.Gdb;
-                            if (isMinGWOrCygwin && version != null && IsUnsupportedWindowsGdbVersion(version))
+                            if (isMinGWOrCygwin && _gdbVersion != null && IsUnsupportedWindowsGdbVersion(_gdbVersion))
```
gregg-miskelly, line 1007: >_gdbVersion != null [](http://example.com/codeflow?start=51&length=19)

Your code initializes `_gdbVersion` to the empty string, so this will never be true.

gregg-miskelly:
>                         _initializationLog.AddLast(line);

Rather than scraping the initialization log after the fact, I think we should just process the line here. Something like:

```
if (string.IsNullOrEmpty(_gdbVersion))
{
    TryInitializeGdbVersion(line);
}
```

---
Refers to: src/MICore/Debugger.cs:961 in 46756e5. [](commit_id = 46756e52bca7c620c0cbd974b8ec77072ca3632b, deletion_comment = False)

gregg-miskelly (changes requested):
See other comments

src/MICore/Debugger.cs
intel-rganesh: _gdbVersion.Split('.').FirstOrDefault() should either return a number, string or empty. If the TryParse fails, we would set the majorVersion to 0. I do not suspect an exception here, but I can wrap it inside a try catch block and log a warning.

src/MICore/Debugger.cs
intel-rganesh: It can be set to null inside GdbVersionFromLog when _initializationLog == null.

src/MICore/Debugger.cs
```diff
@@ -1037,21 +1054,18 @@ public void OnDebuggerProcessExit(/*OPTIONAL*/ string exitCode)
...
             {
-                // Second set of parenthesis looks for a Cygwin-specific version number
-                // Cygwin example: GNU gdb (GDB) (Cygwin 7.11.1-2) 7.11.1
-                // MinGW example:  GNU gdb (GDB) 8.0.1
-                Match match = Regex.Match(line, "GNU gdb \\(GDB\\) (?:\\(Cygwin (\\d+[\\d\\.-]*)\\) )?(\\d+[\\d\\.-]*)");
-                if (match.Success)
-                {
-                    return match.Groups[1].Success ? match.Groups[1].Value : match.Groups[2].Value;
-                }
+                _gdbVersion =  match.Groups[1].Success ? match.Groups[1].Value : match.Groups[2].Value;
```
gregg-miskelly, line 1067: >_gdbVersion [](http://example.com/codeflow?start=16&length=11)

I would move the code for `SetMajorGdbVersion` here

src/MICore/Debugger.cs
gregg-miskelly: No reason have a try/catch, I was instead just suggesting that we should use the same major version if `_gdbVersion` is missing, then if TryParse returned false. So either initialize `majorVersion` to 0, or check the return value, and if it is false, assign it -1. Since the command factory defaults it to 0, I would probably suggest that.

intel-rganesh: Done!

src/MICore/Debugger.cs
```diff
@@ -492,8 +494,6 @@ public async Task WaitForConsoleDebuggerInitialize(CancellationToken token)
             {
                 _consoleDebuggerInitializeCompletionSource = null;
 
-                // We no longer care about keeping these, so empty them out
-                _initializationLog = null;
```
gregg-miskelly, line 496: You want to put this back

src/MICore/Debugger.cs
```diff
@@ -978,24 +982,16 @@ public void OnDebuggerProcessExit(/*OPTIONAL*/ string exitCode)
...
-                            }
-                            MICommandFactory.MajorVersion = majorVersion;
-
+                            
                             // We can't use IsMinGW or IsCygwin because we never connected to the debugger
                             bool isMinGWOrCygwin = _launchOptions is LocalLaunchOptions &&
                                     PlatformUtilities.IsWindows() &&
                                     this.MICommandFactory.Mode == MIMode.Gdb;
-                            if (isMinGWOrCygwin && version != null && IsUnsupportedWindowsGdbVersion(version))
+                            if (isMinGWOrCygwin && IsUnsupportedWindowsGdbVersion(_gdbVersion))
```
gregg-miskelly, line 990: >IsUnsupportedWindowsGdbVersion [](http://example.com/codeflow?start=51&length=30)

`&& !string.IsNullOrEmpty(_gdbVersion)`

gregg-miskelly: Nevermind. Not required

gregg-miskelly (changes requested):
Aside from putting back line 495 now, looks good to me.

src/MICore/Debugger.cs
intel-rganesh: I was in two minds on this one since the logs weren't too long and I wasn't sure if we really wanted to clear them. But given we could fit our logic at the right place (thanks for the tip), and it was cleared previously, I assume it must have been done for a reason. So reverting to the old state!

gregg-miskelly (approved):

WardenGnaw (approved):

PR #1442, opened 2024-02-29
Remove `isabase` from IntelGTRegisters check

sbobko:
For compatibility with future IntelGT platforms, it is necessary to remove the check for the presence of the virtual register of `isabase`.


WardenGnaw (approved):

WardenGnaw:
@sbobko Thank you for your contribution. We will be aiming to put this in the next version of the VS Code C/C++ Extension and Visual Studio 2022 17.10.

PR #1455, opened 2024-05-27
Migrate MIEngine to .NET 8

WardenGnaw:
This PR upgrades OpenDebugAD7, MakePIAPortable, and CppTests/DAR from .NET 6 to .NET 8. 

.NET 6 will be ending on Nov 2024, so this PR moves to the latest LTS, .NET 8.

See https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-and-net-core


gregg-miskelly (approved):
:shipit:

PR #1460, opened 2024-08-08
MIEngine: Disable individual info threads call for each thread created

intel-rganesh:
Changes to disable info thread for each created thread to expedite GPU
debugging. GPUs are likely to have 1000s of threads and calling thread
info for each thread created is very time consuming. 
We call SetThreadInfoFromResultValue from CollectThreadInfo when we
get thread context.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
+++ b/src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
@@ -408,12 +408,12 @@ namespace Microsoft.MIDebugEngine
                 _callback.OnError(message);
             };
 
-            ThreadCreatedEvent += async delegate (object o, EventArgs args)
+            ThreadCreatedEvent += delegate (object o, EventArgs args)
             {
                 try
                 {
                     ResultEventArgs result = (ResultEventArgs)args;
-                    await ThreadCache.ThreadCreatedEvent(result.Results.FindInt("id"), result.Results.TryFindString("group-id"));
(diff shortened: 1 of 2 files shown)


gregg-miskelly:
There are at least parts of this change that we don't want for CPU debugging -- with this change we will no longer raise up thread create events to the rest of the debugger when GDB raises a thread create, which would be a user-visible change at the least, and I am not sure if it would break anything.

For GPU debugging, I am also a bit worried this change isn't enough -- we will still wind up creating thread objects in the UI for all the GPU threads. Is that actually usable in either VS or VS Code? For example in the VS Code call stack window, or in VS in the parallel stacks, or threads window? I would have thought you would just want to show a small number (maybe just 1 virtual thread?) with some custom UI to select the thread.

intel-rganesh:
For GPU debugging, the changes are to ensure we do not call thread-info for each thread. The thread created event is still called.
Any NewThread should be added during GetThreadContext.
I do see the Thread window and Parallel Stacks working fine (although it is slow).

image

Our use case is to show all the stopped GPU threads (similar to Threads in VS), lanes, be able to switch to different GPU threads/lanes and inspect variables.

The views function when we exit the kernel code and step in the CPU as well. 

image

prakashregistermails (commented):
The GetThread(id) was also doing a GetThreads() call, which was causing a lag, was it not observed?

intel-rganesh:
> The GetThread(id) was also doing a GetThreads() call, which was causing a lag, was it not observed?

This was very insignificant compared to ThreadInfo calls in ThreadCreatedEvent. GetThread(id) was called once for a breakpoint but ThreadCreatedEvent is called for each hardware thread. This can be in thousands.

PR #1461, opened 2024-08-28
Migrate ADO Release to 1ES ADO Pipeline

WardenGnaw:
This PR address the issue that ADO Release is not using a governed template. This PR adds the release stage as part of the official build

### DIFF
--- a/eng/pipelines/VS-release.yml
+++ b/eng/pipelines/VS-release.yml
@@ -1,7 +1,9 @@
 ---
 name: $(Date:yyyMMdd).$(Rev:r)
+
 variables:
-  - group: TSDTUSR
+- name: TeamName
+  value: MDDDebugger
 
 resources:
   repositories:
@@ -21,7 +23,7 @@ extends:
         name: VSEngSS-MicroBuild2022-1ES
         os: windows 
     stages:
-    - stage: stage
+    - stage: BuildVSReleasePackage
       displayName: VS_Release
       jobs:
       - job: Phase_1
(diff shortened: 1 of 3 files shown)


eng/pipelines/VS-Insertion-release.yml
```diff
@@ -0,0 +1,81 @@
...
+            Write-Host "##vso[task.setvariable variable=MDDPackageVersion;]$version"
+          displayName: 'Set MDDPackage Version'
+
+        - task: ms-vseng.MicroBuildShipTasks.55100717-a81d-45ea-a363-b8fe3ec375ad.MicroBuildInsertVsPayload@4
+          displayName: 'Insert VS Payload'
+          inputs:
+            TargetBranch: $(TargetBranch)  
+            TeamName: 'VS Debugger Platform'
+            TeamEmail: $(TEAMEMAIL)
+            DefaultConfigValues: 'VS.Redist.Debugger.MDD.MIEngine=$(MDDPackageVersion)'
```
gregg-miskelly, line 50: >DefaultConfigValues: 'VS.Redist.Debugger.MDD.MIEngine=$(MDDPackageVersion)' [](http://example.com/codeflow?start=12&length=75)

Is this still correct? Or there a different way to update src/ConfigData?

gregg-miskelly (approved):
Assuming `DefaultConfigValues` still works, looks good to me.

eng/pipelines/VS-Insertion-release.yml
WardenGnaw: Yep. This modifies `src/ConfigData/Packages/misc.props`

WardenGnaw:
@gregg-miskelly I've discussed with Austen about how we should do this for all the repos. After discussing, it would be preferable that we do the insertion as part of the official build since if it's in its own stage, it's easier to just requeue that if it fails.

gregg-miskelly (approved):

PR #1485, opened 2024-12-19
Update Nuget Packages for Compliance

WardenGnaw:
VS Packages were using System.Text.Json 8.0.3, updated them to use 8.0.4.

Also updated xunit packages and removed code analysis packages to use the built in .NET ones

Updated macOS pool for GitHub Actions but disabled testing since there is no downloadable arm64 lldb-mi to test aganist

### DIFF
--- a/.github/workflows/Build-And-Test.yml
+++ b/.github/workflows/Build-And-Test.yml
@@ -150,7 +150,7 @@ jobs:
         path: ${{ github.workspace }}/bin/DebugAdapterProtocolTests/Debug/CppTests/results.trx
 
   osx_build:
-    runs-on: macos-12
+    runs-on: macos-latest
     steps:
     - name: Checkout
       uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/Build-And-Test.yml
```diff
@@ -150,7 +150,7 @@ jobs:
         path: ${{ github.workspace }}/bin/DebugAdapterProtocolTests/Debug/CppTests/results.trx
 
   osx_build:
-    runs-on: macos-12
+    runs-on: macos-latest
```
gregg-miskelly, line 153: >latest [](http://example.com/codeflow?start=19&length=6)

Should we pick one of the version number pools instead so this doesn't immediately swap over?

gregg-miskelly: Eh. I guess that is okay.

.github/workflows/Build-And-Test.yml
```diff
@@ -168,8 +168,8 @@ jobs:
     - run: |
         ${{ github.workspace }}/eng/Scripts/CI-Build.sh
 
-    - run: |
-        ${{ github.workspace }}/eng/Scripts/CI-Test.sh
+    # - run: |
```
gregg-miskelly, line 171: ># [](http://example.com/codeflow?start=4&length=1)

Add a comment why?

gregg-miskelly (approved):

gregg-miskelly (approved):

PR #1489, opened 2025-02-11
DAR: Allow caller to control VSASSERT path

gregg-miskelly:
For other internal clients, it would be helpful if one could control the VSASSERT path, as this path is also used to write out where dump files are stored on Windows. This PR updates support to enable this.


WardenGnaw (approved):
:shipit:

Issue #1491, opened 2025-02-19
Debugging from VS code rewrites args if containing ISO 8601 compliant date/time string


stertingen:
Passing an ISO 8601 compliant string to `"args"` in VS code's `launch.json` leads to unexpected string conversion.

`launch.json:`
```jsonc
{
    "configurations": [
        {
            "name": "(gdb) Launch",
            "type": "cppdbg",
            "request": "launch",
            "program": "/bin/echo",
            "args": [
                "2022-01-03T03:00:00+01:00",
            ],
            "cwd": "${workspaceFolder}",
        }
    ]
}
```
Launching above debug configuration does not print the original string, but some americanized string instead.

See also: https://github.com/microsoft/vscode-cpptools/issues/13241, https://github.com/microsoft/vscode/issues/238514

I don't speak C#, but I suspect some implicit JSON-to-Datetime and Datetimt-to-JSON serialization here: https://github.com/microsoft/MIEngine/blob/main/src/OpenDebugAD7/AD7DebugSession.cs#L1195

stertingen:
Related to: https://github.com/JamesNK/Newtonsoft.Json/issues/862

gregg-miskelly:
Reopening since this needs a second fix

PR #1492, opened 2025-02-21
Don't try to parse JSON strings as date

stertingen:
When parsing launch options from JSON, the JSON parser's default behavior is – for whatever reason – to parse slightly datetime-ish looking strings as date objects. This PR explicitly forbids this behavior when parsing launch options.

Partially fixes https://github.com/microsoft/MIEngine/issues/1491, https://github.com/microsoft/vscode-cpptools/issues/13241 and https://github.com/microsoft/vscode/issues/238514.

### DIFF
--- a/src/MICore/LaunchOptions.cs
+++ b/src/MICore/LaunchOptions.cs
@@ -1268,6 +1268,10 @@ namespace MICore
                 try
                 {
                     JObject parsedOptions = JsonConvert.DeserializeObject<JObject>(options, new JsonSerializerSettings { DateParseHandling = DateParseHandling.None });
+                    if (parsedOptions is null)
+                    {
+                        throw new InvalidLaunchOptionsException(MICoreResources.Error_UnknownLaunchOptions);
+                    }
 
                     // if the customLauncher element is present then try using the custom launcher implementation from the config store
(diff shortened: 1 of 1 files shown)


stertingen:
Never mind, this PR does not fix anything. It just modifies the string in a more subtle way.

gregg-miskelly:
Thanks for working so hard to fix this!

src/MICore/LaunchOptions.cs
```diff
@@ -1267,7 +1267,7 @@ public static LaunchOptions GetInstance(HostConfigurationStore configStore, stri
             {
                 try
                 {
-                    JObject parsedOptions = JObject.Parse(options);
+                    JObject parsedOptions = JsonConvert.DeserializeObject<JObject>(options, new JsonSerializerSettings { DateParseHandling = DateParseHandling.None });
```
gregg-miskelly, line 1270: I think you need one tweak here:

```suggestion
                    JObject parsedOptions = JsonConvert.DeserializeObject<JObject>(options, new JsonSerializerSettings { DateParseHandling = DateParseHandling.None });
                    if (parsedOptions is null)
                    {
                        throw new InvalidLaunchOptionsException(MICoreResources.Error_UnknownLaunchOptions);
                    }
```

gregg-miskelly (commented):
Aside from that, looks good to me

gregg-miskelly (approved):

PR #1493, opened 2025-02-26
Update VSCodeDebugProtocol library

gregg-miskelly:
This PR updates the version of Microsoft.VisualStudio.Shared.VSCodeDebugProtocol to 17.14.10225.1. This completes the work for #1491.

I also noticed the solution doesn't reference README.md, so this adds a link.


gregg-miskelly:
FYI @stertingen, with this your scenario should now work. Thanks for bringing it to our attention and spending so much time tracking down the cause.

WardenGnaw (approved):

PR #1500, opened 2025-04-04
Enable debuginfod

paulmaybee:
Enable debuginfod by default. Prevents gdb 15 and later from asking to enable debuginfod after attaching.


src/MIDebugEngine/Engine.Impl/DebuggedProcess.cs
```diff
@@ -624,6 +624,7 @@ private async Task<List<LaunchCommand>> GetInitializeCommands()
             if (_launchOptions.DebuggerMIMode == MIMode.Gdb)
             {
                 commands.Add(new LaunchCommand("-interpreter-exec console \"set pagination off\""));
+                commands.Add(new LaunchCommand("set debuginfod enabled on", ignoreFailures:true));
```
WardenGnaw, line 627: Do you want to do this for all versions of GDB or just for GDB >= v15.0

```suggestion
if (MICommandFactory.MajorVersion >= 15)
{
   commands.Add(new LaunchCommand("set debuginfod enabled on", ignoreFailures:true));
}
```

paulmaybee: I was trying to keep it simple. It won't hurt for the older versions of gdb too. the command is ignored if it fails.

WardenGnaw: If its ok with older versions, it should be ok.

WardenGnaw (approved):

gregg-miskelly (approved):

PR #1558, opened 2026-05-07
Natvis intrinsic support

lugerard:
This PR adds two improvements to the NatVis engine for GDB/LLDB.

Strip dll!-qualified module prefixes

NatVis files targeting Windows often qualify type names with a module prefix (e.g. Qt6Cored.dll!) to disambiguate types. These prefixes are meaningless to GDB and LLDB and cause expression evaluation to fail on Linux/macOS. ReplaceNamesInExpression now strips them before any expression is sent to the debugger.

Intrinsic support

<Intrinsic> elements define named inline expressions within a <Type> block that can be called by name (with optional parameters) in any DisplayString or Expand expression in the same block. MIEngine previously ignored them silently, causing visualizers that rely on intrinsics to fall back to raw view. This change collects intrinsics from the <Type> block and resolves calls before expression evaluation. The schema (natvis.xsd) and generated types (NatvisXsdTypes.cs) are updated accordingly.

Added tests.

### DIFF
--- a/src/MIDebugEngine/Natvis.Impl/Natvis.cs
+++ b/src/MIDebugEngine/Natvis.Impl/Natvis.cs
@@ -226,6 +226,7 @@ namespace Microsoft.MIDebugEngine.Natvis
         private static Regex s_variableName = new Regex("[a-zA-Z$_][a-zA-Z$_0-9]*");
         private static Regex s_subfieldNameHere = new Regex(@"\G((\.|->)[a-zA-Z$_][a-zA-Z$_0-9]*)+");
         private static Regex s_expression = new Regex(@"^\{[^\}]*\}");
+        private static readonly Regex s_dllQualifiedPrefix = new Regex(@"\w+(\.\w+)*\.dll!");
         private List<FileInfo> _typeVisualizers;
         private DebuggedProcess _process;
         private HostConfigurationStore _configStore;
@@ -1392,6 +1393,10 @@ namespace Microsoft.MIDebugEngine.Natvis
 
         private string ReplaceNamesInExpression(string expression, IVariableInformation variable, IDictionary<string, string> scopedNames)
         {
+            // Strip Windows dll!-qualified type prefixes (e.g. Qt6Cored.dll!)
+            // for GDB/LLDB compatibility — meaningless outside Windows
+            expression = s_dllQualifiedPrefix.Replace(expression, "");
+
             return ProcessNamesInString(expression, new Substitute[] {
                 (m)=>
                     {
--- a/src/MIDebugEngine/Natvis.Impl/Natvis.cs
+++ b/src/MIDebugEngine/Natvis.Impl/Natvis.cs
@@ -202,6 +202,13 @@ namespace Microsoft.MIDebugEngine.Natvis
             public VisualizerType Visualizer { get; private set; }
             public Dictionary<string, string> ScopedNames { get; private set; }
 
+            /// <summary>
+            /// Intrinsics defined in this type block, keyed by name.
+            /// Stored as IntrinsicType so that Parameter[] is available at call time
+            /// for argument substitution in parametrized intrinsics.
+            /// </summary>
+            public Dictionary<string, IntrinsicType> Intrinsics { get; private set; }
+
             public VisualizerId[] GetUIVisualizers()
             {
                 return this.Visualizer.Items.Where((i) => i is UIVisualizerItemType).Select(i =>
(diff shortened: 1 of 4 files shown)


lugerard:
@microsoft-github-policy-service agree [company="The Qt Company"]

Confidential

lugerard:
@microsoft-github-policy-service agree company="The Qt Company"

Confidential

lugerard:
@waan @greggm  would appreciate a review when you have a moment.

src/MIDebugEngine/Natvis.Impl/natvis.xsd
```diff
@@ -577,6 +577,34 @@
     </xs:simpleContent>
   </xs:complexType>
 
+  <xs:complexType name="ParameterType">
```
gregg-miskelly: >ParameterType [](http://example.com/codeflow?start=24&length=13)

nit: can you rename this to `IntrinsicParameterType` to match with the Visual Studio version of natvis.xsd

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -202,6 +202,13 @@ internal class VisualizerInfo
             public VisualizerType Visualizer { get; private set; }
             public Dictionary<string, string> ScopedNames { get; private set; }
 
+            /// <summary>
+            /// Intrinsics defined in this type block, keyed by name.
+            /// Stored as IntrinsicType so that Parameter[] is available at call time
+            /// for argument substitution in parametrized intrinsics.
+            /// </summary>
+            public Dictionary<string, IntrinsicType> Intrinsics { get; private set; }
```
gregg-miskelly: >private set; [](http://example.com/codeflow?start=71&length=12)

Remove since you are only assigning in the constructor

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -220,12 +227,26 @@ public VisualizerInfo(VisualizerType viz, TypeName name)
...
+                        }
+                    }
+                }
             }
         }
 
         private static Regex s_variableName = new Regex("[a-zA-Z$_][a-zA-Z$_0-9]*");
         private static Regex s_subfieldNameHere = new Regex(@"\G((\.|->)[a-zA-Z$_][a-zA-Z$_0-9]*)+");
         private static Regex s_expression = new Regex(@"^\{[^\}]*\}");
+        private static readonly Regex s_dllQualifiedPrefix = new Regex(@"\w+(\.\w+)*\.dll!");
```
gregg-miskelly: >s_dllQualifiedPrefix = new Regex(@"\w+(\.\w+)*\.dll!"); [](http://example.com/codeflow?start=38&length=55)

I think you want this to be `Regex s_moduleQuerifiedPrefix = new Regex(@"\w+(?:\.\w+)*\.(?:dll|exe)!", RegexOptions.IgnoreCase);`

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -220,12 +227,26 @@ public VisualizerInfo(VisualizerType viz, TypeName name)
...
+        private static readonly Regex s_intrinsicCallPattern = new Regex(@"\b([A-Za-z_]\w*)\s*\(");
```
gregg-miskelly: >[A-Za-z_]\w* [](http://example.com/codeflow?start=78&length=12)

Use `\w+` so you can handle non-Latin characters. That will allow digits at the start of the identifier, but should be okay since they will not be in your dictionary.

lugerard: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
lugerard: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
lugerard: done

src/MIDebugEngine/Natvis.Impl/natvis.xsd
lugerard: done

WardenGnaw:
@lugerard I rebased this PR on top of main, but from the builds theres a couple of code analysis build failures, do you mind fixing them?

lugerard:
> @lugerard I rebased this PR on top of main, but from the builds theres a couple of code analysis build failures, do you mind fixing them?

@WardenGnaw: Fixed in the latest push, the build failures were caused by NatvisIntrinsicTest.cs not being updated when ParameterType was renamed to IntrinsicParameterType in the review-comments commit. Could you approve the workflow run so CI can verify?

WardenGnaw (approved):

WardenGnaw:
@lugerard Thank you for the contribution!

Which version of MIEngine are you looking to see these changes in?
VS Code C++ Extension, Visual Studio 2026 or both?

lugerard:
> @lugerard Thank you for the contribution!
> 
> Which version of MIEngine are you looking to see these changes in? VS Code C++ Extension, Visual Studio 2026 or both?

Both, ideally. The fix is relevant anywhere MIEngine bridges GDB/LLDB. Do you have a rough idea of when we could expect the change to be available in each?

WardenGnaw:
Would most likely land in C/C++ Extension for the June release and VS 2026 18.8 or 18.9, aka the releases in June

lugerard:
> Would most likely land in C/C++ Extension for the June release and VS 2026 18.8 or 18.9, aka the releases in June

That's great! Thanks.

PR #1573, opened 2026-05-29
Natvis: strip address prefix from string pointer DisplayStrings"

lugerard:
GDB and LLDB prefix string pointer results with the pointer address when evaluating ,sub, ,su or ,sb format specifiers. This causes the address to appear inline in the display value.

Add StripFormatSpecifier() and ExtractFormatSpecifier() to parse format specifiers from NatVis expressions. Add CleanUtf16StringValue() to strip the address and surrounding u"..." quotes for ,sub / ,su, and CleanAsciiStringValue() to strip the address for ,sb. Call each in FormatValue for the corresponding specifier. All new functions include unit tests.

### DIFF
--- a/src/MIDebugEngine/Natvis.Impl/Natvis.cs
+++ b/src/MIDebugEngine/Natvis.Impl/Natvis.cs
@@ -247,6 +247,8 @@ namespace Microsoft.MIDebugEngine.Natvis
         private static Regex s_expression = new Regex(@"^\{[^\}]*\}");
         private static readonly Regex s_moduleQualifiedPrefix = new Regex(@"\w+(?:\.\w+)*\.(?:dll|exe)!", RegexOptions.IgnoreCase);
         private static readonly Regex s_intrinsicCallPattern = new Regex(@"\b(\w+)\s*\(");
+        // Matches the leading "0x<hex> " address that GDB/LLDB prepends when displaying a string pointer value.
+        private static readonly Regex s_addressPrefix = new Regex(@"^0x[0-9a-fA-F]+\s+");
         private List<FileInfo> _typeVisualizers;
         private DebuggedProcess _process;
         private HostConfigurationStore _configStore;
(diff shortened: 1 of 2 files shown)


src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1460,6 +1468,93 @@ internal static List<string> SplitArguments(string argsText)
...
+        }
+
+        /// <summary>
+        /// Strips a Visual Studio NatVis format specifier (e.g. ",sub", ",d", ",Xb",
+        /// ",view(name)na") from the end of an expression.  Format specifiers follow the
+        /// last top-level comma (i.e. a comma not nested inside any parentheses or
+        /// square brackets).  GDB and LLDB do not understand these specifiers; leaving them
+        /// in place causes expression evaluation to fail.
+        /// </summary>
+        internal static string StripFormatSpecifier(string expression)
```
WardenGnaw: `StripFormatSpecifier` only looks to be used in `NatvisFormatSpecifierTest`. Can we remove this method and the associated Tests that are calling it?

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1460,6 +1468,93 @@ internal static List<string> SplitArguments(string argsText)
...
+            }
+            return value;
+        }
+
+        /// <summary>
+        /// Cleans up the raw value that GDB/LLDB returns for a <c>char*</c> expression
+        /// (i.e. one evaluated with the <c>,sb</c> format specifier).
+        /// GDB and LLDB prefix the string with the pointer address, e.g.
+        ///   <c>0x00007fff5fbff6c0 "Hello"</c>
+        /// This method strips the address, leaving the quoted string content.
```
WardenGnaw: Why does `CleanUtf16StringValue` removes the quotes but not `CleanAsciiStringValue`?

src/MIDebugEngine/Natvis.Impl/Natvis.cs
lugerard: Removed along with its tests.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
lugerard: CleanAsciiStringValue should strip quotes just like CleanUtf16StringValue, since VS returns bare text for both ,sb and ,sub. Fixed in this updated version. Thanks.

WardenGnaw (approved):

PR #1579, opened 2026-06-10
Refactor Docker transport into shared base classes and strategy pattern

WardenGnaw:
Refactoring the DockerContainer code to be extensible for other Container tools such as Podman.

Key changes:
- `ContainerTransportSettingsBase` shared abstract base eliminates duplication in transport settings (exe name, host flag, command format)
- `IContainerDiscoveryStrategy` interface with Docker implementation keeps runtime-specific logic out of the ViewModel
- `ContainerRuntimeType` enum threaded through port picker -> ConnectionManager -> dialog for future extensibility
- XAML bindings changed from static resources to ViewModel properties so labels can vary per runtime

### DIFF
--- a/src/SSHDebugPS/ContainerRuntimeType.cs
+++ b/src/SSHDebugPS/ContainerRuntimeType.cs
@@ -1,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.SSHDebugPS
+{
+    /// <summary>
+    /// Identifies the container runtime to query when discovering containers.
+    /// </summary>
+    public enum ContainerRuntimeType
+    {
+        Unknown,
+        Docker
+    }
+}
--- a/src/SSHDebugPS/Docker/DockerContainerInstance.cs
+++ b/src/SSHDebugPS/Docker/DockerContainerInstance.cs
@@ -37,7 +37,7 @@ namespace Microsoft.SSHDebugPS.Docker
             return instance != null;
         }
 
-        private DockerContainerInstance() { }
+        protected DockerContainerInstance() { }
 
         #region JsonProperties
 
@@ -48,19 +48,19 @@ namespace Microsoft.SSHDebugPS.Docker
         public override string Name { get; set; }
 
         [JsonProperty(nameof(Image))]
-        public string Image { get; private set; }
+        public virtual string Image { get; protected set; }
 
         [JsonProperty(nameof(Ports))]
-        public string Ports { get; set; }
+        public virtual string Ports { get; set; }
 
         [JsonProperty(nameof(Command))]
-        public string Command { get; private set; }
+        public virtual string Command { get; protected set; }
 
         [JsonProperty(nameof(Status))]
-        public string Status { get; private set; }
+        public virtual string Status { get; protected set; }
 
         [JsonProperty("CreatedAt")]
-        public string Created { get; private set; }
+        public virtual string Created { get; protected set; }
 
         [JsonIgnore]
         public string Platform { get; set; }
--- a/src/SSHDebugPS/Docker/DockerDiscoveryStrategy.cs
+++ b/src/SSHDebugPS/Docker/DockerDiscoveryStrategy.cs
@@ -1,0 +1,77 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using Microsoft.SSHDebugPS.Docker;
+using Microsoft.SSHDebugPS.UI;
+
+namespace Microsoft.SSHDebugPS
+{
+    internal class DockerDiscoveryStrategy : IContainerDiscoveryStrategy
+    {
+        private const string unknownOS = "Unknown";
+
+        public string ConnectionLabel => UIResources.ConnectionLabel;
+        public string HostnameLabel => UIResources.HostnameLabel;
+        public string HostnameTip => UIResources.HostnameTip;
+        public string ConnectionToolTip => UIResources.ConnectionToolTip;
+        public string HostnameAutomationName => UIResources.HostnameAutomationName;
+
+        public IEnumerable<DockerContainerInstance> GetLocalContainers(string hostname, out int totalContainers)
+        {
+            return DockerHelper.GetLocalDockerContainers(hostname, out totalContainers);
+        }
+
+        public IEnumerable<DockerContainerInstance> GetRemoteContainers(IConnection connection, string hostname, out int totalContainers)
+        {
+            return DockerHelper.GetRemoteDockerContainers(connection, hostname, out totalContainers);
+        }
+
+        public void AssignPlatforms(IEnumerable<DockerContainerInstance> containers, string hostname)
+        {
+            if (!containers.Any())
+                return;
+
+            string serverOS;
+            if (DockerHelper.TryGetServerOS(hostname, out serverOS))
+            {
+                bool lcow;
+                DockerHelper.TryGetLCOW(hostname, out lcow);
+                TextInfo textInfo = new CultureInfo("en-US", false).TextInfo;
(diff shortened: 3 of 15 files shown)


src/SSHDebugPS/ContainerRuntimeType.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.SSHDebugPS
+{
+    /// <summary>
+    /// Identifies the container runtime to query when discovering containers.
+    /// </summary>
+    public enum ContainerRuntimeType
+    {
+        Unknown,
```
gregg-miskelly, line 11: >Unknown, [](http://example.com/codeflow?start=8&length=8)

This doesn't seem to be used. Should we remove it or add a TODO comment to remove it when extending?

src/SSHDebugPS/Docker/DockerContainerInstance.cs
```diff
@@ -37,7 +37,7 @@ public static bool TryCreate(string json, out DockerContainerInstance instance)
             return instance != null;
         }
 
-        private DockerContainerInstance() { }
+        protected DockerContainerInstance() { }
```
gregg-miskelly, line 40: >protected [](http://example.com/codeflow?start=8&length=9)

Are the changes to this class (aside from adding `virtual`) needed? Seems like this class could actually be `sealed` unless I am missing something

src/SSHDebugPS/Docker/DockerDiscoveryStrategy.cs
```diff
@@ -0,0 +1,77 @@
...
+            if (!containers.Any())
+                return;
+
+            string serverOS;
+            if (DockerHelper.TryGetServerOS(hostname, out serverOS))
+            {
+                bool lcow;
+                DockerHelper.TryGetLCOW(hostname, out lcow);
+                TextInfo textInfo = new CultureInfo("en-US", false).TextInfo;
+                serverOS = textInfo.ToTitleCase(serverOS);
```
gregg-miskelly: Why do this instead of making the `serverOS.Contains` calls `OrdinalIgnoreCase`?

gregg-miskelly: Or maybe a regex like `\bWindows\b` so it isn't picking up the middle of the word

src/SSHDebugPS/Docker/TransportSettings/DockerTransportSettings.cs
```diff
@@ -1,78 +1,12 @@
...
-        #region IPipeTransportSettings
-
-        public string CommandArgs => GenerateExeCommandArgs();
-
-        public string Command => HostIsUnix ? UnixExe : WindowsExe;
-        #endregion
-    }
-
-    internal class DockerCommandSettings : DockerTransportSettingsBase
+    internal class DockerCommandSettings : ContainerCommandSettings
```
gregg-miskelly: >class [](http://example.com/codeflow?start=13&length=5)

Suggestion: it might be helpful to seal all of the classes that we don't expect to have inheritance now that there is a bit a class hierarchy

src/SSHDebugPS/Docker/DockerPortPicker.cs
```diff
@@ -33,12 +33,13 @@ public class DockerWindowsPortPicker : DockerPortPickerBase
     public abstract class DockerPortPickerBase : IDebugPortPicker
     {
         internal abstract bool SupportSSHConnections { get; }
+        internal virtual ContainerRuntimeType RuntimeType => ContainerRuntimeType.Docker;
```
gregg-miskelly, line 36: >virtual [](http://example.com/codeflow?start=17&length=7)

Will other containers inherit from this class? If so, should we rename it? If not, should this just be a constant?

src/SSHDebugPS/UI/ContainerPickerDialogWindow.xaml.cs
```diff
@@ -25,9 +25,9 @@ namespace Microsoft.SSHDebugPS.UI
     /// </summary>
     public partial class ContainerPickerDialogWindow : DialogWindow
     {
-        public ContainerPickerDialogWindow(bool supportSSHConnections)
+        public ContainerPickerDialogWindow(bool supportSSHConnections, ContainerRuntimeType runtimeType = ContainerRuntimeType.Docker)
```
gregg-miskelly: > = ContainerRuntimeType.Docker [](http://example.com/codeflow?start=103&length=30)

Suggestion: remove default value

src/SSHDebugPS/UI/ViewModels/ContainerPickerViewModel.cs
```diff
@@ -22,10 +22,11 @@ public class ContainerPickerViewModel : INotifyPropertyChanged
     {
         private Lazy<bool> _sshAvailable;
 
-        public ContainerPickerViewModel(bool supportSSHConnections)
+        public ContainerPickerViewModel(bool supportSSHConnections, ContainerRuntimeType runtimeType = ContainerRuntimeType.Docker)
```
gregg-miskelly: > = ContainerRuntimeType.Docker [](http://example.com/codeflow?start=100&length=30)

Same suggestion

gregg-miskelly (commented):
Otherwise LGTM

src/SSHDebugPS/ContainerRuntimeType.cs
WardenGnaw: I was debating only having enum or just `Docker` before adding `Podman`. Is it better to not have a value if for some reason someone wanted an invalid value?

src/SSHDebugPS/Docker/DockerContainerInstance.cs
WardenGnaw: Sorry, I am holding back the podman changes to reduce the amount of code to review, but it might have caused confusion.

I am going to have a `PodmanContainerInstance` inheriting DockerContainerInstance.

src/SSHDebugPS/Docker/DockerDiscoveryStrategy.cs
WardenGnaw: Copy pasted the original code, but I can make this improvement :)

src/SSHDebugPS/Docker/DockerPortPicker.cs
WardenGnaw: Planning on Podman version inheriting this one too.

src/SSHDebugPS/Docker/DockerDiscoveryStrategy.cs
```diff
@@ -0,0 +1,77 @@
...
+                DockerHelper.TryGetLCOW(hostname, out lcow);
+
+                if (lcow && serverOS.IndexOf("windows", StringComparison.OrdinalIgnoreCase) >= 0)
+                {
+                    foreach (DockerContainerInstance container in containers)
+                    {
+                        string containerPlatform = string.Empty;
+                        if (DockerHelper.TryGetContainerPlatform(hostname, container.Name, out containerPlatform))
+                        {
+                            container.Platform = new CultureInfo("en-US", false).TextInfo.ToTitleCase(containerPlatform);
```
gregg-miskelly: >CultureInfo("en-US", false) [](http://example.com/codeflow?start=53&length=27)

I didn't think about this originally, but shouldn't this be `CultureInfo.CurrentCulture` instead of hard-coding en-US?

gregg-miskelly (approved):
Otherwise looks good to me

src/SSHDebugPS/Docker/DockerDiscoveryStrategy.cs
WardenGnaw: Did you mean `CultureInfo.InvariantCulture`? This is the output from `docker inspect -f "{json .Platform}"` which should be `windows` or `linux`

gregg-miskelly: Got it. In that case, nevermind. 

Though maybe we want to put back the local variable so we don't recreate the CulutureInfo object for each container?

gregg-miskelly (approved):

PR #1580, opened 2026-06-10
Natvis: implement view() specifier, IncludeView/ExcludeView filtering, and CustomListItems loop engine

lugerard:
This PR enables MIEngine to evaluate two natvis features:
view-based filtering (`IncludeView`/`ExcludeView` on `DisplayString`
and `Expand` elements, activated via the `view()` format specifier) and
`<CustomListItems>` loop bodies (`<Loop>`, `<Item>`, `<Exec>`,
`<If>`/`<ElseIf>`/`<Else>`, `<Break>`), which are the standard way to walk
pointer-based data structures such as linked lists and trees.

The two commits are bundled together because `<CustomListItems>` loop bodies
can contain `IncludeView`/`ExcludeView` guards on their `<Item>` elements,
so commit 1 is a functional prerequisite for commit 2.

Commit 1 adds `view()` specifier parsing and `IncludeView`/`ExcludeView`
filtering, plus fixes to `EvalCondition` error handling and expression
pre-processing that are prerequisites for correct evaluation on GDB/LLDB.

Commit 2 implements the `<CustomListItems>` loop engine, including
expression-growth prevention after `<Exec>` assignments and a `$i`
word-boundary fix.

All new code is covered by unit tests.


src/MIDebugEngine/Natvis.Impl/NatvisXsdTypes.cs
```diff
@@ -618,6 +657,232 @@ public bool MaxItemsPerViewSpecified {
         }
     }
     
+    // ---- CustomListItems loop-body types ------------------------------------
+    //
+    // These types represent the body of a <Loop> element inside <CustomListItems>.
+    // The loop body is an ordered sequence of Item, Break, Exec, If, and Else elements.
+    // Nested Loop elements are also supported inside If/Else bodies.
+
```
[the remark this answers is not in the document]
gregg-miskelly: Apparently we stopped generating this file sometime back from the natvis.xsd. We should fix this, but it is too big of a change to do so in this PR.

src/MIDebugEngine/Natvis.Impl/NatvisXsdTypes.cs
```diff
@@ -618,6 +657,232 @@ public bool MaxItemsPerViewSpecified {
...
+        /// <remarks/>
+        [System.Xml.Serialization.XmlAttributeAttribute()]
+        public string Condition {
+            get { return this.conditionField; }
+            set { this.conditionField = value; }
+        }
+
+        /// <remarks/>
+        [System.Xml.Serialization.XmlElementAttribute("Item",   typeof(CustomListLoopItemType))]
```
gregg-miskelly: >CustomListLoopItemType [](http://example.com/codeflow?start=71&length=22)

Can you rename this to `CustomListItemType` to match the main natvis.xsd

src/MIDebugEngine/Natvis.Impl/NatvisXsdTypes.cs
```diff
@@ -618,6 +657,232 @@ public bool MaxItemsPerViewSpecified {
...
+        [System.Xml.Serialization.XmlElementAttribute("Break",  typeof(CustomListBreakType))]
```
gregg-miskelly: >CustomListBreakType [](http://example.com/codeflow?start=71&length=19)

Similar request for the rest. Here is how they are declared in the .xsd:
```xml
      <xs:element name="Loop" minOccurs="0" maxOccurs="unbounded" type="LoopType"></xs:element>
      <xs:sequence minOccurs="0" maxOccurs="unbounded">
        <xs:element name="If" minOccurs="1" maxOccurs="1" type="IfType"></xs:element>
        <xs:element name="Elseif" minOccurs="0" maxOccurs="unbounded" type="IfType"></xs:element>
        <xs:element name="Else" minOccurs="0" maxOccurs="1" type="ElseType"></xs:element>
      </xs:sequence>
      <xs:element name="Exec" minOccurs="0" maxOccurs="unbounded" type="ExecType"></xs:element>
      <xs:element name="Break" minOccurs="0" maxOccurs="unbounded" type="BreakType"></xs:element>
```

src/MIDebugEngine/Natvis.Impl/NatvisXsdTypes.cs
```diff
@@ -618,6 +657,232 @@ public bool MaxItemsPerViewSpecified {
...
+        [System.Xml.Serialization.XmlElementAttribute("Exec",   typeof(CustomListExecType))]
+        [System.Xml.Serialization.XmlElementAttribute("If",     typeof(CustomListIfType))]
+        [System.Xml.Serialization.XmlElementAttribute("ElseIf", typeof(CustomListElseIfType))]
```
gregg-miskelly: >ElseIf [](http://example.com/codeflow?start=55&length=6)

The casing of this is incorrect. It should be `Elseif`

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1771,6 +1993,287 @@ private string GetDisplayNameFromArrayIndex(uint arrayIndex, int rank, uint[] di
...
+                    string condExpr = SubstituteLocalVars(br.Condition, localVars);
+                    if (EvalCondition(condExpr, variable, visualizer.ScopedNames, visualizer.Intrinsics))
+                        ctx.Done = true;
+                }
+                else if (elem is CustomListLoopItemType li)
+                {
+                    // <Item Name="..." Condition="...">expr</Item>: emit a child variable.
+                    if (li.Condition != null)
+                    {
+                        string condExpr = SubstituteLocalVars(li.Condition, localVars);
```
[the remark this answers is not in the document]
gregg-miskelly: Doesn't seem to be required

gregg-miskelly:
Can you add an end-to-end test that uses this? Example: https://github.com/microsoft/MIEngine/pull/1559

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1771,6 +1993,287 @@ private string GetDisplayNameFromArrayIndex(uint arrayIndex, int rank, uint[] di
...
+                                string loopCond = SubstituteLocalVars(nestedLoop.Condition, localVars);
+                                if (!EvalCondition(loopCond, variable, visualizer.ScopedNames, visualizer.Intrinsics))
+                                    break;
+                            }
+                            bool innerProgress = ExecuteCustomListBody(nestedLoop.Items, ctx, variable, visualizer, localVars, children);
+                            if (!innerProgress && !ctx.Done) break;
+                            progress = true;
+                        }
+                    }
+                }
```
gregg-miskelly, line 2349: >} [](http://example.com/codeflow?start=16&length=1)

else -> write a warning for unexpected element?

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1771,6 +1993,287 @@ private string GetDisplayNameFromArrayIndex(uint arrayIndex, int rank, uint[] di
...
+                            if (!string.IsNullOrEmpty(nestedLoop.Condition))
+                            {
+                                string loopCond = SubstituteLocalVars(nestedLoop.Condition, localVars);
+                                if (!EvalCondition(loopCond, variable, visualizer.ScopedNames, visualizer.Intrinsics))
+                                    break;
+                            }
+                            bool innerProgress = ExecuteCustomListBody(nestedLoop.Items, ctx, variable, visualizer, localVars, children);
+                            if (!innerProgress && !ctx.Done) break;
+                            progress = true;
+                        }
```
gregg-miskelly: Can we share more code with the top level loop?

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1771,6 +1993,287 @@ private string GetDisplayNameFromArrayIndex(uint arrayIndex, int rank, uint[] di
...
+                {
+                    // <Exec Condition="...">varName = expr</Exec>: update a local variable.
+                    if (exec.Condition != null)
+                    {
+                        string condExpr = SubstituteLocalVars(exec.Condition, localVars);
+                        if (!EvalCondition(condExpr, variable, visualizer.ScopedNames, visualizer.Intrinsics))
+                            continue;
+                    }
+                    string execExpr = SubstituteLocalVars(exec.Value?.Trim() ?? "", localVars);
+                    string updatedVar = ApplyExecToLocalVars(execExpr, localVars);
```
gregg-miskelly: >ApplyExecToLocalVars [](http://example.com/codeflow?start=40&length=20)

Not sure if you want to handle this, but the .natvis files that ship with VS will often assign to multiple variables in an `<Exec>` block. Example: `<Exec>++idx, ++statptr</Exec>`. Feels like we need to at least log a warning, but it probably isn't too hard to support.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1771,6 +1993,287 @@ private string GetDisplayNameFromArrayIndex(uint arrayIndex, int rank, uint[] di
...
+                        {
+                            string normalized = GetExpressionValue(localVars[updatedVar], variable, visualizer.ScopedNames, visualizer.Intrinsics);
+                            if (!string.IsNullOrEmpty(normalized) &&
+                                !normalized.TrimStart().StartsWith("0x", StringComparison.OrdinalIgnoreCase))
+                            {
+                                localVars[updatedVar] = normalized;
+                            }
+                        }
+                        catch (Exception) { /* keep the expression as-is if evaluation fails */ }
+                    }
```
gregg-miskelly, line 2291: >} [](http://example.com/codeflow?start=20&length=1)

Should we be logging a warning here and breaking?

gregg-miskelly (changes requested):
Thanks for doing this. Aside from that, this seems about right to me.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -949,13 +979,96 @@ private IVariableInformation[] ExpandVisualized(IVariableInformation variable)
...
+                    var localVars = new Dictionary<string, string>(StringComparer.Ordinal);
+                    if (customList.Items != null)
+                    {
+                        foreach (var v in customList.Items)
+                        {
+                            if (string.IsNullOrEmpty(v.Name)) continue;
+                            string initVal = v.InitialValue ?? "0";
+                            // Resolve field names, template parameters and intrinsics in the initial value.
+                            localVars[v.Name] = ReplaceNamesInExpression(initVal, variable, visualizer.ScopedNames, visualizer.Intrinsics);
+                        }
```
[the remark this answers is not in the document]
lugerard: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1771,6 +1993,287 @@ private string GetDisplayNameFromArrayIndex(uint arrayIndex, int rank, uint[] di
...
+                    if (updatedVar != null)
+                    {
+                        try
+                        {
+                            string normalized = GetExpressionValue(localVars[updatedVar], variable, visualizer.ScopedNames, visualizer.Intrinsics);
+                            if (!string.IsNullOrEmpty(normalized) &&
+                                !normalized.TrimStart().StartsWith("0x", StringComparison.OrdinalIgnoreCase))
+                            {
+                                localVars[updatedVar] = normalized;
+                            }
```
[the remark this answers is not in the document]
lugerard: done

src/MIDebugEngine/Natvis.Impl/NatvisXsdTypes.cs
lugerard: done

src/MIDebugEngine/Natvis.Impl/NatvisXsdTypes.cs
lugerard: done

src/MIDebugEngine/Natvis.Impl/NatvisXsdTypes.cs
lugerard: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
lugerard: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
lugerard: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
lugerard: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
lugerard: done

lugerard:
> Can you add an end-to-end test that uses this? Example: #1559

done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1771,6 +1984,420 @@ private string GetDisplayNameFromArrayIndex(uint arrayIndex, int rank, uint[] di
...
+                    }
+                    ctx.GlobalIndex++;
+                    progress = true;
+
+                    // Check whether the page is now full.
+                    if (ctx.Emitted >= MAX_EXPAND && ctx.GlobalIndex < ctx.TotalSize)
+                    {
+                        children.Add(new PaginatedVisualizerWrapper(
+                            ResourceStrings.MoreView, _process.Engine, variable,
+                            visualizer, isVisualizerView: true, ctx.StartIndex + MAX_EXPAND));
```
gregg-miskelly: I am not quite understanding the "more elements" behavior here. Don't we need to save state from the expansion so that we can keep going where we left off?

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1771,6 +1984,420 @@ private string GetDisplayNameFromArrayIndex(uint arrayIndex, int rank, uint[] di
...
+                    // Pass the raw <Exec> text: ApplyExecToLocalVars detects the assigned variable
+                    // name(s) and substitutes local vars on the right-hand side itself. A single
+                    // <Exec> may update several variables, comma-separated (e.g. "++idx, ++statptr").
+                    var updatedVars = ApplyExecToLocalVars(exec.Value?.Trim() ?? "", localVars, out List<string> unhandledExec);
+                    foreach (string seg in unhandledExec)
+                    {
+                        // A segment we could not apply (unsupported form, or an undeclared
+                        // left-hand side) would otherwise silently do nothing; log it so the
+                        // omission is visible rather than producing a wrong/stuck traversal.
+                        _process.Logger.NatvisLogger?.WriteLine(LogLevel.Warning, "CustomListItems <Exec> segment not applied (unsupported expression or undeclared variable): " + seg);
```
gregg-miskelly: >_process [](http://example.com/codeflow?start=24&length=8)

Stop looping?

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1771,6 +1984,420 @@ private string GetDisplayNameFromArrayIndex(uint arrayIndex, int rank, uint[] di
...
+                    foreach (string updatedVar in updatedVars)
+                    {
+                        // Normalise each updated variable to prevent unbounded growth across
+                        // iterations: after each i++ the expression would otherwise grow as
+                        // "(((0)+1)+1)+1...". Evaluate it and, ONLY when the result is a plain
+                        // integer literal (the counter case), store that literal so the next
+                        // iteration starts compact. Anything else must keep its expression:
+                        // GetExpressionValue runs FormatDisplayString, so a pointer/struct local
+                        // is rendered as "0x..." or a visualizer display string — storing that
+                        // (a type-less address or non-C++ text) would corrupt later substitutions.
```
gregg-miskelly: This feels like it isn't going to scale past really tiny collections. Can you do some perf testing with a link list of say 500 elements?

Assuming this is a problem, my thought:
- Just handle integer types, and pointers, for pointers, we can save the hex address and cast it to the "type" value (ex: (MyStruct*)0x1234)
- For anything else: maybe add an error row after the first few elements

src/MIDebugEngine/Natvis.Impl/Natvis.cs
lugerard: The [More...] node only stored the next start index: expanding it re-ran the loop from the beginning and skipped the items already shown, so each page re-walked everything before it. It now saves the state instead, as you suggest: the loop variables (compact literals after the fix from your other comment) and the position, and expansion resumes from there, like the LinkedListItems/TreeItems continuations. Pages of the 500-element list now take a flat ~0.13s each regardless of depth.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
lugerard: done.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
lugerard: Confirmed with a 500-element linked list (lldb): each page was slower than the last (0.17s → 0.90s over the first five) and page 6 hung outright once the expression passed ~250 nested levels. Implemented your suggestion. The full list now expands in ~6s (10 pages, longest 1.1s). The residual per-page growth is the [More...] re-walk from your other comment.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -949,13 +1055,92 @@ private IVariableInformation[] ExpandVisualized(IVariableInformation variable)
...
+                    if (customList.Items1 != null)
+                    {
+                        foreach (var sz in customList.Items1)
+                        {
+                            if (!EvalCondition(sz.Condition, variable, visualizer.ScopedNames, visualizer.Intrinsics)) continue;
+                            try
+                            {
+                                string szExpr = SubstituteLocalVars(sz.Value?.Trim() ?? "0", localVars);
+                                string szVal = GetExpressionValue(szExpr, variable, visualizer.ScopedNames, visualizer.Intrinsics);
+                                totalSize = MICore.Debugger.ParseUint(szVal, throwOnError: false);
```
gregg-miskelly: >false [](http://example.com/codeflow?start=91&length=5)

This means that the size will be zero if the value isn't a uint. Is that what you want?

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1771,6 +2065,486 @@ private string GetDisplayNameFromArrayIndex(uint arrayIndex, int rank, uint[] di
...
+        /// <summary>
+        /// True when <paramref name="value"/> is a plain integer literal (optionally signed). Used
+        /// to decide whether an &lt;Exec&gt; result may be substituted back in place of its
+        /// expression: pointers ("0x...") and visualizer display strings are not scalar literals and
+        /// must keep their expression, or later substitutions/evaluations would be corrupted.
+        /// </summary>
+        internal static bool IsScalarLiteral(string value)
+        {
+            return !string.IsNullOrEmpty(value) && s_integerLiteral.IsMatch(value);
+        }
```
gregg-miskelly: This method doesn't seem to be used aside from unit testing

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1771,6 +2065,486 @@ private string GetDisplayNameFromArrayIndex(uint arrayIndex, int rank, uint[] di
...
+                        // debugger's expression limits. The variable is evaluated raw (no natvis
+                        // formatting) and stored back as a literal when possible: integers/bools
+                        // as-is, pointers as the address cast to their type ("(MyStruct *)0x1234",
+                        // stable while the process is stopped). Other types (e.g. a by-value
+                        // iterator) cannot be stored back as C++ text; a few such steps are
+                        // tolerated, then an error row is emitted and the loop stops rather than
+                        // degrade item by item.
+                        try
+                        {
+                            IVariableInformation updated = GetExpression(localVars[updatedVar], variable, visualizer.ScopedNames, null, visualizer.Intrinsics);
```
gregg-miskelly, line 2266: >GetExpression [](http://example.com/codeflow?start=59&length=13)

`MakeCompactLiteral` depends on integers coming back in decimal instead of hex, but `VaribaleInfo.SyncEval` is going to use the engine's current radix. So you need to either change MakeCompactLiteral to be less picky, or you need GetExpression to take an optional radix argument and push that though

lugerard: Made MakeCompactLiteral less picky: hex integers are stored back with their type cast ((int)0xffffffd6), which also keeps negative values negative; a bare hex literal is unsigned. Pointers are recognized by their type name first. Verified paging of the 500-element list under hex radix; the integer-literal cases are unit-tested.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
lugerard: Indeed.  Leftover after MakeCompactLiteral replaced it. Removed, along with its unit tests.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
lugerard: No. Fixed so it falls back to Break-driven termination instead of an empty expansion.

lugerard:
@gregg-miskelly All the points from your last review are addressed in the latest commits. Is there anything still blocking, or is it ready for another pass?

PR #1581, opened 2026-06-15
Use internal NuGet feed for release builds

WardenGnaw:
Release pipelines (VS, VSCode, DebuggerTesting) now use the MIEngine-Dependencies Azure Artifacts feed which upstreams nuget.org. This resolves NuGet restore failures caused by CFSClean network isolation policies blocking direct access to api.nuget.org.

CI builds continue using the public nuget.org feed directly.


gregg-miskelly:
Do we want PR builds to also use this feed so that we will get notified when we need to fix annoying upstream errors? Or would you rather deal with that as part of the release process?

gregg-miskelly (approved):

gregg-miskelly (approved):

gregg-miskelly (approved):

PR #1582, opened 2026-06-16
Add Podman container support for Attach to Process

WardenGnaw:
Adds Podman as a option in the Attach to Process dialog, enabling developers to discover and debug processes inside Podman containers.

TODO:
- [ ] The Podman port supplier also needs to be registered in vsdbg's VsIntegration.pkgdef for .NET debugging.


gregg-miskelly:
> "1"="{267B1341-AC92-44DC-94DF-2EE4205DD17E}"

Did you want to register LLDB also?

---
Refers to: src/MIDebugEngine/Microsoft.MIDebugEngine.pkgdef:87 in 18945f6. [](commit_id = 18945f6da49c752c98ae664afda61cb8a1f5a5c5, deletion_comment = False)

src/SSHDebugPS/ContainerRuntimeType.cs
```diff
@@ -9,6 +9,7 @@ namespace Microsoft.SSHDebugPS
     public enum ContainerRuntimeType
     {
         Unknown,
```
gregg-miskelly: >Unknown [](http://example.com/codeflow?start=8&length=7)

Can we remove `Unknown` or is it used?

src/SSHDebugPS/Docker/ContainerInstance.cs
```diff
@@ -67,10 +67,55 @@ protected DockerContainerInstance() { }
...
+
+        public static bool operator !=(ContainerInstance left, ContainerInstance right)
+        {
+            return !(left == right);
+        }
+
+        public bool Equals(IContainerInstance instance)
         {
-            if (instance is DockerContainerInstance other)
+            if (!ReferenceEquals(null, instance) && instance is ContainerInstance container)
```
gregg-miskelly: >!ReferenceEquals(null, instance) && i [](http://example.com/codeflow?start=16&length=37)

nit: this is redundant with the `is` check

src/SSHDebugPS/Docker/ContainerInstance.cs
```diff
@@ -67,10 +67,55 @@ protected DockerContainerInstance() { }
...
+
+        #region Helper Methods
+
+        // Container names: only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed. It is also case sensitive
+        protected virtual bool EqualsInternal(ContainerInstance instance)
+        {
+            if (instance is ContainerInstance other)
             {
                 // the id can be a partial on a container
                 return String.Equals(Id, other.Id, StringComparison.Ordinal) ||
```
gregg-miskelly: >return [](http://example.com/codeflow?start=16&length=6)

Since this type isn't sealed, should we also verify the types are equal?

gregg-miskelly: Looks like we could also make this method non-virtual if we did that.

Feels like we should either add the type check, or just make this method abstract and move this check into all the implementations.

src/SSHDebugPS/Podman/PodmanConnection.cs
```diff
@@ -0,0 +1,200 @@
...
+using System.Threading;
+using System.Diagnostics;
+using Microsoft.SSHDebugPS.Docker;
+using Microsoft.SSHDebugPS.Utilities;
+using Microsoft.VisualStudio.Debugger.Interop.UnixPortSupplier;
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.SSHDebugPS.Podman
+{
+    internal class PodmanConnection : PipeConnection
```
gregg-miskelly: >class [](http://example.com/codeflow?start=13&length=5)

nit: `sealed` where we can

src/SSHDebugPS/Podman/PodmanConnection.cs
```diff
@@ -0,0 +1,200 @@
...
+            return DockerConnection.CreateConnectionString(containerName, remoteConnectionName, hostName);
+        }
+
+        internal static bool TryConvertConnectionStringToSettings(string connectionString, out PodmanContainerTransportSettings settings, out Connection remoteConnection)
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+            remoteConnection = null;
+            settings = null;
+
+            string containerName = string.Empty;
```
gregg-miskelly: >string [](http://example.com/codeflow?start=12&length=6)

nit: Should we just call the Docker one and then convert the DockerTransportSettings to the podman version? Or are there side effects in the constructor?

gregg-miskelly (approved):
Otherwise LGTM

gregg-miskelly (approved):

PR #1587, opened 2026-06-18
Copilot instructions + test filtering fix

gregg-miskelly:
This PR:
- Adds copilot-instructions.md with links to documentation
- Adds a READMD.md to the various DebugEngineHost implementations to make that more clear
- Fixes a bug with how `DependsOnTest` works which made it so that any test with that attribute couldn't be run with a test filter

### DIFF
--- a/docs/Building-outside-of-VS-for-AI.md
+++ b/docs/Building-outside-of-VS-for-AI.md
@@ -1,0 +1,56 @@
+# Building MIEngine outside of Visual Studio
+
+⚠️ **Do not use these instructions when working inside Visual Studio.** When VS is hosting the conversation (e.g. via the in-IDE Copilot chat / `debugger_*` tools), prefer the IDE's built-in build functions, or functions like `debugger_launch` which have an implicit build. Those commands set up the VS environment correctly and update the Error List for you. The scripts below are for command-line / CI / VS Code scenarios where no IDE is available.
+
+## Solutions
+
+There are two solutions, picked by what you are targeting:
+
+| Solution | Use when… | How to build |
+| --- | --- | --- |
+| `src/MIDebugEngine.sln` | Building the **Visual Studio extension** (full set of projects, including the VSIX, COM/PIA references, IOSDebugLauncher, MIDebugPackage, SSHDebugPS, …). | `msbuild` from a **VS Developer Command Prompt** (Dev 17+ with the *Visual Studio extension development* workload), via `eng/Scripts/CI-Build.ps1`. **Windows only.** |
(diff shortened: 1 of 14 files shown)


docs/Building-outside-of-VS-for-AI.md
```diff
@@ -0,0 +1,56 @@
...
+## Command-line build
+
+### Windows (full VS extension build)
+
+From a VS Developer Command Prompt or after putting `MSBuild.exe` on `PATH`:
+
+```powershell
+# Debug, VS extension flavor (default)
+eng\Scripts\CI-Build.ps1 -Configuration Debug -TargetPlatform vs
+
```
WardenGnaw, line 25: This section is wrong, it will need to use `tools/Setup.cmd` to patch an instance of VS. And it do a UAC prompt. 

This just creates the VS Code version with OpenDebugAD7 and setups the tests in the right location.

docs/Building-outside-of-VS-for-AI.md
```diff
@@ -0,0 +1,56 @@
...
+## Outputs
+
+- `bin/<Configuration>/...` — primary binaries from `msbuild`.
+- `bin/<Configuration>/vscode/...` — the VS Code-flavor host (`Microsoft.DebugEngineHost.dll` for VSCode, `WindowsDebugLauncher.exe`, etc.).
+- `bin/DebugAdapterProtocolTests/<Configuration>/extension/debugAdapters/` — the staged VS Code debug adapter used by the CppTests.
+
+## Targeting
+
+The .NET SDK target is `net8.0` (CI installs `dotnet-version: 8.0.x`). Some projects multi-target .NET Framework for the in-VS host. Don't change target frameworks without checking both solutions still build.
```
WardenGnaw: Do we want to state net8.0 or can we be vague and be like The VS Code version uses .NET however, some projects multi-target .NET Framework. Don't touch those :D

WardenGnaw (changes requested):

docs/Building-outside-of-VS-for-AI.md
gregg-miskelly: Discussed offline - we don't want AI to do the patching, so I will leave this as-is

WardenGnaw: Discussed offline, doesnt make sense for AI to do the patching part since its a one time thing.

docs/Building-outside-of-VS-for-AI.md
gregg-miskelly: Sounds good

WardenGnaw (approved):

PR #1595, opened 2026-06-22
Prepare for nullable reference types in MICore

gregg-miskelly:
This PR contains work to prepare for enabling nullable reference types in MICore. Changes:

- Correct two incorrect annotations in DebugEngineHost
- Refactor MICommandFactory to use constructor injection (readonly _debugger field)
- Move MICommandFactory.GetInstance() from DebuggedProcess to Debugger constructor
- Fix bug: firstException should only be set when null (was checking != null)
- Add null safety: _transport?.Close(), ThreadCreatedEvent?.Invoke, ThreadExitedEvent?.Invoke
- Add null throw in SendToTransport for null transport
- Use GetTargetProcessExitedReason() instead of raw _closeMessage
- Null-safe access to _initialErrors/_initializationLog in OnDebuggerProcessExit
- Pass EventArgs.Empty instead of null for DebuggerExitEvent
- Use pattern matching for IsModuleLoad check
- Make _commandLock readonly, initialize _lastCommandText

### DIFF
--- a/src/DebugEngineHost.Common/HostLogChannel.cs
+++ b/src/DebugEngineHost.Common/HostLogChannel.cs
@@ -50,15 +50,15 @@ namespace Microsoft.DebugEngineHost
 
     public class HostLogChannel : ILogChannel
     {
-        private readonly Action<string> _log;
+        private readonly Action<string>? _log;
         private StreamWriter? _logFile;
         private LogLevel _minLevelToBeLogged;
 
         private readonly object _lock = new object();
 
-        private HostLogChannel() { _log = null!; }
+        private HostLogChannel() { }
 
-        public HostLogChannel(Action<string> logAction, string? file, LogLevel logLevel)
+        public HostLogChannel(Action<string>? logAction, string? file, LogLevel logLevel)
         {
             _log = logAction;
 
--- a/src/MICore/Debugger.cs
+++ b/src/MICore/Debugger.cs
@@ -1619,7 +1619,13 @@ namespace MICore
 
         private void SendToTransport(string cmd)
         {
-            ITransport transport = _transport ?? throw new InvalidOperationException();
+            ITransport transport = _transport;
+            if (transport is null)
+            {
+                Debug.Fail("Invalid: `SendToTransport` called before `Init`");
+                throw new InvalidOperationException();
+            }
+
             transport.Send(cmd);
 
             // https://github.com/Microsoft/MIEngine/issues/616 :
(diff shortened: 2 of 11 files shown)


src/DebugEngineHost.Common/HostLogChannel.cs
```diff
@@ -50,15 +50,15 @@ public interface ILogChannel
...
     {
-        private readonly Action<string> _log;
+        private readonly Action<string>? _log;
         private StreamWriter? _logFile;
         private LogLevel _minLevelToBeLogged;
 
         private readonly object _lock = new object();
 
-        private HostLogChannel() { _log = null!; }
+        private HostLogChannel() { }
```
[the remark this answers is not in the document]
gregg-miskelly: You are incorrect

src/MICore/Debugger.cs
```diff
@@ -1619,14 +1619,15 @@ private async void PostCommand(string cmd)
 
         private void SendToTransport(string cmd)
         {
-            _transport.Send(cmd);
+            ITransport transport = _transport ?? throw new InvalidOperationException();
+            transport.Send(cmd);
```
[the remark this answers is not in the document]
gregg-miskelly: Don't care

src/MICore/Debugger.cs
```diff
@@ -124,12 +124,12 @@ public StoppingEventArgs(Results results, BreakRequest asyncRequest = BreakReque
         }
 
         private ITransport _transport;
```
WardenGnaw, line 126: There are changes for `_transport?` but we are not declaring it as `ITransport?`?

src/MICore/Debugger.cs
```diff
@@ -1619,14 +1619,15 @@ private async void PostCommand(string cmd)
 
         private void SendToTransport(string cmd)
         {
-            _transport.Send(cmd);
+            ITransport transport = _transport ?? throw new InvalidOperationException();
```
WardenGnaw: Is this going to cause a crash if we get called via `PostCommand` and somehow _transport is null?

WardenGnaw (approved):

src/MICore/Debugger.cs
gregg-miskelly: It's because this PR doesn't actually enable nullability. That comes next. But to try and make that PR easier to review, I had AI split out the changes that had actual behavior change into their own PR.

src/MICore/Debugger.cs
gregg-miskelly: Correct. This isn't a new behavior -- it would have crashed before (with a NullReferenceException) also if something called `PostCommand` before Init. I just changed the exception type. Let me change this to an assert+throw though so we get slightly better diagnostics in debug builds.

PR #1596, opened 2026-06-26
Fix UI-thread hang when resuming the target during launch

WardenGnaw:
AD7Engine.ContinueFromSynchronousEvent (the AD7ProgramCreateEvent path) dispatched ResumeFromLaunch through the blocking WorkerThread.RunOperation, coupling the VS UI/SDM thread to debugger I/O. Over a slow or remote (SSH) transport, AD7Engine.ContinueFromSynchronousEvent + ResumeFromLaunch could stall for a long time or indefinitely, producing a Watson hang.

- Add WorkerThread.PostAsyncOperation, a non-blocking dispatch that claims the running-op slot (so later AD7 operations still serialize behind it) but returns without waiting for completion. Faults from the operation are routed to an onError callback from both the synchronous and async completion paths.
- ContinueFromSynchronousEvent now posts ResumeFromLaunch via PostAsyncOperation and returns S_OK immediately, so the UI thread is no longer blocked. Resume faults report through SendStartDebuggingError + Terminate via onError, and synchronous failures (OnLoadComplete or the dispatch itself) are still handled locally, since the SDM drops errors returned from this method.

- [x] Built and tested locally
  - [x] Debugged through WSL Linux gdb and the async resume completed.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/OperationThread.cs
+++ b/src/MIDebugEngine/Engine.Impl/OperationThread.cs
@@ -37,6 +37,12 @@ namespace Microsoft.MIDebugEngine
             /// Delegate that was added via 'RunOperation'. Is of type 'Operation' or 'AsyncOperation'
             /// </summary>
             public readonly Delegate Target;
+
+            /// <summary>
+            /// Handler invoked on the worker thread if the operation faults. Only set for PostAsyncOperation.
+            /// </summary>
+            public Action<Exception> ErrorHandler;
+
             public ExceptionDispatchInfo ExceptionDispatchInfo;
             public Task Task;
             private bool _isStarted;
@@ -121,6 +127,20 @@ namespace Microsoft.MIDebugEngine
             SetOperationInternalWithProgress(op, text, canTokenSource);
         }
 
+        /// <summary>
+        /// Send an async operation to the worker thread, claiming the running-op slot but returning without
+        /// waiting for it to complete. Later operations still serialize behind it. Faults are reported to
+        /// <paramref name="onError"/>.
+        /// </summary>
+        public void PostAsyncOperation(AsyncOperation op, Action<Exception> onError)
+        {
+            if (op == null)
+                throw new ArgumentNullException(nameof(op));
+            if (onError == null)
+                throw new ArgumentNullException(nameof(onError));
+
+            PostAsyncOperationInternal(op, onError);
+        }
 
         public void Close()
         {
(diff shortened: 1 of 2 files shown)


src/MIDebugEngine/Engine.Impl/OperationThread.cs
```diff
@@ -189,6 +210,27 @@ internal void SetOperationInternalWithProgress(AsyncProgressOperation op, string
...
+            // If this is called on the Worker thread it will deadlock
+            Debug.Assert(!IsPollThread());
+
+            while (true)
+            {
+                if (_isClosed)
+                    throw new ObjectDisposedException("WorkerThread");
+
+                // Wait for the slot so this serializes behind any in-flight operation.
+                _runningOpCompleteEvent.WaitOne();
```
gregg-miskelly: Instead of waiting for any in-flight operation, would it make more sense to add a `Queue<(AsyncOperation, Action<Exception>>`? I say that because we are calling this method `Post`, but it is only sort of a post.

src/MIDebugEngine/Engine.Impl/OperationThread.cs
```diff
@@ -333,11 +396,20 @@ private void ThreadFunc()
 
                         if (!completeAsync)
                         {
+                            // Capture the fault before clearing the slot so a synchronous throw is still reported.
+                            Action<Exception> errorHandler = runningOp.ErrorHandler;
+                            ExceptionDispatchInfo exceptionDispatchInfo = runningOp.ExceptionDispatchInfo;
```
gregg-miskelly, line 476: This is probably moot if we implement my other suggestion, but why do this here instead of on line 386?

gregg-miskelly: Eh, I guess that it is better to use `ExceptionDispatchInfo` for consistency.

gregg-miskelly (approved):
Looks good aside from the one suggestion

gregg-miskelly:
>                 }

Should probably drain the async operation queue also

---
Refers to: src/MIDebugEngine/Engine.Impl/OperationThread.cs:184 in d575ea9. [](commit_id = d575ea9a168f993dd70a4ffef9a389f60bee751c, deletion_comment = False)

gregg-miskelly (approved):
Aside from `Close`, LGTM

WardenGnaw:
Had to also fix thie issue with Background actions are getting stuck waiting in the queue and may never running because thread went back to sleep.

gregg-miskelly (approved):

WardenGnaw:
@microsoft-github-policy-service rerun

PR #1604, opened 2026-06-30
Fix race condition monitor thread in RegistryMonitor

### LINKED ISSUE #1593: Race condition in src/DebugEngineHost/RegistryMonitor.cs could leave thread running
A Copilot review pointed out that there is a race in src/DebugEngineHost/RegistryMonitor.cs . If `Start` is called and then `Stop` is called before the `Monitor` thread gets a chance to actually start, then the `Monitor` thread will run forever.

SachinM123:
This PR fixes a race condition in Microsoft.DebugEngineHost.RegisteryMonitor where if `Stop` was called before the monitor thread started, it would never shutdown.

This fixes https://github.com/microsoft/MIEngine/issues/1593


SachinM123:
@microsoft-github-policy-service agree

src/DebugEngineHost/RegistryMonitor.cs
```diff
@@ -97,7 +106,8 @@ private void Monitor()
             bool stopped = false;
             try
             {
-                _stoppedEvent = new AutoResetEvent(false);
+                _nativsLogger?.WriteLine(LogLevel.Verbose, "RegistryMonitor: Monitor thread starting (tid={0})", Thread.CurrentThread.ManagedThreadId);
```
gregg-miskelly: Remove the extra logging please

src/DebugEngineHost/RegistryMonitor.cs
```diff
@@ -133,9 +148,12 @@ private void Monitor()
             }
             finally
             {
-                _stoppedEvent?.Dispose();
-                _stoppedEvent = null;
-
+                _nativsLogger?.WriteLine(LogLevel.Verbose, "RegistryMonitor: Monitor thread exiting (tid={0})", Thread.CurrentThread.ManagedThreadId);
```
gregg-miskelly: Same

src/DebugEngineHost/RegistryMonitor.cs
```diff
@@ -111,7 +121,12 @@ private void Monitor()
                     {
                         while (!stopped)
                         {
-                            int waitResult = WaitHandle.WaitAny(new WaitHandle[] { _stoppedEvent, registryChangedEvent });
+                            AutoResetEvent? stoppedEvent;
+                            lock (_stopLock)
```
gregg-miskelly: Suggestion:
```
AutoResetEvent? stoppedEvent = _stoppedEvent;
Debug.Assert(stoppedEvent is not null, "Should be impossible - this code can only run after `Start` is called.");
```

Reasons:
1. It is better to assert then to silently suppress nullability warnings
2. There is no actual need to acquire the lock since this thread is guaranteed by `Start` to not get created until after `_stoppedEvent` is assigned

gregg-miskelly (changes requested):
A few small things to fix

src/DebugEngineHost/RegistryMonitor.cs
SachinM123: done

src/DebugEngineHost/RegistryMonitor.cs
SachinM123: done

src/DebugEngineHost/RegistryMonitor.cs
SachinM123: done

src/DebugEngineHost/RegistryMonitor.cs
```diff
@@ -111,7 +133,9 @@ private void Monitor()
                     {
                         while (!stopped)
                         {
-                            int waitResult = WaitHandle.WaitAny(new WaitHandle[] { _stoppedEvent, registryChangedEvent });
+                            AutoResetEvent? stoppedEvent = _stoppedEvent;
+                            Debug.Assert(stoppedEvent is not null, "Should be impossible - this code can only run after `Start` is called.");
+                            int waitResult = WaitHandle.WaitAny(new WaitHandle[] { stoppedEvent!, registryChangedEvent });
```
gregg-miskelly: >! [](http://example.com/codeflow?start=95&length=1)

nit: remove the `!` -- shouldn't be needed now that you added the assert

src/DebugEngineHost/RegistryMonitor.cs
```diff
@@ -97,7 +106,20 @@ private void Monitor()
...
+                {
+                    if (_isStopped)
+                    {
+                        return;
+                    }
+
+                    if (_stoppedEvent == null)
+                    {
+                        _stoppedEvent = new AutoResetEvent(false);
+                    }
```
gregg-miskelly: This shouldn't be possible. I would suggest just deleting it for clarity

SachinM123: done

src/DebugEngineHost/RegistryMonitor.cs
SachinM123: done

src/DebugEngineHost/RegistryMonitor.cs
```diff
@@ -133,9 +152,12 @@ private void Monitor()
             }
             finally
             {
-                _stoppedEvent?.Dispose();
-                _stoppedEvent = null;
-
+                _nativsLogger?.WriteLine(LogLevel.Verbose, "RegistryMonitor: Monitor thread exiting (tid={0})", Thread.CurrentThread.ManagedThreadId);
```
gregg-miskelly: Last thing: remove this log line

gregg-miskelly (approved):

gregg-miskelly:
@WardenGnaw Do you want to have a quick look before I merge this?

WardenGnaw (approved):

PR #1605, opened 2026-07-02
handle synthetic frames produced by GDB FrameFilter without a level

tzcnt:
Synthetic frames produced by GDB's FrameFilter, e.g. async stack backtraces for C++20 coroutines produced by a [GDB script](https://github.com/tzcnt/tmc-examples/blob/backtrace/coro_backtrace_gdb.py) do not have a Level, and cannot be assigned a Level with the current version of GDB. Attempting to set a breakpoint inside of an async stack then causes an error when walking the synthetic backtrace:

```
ERROR: Error while trying to enter break state. Debugging will now stop. Unrecognized format of field "level" in result: {addr=0x000055555555fef0,func=[async] main::$_0::operator()(unsigned long) const [clone .destroy],file=<file>,line=80,arch=i386:x86-64}
```

This PR updates MIEngine’s callstack and expression-evaluation paths to tolerate GDB FrameFilter “synthetic” frames that omit the `level` field, avoiding stack-walk failures and guarding frame-relative MI commands (locals/args/registers/eval) when the frame is not addressable by level.

**Changes:**
- Make `ThreadContext.Level` nullable and update stack walking/parsing to allow frames without `level`.
- Add defensive guards across locals/args/registers/expression evaluation paths when `Level == null`.
- Introduce a user-facing resource string for unsupported evaluation scenarios in non-level-bearing frames.

### DIFF
--- a/src/MIDebugEngine/AD7.Impl/AD7Thread.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Thread.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using Microsoft.VisualStudio.Debugger.Interop;
 using System.Diagnostics;
@@ -127,21 +128,23 @@ namespace Microsoft.MIDebugEngine
                 int numStackFrames = stackFrames != null ? stackFrames.Count : 0;
                 FRAMEINFO[] frameInfoArray;
 
-                if (numStackFrames == 0)
+                // -stack-list-arguments is a frame-relative command. If we walked no frames, or
+                // none of them carries a level (e.g. every frame is a GDB synthetic frame-filter
+                // frame), there is nothing addressable to query.
+                if (numStackFrames == 0 || !stackFrames.Any(f => f.Level != null))
                 {
-                    // failed to walk any frames. Return an empty stack.
+                    // failed to walk any real frames. Return an empty stack.
                     frameInfoArray = new FRAMEINFO[0];
                 }
                 else
                 {
-                    // -stack-list-arguments takes a low/high *frame number* range. When a Python
-                    // frame filter is active, those numbers index the decorated stack (synthetic
-                    // frames are counted but carry no level), which is the same order as
-                    // stackFrames here. So request the whole [0, count-1] range rather than a
-                    // range of levels: a level-based range would stop at the first synthetic
-                    // frame and miss every real frame below it. Synthetic frames in the range
-                    // have no arguments and are skipped in GetParameterInfoOnly, and results are
-                    // matched back to frames by level below, so ordering within the range is moot.
+                    // -stack-list-arguments takes a low/high *frame index* range. When a GDB Python
+                    // frame filter is active, those numbers index the decorated stack, not the GDB frame
+                    // level. Thus, synthetic frames are indexed even though they carry no level, and we
+                    // must request the whole [0, count-1] decorated range.
+                    // Synthetic frames in the range have no arguments and are skipped in
+                    // GetParameterInfoOnly, and results are matched back to frames by level below,
+                    // so ordering within the range is moot.
                     uint low = 0;
                     uint high = (uint)(stackFrames.Count - 1);
                     FilterUnknownFrames(stackFrames);
--- a/src/MIDebugEngine/AD7.Impl/AD7Thread.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Thread.cs
@@ -128,12 +128,9 @@ namespace Microsoft.MIDebugEngine
                 int numStackFrames = stackFrames != null ? stackFrames.Count : 0;
                 FRAMEINFO[] frameInfoArray;
 
-                // -stack-list-arguments is a frame-relative command. If we walked no frames, or
-                // none of them carries a level (e.g. every frame is a GDB synthetic frame-filter
-                // frame), there is nothing addressable to query.
-                if (numStackFrames == 0 || !stackFrames.Any(f => f.Level != null))
+                if (numStackFrames == 0)
                 {
-                    // failed to walk any real frames. Return an empty stack.
+                    // failed to walk any frames. Return an empty stack.
                     frameInfoArray = new FRAMEINFO[0];
                 }
                 else
@@ -152,7 +149,8 @@ namespace Microsoft.MIDebugEngine
                     frameInfoArray = new FRAMEINFO[numStackFrames];
                     List<ArgumentList> parameters = null;
 
-                    if ((dwFieldSpec & enum_FRAMEINFO_FLAGS.FIF_FUNCNAME_ARGS) != 0 && !_engine.DebuggedProcess.MICommandFactory.SupportsFrameFormatting)
+                    if ((dwFieldSpec & enum_FRAMEINFO_FLAGS.FIF_FUNCNAME_ARGS) != 0 && !_engine.DebuggedProcess.MICommandFactory.SupportsFrameFormatting
+                        && stackFrames.Any(f => f.Level != null))
                     {
                         _engine.DebuggedProcess.WorkerThread.RunOperation(async () => parameters = await _engine.DebuggedProcess.GetParameterInfoOnly(this, (dwFieldSpec & enum_FRAMEINFO_FLAGS.FIF_FUNCNAME_ARGS_VALUES) != 0,
                             (dwFieldSpec & enum_FRAMEINFO_FLAGS.FIF_FUNCNAME_ARGS_TYPES) != 0, low, high));
--- a/src/MIDebugEngine/AD7.Impl/AD7Thread.cs
+++ b/src/MIDebugEngine/AD7.Impl/AD7Thread.cs
@@ -134,8 +134,16 @@ namespace Microsoft.MIDebugEngine
                 }
                 else
                 {
-                    uint low = stackFrames[0].Level;
-                    uint high = stackFrames[stackFrames.Count - 1].Level;
+                    // -stack-list-arguments takes a low/high *frame number* range. When a Python
+                    // frame filter is active, those numbers index the decorated stack (synthetic
+                    // frames are counted but carry no level), which is the same order as
+                    // stackFrames here. So request the whole [0, count-1] range rather than a
+                    // range of levels: a level-based range would stop at the first synthetic
+                    // frame and miss every real frame below it. Synthetic frames in the range
+                    // have no arguments and are skipped in GetParameterInfoOnly, and results are
+                    // matched back to frames by level below, so ordering within the range is moot.
+                    uint low = 0;
+                    uint high = (uint)(stackFrames.Count - 1);
                     FilterUnknownFrames(stackFrames);
                     numStackFrames = stackFrames.Count;
                     frameInfoArray = new FRAMEINFO[numStackFrames];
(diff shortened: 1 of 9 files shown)


gregg-miskelly:
To make sure I understand correctly, it isn't possible to evaluate anything in these frames without levels correct?

If so, I would suggest a slightly different fix -- in `ThreadContext`, make [`Level`](https://github.com/microsoft/MIEngine/blob/4aef1f46e460748d6a98aca5156371701f259a75/src/MIDebugEngine/Engine.Impl/Structures.cs?plain=1#L35) a `uint?` instead of a `uint` and then fixup the various spots that are checking the level to early out if `Level` is null.

gregg-miskelly:
I should have said: thanks for taking the time to send this in!

tzcnt:
Synthetic frames from the Python frame filter have no underlying debugger frame so GDB emits no level for them (the original crash), and they can never be selected via --frame N for the target of an MI command (-stack-list-variables, -stack-list-arguments, registers, etc). The only thing you can do with them is show the actively executing line in each one, and show them in the stack trace.

Note that LLDB does not have this problem. LLDB synthetic frames get levels, and after https://github.com/llvm/llvm-project/pull/178575 they should be able to render the frame variables as well without any further MIEngine patch (I need to update my LLDB and test).

I've made the update as you suggested. As a result, querying a GDB synthetic async frame now returns 0 locals / 0 registers, rather than a neighboring real frame's data. Here's an example of the resulting display, with one of the synthetic async frames selected (no locals / no registers):

image

src/MIDebugEngine/AD7.Impl/AD7Thread.cs
```diff
@@ -134,8 +134,16 @@ int IDebugThread2.EnumFrameInfo(enum_FRAMEINFO_FLAGS dwFieldSpec, uint nRadix, o
                 }
                 else
                 {
-                    uint low = stackFrames[0].Level;
-                    uint high = stackFrames[stackFrames.Count - 1].Level;
+                    // -stack-list-arguments takes a low/high *frame number* range. When a Python
```
gregg-miskelly: Suggestion. Replace lines 130-146 with something like this (I didn't compile this, so sorry if it isn't quite right

```C#
                // Synthetic frames injected by a Python frame filter carry no level; only real
                // (level-bearing) frames can be targeted by frame-relative MI commands.
                IEnumerable<ThreadContext> realFrames = stackFrames?.Where(f => f.Level != null);
                ThreadContext firstFrame = realFrames?.FirstOrDefault();
                if (firstFrame is null)
                {
                    // failed to walk any real frames. Return an empty stack.
                    frameInfoArray = Array.Empty<FRAMEINFO>();
                }
                else
                {
                    uint low = firstFrame.Level;
                    uint high = realFrames.Last().Level;
```

So we are dependent on the way GDB/lldb-mi calculate levels, and so we don't go down the `else` block if somehow none of the frames have levels

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -633,7 +639,15 @@ internal async Task Eval(uint radix, enum_EVALFLAGS dwFlags = 0, DAPEvalFlags dw
...
 
                     int threadId = Client.GetDebuggedThread().Id;
-                    uint frameLevel = _ctx.Level;
+
+                    // Synthetic frames injected by a Python frame filter have no level and provide
+                    // no evaluation context. This is normally unreachable (annotated frames expose
+                    // no expression context), but guard defensively rather than crash on a null level.
+                    if (_ctx.Level == null)
+                    {
+                        return;
```
gregg-miskelly, line 649: Before returning, call `SetAsError(ResourceStrings.ExpressionEvalUnsupportedInFrame)`

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -585,10 +585,16 @@ public string EvalDependentExpression(string expr)
         {
             this.VerifyNotDisposed();
 
+            if (_ctx.Level == null)
+            {
+                return null;
```
gregg-miskelly: Callers don't expect this method to return `null`. So I would suggest instead acting like `-data-evaluate-expression` failed:

1. Add a new `ExpressionEvalUnsupportedInFrame` resource string -- add a new resource string to ResourceStrings.resx with a value like "Expression evaluation is unavailable in the current frame"
2. Add a `GetEvalUnsupportedException method:

```C#
private Exception GetEvalUnsupportedInFrameException() => new UnexpectedMIResultException(_debuggedProcess.MICommandFactory.Name, "-data-evaluate-expression", ResourceStrings.ExpressionEvalUnsupportedInFrame);
```
3. Change `return null` to `throw GetEvalUnsupportedInFrameException()`

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -957,10 +971,15 @@ public void Assign(string expression)
         {
             this.VerifyNotDisposed();
 
+            if (_ctx.Level == null)
+            {
+                return;
```
gregg-miskelly: Change to `throw GetEvalUnsupportedInFrameException();`

gregg-miskelly (changes requested):
Otherwise LGTM. Thanks for doing this!

src/MIDebugEngine/Engine.Impl/Variables.cs
tzcnt: Done in https://github.com/microsoft/MIEngine/pull/1605/changes/f17630b2a199eafe877f30858f1959581b23f378

src/MIDebugEngine/Engine.Impl/Variables.cs
tzcnt: Done in https://github.com/microsoft/MIEngine/pull/1605/changes/f17630b2a199eafe877f30858f1959581b23f378

src/MIDebugEngine/Engine.Impl/Variables.cs
tzcnt: Done in https://github.com/microsoft/MIEngine/pull/1605/changes/f17630b2a199eafe877f30858f1959581b23f378

src/MIDebugEngine/AD7.Impl/AD7Thread.cs
tzcnt: I've added the no-real-frames guard in https://github.com/microsoft/MIEngine/pull/1605/commits/2a9193e1161457bba7d257b34931e5abb0e9d80c. However, I kept `high = stackFrames.Count - 1` rather than `realFrames.Last().Level`. With frame filters active, GDB 17.2 returns the full list of frames (including synthetic) from both:
- `-stack-list-frames`: synthetic frames appear in the list, with no `level`
- `-stack-list-arguments`: synthetic frames appear in the list, with empty `args`

The `-stack-list-arguments` low-frame / high-frame parameters are indexes into this full list, not frame levels. There's no guarantee that real frames appear at any particular end of the range (in my example, they appear at the bottom and top of the range, with synthetic frames in the middle), so we need to scan the entire list.

gregg-miskelly (approved):

src/MIDebugEngine/Natvis.Impl/VisualizationCache.cs
```diff
@@ -21,7 +21,10 @@ public VisualizerKey(IVariableInformation variable)
             {
                 _name = variable.FullName();
                 _threadId = variable.Client.Id;
-                _level = (int)variable.ThreadContext.Level;
+                // Variables only ever come from real (level-bearing) frames; synthetic frames
+                // from a Python frame filter expose no locals to visualize. Fall back to 0 to
+                // keep the key total in the null case rather than throwing.
+                _level = (int)(variable.ThreadContext.Level ?? 0);
```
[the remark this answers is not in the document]
gregg-miskelly: This should never happen, so I think it is fine

src/MIDebugEngine/AD7.Impl/AD7Thread.cs
```diff
@@ -127,15 +128,25 @@ int IDebugThread2.EnumFrameInfo(enum_FRAMEINFO_FLAGS dwFieldSpec, uint nRadix, o
...
 
-                if (numStackFrames == 0)
+                // -stack-list-arguments is a frame-relative command. If we walked no frames, or
+                // none of them carries a level (e.g. every frame is a GDB synthetic frame-filter
+                // frame), there is nothing addressable to query.
+                if (numStackFrames == 0 || !stackFrames.Any(f => f.Level != null))
                 {
-                    // failed to walk any frames. Return an empty stack.
+                    // failed to walk any real frames. Return an empty stack.
                     frameInfoArray = new FRAMEINFO[0];
```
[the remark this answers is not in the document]
gregg-miskelly: Oh, good call. The `!stackFrames.Any(f => f.Level != null)` should be around lines 141-159

src/MIDebugEngine/AD7.Impl/AD7Thread.cs
```diff
@@ -127,15 +128,25 @@ int IDebugThread2.EnumFrameInfo(enum_FRAMEINFO_FLAGS dwFieldSpec, uint nRadix, o
...
+                    // level. Thus, synthetic frames are indexed even though they carry no level, and we
+                    // must request the whole [0, count-1] decorated range.
+                    // Synthetic frames in the range have no arguments and are skipped in
+                    // GetParameterInfoOnly, and results are matched back to frames by level below,
+                    // so ordering within the range is moot.
+                    uint low = 0;
+                    uint high = (uint)(stackFrames.Count - 1);
                     FilterUnknownFrames(stackFrames);
                     numStackFrames = stackFrames.Count;
                     frameInfoArray = new FRAMEINFO[numStackFrames];
```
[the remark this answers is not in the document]
gregg-miskelly: By design, and not a change in behavior

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -957,10 +974,15 @@ public void Assign(string expression)
         {
             this.VerifyNotDisposed();
 
+            if (_ctx.Level == null)
+            {
+                throw GetEvalUnsupportedInFrameException();
+            }
```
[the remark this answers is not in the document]
gregg-miskelly: Not a huge deal, but yes, I guess `GetEvalUnsupportedInFrameException` might as well take a string for the command name.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -585,10 +585,16 @@ public string EvalDependentExpression(string expr)
...
+            uint frameLevel = _ctx.Level.Value;
```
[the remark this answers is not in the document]
gregg-miskelly: The PR description is a bit out of date -- who cares

src/MIDebugEngine/Natvis.Impl/VisualizationCache.cs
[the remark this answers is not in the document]
gregg-miskelly: But it probably is worth an assert. Something like: `Debug.Assert(variable.ThreadContext.Level.HasValue, "How are we getting a variable from a synthetic thread context?");`

gregg-miskelly (changes requested):
Sorry, Copilot noticed a few small changes that seem to be worth addressing. Please fix and I will merge this. Thanks again!

tzcnt:
All should be addressed in the latest commit.

gregg-miskelly (approved):

PR #1606, opened 2026-07-08
Natvis: resolve typedef and base-class type names when rule lookup misses

lugerard:
GDB/LLDB report a variable's type under its declared name, so a typedef
alias or a subclass of a visualized type misses the wildcard rule of its
underlying type and falls back to raw struct display.
**Commit 1** adds a fallback to FindType: on a miss, ask the debugger to
resolve the name (gdb: whatis/ptype; lldb: type lookup) and retry the
rule lookup. Results are cached per session, and names are validated
before entering a console command.
**Commit 2** fixes base-class child classification for lldb, which
reports base-class children with a value where detection expected none;
without it, rules matched through a base class fail to evaluate
inherited members.
**Commit 3** adds an end-to-end test: a typedef alias and a subclass in
the natvis debuggee, verified to display through the underlying type's
visualizer.
Verified live on macOS/LLDB; the reply parser is unit-tested against
real gdb 16.3 output.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -297,10 +297,17 @@ namespace Microsoft.MIDebugEngine
 
             int index;
 
-            if (!results.Contains("value") && (Name == TypeName || (Name.Contains("::") && name == null)))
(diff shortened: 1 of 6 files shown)


lugerard:
@gregg-miskelly: could you take a first look at this PR?

PR #1612, opened 2026-07-21
Implement natvis 'na' modifier and $T substitution

SachinM123:
Add support for the natvis 'na' (no-address) modifier to strip MI's leading address prefix from string values. This requires tracking the format specifier through variable initialization and applying cleanup at value retrieval points.

Also implement template parameter substitution ($T1, $T2, etc.) within natvis brace expressions before format specifier extraction.

### DIFF
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -5,6 +5,7 @@ using MICore;
 using Microsoft.VisualStudio.Debugger.Interop;
 using Microsoft.VisualStudio.Debugger.Interop.DAP;
 using System;
+using Microsoft.DebugEngineHost;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
@@ -272,6 +273,17 @@ namespace Microsoft.MIDebugEngine
         {
             TypeName = results.TryFindString("type");
             Value = results.TryFindString("value");
+            // Diagnostic: log raw tuple child value before any cleanup
+            _debuggedProcess.Logger?.WriteLine(LogLevel.Verbose, FormattableString.Invariant($"TupleCtor: name={Name}, exp={results.TryFindString("exp")}, type={TypeName}, format={_format}, formatHasNa={_formatHasNa}, rawValue={results.TryFindString("value")}"));
+            // Only strip the leading MI address prefix ("0x... \"\"") when the natvis
+            // format included the 'na' modifier. 
+            if (_formatHasNa && !string.IsNullOrEmpty(Value) && Regex.IsMatch(Value, "^0x[0-9a-fA-F]+\\s+"))
+            {
+                // Recognize typical GDB prefix: 0x<hex> <string>
+                string before = Value;
+                Value = Regex.Replace(Value, "^0x[0-9a-fA-F]+\\s+", "");
+                _debuggedProcess.Logger?.WriteLine(LogLevel.Verbose, FormattableString.Invariant($"TupleCtor: stripped address prefix: before={before}, after={Value}"));
+            }
             Name = name ?? results.FindString("exp");
             if (results.Contains("dynamic"))
             {
@@ -332,6 +344,8 @@ namespace Microsoft.MIDebugEngine
             _internalName = results.FindString("name");
             IsChild = true;
             _format = parent._format; // inherit formatting
+            // inherit whether the parent's format included the 'na' modifier
+            _formatHasNa = parent._formatHasNa;
             _parent = parent.VariableNodeType == NodeType.AccessQualifier ? parent._parent : parent;
             this.PropertyInfoFlags = parent.PropertyInfoFlags;
         }
@@ -375,6 +389,9 @@ namespace Microsoft.MIDebugEngine
         private DeferedFormatExpression _deferedFormatExpression;
         private IVariableInformation _parent;
         private string _format;
+        // Indicates the original format specifier included the natvis "na" modifier
+        // (used to decide whether to strip MI's leading address prefix from string values)
+        private bool _formatHasNa = false;
         private string _strippedName;  // "Name" stripped of format specifiers
         private string _fullname;
 
@@ -416,6 +433,9 @@ namespace Microsoft.MIDebugEngine
 
             // Find the format specifier expression
             string expFS = exp.Substring(lastComma + 1).Trim();
+            // Detect whether the natvis 'na' modifier is present in the original format specifier.
+            // We must detect this before we strip modifiers below.
+            _formatHasNa = expFS.IndexOf("na", StringComparison.Ordinal) >= 0;
 
             // Strip off modifiers that may be included together with another format specifier, e.g. 'nvoXb' is a valid format specifier, but we only care about the 'Xb' part
             // This is not quite the right fix -- really the below switch statement should be a series of if statements. But since none of the supported format specifiers
@@ -698,6 +718,11 @@ namespace Microsoft.MIDebugEngine
                             _attribsFetched = true;
                         }
                         Value = results.TryFindString("value");
+                        // If natvis requested 'na', strip MI's leading address prefix
+                        if (_formatHasNa && !string.IsNullOrEmpty(Value) && Regex.IsMatch(Value, "^0x[0-9a-fA-F]+\\s+"))
+                        {
+                            Value = Regex.Replace(Value, "^0x[0-9a-fA-F]+\\s+", "");
+                        }
                         if ((string.IsNullOrEmpty(Value) || _format != null) && !string.IsNullOrEmpty(_internalName))
                         {
                             if (_format != null)
@@ -711,6 +736,11 @@ namespace Microsoft.MIDebugEngine
                                 if (results.ResultClass == ResultClass.done)
                                 {
                                     Value = results.FindString("value");
+                                    // If natvis requested 'na', strip MI's leading address prefix
+                                    if (_formatHasNa && !string.IsNullOrEmpty(Value) && Regex.IsMatch(Value, "^0x[0-9a-fA-F]+\\s+"))
+                                    {
+                                        Value = Regex.Replace(Value, "^0x[0-9a-fA-F]+\\s+", "");
+                                    }
                                 }
                                 else if (results.ResultClass == ResultClass.error)
                                 {
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -93,6 +93,18 @@ namespace Microsoft.MIDebugEngine
 
         static readonly Lazy<Regex> s_addressPattern = new Lazy<Regex>(() => new Regex(@"^(0x[0-9a-fA-F]+)\b"));
 
+        static readonly Regex s_naPattern = new Regex(@"^0x[0-9a-fA-F]+\s+(.)");
+
+        private static string StripLeadingAddress(string value, bool formatNa)
+        {
+            if(!formatNa || string.IsNullOrEmpty(value))
+            {
+                return value;
+            }
+
+            return s_naPattern.Replace(value, "$1");
+        }
+
         public string Address()
         {
             // ask GDB to evaluate "&expression"
@@ -239,7 +251,7 @@ namespace Microsoft.MIDebugEngine
             : this(ctx, engine, thread)
         {
             // strip off formatting string
-            _strippedName = ProcessFormatSpecifiers(expr, out _format);
+            _strippedName = ProcessFormatSpecifiers(expr, out _format, out _formatHasNa);
             Name = displayName;
             IsParameter = isParameter;
             _parent = null;
@@ -251,7 +263,7 @@ namespace Microsoft.MIDebugEngine
             : this(parent.ThreadContext, engine, parent.Client)
         {
             // strip off formatting string
-            _strippedName = ProcessFormatSpecifiers(expr, out _format);
+            _strippedName = ProcessFormatSpecifiers(expr, out _format, out _formatHasNa);
             Name = displayName ?? expr;
             _parent = parent;
             VariableNodeType = NodeType.Synthetic;
@@ -262,7 +274,7 @@ namespace Microsoft.MIDebugEngine
             : this(parent._ctx, parent._engine, parent.Client)
         {
             // strip off formatting string
-            _strippedName = ProcessFormatSpecifiers(expr, out _format);
+            _strippedName = ProcessFormatSpecifiers(expr, out _format, out _formatHasNa);
             Name = expr;
             VariableNodeType = NodeType.Root;
         }
@@ -273,17 +285,10 @@ namespace Microsoft.MIDebugEngine
         {
             TypeName = results.TryFindString("type");
             Value = results.TryFindString("value");
-            // Diagnostic: log raw tuple child value before any cleanup
-            _debuggedProcess.Logger?.WriteLine(LogLevel.Verbose, FormattableString.Invariant($"TupleCtor: name={Name}, exp={results.TryFindString("exp")}, type={TypeName}, format={_format}, formatHasNa={_formatHasNa}, rawValue={results.TryFindString("value")}"));
-            // Only strip the leading MI address prefix ("0x... \"\"") when the natvis
-            // format included the 'na' modifier. 
-            if (_formatHasNa && !string.IsNullOrEmpty(Value) && Regex.IsMatch(Value, "^0x[0-9a-fA-F]+\\s+"))
-            {
-                // Recognize typical GDB prefix: 0x<hex> <string>
-                string before = Value;
-                Value = Regex.Replace(Value, "^0x[0-9a-fA-F]+\\s+", "");
-                _debuggedProcess.Logger?.WriteLine(LogLevel.Verbose, FormattableString.Invariant($"TupleCtor: stripped address prefix: before={before}, after={Value}"));
-            }
+            // Only strip the leading MI address prefix ("0x... ") when the natvis format included the 'na' modifier.
+            Value = StripLeadingAddress(Value, _formatHasNa);
+
+
             Name = name ?? results.FindString("exp");
             if (results.Contains("dynamic"))
             {
@@ -418,8 +423,9 @@ namespace Microsoft.MIDebugEngine
 
         private static Regex s_isFunction = new Regex(@".+\(.*\).*");
 
-        private string ProcessFormatSpecifiers(string exp, out string formatSpecifier)
+        private string ProcessFormatSpecifiers(string exp, out string formatSpecifier, out bool formatNa)
         {
+            formatNa = false;
             formatSpecifier = null; // will be used with -var-set-format
 
             if (EngineUtils.IsConsoleExecCmd(exp, out string _, out string _))
@@ -436,6 +442,7 @@ namespace Microsoft.MIDebugEngine
             // Detect whether the natvis 'na' modifier is present in the original format specifier.
             // We must detect this before we strip modifiers below.
             _formatHasNa = expFS.IndexOf("na", StringComparison.Ordinal) >= 0;
+            formatNa = _formatHasNa;
 
             // Strip off modifiers that may be included together with another format specifier, e.g. 'nvoXb' is a valid format specifier, but we only care about the 'Xb' part
             // This is not quite the right fix -- really the below switch statement should be a series of if statements. But since none of the supported format specifiers
@@ -719,10 +726,7 @@ namespace Microsoft.MIDebugEngine
                         }
                         Value = results.TryFindString("value");
                         // If natvis requested 'na', strip MI's leading address prefix
-                        if (_formatHasNa && !string.IsNullOrEmpty(Value) && Regex.IsMatch(Value, "^0x[0-9a-fA-F]+\\s+"))
-                        {
-                            Value = Regex.Replace(Value, "^0x[0-9a-fA-F]+\\s+", "");
-                        }
+                        Value = StripLeadingAddress(Value, _formatHasNa);
                         if ((string.IsNullOrEmpty(Value) || _format != null) && !string.IsNullOrEmpty(_internalName))
                         {
                             if (_format != null)
@@ -737,10 +741,7 @@ namespace Microsoft.MIDebugEngine
                                 {
                                     Value = results.FindString("value");
                                     // If natvis requested 'na', strip MI's leading address prefix
-                                    if (_formatHasNa && !string.IsNullOrEmpty(Value) && Regex.IsMatch(Value, "^0x[0-9a-fA-F]+\\s+"))
-                                    {
-                                        Value = Regex.Replace(Value, "^0x[0-9a-fA-F]+\\s+", "");
-                                    }
+                                    Value = StripLeadingAddress(Value, _formatHasNa);
                                 }
                                 else if (results.ResultClass == ResultClass.error)
                                 {
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -441,8 +441,7 @@ namespace Microsoft.MIDebugEngine
             string expFS = exp.Substring(lastComma + 1).Trim();
             // Detect whether the natvis 'na' modifier is present in the original format specifier.
             // We must detect this before we strip modifiers below.
-            _formatHasNa = expFS.IndexOf("na", StringComparison.Ordinal) >= 0;
-            formatNa = _formatHasNa;
+            formatNa = expFS.IndexOf("na", StringComparison.Ordinal) >= 0;
 
             // Strip off modifiers that may be included together with another format specifier, e.g. 'nvoXb' is a valid format specifier, but we only care about the 'Xb' part
             // This is not quite the right fix -- really the below switch statement should be a series of if statements. But since none of the supported format specifiers
--- a/src/MIDebugEngine/Engine.Impl/Variables.cs
+++ b/src/MIDebugEngine/Engine.Impl/Variables.cs
@@ -95,9 +95,9 @@ namespace Microsoft.MIDebugEngine
 
         static readonly Regex s_naPattern = new Regex(@"^0x[0-9a-fA-F]+\s+(.)");
 
-        private static string StripLeadingAddress(string value, bool formatNa)
+        internal static string StripLeadingAddress(string value)
         {
-            if(!formatNa || string.IsNullOrEmpty(value))
+            if (string.IsNullOrEmpty(value))
             {
                 return value;
             }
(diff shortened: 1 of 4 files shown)


src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -272,6 +273,17 @@ private VariableInformation(TupleValue results, VariableInformation parent, stri
         {
             TypeName = results.TryFindString("type");
             Value = results.TryFindString("value");
+            // Diagnostic: log raw tuple child value before any cleanup
+            _debuggedProcess.Logger?.WriteLine(LogLevel.Verbose, FormattableString.Invariant($"TupleCtor: name={Name}, exp={results.TryFindString("exp")}, type={TypeName}, format={_format}, formatHasNa={_formatHasNa}, rawValue={results.TryFindString("value")}"));
```
gregg-miskelly: These should already be in the MI log. Is there value in keeping them?

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -272,6 +273,17 @@ private VariableInformation(TupleValue results, VariableInformation parent, stri
...
+            // Only strip the leading MI address prefix ("0x... \"\"") when the natvis
+            // format included the 'na' modifier. 
+            if (_formatHasNa && !string.IsNullOrEmpty(Value) && Regex.IsMatch(Value, "^0x[0-9a-fA-F]+\\s+"))
```
gregg-miskelly: >" [](http://example.com/codeflow?start=105&length=1)

You want a non-space character after the spaces

gregg-miskelly: >" [](http://example.com/codeflow?start=85&length=1)

Since we will be using the Regex often, probably better to move it to a static variable. See `s_addressPattern` for an example, though I don't think you need to bother making this a `Lazy`

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -272,6 +273,17 @@ private VariableInformation(TupleValue results, VariableInformation parent, stri
...
+            {
+                // Recognize typical GDB prefix: 0x<hex> <string>
+                string before = Value;
+                Value = Regex.Replace(Value, "^0x[0-9a-fA-F]+\\s+", "");
+                _debuggedProcess.Logger?.WriteLine(LogLevel.Verbose, FormattableString.Invariant($"TupleCtor: stripped address prefix: before={before}, after={Value}"));
```
gregg-miskelly: >_debuggedProcess [](http://example.com/codeflow?start=16&length=16)

Remove

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -416,6 +433,9 @@ private string ProcessFormatSpecifiers(string exp, out string formatSpecifier)
 
             // Find the format specifier expression
             string expFS = exp.Substring(lastComma + 1).Trim();
+            // Detect whether the natvis 'na' modifier is present in the original format specifier.
+            // We must detect this before we strip modifiers below.
+            _formatHasNa = expFS.IndexOf("na", StringComparison.Ordinal) >= 0;
```
gregg-miskelly: >_formatHasNa [](http://example.com/codeflow?start=12&length=12)

The other outputs from this method are return as out params/return value, so I you want this as an `out bool formatNa` to go along them them.

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -698,6 +718,11 @@ internal async Task Eval(uint radix, enum_EVALFLAGS dwFlags = 0, DAPEvalFlags dw
                             _attribsFetched = true;
                         }
                         Value = results.TryFindString("value");
+                        // If natvis requested 'na', strip MI's leading address prefix
+                        if (_formatHasNa && !string.IsNullOrEmpty(Value) && Regex.IsMatch(Value, "^0x[0-9a-fA-F]+\\s+"))
```
gregg-miskelly: >"^0x[0-9a-fA-F]+\\s+" [](http://example.com/codeflow?start=97&length=21)

Same suggestions on the other usage of the Regex

gregg-miskelly: Actually, better yet: can you extract this into a helper method so you don't repeat it three times?

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1283,12 +1283,24 @@ private string FormatValue(string format, IVariableInformation variable, IDictio
                     if (m.Success)
                     {
                         string rawExpr = format.Substring(i + 1, m.Length - 2);
+                        // Substitute template parameter macros ($T1, $T2, ...) inside the whole brace
+                        // expression (this covers both the expression and any trailing format specifier)
+                        if (scopedNames != null)
+                        {
+                            rawExpr = Regex.Replace(rawExpr, "\\$T\\d+", (Match mt) =>
```
gregg-miskelly: >"\\$T\\d+" [](http://example.com/codeflow?start=61&length=10)

nit: this would be easier to read as `@"\$T\d+"`

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1283,12 +1283,24 @@ private string FormatValue(string format, IVariableInformation variable, IDictio
...
+                            {
+                                if (scopedNames.TryGetValue(mt.Value, out string replacement))
+                                    return replacement;
+                                return mt.Value;
+                            });
+                        }
+                        bool hasNa = HasNaModifier(rawExpr);
```
gregg-miskelly: >bool hasNa = HasNaModifier(rawExpr); [](http://example.com/codeflow?start=24&length=36)

Instead of duplicating the work of `ExtractFormatSpecifier`, have `ExtractFormatSpecifier` return this as an out param

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1283,12 +1283,24 @@ private string FormatValue(string format, IVariableInformation variable, IDictio
...
                         string spec = ExtractFormatSpecifier(rawExpr);
                         string exprValue = GetExpressionValue(rawExpr, variable, scopedNames, intrinsics);
-                        if (spec == "sub" || spec == "su")
-                            exprValue = CleanUtf16StringValue(exprValue);
-                        else if (spec == "sb")
-                            exprValue = CleanAsciiStringValue(exprValue);
```
gregg-miskelly, line 1291: Did you mean to loose this `sub`/`su`/`sb` code?

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1502,27 +1514,33 @@ internal static string ExtractFormatSpecifier(string expression)
...
             // Strip leading "0x<hex> " address prefix emitted by GDB/LLDB.
             value = s_addressPrefix.Replace(value, "");
-            // Strip surrounding u"..." or U"..." quotes.
-            if (value.Length >= 3 &&
-                (value.StartsWith("u\"", StringComparison.Ordinal) || value.StartsWith("U\"", StringComparison.Ordinal)))
-            {
-                value = value.EndsWith("\"", StringComparison.Ordinal)
-                    ? value.Substring(2, value.Length - 3)
-                    : value.Substring(2);
-            }
```
gregg-miskelly, line 1525: `,sub` is supposed to remove the quotes/prefix

gregg-miskelly: Though this changes looks like it is correct for `,su`

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1283,12 +1283,24 @@ private string FormatValue(string format, IVariableInformation variable, IDictio
...
+                        // expression (this covers both the expression and any trailing format specifier)
+                        if (scopedNames != null)
+                        {
+                            rawExpr = Regex.Replace(rawExpr, "\\$T\\d+", (Match mt) =>
+                            {
+                                if (scopedNames.TryGetValue(mt.Value, out string replacement))
+                                    return replacement;
+                                return mt.Value;
+                            });
+                        }
```
gregg-miskelly: If I am reading the code correctly, this is already at least partially done by `ReplaceNamesInExpression`, which is called by `GetExpressionValue`, so I don't think you want to do this over the entire expression.

Can you trace through why `ReplaceNamesInExpression` isn't doing this already? (lines 1713-1717).

gregg-miskelly:
Can you add a test for your changes? You may also need to adjust a test baseline for `su` since the code was doing the wrong thing there.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
gregg-miskelly: Actually, is `su` needed? Variabel.cs line 477 already handles `su` correctly.

SachinM123:
> Can you add a test for your changes? You may also need to adjust a test baseline for `su` since the code was doing the wrong thing there.

The test seems to get properly handled via TestDisplayString in NatvisTests to me

src/MIDebugEngine/Natvis.Impl/Natvis.cs
SachinM123: Yeah it seems like ReplaceNamesInExpression should properly handle the substitution, and it seems like it gets called at the right time, so I have no answer for why it isn't doing it already

src/MIDebugEngine/Natvis.Impl/Natvis.cs
SachinM123: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
SachinM123: done

src/MIDebugEngine/Engine.Impl/Variables.cs
SachinM123: done; helper method created

src/MIDebugEngine/Engine.Impl/Variables.cs
SachinM123: done

src/MIDebugEngine/Engine.Impl/Variables.cs
SachinM123: done

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -272,6 +273,17 @@ private VariableInformation(TupleValue results, VariableInformation parent, stri
...
+            _debuggedProcess.Logger?.WriteLine(LogLevel.Verbose, FormattableString.Invariant($"TupleCtor: name={Name}, exp={results.TryFindString("exp")}, type={TypeName}, format={_format}, formatHasNa={_formatHasNa}, rawValue={results.TryFindString("value")}"));
+            // Only strip the leading MI address prefix ("0x... \"\"") when the natvis
+            // format included the 'na' modifier. 
+            if (_formatHasNa && !string.IsNullOrEmpty(Value) && Regex.IsMatch(Value, "^0x[0-9a-fA-F]+\\s+"))
+            {
+                // Recognize typical GDB prefix: 0x<hex> <string>
+                string before = Value;
+                Value = Regex.Replace(Value, "^0x[0-9a-fA-F]+\\s+", "");
+                _debuggedProcess.Logger?.WriteLine(LogLevel.Verbose, FormattableString.Invariant($"TupleCtor: stripped address prefix: before={before}, after={Value}"));
+            }
```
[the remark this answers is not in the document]
SachinM123: change accepted

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1531,9 +1549,9 @@ internal static string CleanUtf16StringValue(string value)
         /// (i.e. one evaluated with the <c>,sb</c> format specifier).
         /// GDB and LLDB prefix the string with the pointer address, e.g.
         ///   <c>0x00007fff5fbff6c0 "Hello"</c>
-        /// This method strips the address and the surrounding <c>"…"</c> quotes so that
-        /// the NatVis DisplayString shows just the string content (matching VS behaviour,
-        /// where <c>{ptr,sb}</c> evaluates to bare text without quotes).
+        /// This method strips the leading address prefix that GDB/LLDB emits ("0x... ").
+        /// It does NOT remove surrounding quotes; callers should decide whether quotes
+        /// should be removed based on the caller's context.
```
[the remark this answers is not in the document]
SachinM123: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1283,12 +1283,24 @@ private string FormatValue(string format, IVariableInformation variable, IDictio
...
+                            rawExpr = Regex.Replace(rawExpr, "\\$T\\d+", (Match mt) =>
+                            {
+                                if (scopedNames.TryGetValue(mt.Value, out string replacement))
+                                    return replacement;
+                                return mt.Value;
+                            });
+                        }
+                        bool hasNa = HasNaModifier(rawExpr);
                         string spec = ExtractFormatSpecifier(rawExpr);
                         string exprValue = GetExpressionValue(rawExpr, variable, scopedNames, intrinsics);
```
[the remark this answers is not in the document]
SachinM123: change accepted

src/MIDebugEngine/Engine.Impl/Variables.cs
SachinM123: created helper method

SachinM123: done

src/MIDebugEngine/Engine.Impl/Variables.cs
SachinM123: removed

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -272,6 +285,10 @@ private VariableInformation(TupleValue results, VariableInformation parent, stri
         {
             TypeName = results.TryFindString("type");
             Value = results.TryFindString("value");
+            // Only strip the leading MI address prefix ("0x... ") when the natvis format included the 'na' modifier.
+            Value = StripLeadingAddress(Value, _formatHasNa);
```
gregg-miskelly: >StripLeadingAddress [](http://example.com/codeflow?start=20&length=19)

nit: Either change to:
```
if (_formatHasNa)
{
   Value = StripLeadingAddress(Value);
}
```
Or rename to `MaybeStripLeadingAddress` 

And get rid of the comment and extra trailing whitespace

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1283,12 +1283,23 @@ private string FormatValue(string format, IVariableInformation variable, IDictio
...
+                        bool hasNa = HasNaModifier(rawExpr);
                         string exprValue = GetExpressionValue(rawExpr, variable, scopedNames, intrinsics);
-                        if (spec == "sub" || spec == "su")
-                            exprValue = CleanUtf16StringValue(exprValue);
-                        else if (spec == "sb")
-                            exprValue = CleanAsciiStringValue(exprValue);
+                        if (hasNa && !string.IsNullOrEmpty(exprValue))
+                        {
+                            exprValue = s_addressPrefix.Replace(exprValue, "");
+                        }
```
gregg-miskelly: Suggestion: Make `VariableInformation.StripLeadingAddress` public so you can use it here

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1494,35 +1505,46 @@ private static int FindLastTopLevelComma(string expression)
         /// <see cref="VariableInformation.ProcessFormatSpecifiers"/>: modifiers "nvo", "na",
         /// "nr", "nd" are stripped before returning.  Returns null when no specifier is present.
         /// </summary>
-        internal static string ExtractFormatSpecifier(string expression)
+        internal static string ExtractFormatSpecifier(string expression, out bool hasNa)
```
gregg-miskelly, line 1512: >ExtractFormatSpecifier [](http://example.com/codeflow?start=31&length=22)

You need to update src/MIDebugEngineUnitTests/NatvisFormatSpecifierTest.cs for this change

src/MIDebugEngine/Engine.Impl/Variables.cs
gregg-miskelly: >_formatHasNa [](http://example.com/codeflow?start=47&length=12)

You are reading this before you assigned to it

SachinM123: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
SachinM123: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
SachinM123: done

src/MIDebugEngine/Engine.Impl/Variables.cs
SachinM123: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1502,27 +1514,33 @@ internal static string ExtractFormatSpecifier(string expression)
...
-            // Strip surrounding u"..." or U"..." quotes.
-            if (value.Length >= 3 &&
-                (value.StartsWith("u\"", StringComparison.Ordinal) || value.StartsWith("U\"", StringComparison.Ordinal)))
-            {
-                value = value.EndsWith("\"", StringComparison.Ordinal)
-                    ? value.Substring(2, value.Length - 3)
-                    : value.Substring(2);
-            }
             return value;
         }
```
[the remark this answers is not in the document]
SachinM123: change reverted

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -416,6 +439,10 @@ private string ProcessFormatSpecifiers(string exp, out string formatSpecifier)
 
             // Find the format specifier expression
             string expFS = exp.Substring(lastComma + 1).Trim();
+            // Detect whether the natvis 'na' modifier is present in the original format specifier.
+            // We must detect this before we strip modifiers below.
+            _formatHasNa = expFS.IndexOf("na", StringComparison.Ordinal) >= 0;
```
gregg-miskelly: >_formatHasNa [](http://example.com/codeflow?start=12&length=12)

Not that it really matters, but to be the same as the rest of the outputs of this function, you want to just assign to the out param (which in all cases currently is just a ref to `_formatHasNa`, but you want to keep that consistency for the future)

src/MIDebugEngine/Natvis.Impl/Natvis.cs
gregg-miskelly: This problem seems to still be here

src/MIDebugEngine/Natvis.Impl/Natvis.cs
gregg-miskelly: We should debug through it and see why it doesn't work. Let me know if you need help.

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1494,30 +1505,48 @@ private static int FindLastTopLevelComma(string expression)
...
+        /// whether the original expression asked for the "na" behavior.
+        /// </summary>
+        private static bool HasNaModifier(string expression)
+        {
+            int commaPos = FindLastTopLevelComma(expression);
+            if (commaPos < 0) return false;
+            string tail = expression.Substring(commaPos + 1);
+            return tail.IndexOf("na", StringComparison.Ordinal) >= 0;
+        }
+
```
gregg-miskelly, line 1524: This method should be deleted in favor of just using `ExtractFormatSpecifier`

SachinM123: done

src/MIDebugEngine/Engine.Impl/Variables.cs
SachinM123: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1283,12 +1283,23 @@ private string FormatValue(string format, IVariableInformation variable, IDictio
...
+                                    return replacement;
+                                return mt.Value;
+                            });
+                        }
+                        ExtractFormatSpecifier(rawExpr, out bool hasNa);
                         string exprValue = GetExpressionValue(rawExpr, variable, scopedNames, intrinsics);
-                        if (spec == "sub" || spec == "su")
-                            exprValue = CleanUtf16StringValue(exprValue);
-                        else if (spec == "sb")
-                            exprValue = CleanAsciiStringValue(exprValue);
```
gregg-miskelly, line 1291: This is still not addressed - you want to put back the old code for `sub`/`sb`

src/MIDebugEngine/Natvis.Impl/Natvis.cs
SachinM123: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
SachinM123: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
SachinM123: I traced through it and I think I understand why it works. it goes from
1.	GetExpressionValue (line 1287) calls ReplaceNamesInExpression (line 1739)
2.	ReplaceNamesInExpression (lines 1685-1720) calls ProcessNamesInString (line 1694)
3.	ProcessNamesInString (lines 1321-1375) uses regex pattern s_variableName = [a-zA-Z$_][a-zA-Z$_0-9]* to match identifiers
4.	Pattern Matching: $T1, $T2, etc. are correctly matched because:
- $ is in the first character class [a-zA-Z$_]
- 	T1, T2, etc. match [a-zA-Z$_0-9]*
5.	Substitution: The matched $Tx values are looked up in scopedNames dictionary (lines 1713-1717) and replaced
and so the duplicate code wasn't needed because the manual Regex.Replace(rawExpr, @"\$T\d+", ...) was redundant because ProcessNamesInString already does this correctly
- The removal was correct, and tests passing confirms the original flow works

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1494,30 +1498,40 @@ private static int FindLastTopLevelComma(string expression)
...
+        internal static string ExtractFormatSpecifier(string expression, out bool hasNa)
         {
+            hasNa = false;
             int commaPos = FindLastTopLevelComma(expression);
             if (commaPos < 0) return null;
+
+            string tail = expression.Substring(commaPos + 1).Trim();
+            hasNa = tail.IndexOf("na", StringComparison.Ordinal) >= 0;
+
             return expression.Substring(commaPos + 1).Trim()
```
gregg-miskelly: >expression.Substring(commaPos + 1).Trim() [](http://example.com/codeflow?start=19&length=41)

Replace with `tail`

src/MIDebugEngine/Engine.Impl/Variables.cs
```diff
@@ -92,6 +93,18 @@ internal sealed class VariableInformation : IVariableInformation
...
+
+        internal static string StripLeadingAddress(string value)
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                return value;
+            }
+
+            return s_naPattern.Replace(value, "$1");
+        }
```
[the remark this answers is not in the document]
gregg-miskelly: False. `0x1234 "Hello"` becomes `"Hello"`

src/MIDebugEngine/Natvis.Impl/Natvis.cs
SachinM123: done

gregg-miskelly:
>                         if (spec == "sub" || spec == "su")

Can you restore your code to delete `|| spec == "su"`

---
Refers to: src/MIDebugEngine/Natvis.Impl/Natvis.cs:1288 in 596849a. [](commit_id = 596849aec482bef9dc556cfa510d5ee7a4eb360b, deletion_comment = False)

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1502,13 +1502,31 @@ internal static string ExtractFormatSpecifier(string expression)
...
+            int commaPos = FindLastTopLevelComma(expression);
+            if (commaPos < 0) return null;
+
+            string tail = expression.Substring(commaPos + 1).Trim();
+            hasNa = tail.IndexOf("na", StringComparison.Ordinal) >= 0;
+
+            return tail
+                .Replace("nvo", "").Replace("na", "").Replace("nr", "").Replace("nd", "");
+        }
+
```
gregg-miskelly, line 1524: Am I missing something? It looks like nothing is calling your new method

SachinM123: now calls the method string spec = ExtractFormatSpecifier(rawExpr, out bool hasNa); at 1286 natvis.cs

src/MIDebugEngine/Natvis.Impl/Natvis.cs
```diff
@@ -1283,12 +1283,16 @@ private string FormatValue(string format, IVariableInformation variable, IDictio
...
                         string rawExpr = format.Substring(i + 1, m.Length - 2);
-                        string spec = ExtractFormatSpecifier(rawExpr);
+                        string spec = ExtractFormatSpecifier(rawExpr, out bool hasNa);
                         string exprValue = GetExpressionValue(rawExpr, variable, scopedNames, intrinsics);
-                        if (spec == "sub" || spec == "su")
+                        if (spec == "sub")
                             exprValue = CleanUtf16StringValue(exprValue);
                         else if (spec == "sb")
                             exprValue = CleanAsciiStringValue(exprValue);
+                        if (hasNa)
```
gregg-miskelly: >if [](http://example.com/codeflow?start=24&length=2)

This should be `else if` (since a string containing a hex number would otherwise be double escaped)

src/MIDebugEngineUnitTests/NatvisFormatStringTest.cs
```diff
@@ -0,0 +1,85 @@
...
+using Microsoft.MIDebugEngine;
+
+namespace MIDebugEngineUnitTests
+{
+    /// <summary>
+    /// unit tests for natvis format string processing with format specifiers
+    /// tests the handling of format specifiers like, 'na', sub (UTF-16), sb (ASCII).
+    /// </summary>
+    public class NatvisFormatStringTest
+    {
```
gregg-miskelly: >{ [](http://example.com/codeflow?start=4&length=1)

Move these into NatvisFormatSpecifierTest

gregg-miskelly:
Can you add a CppTests based test using the scenario in the bug you are addressing? Example: https://github.com/Microsoft/MIEngine/pull/1559

src/MIDebugEngineUnitTests/NatvisFormatStringTest.cs
SachinM123: done

src/MIDebugEngine/Natvis.Impl/Natvis.cs
SachinM123: done

PR #1561, opened 2026-05-19
Add configurable debuginfod settings to prevent GDB hangs

WardenGnaw:
## Why is this change being made?

Fixes https://github.com/microsoft/vscode-cpptools/issues/14458

MIEngine commit f169b02 added `set debuginfod enabled on` unconditionally in GetInitializeCommands(). On systems where DEBUGINFOD_URLS points to an unreachable server (e.g. debuginfod.ubuntu.com which is currently down), subsequent GDB commands block for 90+ seconds waiting for the network request to time out, causing GDB to appear hung on launch.

## Summarize what changed

- Added a `debuginfod` launch option (JSON: `{ enabled: bool, timeout: int }` XML: `EnableDebuginfod`/`DebuginfodTimeout` attributes) defaulting to enabled with a 30-second timeout.
- When enabled (timeout > 0): sets DEBUGINFOD_TIMEOUT and DEBUGINFOD_MAXTIME environment variables on the GDB process to cap network requests.
- When enabled (timeout = 0): no timeout override is applied; GDB/libdebuginfod defaults are used.
- When disabled: clears DEBUGINFOD_URLS on the GDB process so debuginfod connections are never attempted, even if the system has the env var set.
- `set debuginfod enabled on` is now conditional on the EnableDebuginfod setting.
- Added XSD schema validation (minInclusive=0) and JSON schema (minimum: 0).
- Negative timeout values are clamped to the default (30) at runtime.

## How was the change tested?

- [x] Changes reviewed by Copilot CLI (using /review)
- [x] All 36 MICoreUnitTests pass
- [x] Integration tests (DebuginfodTests) pass in WSL with GDB 12.1: - DebuginfodDisabledDoesNotHang: steps into std::regex with unreachable server, verifies launch completes in < 10s - DebuginfodTimeoutPreventsHang: steps into std::regex with 5s timeout, verifies launch completes in < 30s
- [x] All existing CppTests pass (Sample, Execution, Breakpoint, Environment)
- [x] Verified via TCP listener that debuginfod URLs are actually queried during step-into-library operations

PR #1613, opened 2026-08-10
Backport configurable debuginfod settings for VS 17.14

WardenGnaw:
Backports MIEngine PR #1561 onto the MIEngine version used by Visual Studio 17.14.

Linux remote debugging can stall for several minutes before attaching because the older MIEngine unconditionally enables GDB debuginfod downloads. This change adds configuration for enabling debuginfod and controlling its download timeout.

Changes

• Backport configurable debuginfod settings from #1561.
• Update the MIEngine version from 17.12 to 17.14.
• Bring the applicable release-pipeline configuration in sync with main for producing the insertion build.


loc/lci/OpenFolderSchema.json.lci
```diff
@@ -338,6 +338,60 @@
           <Cmt Name="LcxAdmin"><![CDATA[{Locked}]]></Cmt>
         </Cmts>
       </Item>
+      <Item ItemId=";debugExtensions.cppdbg.schema.definitions.cpp_schema.properties.debuginfod.description" ItemType="0" PsrId="306" InstFlg="true" Leaf="true">
```
gregg-miskelly, line 341: >Item [](http://example.com/codeflow?start=7&length=4)

You want to also port all the OpenFolderSchema.json.lcl files

gregg-miskelly: And probably all the other changes to this file too?

gregg-miskelly: And OpenFolderSchema.json.lci?

src/MICore/LaunchOptions.cs
```diff
@@ -1211,6 +1224,81 @@ public UnknownBreakpointHandling UnknownBreakpointHandling
             }
         }
 
+        private bool _enableDebuginfod = true;
```
gregg-miskelly: >_enableDebuginfod [](http://example.com/codeflow?start=21&length=17)

Do we want it or enabled or disabled by default?

WardenGnaw: Shoot. I also need https://github.com/microsoft/MIEngine/pull/1574 . Great catch.

loc/lci/OpenFolderSchema.json.lci
gregg-miskelly: I guess the answer to this depends on if we want it enabled or disabled by default, as it looks like we didn't fix localization during the time what it was enabled by default

WardenGnaw: I had the loc changes in that one commit that made it default `false` which I just cherry-picked over`

gregg-miskelly (approved):

PR #1617, opened 2026-08-19
Keep the debug session alive when one of several gdb inferiors exits

### LINKED ISSUE #13893: When debugging multiple inferiors with gdb, if one inferior exits the full debug session stops
### Environment

- OS and version: Arch Linux with gdb 16.3
- VS Code: 1.103.1
- C/C++ extension: 1.26.3
- OS and version of remote machine (if applicable): n/a
- GDB / LLDB version: 16.3

### Bug Summary and Steps to Reproduce

Bug Summary:

When debugging multiple inferiors in gdb, if one inferior exits the vscode debug session ends.

Steps to reproduce:
1. git clone this repo: https://github.com/lostgoat/multiinferior-repro
2. Set a breakpoint on the "parent exiting" line
3. Start the "(gdb) Launch" debug task

When running gdb with the same configuration via commandline everything works correctly. Execution pauses after the child exits, and then one can use `inferior <id>` to switch back to the parent and keep debugging. 

### Debugger Configurations

```shell
{
            "name": "(gdb) Launch",
            "type": "cppdbg",
            "preLaunchTask": "build test app",
            "request": "launch",
            "program": "${workspaceFolder}/test.out",
            "args": [],
            "stopAtEntry": false,
            "cwd": "${fileDirname}",
            "environment": [],
            "externalConsole": false,
            "MIMode": "gdb",
            "setupCommands": [
                {
                    "description": "Enable pretty-printing for gdb",
                    "text": "-enable-pretty-printing",
                    "ignoreFailures": true
                },
                {
                    "description": "Keep both sides of fork",
                    "text": "set detach-on-fork off"
                },
                {
                    "description": "Allow multiple inferiors to run",
                    "text": "set schedule-multiple on"
                }
            ],
            "logging": { "engineLogging": true, "trace": true, "traceResponse": true }
        },
```

### Debugger Logs

```shell
Output window:

Starting test
parent running 
waiting for child: 965664
parent waitpid returned: 0 - 0
child started
child looping: 0
child looping: 1
child looping: 2
child looping: 3
parent waitpid returned: 0 - 0
child looping: 4
child looping: 5
child looping: 6
child looping: 7
parent waitpid returned: 0 - 0
child looping: 8
child looping: 9
child exiting
```

### Other Extensions

_No response_

### Additional Information

_No response_

lostgoat:
gdb sends *stopped,reason="exited-normally" whenever any inferior exits, with nothing in the record identifying which one, so MIEngine treated the exit of a forked child as the end of the debuggee and sent -gdb-exit. In a multi-inferior session (set detach-on-fork off, set schedule-multiple on) the session now survives: gdb sends =thread-group-exited for the dying inferior first, so a non-empty _debuggeePids means other inferiors are still alive. In that case reflect gdb's all-stop state, select a surviving thread (gdb leaves the exited inferior current, which fails any command that doesn't name a thread), and report a break instead of tearing down.

Fixes microsoft/vscode-cpptools#13893


lostgoat:
This commit was co-authored by claude, so I don't think I'll be able to agree to the `Originality of Work.` section.

I'm going to close this MR in favour of an issue report and maybe you can point copilot at that. The fix should be straightforward. 

https://github.com/microsoft/MIEngine/issues/1618

